PR #1, opened 2018-12-17
Added test/runtests.jl and added to .gitignore

charleskawczynski:


test/runtests.jl
```diff
@@ -0,0 +1,54 @@
+using Test
+
+function is_test_folder(f, path_separator)
+  a = split(f, path_separator)
+  dir = join(a[1:end-1], path_separator)
```
lcw: You may want to look at `joinpath` it takes care of the path separator for you.

test/runtests.jl
```diff
@@ -0,0 +1,54 @@
...
+  d = split(dir, path_separator)[end]
```
lcw: There will be a `splitpath` in future versions of julia https://github.com/JuliaLang/julia/issues/24477

test/runtests.jl
charleskawczynski: Ah, ok, I'll update.

test/runtests.jl
charleskawczynski: Thanks, I wasn't crazy about this function-- I'll update this in my next PR.

lcw: no problem; also see the `isdir` function.

lcw: also `basename`

PR #5, opened 2018-12-19
Placing the Planet parameters into a module

jkozdon:
Placing the Planet parameters into a module

Here we are putting all the planet parameters into a module. Doing this
leads to several benefits:

# Namespace
The user need not load all the parameters but only the ones they need.
For instance one can do
```julia
julia> using PlanetParameters: gas_constant
```
which will make on the `gas_constant` in the local name space

# Documentation
We have also added documentation for the paremeters:
```julia
julia> ?gas_constant
help?> gas_constant
search: gas_constant

  gas_constant

  Universal gas constant (J/mol/K)

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> gas_constant
  gas_constant = 8.3144598 [J/mol/K]
```

# Typing for used
Using `Irrational` as a base type, we have introduced a new type
`Parameter`, which allows us to ensure that undesired promotion does not
occur. For instance consider the following code:
```julia
julia> y = Float32(2.3) * gas_constant
19.123257f0
julia> typeof(y)
Float32
```
Note the that return type is a `Float32`. If `gas_constant` was a
`Float32` the return type would have been a `Float32`:
```julia
julia> gas_constant64 = 8.3144598
8.3144598

julia> typeof(gas_constant64)
Float64

julia> y = Float32(2.3) * gas_constant64
19.12325714353566

julia> typeof(y)
Float64
```

# Generated Code
In the generated code the values are placed in with the required types
(they are not variables):
```julia
julia> @code_typed Float32(1) + gas_constant
CodeInfo(
313 1 ─ %1 = (Base.add_float)(x, 8.31446)::Float32
    └──      return %1
) => Float32

julia> @code_typed 1.0 + gas_constant
CodeInfo(
313 1 ─ %1 = (Base.add_float)(x, 8.31446)::Float64
    └──      return %1
) => Float64
```


jkozdon:
FYI: @lcw

tapios:
Is there a simple way to export all parameters in planet_constants.jl within the module, without listing each explicitly?

glwagner:
The `Parameter` type seems nice!

@tapios, I think

```
julia> using PlanetParameters
```

without specifying any particular names will do the trick.

fxgiraldo:
That’s correct. Modules in Julia are similar in this respect to modules in modern Fortran.


Frank

tapios:
A couple more thoughts on the parameters:

1) We will need to specify more parameters that are not necessarily as widely used throughout the code, e.g., chemical reaction rates. We should have separate modules for those, but it would be good for them to also use the parameter type. Hence, can we make the parameter type global in some more central place?

2) Can we give the parameter type a third argument 'definition' that contains the definition of the parameter, e.g., 

@parameter gas_constant 8.3144598     "J/mol/K"   "Universal gas constant"

This would have the advantage that the definition is visually where the parameter is defined. 

3) Can the expanded definition of the parameter type under 2) be used to automatically generate the jldoctest code? That way, we would just need to change the definition of the parameter in the line where the parameter is declared, without having to touch the documentation manually.

4) Once the parameter definition is moved out of the constants file under 1), it would seem preferable to put the definitions of the constants into the same file as the module declaration. If there is no way to export all parameters with one command, at least it would mean that parameter definitions/additions can be made in one file only, rather than having to change, e.g., 'gas_constant' in 2 files (and in 3 different places, as it currently is, if the documentation is manually generated).

Generally, points 2-4 aim at increasing clarity, and reducing the number of lines that need to be changed if a parameter is changed to as few as possible (ideally, only one LOC).

jkozdon:
Glad folks like the idea.

@tapios I think that most of these can all be addressed fairly easily. Wanted to get something out there to see if there was traction on the idea, then keep blowing things out. @lcw and I were talking, and there are a number of other uses we can see for this as well (such as ODE solver coefficients and the like).

Going to work some more on this today and we can iterate on the API / structure.

PR #7, opened 2018-12-20
Parameters modules

jkozdon:
I have broken the parameters functionality into a module and refactored the Planet_parameters to use this functionality. This address some of the issues raised by @tapios in #5.

I have also included a dependency on [Unitful](https://github.com/ajkeller34/Unitful.jl) so that the parameters can be typed with units. When used though, the parameters tend to be extracted as some form of a float.


tapios:
Thanks, @jkozdon. I like the incorporation of definition and documentation generation. From a quick glance, I am not sure what unitful adds for us. The units as typed in their syntax are hard to read (glued to the numerical values), and we do not really need them in the code. It also forces the otherwise unnecessary use of getval. My preference would be to stick the units back into the definition (documentation) and work without unitful.

tapios:
I'd also suggest to call the ParametersType directory just something like Types, or CLIMATypes, and similarly for the module. Other type declarations are likely to come, and this can be a central place for all of them.

glwagner:
How about something like `ClimaUtils` for 'miscellaneous' functionality that is meant to be shared across all Clima code? `PlanetParameters` doesn't just define a type, for example.

Two cents on the word 'parameters' (I hope this is helpful) --- to me, a 'parameter' is more general than what's being referred to here (a parameter might be any scalar that a model depends on, like grid resolution, time step, etc). The objects referred to here are things that have units and can be measured. Perhaps 'PhysicalConstant` is better, and `@physconst`, or something like that? Just a thought.

tapios:
@glwagner I prefer parameter here. Only a few of these are actually physical constants (in that they cannot change). Most of them are parameter that can change (e.g., for other planets). In that sense, they are like other adjustable parameters in the model (just with much smaller uncertainties!). I was hoping that the planet in PlanetParameters makes it clear that this is characterizing the planet, rather than, say, the model configuration.

PR #16, opened 2018-12-21
Typo in PlanetParameters

jkozdon:
In the PlanetParameters `T_trip` in the definition of `T_0` should be `T_triple`


tapios:
> @tapios Can you verify that this fix is correct.

Yes, that's correct. Thanks!

jkozdon:
Thanks. Trying to add some testing to prevent this in the future.

PR #18, opened 2018-12-21
Adding test for ParametersType and PlanetParameters

jkozdon:
Adding in some test for the ParametersType and PlanetParameters modules.


tapios:
I wouldn't include this test:

@test Float64(kappa_d) == Float64(2//7)

We may one day want to use the model for an atmosphere that is not diatomic, without triggering an error. 

You could test L_f0  = L_s0 - L_v0, which should hold universally.

PR #27, opened 2019-01-04
Adding MPI (and cmake) dependency to README.md

jkozdon:
Here we are adding some information concerning the `MPI.jl` and `cmake`
dependency of the CLIMA-atmos Dycore. We also offer some advice on how to begin
tackling MPI related build issues.


jkozdon:
Duplicated in #28

Issue #30, opened 2019-01-04
Utilities module


jkozdon:
There was a discussion to move the Parameters type to a general utilities module. Just recording that idea here.

tapios:
I support a utilities module. I also suggest going back to ParametersType, to make it clearer that this is a type, not a module defining parameters (Julia conventions notwithstanding, I find this clearer).

Issue #20, opened 2018-12-21
Thermodynamics additions/checks needed


tapios:
- Thermodynamics.jl should be checked carefully

- Tests should be added. For example, we must have

air_pressure([1, 1], [1, 1], [1, 0], [0, 0]) = [R_v, R_d]
latent_heat_vapor(T_0) = L_v0
latent_heat_fusion(T_0) = L_f0
latent_heat_sublim(T_0) = L_s0
sat_vapor_press_liquid(T_triple) = sat_vapor_press_triple
sat_vapor_press_ice(T_triple) = sat_vapor_press_triple

- As the dycore evolves, we should make sure the data structure for moisture variables (e.g., q_t, q_l, q_i) in the thermodynamics module is consistent with how the thermodynamic functions are called from the dycore and parameterizations. We may need to adapt these.

jkozdon:
addressed by #32

PR #32, opened 2019-01-08
Add Utilities package

lcw:
This creates the Utilities package as suggested in issue #30 and places
the MoistThermodynamics module into it.  Tests are also added for the
MoistThermodynamics module.

The next step in addressing issue #30 would be to move Parameters into
this package.


jkozdon:
Looks good to me. Thanks!

jkozdon:
This also addresses #20

PR #33, opened 2019-01-09
Renaming `Parameters` to `ParametersType`

jkozdon:
This address @tapios comment in \#30 concerning the changing the `Parameters`
package to `ParametersType`


tapios:
Looks good. We probably also should rename MoistThermodynamics to MoistThermo to reduce typing in constructs such as MoistThermodynamics.sat_shum.

PR #34, opened 2019-01-13
[AtmosDycore] Major Refactor of AtmosDycore

jkozdon:
The major motivation behind this refactoring is the split apart in a more coherent manner the problem setup, the state of the simulation, and the runtime configurations.

There is now a `CLIMAAtmosDycore.Runner` struct which contains separate runners for the spatial and temporal discretizations. We have tried to make this sufficiently expressive to not cripple future spatial and temporal discretizations, but also kept it simple enough that users of the `CLIMAAtmosDycore` need not know the implementation details. We have also sought to ensure that different spatial and temporal discretizations are in independent "sandboxes", and generally speaking cross submodule access should be done with collection/array-like access,e.g. `runner[:time]` and `runner[:Q]`

Currently we have one spatial and temporal runner: a "vanilla" DG spatial discretization and a low storage Runge-Kutta method. Each of these has three primary structs:

**`Parameters`**
The `Parameters` `struct` contains all the data necessary to initialize a simulation apart from the solution vector and initial condition. This can be thought of as the "input" file for a fresh simulation

**`Configuration`**
The `Configuration` `struct` contains all the runtime configuration that is not required to be checkpointed for simulation restart. This is all data that once the parameters and MPI configuration are known can be regenerated. For example, MPI buffers, metric terms, RK coefficients would all be part of the configuration since they can be generated as needed (e.g., do not change in time).

**`State`**
The `State` `struct` is all of the dynamic data for a simulation that must be saved once checkpointing is implemented. This contains for instance the solution vector, time step, simulation time, etc..

**Note:** The design decision has been made that time is kept track of by the space discretization. The reason for this is that the solution is at a given time and the time stepper can be used to advance the solution.


jkozdon:
cc/ @lcw 

I just added a commit with the commit message below. If we like this approach, I/we should update the pull request description (possibly squashing all the commits too).

---
[AtmosDycore] Switching to Runner (from 3 types)

Upon further reflection, since everything is now in submodules it seemed
best to remove the `Parameters`, `Configuration`, and `State` types at
the user level. It caused some awkward calling conventions for the
accessor functions since either the caller needed to know how something
was stored, or had to awkwardly pass through all three variables.

To avoid this clunkiness I have introduced a new type `Runner` for the
parent and submodules. The parent module Runner just contains the two
space and time runners which should be accessed via indexing
(`runner[:space]` or `runner[:time]`, strings can also be used). For
concrete space and time discretization submodules the parameters,
configuration, and state still seems like a reasonable/useful
abstraction for organization and to facilitate (yet-to-be-implemented)
checkpointing.

lcw:
@jkozdon The new runner seems like the way forward.  Do you want me to squash and make a new pull request?

jkozdon:
After discussion with @lcw, this pull-request is being accepted.

PR #38, opened 2019-01-17
[AtmosDycore] Fix deadlock in wall time callback

lcw:
Ensure that all ranks call a wall time callback.  Before only some ranks
would call the callback and some would execute the next step.  When this
happened a reduction (e.g, L2 error calculation) in the callback would
cause a deadlock.


jkozdon:
Thanks!

PR #39, opened 2019-01-18
[AtmosDycore] Implementation of tracers and (passive) moist variables

jkozdon:
This primarily implements tracers and moist variables for the AtmosDycore on both the CPU and GPU. Right now the moist variables have no physics and are just passively advected. As a test we make sure that if the tracers are initialized to be the scaled density that they remain this through out the simulation.

A few other things that are caught up in commits with this pull-request:

- In `VanillaEuler` the  Euler state variable storage order now matches the input function ordering
- In `VanillaEuler` the `facerhs!` on the CPU is now unified for 2- and 3-D
- In `GenericCallbacks` there is a comment fix for `EveryXWallTimeSeconds`
- In `VanillaEuler` an unneeded `MJI` array is removed on the CPU
- In `VanillaEuler` some postscript `L` are removed in the device arrays for the GPU kernel wrapper functions
- In `VanillaEuler` the `Val`s in the MPI buffer functions are removed

### DIFF
--- a/src/CLIMA-atmos/Dycore/TODO.md
+++ b/src/CLIMA-atmos/Dycore/TODO.md
@@ -1,13 +1,14 @@
 # TODO:
 
- - Consider how to do logging with MPI
- - Restart files
- - pretty print time info (and track performance issues)
- - asynchronous GPU memory/compute
- - asynchronous i/o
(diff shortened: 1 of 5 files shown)


jkozdon:
Looks good. Thanks @lcw

PR #43, opened 2019-01-23
Acceptable Unicode & modified CodingConventions

charleskawczynski:
Added a list of acceptable Unicode, and modified the CodingConventions doc, so that Unicode doesn't go astray.


docs/AcceptableUnicode.md
```diff
@@ -0,0 +1,92 @@
...
+∂ # (partial)
+∇ # (nabla/del), note that nabla and del are indistinguishable
+∀ # (forall)
+≥ # (greater than equal to)
+≤ # (less than equal to)
+
+## Indistinguishable characters
+Some characters are indistinguishable. For example:
+
+  ∇ = 1 # nabla
```
jkozdon: So are nabla and del acceptable? (It's unclear in the document)

docs/AcceptableUnicode.md
```diff
@@ -0,0 +1,92 @@
...
+ζ # (zeta)
+ϕ # (psi)
+φ # (varphi)
+
+## Acceptable upper-case Greek letters:
+Δ # (Delta)
+∑ # (Sigma)
+Γ # (Gamma)
+Ω # (Omega)
+Φ # (Phi)
```
jkozdon: I find this hard to distinguish from the lowercase is some fonts.

docs/AcceptableUnicode.md
```diff
@@ -0,0 +1,92 @@
...
+Ψ # (Psi)
+
+## Acceptable mathematical symbols:
+∫ # (int)
+∬ # (iint)
+∭ # (iiint)
+∈ # (in)
```
jkozdon: Are we avoiding:
```
∉ # \notin
≠ # \ne
≉ # \napprox
```
The later two are hard to distinguish I think

jkozdon: Also both `∈ # (in)` and `ϵ # (epsilon)` can be hard to distinguish in some fonts.

docs/AcceptableUnicode.md
```diff
@@ -0,0 +1,92 @@
...
+avoiding problematic Unicode usage. Here is a list of acceptable
+and forbidden Unicode characters and accents. Please note that
+we forbid the use of accents, because this can lead to visually
+ambiguous characters (see below).
+
+## Acceptable lower-case Greek letters:
+α # (alpha)
+β # (beta)
+δ # (delta)
+ϵ # (epsilon)
```
jkozdon: What about `ε # \varepsilon`?

docs/CodingConventions.md
```diff
@@ -50,16 +50,14 @@ A list of recommended coding conventions.
...
+## * Why do we limit our Unicode use?
 
  - Some characters are visibly indistinguishable. Capital "a" and capital alpha are visibly indistinguishable, but are recognized as separate characters (e.g., search distinguishable).
 
  - Some characters are difficult to read. Sometimes, the overline/overdot/hats overlap with characters making them difficult to see.
 
  - Portability issues. Unicode does not render in Jupyter notebook natively (on OSX).
 
- - If it does improve readability enough, and are not worried about portability, we may introduce a list of permissable characters that are commonly used.
+ - If it does improve readability enough, and are not worried about portability, we may introduce a list of permissible characters that are commonly used.
```
jkozdon: Do we apply the conventions across all internals of CLIMA, or just the more public facing bits?

glwagner: My 2 cents: I think anything that improves the readability of internals is "good", while API / user-facing things should be designed with ease of use. So it might make sense to allow arbitrary characters for code that is typed once or twice and read many times, and restrict API code to symbols that can be "easily typed" by a mainstream user.

docs/AcceptableUnicode.md
charleskawczynski: I'll add this, thanks!

docs/AcceptableUnicode.md
charleskawczynski: Yes, `Phi` and `psi` are similar.. Maybe we should remove `Phi`?

docs/AcceptableUnicode.md
charleskawczynski: I would suggest acceptable. I'll make this more clear.

docs/AcceptableUnicode.md
charleskawczynski: Yeah, I suggest that we avoid those negated symbols. I have an umbrella statement at the end of the doc, I'll move it to the beginning. And good point about `∈ # (in)` and `ϵ # (epsilon)`. I'm torn. My vote would be to ban `in` to be more consistent with allowing greek letters.

docs/AcceptableUnicode.md
charleskawczynski: Any suggestion about `\nu` and `\upsilon` @jkozdon ?

docs/AcceptableUnicode.md
charleskawczynski: Yes, the `\Phi` and `\psi` are very similar. I can remove `\Phi`.

docs/CodingConventions.md
charleskawczynski: I suggest these rules be followed across all internals. We can put more restrictions on interfaces if necessary, but I think that this list should provide fundamental readability standards.

docs/AcceptableUnicode.md
charleskawczynski: Can you give an example?

jkozdon: I meant \phi and \Phi (lowercase and uppercase phi)

charleskawczynski: Oh, okay. Regardless, I removed `\Phi` from the list for now.

jkozdon:
Might be good to have JuliaLab folks weigh in on this too (so we do not stray too far from Julia conventions)

peterahrens:
I agree that unicode modifier characters like vec and hat seem to display poorly, and this list is a great way to avoid look-alikes. I have no objections to merging this, as it seems like the kind of document which can change as programmers need new characters.

docs/CodingConventions.md
peterahrens: I think we can do both. If you use a unicode function name in a public facing interface, just also have an ascii analogue and document them together
```julia
"""
    ∇(f)
Returns the gradient of some function `f`.
See also: [`gradient`](@ref).
"""
∇(f) = some(complicated(stuff(f)))

"""
    gradient(f)
An alias for [`∇`](@ref).
"""
gradient(f) = ∇(f)
```

peterahrens:
Regarding the coding conventions at large, I have three suggestions:

1. Consider that Julia code very often uses four spaces for whitespace. At least, all of the code in Julia Base does. Feel free to ignore this point.

2. On the "Single access point" bullet, I would also add that "Any time a chunk of code is used more than once, or when several similar versions exist across the codebase, consider generalizing this functionality and using a new function to avoid replicating code"

3. on the "2 or more, use a "for"" rule. This is a good rule, but Julia lets us go farther.

  If indices are written out one-by-one like "a[1] = 4 a[2] = 2 a[3] = 6", this is usually done instead of a for loop on purpose. If each index will be handled differently and/or has some semantic meaning (like index `1` means pressure and index `2` means weight or something), consider using a `struct` or `SVector` to represent the array instead of writing it all out, or perhaps using separate arrays for separate quantities. See #45 for an example.

  If the loop has been manually unrolled for performance reasons, I would suggest that a for loop first be used with constant loop bounds (the compiler is pretty smart about whether it's going to be worth it to unroll a loop with constant bounds). If that's not enough and you need to manually unroll, let me know and I can add a quick macro to the CLIMA utilities that manually unrolls loop expressions so that readers don't have to suffer.

I understand my suggestions are not hard and fast rules (maybe they aren't really style). Just some things I thought about as I read the guide.

tapios:
Thanks, @peterahrens! These all seem like really good suggestions to me.

charleskawczynski:
I agree, thanks Peter! I may remove this 2-or more rule since it may not be generally true. I can update the spaces. Fortunately, we're not using Python so, this is not super critical IMO.

Issue #47, opened 2019-02-02
Remove `getindex` and `setindex!`


jkozdon:
Switch to using functions and not `getindex` and `setindex!` in the AtmosDycore

```julia
time(runner)
settime!(runner, time)
```

instead of

```julia
runner[:time]
runner[:time] = time
```

jkozdon:
The Julian style is that really `getindex` and `setindex!` should be for collection type accesses, not necessarily to access members of a struct. I was trying to hide the implementation of things like `time` from the user, so did not want to assume that `runner.time` was valid. Another option would be to use `getproperty`

jkozdon:
fixed by #54

PR #54, opened 2019-02-04
[AtmosDycore] Remove `getindex` and `setindex!`

jkozdon:
In the AtmosDycore I had originally set up the accessor functions to use
`getindex` and `setindex!`, but this is out of line with the julia
coding style, so we are switching to using straight accessors and member
access.


jkozdon:
This closes #47

Issue #41, opened 2019-01-19
MoistThermodynamics performance improvements


tapios:
The MoistThermodynamics module will need to be revised to improve performance:

* Use something like `Cassette.jl` to swap out things like `exp` with CUDA-native versions on GPUs
* Make sure all functions are type generic 
* The function `phase_partitioning_eq!`, which has two output arguments, is wrapped in an explicit loop to handle two arrays on output. Other functions with one output argument can be broadcast. Make more consistent?
* Improve initial conditions for saturation adjustment so that secant method for root finding converges reliably

PR #56, opened 2019-02-05
Roots -> in-house RootSolvers in Utilities

charleskawczynski:
@tapios mentioned that the `Roots.jl` package seemed to have some bugs. I converted the secant method solver in SCAMPy to Julia and included it in `Utilities`. This also has the benefit that it returns a Bool indicating convergence, which may be useful.

### DIFF
--- a/src/Utilities/src/MoistThermodynamics.jl
+++ b/src/Utilities/src/MoistThermodynamics.jl
@@ -12,9 +12,6 @@ using PlanetParameters
 
 using ..RootSolvers
 
-# Uses Roots.jl from JuliaMath in the saturation adjustment function
-using Roots
-
 # Atmospheric equation of state
 export air_pressure, air_temperature, air_density
 
@@ -418,12 +415,12 @@ function saturation_adjustment(E_int, ρ, q_t, T_init = T_triple)
     args = ()
     T_0, T_1 = T_init, T_init*1.01
     roots_equation(x) = internal_energy_sat(x, ρ, q_t) - E_int
-    T, converged = RootSolvers.find_zero(roots_equation,
-                                         InitialGuessSecant(T_0, T_1),
-                                         args,
-                                         IterParams(tol_abs, iter_max),
-                                         SecantMethod()
-                                         )
+    T, converged = find_zero(roots_equation,
+                             T_0, T_1,
+                             args,
+                             IterParams(tol_abs, iter_max),
+                             SecantMethod()
+                             )
     return T
 
 end
--- a/src/Utilities/src/MoistThermodynamics.jl
+++ b/src/Utilities/src/MoistThermodynamics.jl
@@ -5,7 +5,6 @@ Moist thermodynamic functions, e.g., for air pressure (atmosphere equation
 of state), latent heats of phase transitions, saturation vapor pressures, and
 saturation specific humidities
 """
-
 module MoistThermodynamics
 
 using PlanetParameters
@@ -107,9 +106,9 @@ function cv_m(q_t=0, q_l=0, q_i=0)
 end
 
 """
-    air_temperature(E_int[, q_t=0, q_l=0, q_i=0])
+    air_temperature(e_int[, q_t=0, q_l=0, q_i=0])
 
-Return the air temperature, given the internal energy `E_int` per unit mass,
+Return the air temperature, given the internal energy `e_int` per unit mass,
 and, optionally, the total specific humidity `q_t`, the liquid specific humidity
 `q_l`, and the ice specific humidity `q_i`.
 """
@@ -335,7 +334,7 @@ temperature `T`, and the saturation vapor pressure `p_vs`.
 """
 function sat_shum_from_pressure(ρ, T, p_vs)
 
-    return p_vs / (ρ * R_v * T);
+    return min(eltype(ρ)(1.), p_vs / (ρ * R_v * T))
 
 end
 
@@ -401,22 +400,23 @@ function phase_partitioning_eq(T, ρ, q_t)
 end
 
 """
-    saturation_adjustment(E_int, ρ, q_t[, T_init = T_triple])
+    saturation_adjustment(e_int, ρ, q_t[, T_init = T_triple])
 
-Return the temperature that is consistent with the internal energy `E_int`,
+Return the temperature that is consistent with the internal energy `e_int`,
 (moist-)air density `ρ`, and total specific humidity `q_t`.
 
 The optional input value of the temperature `T_init` is taken as the initial
 value of the saturation adjustment iterations.
 """
-function saturation_adjustment(E_int, ρ, q_t, T_init = T_triple)
+function saturation_adjustment(e_int, ρ, q_t, T_init = T_triple)
     tol_abs = 1e-3*cv_d
-    iter_max = 100
-    args = ()
-    T_0, T_1 = T_init, T_init*1.01
-    roots_equation(x) = internal_energy_sat(x, ρ, q_t) - E_int
+    iter_max = 10
+    args = (ρ, q_t, e_int)
+    T0 = max(T_min, air_temperature(e_int, q_t, 0.0, 0.0))
+    T1 = air_temperature(e_int, q_t, 0.0, q_t)
+    roots_equation(x, ρ, q_t, e_int) = internal_energy_sat(x, ρ, q_t) - e_int
     T, converged = find_zero(roots_equation,
-                             T_0, T_1,
+                             T0, T1,
                              args,
                              IterParams(tol_abs, iter_max),
                              SecantMethod()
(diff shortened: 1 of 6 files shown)


src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -411,10 +413,17 @@ The optional input value of the temperature `T_init` is taken as the initial
...
 """
 function saturation_adjustment(E_int, ρ, q_t, T_init = T_triple)
-
-    T = find_zero(x -> internal_energy_sat(x, ρ, q_t) - E_int, T_init,
-            Order1(), atol=1e-3*cv_d)
-
+    tol_abs = 1e-3*cv_d
+    iter_max = 100
+    args = ()
+    T_0, T_1 = T_init, T_init*1.01
```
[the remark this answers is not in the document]
charleskawczynski: I'm working on this now. I think that I've got an idea of how to address this and #41 in a pretty simple and encapsulated way.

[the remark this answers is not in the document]
charleskawczynski: It turns out that allowing the function to operate on single `AbstractFloat`'s and broadcast over `Array`'s, or `CuArray`'s, is difficult because of the type system hierarchy (a figure of the CPU types are here: https://en.wikibooks.org/wiki/Introducing_Julia/Types#/), and the CuArray type structure is built from the top down:

```
julia> supertype(CuArray)
GPUArrays.GPUArray
julia> supertype(supertype(CuArray))
DenseArray
julia> supertype(supertype(supertype(CuArray)))
AbstractArray
```

charleskawczynski:
I think that we're in good shape now. I added a GPU test case in `runtests.jl`, and it has passed locally on my laptop, but we should probably move it once we get GPU CI up and running (sooner the better).

charleskawczynski:
Thanks to @vchuravy, I was able to add a closure implementation of find_zero (which does not take `args`). Again, I tested this locally. The closures implementation is supported on GPUs so long the closure is formed with "simple" data (e.g, defined with `const`). The `args` implementation can handle more complex cases.

I'm fine with closing this unless there are any concerns.

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -411,10 +413,17 @@ The optional input value of the temperature `T_init` is taken as the initial
 value of the saturation adjustment iterations.
 """
 function saturation_adjustment(E_int, ρ, q_t, T_init = T_triple)
-
-    T = find_zero(x -> internal_energy_sat(x, ρ, q_t) - E_int, T_init,
-            Order1(), atol=1e-3*cv_d)
-
+    tol_abs = 1e-3*cv_d
+    iter_max = 100
```
tapios: We should limit the number of iterations to < 10 or so.

src/Utilities/src/MoistThermodynamics.jl
tapios: air_temperature(E_int, q_t, 0, 0) and air_temperature(E_int, q_t, 0, q_t) should bracket the zero and would make good initial values. (Please check.)

src/Utilities/src/MoistThermodynamics.jl
charleskawczynski: Ok, done (set `iter_max = 10`).

src/Utilities/src/MoistThermodynamics.jl
charleskawczynski: I tried
`T_0 = air_temperature(E_int, q_t, 0.0, 0.0)`
`T_1 = air_temperature(E_int, q_t, 0.0, q_t)`
and I'm getting a domain error in the saturation adjustment. I tried printing these temperatures, and I'm getting 
`T_0 = -473.59...`
`T_1 = 200.04...`
It looks like there is something wrong with `air_temperature`.

charleskawczynski: I realize now, that we're not using `T_init = T_triple`. If I replace `T_0` with `T_init`, the tests pass, but I don't know that IC will sufficiently bound the equilibrium solution in general.

tapios: air_temperature should be ok. You can just get negative values if E_int is small and q_t is relatively large. For the initial values for saturation adjustment, you can just replace the T_0 call by `T_0 = max(0, air_temperature(E_int, q_t, 0.0, 0.0))`, to prevent T_0 from going below zero.

charleskawczynski: Hmm, getting `NaN`'s with `T_0 = max(0.0, air_temperature(E_int, q_t, 0.0, 0.0))`. I'm going to look at the equations more closely over the weekend and see if I can figure this out.

src/Utilities/test/runtests.jl
```diff
@@ -85,3 +86,74 @@ using LinearAlgebra
...
+catch
+  false
+end
+@testset "CUDA RootSolvers" begin
+  if HAVE_CUDA
+    x_ca = cu(rand(5, 5))
+    x_ca_0 = x_ca
+    x_ca_1 = x_ca.+2
+    t = typeof(x_ca[1])
+    x_star2 = t(10000.0)
```
vchuravy: The important thing for the GPU compiler is that:

1. `x_star2` is constant for the scope of the closure, so you can't modify it outside the closure.
2. That it is an immutable & pointer free struct, e.g. doesn't contain a reference to data residing on the GPU, sadly this includes `CuArray`.
3. If it is a global variable, the variable needs to be marked constant

Sometimes it is helpful to use a let block to ensure 1. and 3.

```
let x_star2=x_star2
   f_close(x) = x^2 - x_star2
end
```

tapios:
You might want to try replacing secant method by regula falsi to guarantee convergence, if that is the issue.

On Feb 8, 2019, at 5:53 PM, Charles Kawczynski <notifications@github.com<mailto:notifications@github.com>> wrote:


@charleskawczynski commented on this pull request.

________________________________

In src/Utilities/src/MoistThermodynamics.jl<https://github.com/climate-machine/CLIMA/pull/56#discussion_r255282169>:

> @@ -411,10 +413,17 @@ The optional input value of the temperature `T_init` is taken as the initial
 value of the saturation adjustment iterations.
 """
 function saturation_adjustment(E_int, ρ, q_t, T_init = T_triple)
-
-    T = find_zero(x -> internal_energy_sat(x, ρ, q_t) - E_int, T_init,
-            Order1(), atol=1e-3*cv_d)
-
+    tol_abs = 1e-3*cv_d
+    iter_max = 100
+    args = ()
+    T_0, T_1 = T_init, T_init*1.01


Hmm, getting NaN's with T_0 = max(0.0, air_temperature(E_int, q_t, 0.0, 0.0)). I'm going to look at the equations more closely over the weekend and see if I can figure this out.

src/Utilities/src/RootSolvers.jl
```diff
@@ -0,0 +1,115 @@
...
+    y_1 = f(x_1, args...)
+    Δx = x_1 - x_0
+    if abs(y_1) < tol_abs
+      return x, true
+    end
+  end
+  return x, false
+end
+
+function find_zero(f::F,
```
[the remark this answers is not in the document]
charleskawczynski: Yeah, I realized this last night, the `args` method is more general. I'll remove this method.

charleskawczynski:
I looked into it and secant method is fine, the problem was a `1/T` in the roots equation (for computing q_v^*). If it's okay, I'll just use `T0 = max(1.0, air_temperature(E_int, q_t, 0.0, 0.0))`.

src/Utilities/src/MoistThermodynamics.jl
tapios: Please replace l. 338 in MoistThermodynamics by `return min(eltype(ρ)(1.), p_vs / (ρ * R_v * T));`. Some of the tests had extreme (very unrealistic) values for q_t and T, resulting in specific humidities > 1. This change will bound saturation specific humidities by 1.

tapios:
Better to define a temperature T_min = 150. in PlanetParameters and use that as a lower bound.

Regula falsi might still be a good idea, as it guarantees convergence and is only slightly slower.

On Feb 9, 2019, at 8:22 PM, Charles Kawczynski <notifications@github.com<mailto:notifications@github.com>> wrote:


I looked into it and secant method is fine, the problem was a 1/T in the roots equation (for computing q_v^*). If it's okay, I'll just use T0 = max(1.0, air_temperature(E_int, q_t, 0.0, 0.0)).

tapios:
One other request: Could you please replace E_int in MoistThermodynamics by e_int? I'd like to use lowercase e for specific energies (per unit mass), to distinguish them from the energies per unit volume used in the dynamical equations. I'll update the documentation accordingly.

charleskawczynski:
Alright, I've updated everything we've discussed. I've implemented Regula Falsi method, but its convergence seems a bit slow. So, for now, I've set the saturation adjustment to use Secant method, but we can always change this.

Also, are we okay with leaving `iter_max` defined for the saturation adjustment in MoistThermodynamics?

PR #60, opened 2019-02-09
[AtmosDycore] Bug fix for viscosity on CPU

jkozdon:
In the CPU kernels the viscosity was being set on the tracers and not
the moisture variables

@lcw please confirm this is the correct fix


lcw:
Looks correct to me.  Thanks!

PR #64, opened 2019-02-12
Added SurfaceFluxes package

charleskawczynski:
Added a package for the surface fluxes.

This includes functions that compute friction velocity, buoyancy flux and exchange coefficients at the surface of the atmosphere using Monin-Obhukov similarity theory.

A few parameters were added to PlanetParameters, which this package depends on.

There are a few things I wasn't 100% sure about:

 - The value of the Turbulent Prandtl number (added to PlanetParameters), I saw different values in different references

PR #67, opened 2019-02-13
Add docs

charleskawczynski:
Added a skeleton for the docs.

Right now, there's only a index.md, but I'm hoping that we can start getting used to referencing functions and filling out the documentation as we're developing.


jkozdon:
This looks good to me. Only suggestion would be to rebase the commits on top of master before accepting it to clean up the git history.

jkozdon:
How will this all work with the subpackages? Does each subpackage have its own docs directory or we just have one docs directory and stuff gets added in there?

charleskawczynski:
> How will this all work with the subpackages? Does each subpackage have its own docs directory or we just have one docs directory and stuff gets added in there?

I don't remember reading anything about deployment of subpackages in the Documenter docs. So, I suspect that there may only be one docs directory. I just asked in the Jula slack Documentation channel.

charleskawczynski:
> Only suggestion would be to rebase the commits on top of master before accepting it to clean up the git history.

I'll do the same with #64.

PR #73, opened 2019-02-16
[AtmosDycore] Refactor to use the type system more

### LINKED ISSUE #46: Refactor AtmosDycore Runner naming
Consider refactoring the way the AtmosDycore runners are set up to make it more inline with Julia conventions. For now likely a thin refactoring with possible future major refactoring.

Move from
```julia
runner = Runner(:SpaceStepper, (space_params...), :TimeStepper, (time_params..))
```
to
```julia
space_runnner = SpaceRunner(space_params...)
time_runner = TimeRunner(space_runner, time_params...)
step!/run!/advance!/whatever!(time_runner, time_end, callbacks) #or something like this
```

### LINKED ISSUE #70: Encapsulate the Dycore mesh
Right now, it seems that the dycore mesh consists of many tuples of arrays that define relations between element neighbors, vertices etc.

I think it would be good to make a struct `Mesh` and make these mesh-related arrays properties of `Mesh` so that functions calls can have simpler interfaces. For example, in [L2errornorm](https://github.com/climate-machine/CLIMA/blob/master/src/CLIMA-atmos/Dycore/src/VanillaEuler.jl#L587)

`L2errornorm(::Val{dim}, ::Val{N}, time, Q, vgeo, elems, Qexact)`

would become

`L2errornorm(::Val{dim}, ::Val{N}, time, Q, mesh::Mesh, Qexact)`

This will also greatly simplify the mesh initialization, which should be something of the form:

`mesh = Mesh(params)`

jkozdon:
The AtmosDycore has been refactored into (hopefully) more reusable pieces. The best place to start is looking at the `src/CLIMA-atmos/Dycore/drivers` and see how the new types impact the driver.

The main building block types are now a series of abstract types.

- `AbstractTopology`: Grid type; see `Topologies.jl`
- `AbstractGrid`: Grid type; see `Grids.jl`
- `AbstractAtmosDiscretization`: Spatial discretization Type; see `CLIMAAtmosDycore.jl` and `VanillaAtmosDiscretizations.jl`
- `AbstractAtmosODESolver`: Time discretization Type; see `CLIMAAtmosDycore.jl` and `LSRKMethods.jl`
- `AtmosStateArray`: MPI/FEM aware array type; see 'AtmosStateArrays.jl`

I fully expect some smoothing will happen. And I might have gone a bit overboard in some of the type hierarchy stuff to make sure we get good compiling and automatic dispatch.

Once we have a second version of everything, we should revisit some of the pieces. I have tried to define things in a way that after they are cleaned up and smoothed pieces (such as `Grids`, `Topology`, `AtmosStateArray`) maybe useful more broadly in the CLIMA project.

This closes #46 and closes #70


jkozdon:
pull request #72 and this one will likely clash

charleskawczynski:
> pull request #72 and this one will likely clash. So one will need to be updated after the other.

I figured that might happened (that's why that was the only change). This PR is more important, so merge if you need to and I'll update #72 later.

src/CLIMA-atmos/Dycore/drivers/isentropic_vortex.jl
```diff
@@ -20,12 +22,12 @@ macro hascuda(ex)
   return HAVE_CUDA ? :($(esc(ex))) : :(nothing)
 end
 
-const halfperiod = 5
+using ParametersType
+using PlanetParameters: R_d, cp_d, grav, cv_d
+@parameter gamma_d cp_d/cv_d "Heat capcity ratio of dry air"
+@parameter gdm1 R_d/cv_d "(equivalent to gamma_d-1)"
```
charleskawczynski: These additional parameters could probably be moved to PlanetParameters.jl.

src/CLIMA-atmos/Dycore/drivers/rising_thermal_bubble.jl
```diff
@@ -27,134 +22,159 @@ macro hascuda(ex)
...
 end
 
-meshgenerator(part, numparts, Ne, dim, DFloat) =
-brickmesh(ntuple(j->range(DFloat(0); length=Ne+1, stop=1000), dim),
-          (true, ntuple(j->false, dim-1)...),
-          part=part, numparts=numparts)
+using ParametersType
+using PlanetParameters: R_d, cp_d, grav, cv_d
+@parameter gamma_d cp_d/cv_d "Heat capcity ratio of dry air"
+@parameter gdm1 R_d/cv_d "(equivalent to gamma_d-1)"
```
charleskawczynski: > @parameter gamma_d cp_d/cv_d "Heat capcity ratio of dry air"
@parameter gdm1 R_d/cv_d "(equivalent to gamma_d-1)"

Again, these parameters can be moved to PlanetParamers (or into a hyper-parameter list if that is more appropriate).

charleskawczynski: I'll skip commenting on the remaining ones, but there are several terms like this across many files it seems.

jkozdon: They could be added to `PlanetParameters` if you wanted. (Maybe in a separate pull request?)

Since they are derived parameters, I could go either way. We mainly packed them into `@parameter` types in order to get compile time type conversions.

jkozdon: Thanks for looking through the code.

PR #77, opened 2019-02-21
Simplify test harness and add some AtmosDycore tests

lcw:
This change enables the uses of the Pkg infrastructure to run tests.  While doing this I cleaned up CLIMA's dependencies and added some tests to AtmosDycore.


lcw:
Unfortunately this doesn't seem to play nice with codecov.

lcw:
Codecov now works

lcw:
Cool, should we go ahead and merge this or will it conflict with your branch?

Issue #48, opened 2019-02-02
Stacked elements


jkozdon:
Need to implement stacking of elements for 1-D solves and physics.

charleskawczynski:
I sent Simone a temporary solution for this earlier today,
```
function get_x_coordinate(mesh::T) where T
  (elemtovert, elemtocoord, elemtobndy, faceconnections) = mesh
  return unique(hcat(elemtocoord[1,1,:],elemtocoord[1,6,:]))
end
function get_y_coordinate(mesh::T) where T
  (elemtovert, elemtocoord, elemtobndy, faceconnections) = mesh
  return unique(hcat(elemtocoord[2,1,:],elemtocoord[2,3,:]))
end
function get_z_coordinate(mesh::T) where T
  (elemtovert, elemtocoord, elemtobndy, faceconnections) = mesh
  return unique(hcat(elemtocoord[3,1,:],elemtocoord[3,6,:]))
end
```
but we'll need an efficient implementation of these routines.

jkozdon:
Thanks. Stacked meshes is on the TODO list for the near future which should alleviate some of this.

jkozdon:
The brick version of this was handled in #85.

PR #85, opened 2019-02-27
Stacked Brick Topology

lcw:
This series of commits adds a `StackedBrickTopology` type that is a drop in replacement for `BrickTopology`.  The stacked topology has elements ordered such that elements in a stack are contiguous.


jkozdon:
Closes #48

Issue #90, opened 2019-03-07
CLIMA nightly travis build fails with non-registered packages (e.g., Canary)


charleskawczynski:
I'm often experiencing issues with CLIMA depending on Canary. For example, right now [Travis](https://travis-ci.org/climate-machine/CLIMA/jobs/503222500#L623) is complaining that

```
ERROR: Build path for Canary does not exist: /home/travis/.julia/packages/Canary/L6RGe
```
for the nightly build, but is passing on Julia 1.0.

As far as I can see, none of my commits touched Canary and I never removed Canary from the Clima root Manifest. Any help from @lcw or @jkozdon is greatly appreciated!

lcw:
I am not sure what is going on.  Are your issues only while running on Travis?

charleskawczynski:
Well, I was able to `julia --project; ] instantiate; build` fine locally from Clima root in this branch, so I guess so?

I don't think that this issue is pressing. But I'd like to merge this (maybe after squashing some commits) so long we're okay with not passing the nightly builds.

jkozdon:
I think that this is an issue with the currently nightly build not Canary.

The one unique thing about the Canary dependency is that it is our one non-registered package.

I cannot find an issue on the julialang github page, so if someone has time it could be helpful to spin up a MWE and submit the issue.

I can recreate the problem locally using the currently nightly on my mac <https://julialang.org/downloads/nightlies.html>:

```
jekozdon@chops:~/codes/CLIMA2 ± master
❯ rm -rf ~/.julia

jekozdon@chops:~/codes/CLIMA2 ± master
❯ export JULIA_PROJECT=@.

jekozdon@chops:~/codes/CLIMA2 ± master
❯ export JL_PKG=CLIMA

jekozdon@chops:~/codes/CLIMA2 ± master
❯ /Applications/Julia-1.2.app/Contents/Resources/julia/bin/julia --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\"); else using Pkg; if VERSION >= v\"1.1.0-rc1\"; Pkg.build(verbose=true); else Pkg.build(); end; end"
┌ Error: Error during initialization of module CHOLMOD
│   exception =
│    could not load library "libcholmod"
[...]

PR #93, opened 2019-03-08
Temporary fix for #90 travis build issue with nightly

jkozdon:
This is a tempory fix for \#90 related to currently nightly and Canary dependency. Now we create the "Build path" for Canary before testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,11 @@ before_script:
       brew cask uninstall oclint # Prevent conflict with gcc
       brew install mpich
     fi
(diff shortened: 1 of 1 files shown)


jkozdon:
Alternatively, since we allow for failures on nightly, we could just scrap this.... thoughts?

charleskawczynski:
I think that it would be nice to see if Clima is passing nightly builds so that we can spot issues earlier than later.

Maybe we can add another issue, as a reminder, to remove this fix when Clima becomes Canary-independent.

PR #96, opened 2019-03-12
[AtmosDycore] Add instruction for runs to README

jkozdon:
Adding some example instructions for running the AtmosDycore


jkozdon:
Not sure what I was thinking.... Let's pretend this never happened!

PR #97, opened 2019-03-13
Adding StackedCubedSphereTopology type

jkozdon:
This pull requests adds a `StackedCubedSphereTopology` type.

### DIFF
--- a/src/ClimaAtmos/Dycore/src/vtk.jl
+++ b/src/ClimaAtmos/Dycore/src/vtk.jl
@@ -26,7 +26,7 @@ end
 #=
 This is the 2D WriteMesh routine
 =#
-function writemesh(base_name, x, y; fields=(), realelems=1:size(x)[end])
+function writemesh(base_name, x, y; z=nothing, fields=(), realelems=1:size(x)[end])
   @assert size(x) == size(y)
   (Nqr, Nqs, _) = size(x)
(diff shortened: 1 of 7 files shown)


src/ClimaAtmos/Dycore/src/vtk.jl
```diff
@@ -26,7 +26,7 @@ end
 #=
 This is the 2D WriteMesh routine
 =#
-function writemesh(base_name, x, y; fields=(), realelems=1:size(x)[end])
+function writemesh(base_name, x, y; z=nothing, fields=(), realelems=1:size(x)[end])
   @assert size(x) == size(y)
   (Nqr, Nqs, ~) = size(x)
```
simonbyrne: should the `~` here be a `_`?

lcw: I think so.  I am guessing this is a MATLABism.  I submitted a separate pull request, https://github.com/climate-machine/CLIMA/pull/103, that makes the change you suggested.

simonbyrne: Somewhat annoyingly, it doesn't appear to be documented, but the idea is that `_` can only appear as a variable on the left-hand side of an assignment.

jkozdon: fixed in #103

PR #100, opened 2019-03-13
Remove sub-level manifests

simonbyrne:
This should help simplify some of the resolver issues (as well as reduce noise in the git diffs).


lcw:
@jkozdon is traveling but I am happy to check this out.  After I added MPI to the Project.toml

```
diff --git a/Project.toml b/Project.toml
index 4274ce0..9da5beb 100644
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ Canary = "f104530e-ab06-11e8-1c1e-6d6e4a5d5c91"
 ParametersType = "046d00ba-03b1-11e9-3cca-b5213a01defd"
 PlanetParameters = "50a4fe1a-031f-11e9-2493-0505e0da639e"
 Utilities = "d528e2a8-1396-11e9-0475-bb99f588b544"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"

 [extras]
```

I could run the CPU version of the Dycore with 

```
julia --project=. -e "using Pkg; Pkg.instantiate(); Pkg.API.precompile()"
mpirun -np 2 julia --project  src/ClimaAtmos/Dycore/drivers/isentropic_vortex.jl
```

How do we now launch the GPU version of the Dycore?  To see how we used to do it look [here](https://github.com/climate-machine/CLIMA/blob/a69cbf5c25047ae4054395b7b8515b37ac1e5482/src/ClimaAtmos/Dycore/README.md).

simonbyrne:
Ah, I didn't realise that those Manifests were external endpoints (I'm still finding my way around the codebase).

What is the code in `drivers`? Are these examples, test cases, or code that is going to be folded into a package eventually?

lcw:
Good question.  In my mind, the code in `drivers` are examples right now but we would like to be able to run these (or a modified version) as test cases as well.

We have been discussing with @fxgiraldo how to do integration testing for the dynamical core.  One of the examples we have an analytical solution for but the others we don't.  So do we just make sure the L^2 norm of the solution at a given time doesn't change after a series of commits?

simonbyrne:
Yeah, integration tests for these are hard. If we apply the ["make it work, make it right, make it fast"](http://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) maxim to testing, we should:

1. Test that they work: we could just hook up these examples to the `runtests.jl` (and check that they don't cause errors).
2. Test that they're correct: testing analytical cases, or other known references. We can also check known properties (e.g. conservation of energy), or compare results at different timesteps/scales
3. Test for performance regressions: this is tougher. There are some Julia tools for this (BenchmarkTools.jl) but we will need to think carefully how best to do this.

lcw:
For what it is worth, the current use of the package system was roughly modeled after [ShallowWaterBench](https://github.com/JuliaLabs/ShallowWaterBench).

jkozdon:
I'm really in a place to really run stuff to make sure that things didn't break, but if @lcw is good with it, I am too.

Having a uniform set of dependencies seems like a good thing in principle to me (as shown by the fix I just had to add to #97 with a27656a)

simonbyrne:
How about I keep the `src/ClimaAtmos/Dycore/env/gpu/Manifest.toml` for now, but clean up the rest?

lcw:
Sounds okay to me.

simonbyrne:
Okay, updated.

simonbyrne:
Actually, what if we moved `env/gpu` to the top-level?

lcw:
So will all dependencies that we need for all packages and examples then go into Project.toml?

simonbyrne:
Hmm, let me think about this more: will probably be better as a separate PR.

simonbyrne:
Closed in favour of #107

PR #102, opened 2019-03-13
Added "Phase" type to dispatch sat_vapor_press

### LINKED ISSUE #98: Change string eval to type dispatch
I think we should change https://github.com/climate-machine/CLIMA/blob/master/src/Utilities/src/MoistThermodynamics.jl#L280-L287

to apply type dispatch instead of evaluating a string.

charleskawczynski:
Fixes #98

### DIFF
--- a/src/Utilities/src/MoistThermodynamics.jl
+++ b/src/Utilities/src/MoistThermodynamics.jl
@@ -25,7 +25,7 @@ export latent_heat_vapor, latent_heat_sublim, latent_heat_fusion
 
 # Saturation vapor pressures and specific humidities over liquid and ice
 export Liquid, Ice
-export saturation_vapor_pressure, sat_shum_generic, sat_shum
+export saturation_vapor_pressure, saturation_shum_generic, saturation_shum
 
 # Functions used in thermodynamic equilibrium among phases (liquid and ice
 # determined diagnostically from total water specific humidity)
@@ -262,7 +262,7 @@ function saturation_vapor_pressure(T, LH_0, cp_diff)
 end
 
 """
-    sat_shum_generic(T, ρ[; phase=Liquid()])
+    saturation_shum_generic(T, ρ[; phase=Liquid()])
 
 Compute the saturation specific humidity over a plane surface of
 condensate, given the temperature `T` and the (moist-)air density `ρ`.
@@ -270,16 +270,16 @@ condensate, given the temperature `T` and the (moist-)air density `ρ`.
 The optional argument `phase` can be ``Liquid()`` or ``"ice"`` and indicates
 the condensed phase.
 """
-function sat_shum_generic(T, ρ; phase::Phase=Liquid())
+function saturation_shum_generic(T, ρ; phase::Phase=Liquid())
 
     p_vs = saturation_vapor_pressure(T, phase)
 
-    return sat_shum_from_pressure(ρ, T, p_vs)
+    return saturation_shum_from_pressure(ρ, T, p_vs)
 
 end
 
 """
-    sat_shum(T, ρ[, q_l=0, q_i=0])
+    saturation_shum(T, ρ[, q_l=0, q_i=0])
 
 Compute the saturation specific humidity, given the temperature `T` and
 (moist-)air density `ρ`.
@@ -299,7 +299,7 @@ zero, the saturation specific humidity is that over a mixture of liquid and ice,
 with the fraction of liquid given by temperature dependent `liquid_fraction(T)`
 and the fraction of ice by the complement `1 - liquid_fraction(T)`.
 """
-function sat_shum(T, ρ, q_l=0, q_i=0)
+function saturation_shum(T, ρ, q_l=0, q_i=0)
 
     # get phase partitioning
     _liquid_frac = liquid_fraction(T, q_l, q_i)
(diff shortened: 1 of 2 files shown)


simonbyrne:
Is this only ever going to be two values? If so, would it be simpler to just have an `isliquid` argument and do a branch?

charleskawczynski:
> Is this only ever going to be two values? If so, would it be simpler to just have an `isliquid` argument and do a branch?

There may be 3 phases at most (liquid, ice, vapor), but I need to check what, if any, mixed-phases this will work for.

By `isliquid`, do you mean an if statement?

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -216,32 +217,37 @@ function latent_heat_generic(T, LH_0, cp_diff)
...
 end
 
+abstract type Phase end
+struct Liquid <: Phase end
+struct Ice <: Phase end
+struct Generic <: Phase end
+
 """
-    sat_vapor_press_liquid(T)
+    sat_vapor_press(T, Liquid())
```
simonbyrne: - a more verbose name might be better (`saturation_vapor_pressure` is easier to understand).
- using `temp` instead of `T` (Julia convention has it that `T` is usually a type parameter)
- these methods can all share the one docstring

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -216,32 +217,37 @@ function latent_heat_generic(T, LH_0, cp_diff)
...
 
 Return the saturation vapor pressure over a plane liquid surface at
 temperature `T`.
 """
-function sat_vapor_press_liquid(T)
+function sat_vapor_press(T, phase::Liquid)
 
-    return sat_vapor_press_generic(T, LH_v0, cp_v - cp_l)
+    return sat_vapor_press(T, LH_v0, cp_v - cp_l, Generic())
```
simonbyrne: replace these one-liners with inline form, and you don't need an argument name (unless you're passing it to another function:
```
saturation_vapor_pressure(temp, ::Liquid) = saturation_vapor_pressure(temp, LH_v0, cp_v - cp_l)
```

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -261,7 +267,7 @@ on temperature `T` allows analytic integration of the Clausius-Clapeyron
 relation to obtain the saturation vapor pressure `p_vs` as a function of
 the triple point pressure `press_triple`.
 """
-function sat_vapor_press_generic(T, LH_0, cp_diff)
+function sat_vapor_press(T, LH_0, cp_diff, phase::Generic)
```
simonbyrne: rather than use a `Generic` singleton, we can just make this the 3-argument form of the function.

src/Utilities/src/MoistThermodynamics.jl
charleskawczynski: - Agree with `saturation_vapor_pressure`
- So, temperature shows up a lot and I think that, while `T` is used in Julia for types, it may be more appropriate here to use `T` (or `temperature`) for temperature. My only hesitation on `temp` is confusion with temporary variables (which I usually like using `temp` for). I'd like to hear @tapios weigh in on this.
- Agree with shared docstring

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -24,7 +24,8 @@ export cp_m, cv_m, gas_constant_air
 export latent_heat_vapor, latent_heat_sublim, latent_heat_fusion
 
 # Saturation vapor pressures and specific humidities over liquid and ice
-export sat_vapor_press_liquid, sat_vapor_press_ice, sat_shum_generic, sat_shum
+export Liquid, Ice, Generic
```
simonbyrne: remove `Generic`

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -216,32 +217,20 @@ function latent_heat_generic(T, LH_0, cp_diff)
...
 end
 
-"""
-    sat_vapor_press_liquid(T)
+abstract type Phase end
+struct Liquid <: Phase end
+struct Ice <: Phase end
 
+"""
+    `sat_vapor_pressure(T, Liquid())`
```
simonbyrne: were you going to rename the function?

charleskawczynski: Yes, sorry, I misread and only extended `pres` to `pressure`. Updating now.

tapios:
> Is this only ever going to be two values? If so, would it be simpler to just have an `isliquid` argument and do a branch?

It will only have two values. If Boolean, I’d prefer isice over isliquid (because liquid is default). But I think it is nice to have verbose clarity of liquid vs ice.

src/Utilities/src/MoistThermodynamics.jl
tapios: * Updating to ``saturation_vapor_pressure`` is good
* Prefer to avoid ``temp`` for ambiguity @charleskawczynski mentioned
* One may argue a few centuries of history of ``T`` for temperature may outweigh Julia conventions, unless there is real potential for confusion
* If we want to replace ``T``, use, e.g., ``Ta`` for atmospheric temperature, which is close to CMIP convention for this variable ``ta`` (but I prefer uppercase T because it is closer to how we write the equations)

tapios: * If you change ``sat_`` to ``saturation_`` in the vapor pressure functions, I suggest you do the same for the humidity (``shum``) functions

Issue #108, opened 2019-03-15
Fold Atmos TCC into master


charleskawczynski:

charleskawczynski:
This was moved to SingleColumnModels.jl

PR #109, opened 2019-03-15
WIP: Fold turb conv

charleskawczynski:
For #108 

This PR is to include the infrastructure to solve the EDMF equations. It also includes a simple `Grids` module which we can use for finite-difference method to compare the Julia version of solving the EDMF equations to compare with SCAMPy. Once we're comfortable with the comparison, we can convert to DG method.


charleskawczynski:
Going to break this into smaller PRs

PR #107, opened 2019-03-15
change to submodules instead of subpackages

simonbyrne:
Fixes #106.

The main changes turned out to be:
- adding/removing dependencies is now done in the main Project.toml file
- submodules need to be manually `include`ed from `src/CLIMA.jl`
- instead of `using SubModule` you now do:
  - for package code: `using ..SubModule` (potentially adding more `.`s at the end if you are in nested submodules)
  - for tests: `using CLIMA.SubModule`

Overall it seems a bit easier to maintain. I'll see if I can figure out a neat way to handle the GPU environment.

PR #112, opened 2019-03-18
add gitlab CI

vchuravy:
should wait in #107


vchuravy:
actually we should merge this first so that we can adjust #102 to run GPU integration tests.

simonbyrne:
Ah! too late. I didn't see your comment.

Issue #101, opened 2019-03-13
Verify correctness of Regula Falsi method in RootSolvers


charleskawczynski:
It seems that the Regula Falsi method tests are brittle for the CUDA tests (which operate at lower precision by default). Recording this here in case we are to use this method.

simonbyrne:
The problem seems to be caused by two things:
1) The starting points aren't on either side of the root.
2) This line:
https://github.com/climate-machine/CLIMA/blob/9f93ecad302116870dc58ffaff4bd8e6eac8bd38/src/Utilities/src/RootSolvers.jl#L150
can exhibit catastrophic cancellation when `fb >> fa`, in which case it will simply keep setting `c = a`, without the interval getting smaller.

charleskawczynski:
Closed by #117

PR #117, opened 2019-03-21
Root solver improvements

simonbyrne:
- make interface consistent with other Julia functions, such as
  `isapprox`.
- improve convergence criteria based on input tolerances
- improve Regula Falsi algorithm based on Illinois algorithm
- test that GPU tests are actually running on GPUs


simonbyrne:
I also removed the `args` argument, since that can be handled directly via closures.

src/Utilities/src/RootSolvers.jl
```diff
@@ -2,162 +2,97 @@
...
-    fa = f(a, args...)
-    fb = f(b, args...)
-    c = (a * fb - b * fa)/ (fb - fa)
-    fc = f(c, args...)
-    if abs(fc) < tol_abs
-      return c, true
-    elseif fc * fa < 0
-      b = c
+function find_zero(f::F, x0::T, x1::T, ::RegulaFalsiMethod;
+                   xatol=zero(x0), xrtol=sqrt(eps(T)), yatol=sqrt(eps(T)), maxiters=10_000) where {F, T<:AbstractFloat}
```
charleskawczynski: Should `xatol=zero(x0)` be `xatol=sqrt(eps(T))`?

src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -400,20 +400,11 @@ The optional input value of the temperature `T_init` is taken as the initial
...
-                             T0, T1,
-                             args,
-                             IterParams(tol_abs, iter_max),
-                             SecantMethod()
-                             )
-    return T
-
+  T0 = max(T_min, air_temperature(e_int, q_t, 0.0, 0.0))
+  T1 = air_temperature(e_int, q_t, 0.0, q_t)
+  T, converged = find_zero(x -> internal_energy_sat(x, ρ, q_t) - e_int,
```
charleskawczynski: We should set a lower `maxiter` so that this function call does not hang. I think that Tapio suggested `maxiter = 10`. I almost wonder if this parameter should be passed in as an argument so that we can worry about it upstream.

src/Utilities/src/RootSolvers.jl
simonbyrne: Relative tolerance by default tends to make more sense for this: these are the same defaults for [`isapprox`](https://docs.julialang.org/en/v1/base/math/#Base.isapprox).

src/Utilities/src/MoistThermodynamics.jl
simonbyrne: Probably, I was just trying to keep this consistent.

Do you know if we could use Newton iterations for this?

simonbyrne:
@vchuravy any idea why we're getting a GPU failure on 1.1 but not dev?

The issue seems to be `isapprox`: if I replace that with a simple `abs(x0-x1) < xatol` it works correctly, but not if I do say `abs(x0-x1) < xatol + xrtol*abs(x0)`

simonbyrne:
> Are there any advantages to removing `args` aside from a simpler interface? I guess I'm just wondering if closures will always be the right solution. I'm not super-aware of the overhead for closures (e.g., on GPUs).

In general, it should be more or less the same as passing it as a tuple. Occasionally you need to do the `let` trick when calling, but we can find those by looking at the allocations.

src/Utilities/src/MoistThermodynamics.jl
charleskawczynski: I think that we _could_, but I think that one advantage of the Secant method is that we've got a good idea of realistic physical bounds. That said, we can always implement both and see which one works better. I've typically used Newton iteration in the past and found it effective.

tapios: Secant has been more reliable for us, at essentially the same computational cost as Newton.

tapios: We don't want more than ~10 iterations (and can have larger relative tolerances, see below).

charleskawczynski:
Closes #101

simonbyrne:
I'll wait until #147 is sorted.

src/Utilities/RootSolvers/src/RootSolvers.jl
```diff
@@ -2,162 +2,129 @@
...
-  x = x_1 - y_1 * Δx / (y_1 - y_0) # puts x in outer scope, & saves iter_max = 0 case
-  for i in 1:iter_max
-    x = x_1 - y_1 * Δx / (y_1 - y_0)
-    y_0 = y_1
-    x_0 = x_1
-    x_1 = x
-    y_1 = f(x_1, args...)
-    Δx = x_1 - x_0
-    if abs(y_1) < tol_abs
+function find_zero end
```
charleskawczynski: What's this for/do?

src/Atmos/Parameterizations/SurfaceFluxes/src/SurfaceFluxes.jl
```diff
@@ -128,13 +128,10 @@ function compute_friction_velocity(u_ave, flux, z_0, z_1, β_m, γ_m, tol_abs, i
 
     if (abs(flux) > 0)
       ustar_1 = compute_ustar(ustar_0)
-      args = ()
       ustar, converged = RootSolvers.find_zero(compute_ustar_roots,
```
charleskawczynski:
```suggestion
      ustar, converged = find_zero(compute_ustar_roots,
```

src/Utilities/RootSolvers/src/RootSolvers.jl
simonbyrne: basically it defines a function without any methods: it's handy for adding docs where you want the same docs to be shared by different methods.

simonbyrne:
The problem is now that my fixes to the Regula Falsi method which make it converge actually break it on GPU. Since we're not using it I've just disabled those tests.

Issue #119, opened 2019-03-22
Need more MoistThermo conversion functions


charleskawczynski:
To get Bomex running, we need several more MoistThermodynamics conversion functions:

 - A saturation adjustment for liquid potential temperature
 - All of the functions needed for the liquid potential temperature saturation adjustment
 - A function to compute density potential temperature

tapios:
Why saturation adjustment for liquid potential temperature? I would just convert the initial condition given in liquid potential temperature (and any relaxation terms involving liquid potential energy) to internal energy and work with that.

charleskawczynski:
In Bomex, and several other cases, `q_t` and `\theta_l` are given as initial values. So, if I understand correctly, I need to compute temperature, given `q_t` and `\theta_l`, in order to convert and solve in terms of internal energy.

I'm happy to use a more direct relation between liquid potential temperature and internal energy if there is one.

tapios:
But the initial condition is unsaturated, right? So you can straightforwardly extract temperature (without iteration) and get internal energy from it.

trontrytel:
We won't be able to hide from it for long. DYCOMS case starts saturated.

tapios:
But DYCOMS starts well mixed in theta_l below inversion and is unsaturated above inversion. So you can get what you need from getting temperature from the unsaturated surface, then, e.g., assuming well mixed MSE to get e_int. No need to add code just for that.

charleskawczynski:
Ok, I think that only some functions are needed-- many functions from the current formulation can be used, but the `theta_l-q_t`-based saturation adjustment is nice to have so that we can get temperature. From there we can get all of the other variables.

charleskawczynski:
Closed by #147

PR #120, opened 2019-03-22
[MoistThermo] + exner & dry_pottemp, + exports

charleskawczynski:
This is to partially address #119

### DIFF
--- a/src/Utilities/src/MoistThermodynamics.jl
+++ b/src/Utilities/src/MoistThermodynamics.jl
@@ -425,8 +425,12 @@ and ice specific humidity `q_i`.
 """
 function liquid_ice_pottemp(T, p, q_t=0, q_l=0, q_i=0)
 
-    # potential temperature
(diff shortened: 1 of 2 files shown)


src/Utilities/src/MoistThermodynamics.jl
```diff
@@ -422,21 +422,42 @@ end
...
+Return the dry potential temperature, given the temperature `T`,
+pressure `p`, total specific humidity `q_t`, liquid specific humidity `q_l`,
+and ice specific humidity `q_i`.
+"""
+dry_pottemp(T, p, q_t=0, q_l=0, q_i=0) = T / exner(p, q_t, q_l, q_i)
+
+"""
+    exner(p, q_t=0, q_l=0, q_i=0)
+
+Return the exner function, given the pressure `p`, total specific
```
simonbyrne: should Exner here be capitalised? (only in the docs)

PR #137, opened 2019-03-28
Added saturation excess function (for relaxation of q_l and q_i) and cleaned up MoistThermodynamics 

tapios:
The saturation excess function will make it easy to create relaxation sources/sinks for liquid and ice specific humdities, of the form -(q_l - liquid_faction(..) * saturation_excess(..) )/timescale.


src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -403,8 +419,8 @@ function saturation_adjustment(e_int, ρ, q_t, T_init = T_triple)
     tol_abs = 1e-3*cv_d
     iter_max = 10
     args = (ρ, q_t, e_int)
-    T0 = max(T_min, air_temperature(e_int, q_t, 0.0, 0.0))
-    T1 = air_temperature(e_int, q_t, 0.0, q_t)
+    T0 = max(T_min, air_temperature(e_int, q_t, eltype(q_t)(0.), eltype(q_t)(0.)))
```
charleskawczynski:
```suggestion
    T0 = max(T_min, air_temperature(e_int, q_t, typeof(q_t)(0), typeof(q_t)(0)))
```

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -403,8 +419,8 @@ function saturation_adjustment(e_int, ρ, q_t, T_init = T_triple)
...
+    T1 = air_temperature(e_int, q_t, eltype(q_t)(0.), q_t)
```
charleskawczynski:
```suggestion
    T1 = air_temperature(e_int, q_t, typeof(q_t)(0), q_t)
```

tapios:
> I think these changes look great. The only suggestions I have are to change `eltype` (which is for arrays) to `typeof` which is for concrete types. Also, this will promote the type of `0`, so we don't need `0.`.

Thanks, Charlie. Should the other `eltype`'s in MoistThermodynamics than also be changed to `typeof`'s?

charleskawczynski:
Yes, they can be.

charleskawczynski:
Ah, that failure is my fault. I need to change the test case in surface fluxes to not use `rand`. You can change it if you'd like or I can submit a PR to fix that.

PR #139, opened 2019-03-28
[git] Added file extensions to git ignore

charleskawczynski:
Added the following lines/extensions to the git ignore:
src/**/*.vtk
src/**/*.vtu
src/**/*.swp
src/**/*.png

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@ deps/deps.jl
 docs/build/
 docs/site/
 src/**/Manifest.toml
+src/**/*.vtk
+src/**/*.vtu
+src/**/*.swp
+src/**/*.png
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,3 +7,8 @@ deps/deps.jl
 docs/build/
 docs/site/
 src/**/Manifest.toml
+*.vtk
+*.vtu
+*.swp
```
simonbyrne: I'm happy to keep this here, but it really should go in your personal `.gitignore` file. I think the default path is
`~/.config/git/ignore`

cc: @asraero

asraero: Noted - .gitignore locally has been updated

PR #140, opened 2019-03-29
[SurfaceFluxes] move plot tests to docs

simonbyrne:
These plots will now be built as part of the docs


docs/src/Atmos/SurfaceFluxes.md
```diff
@@ -0,0 +1,114 @@
...
+using CLIMA.SurfaceFluxes.Byun1990
+using Plots, LaTeXStrings
+
+Ri_range = range(-1.2, stop=0.24, length=100)
+scales = [50,200,600,1000,10_000]
+
+z_0 = 1.0
+γ_m, γ_h = 15.0, 9.0
+β_m, β_h = 4.8, 7.8
+Pr_0 = 0.74
```
simonbyrne: are these all reasonable values?

docs/src/Atmos/SurfaceFluxes.md
```diff
@@ -0,0 +1,114 @@
...
+Recreation of Figure 4(b) from Byun (1990)
+
+## `Nishizawa2018`
+
+### Plots
+```@example
+using CLIMA.SurfaceFluxes.Nishizawa2018
+using Plots, LaTeXStrings
+
+L, a, θ, z, flux = rand(5,1)
```
simonbyrne: what are reasonable values for these?

docs/src/Atmos/SurfaceFluxes.md
```diff
@@ -0,0 +1,114 @@
...
+z = z/100
+
+u = range(-0.1, stop=0.1, length=100)
+L = Nishizawa2018.compute_MO_len.(u, θ, flux)
+ζ = z./L
+ϕ_m = Nishizawa2018.compute_ϕ_m.(ζ, L, a)
```
simonbyrne: this function isn't defined. I guess it got renamed somewhere?

docs/src/Atmos/SurfaceFluxes.md
```diff
@@ -0,0 +1,114 @@
...
+```
+![](businger1970_fig1.svg)
+
+Recreation of Figure 1 from Businger et al (1970).
+
+```@example
+using CLIMA.SurfaceFluxes.Nishizawa2018
+using Plots, LaTeXStrings
+
+u_ave, a, θ, flux, z_0 = rand(5,1)
```
simonbyrne: again, what are reasonable non-random values we can use here?

docs/src/Atmos/SurfaceFluxes.md
```diff
@@ -0,0 +1,114 @@
...
+z_0 = z_0/1000
+u_ave = 200+u_ave*10
+Δz = range(10.0, stop=100.0, length=100)
+Ψ_m_tol, tol_abs, iter_max = 1e-3, 1e-3, 10
+u_star = Nishizawa2018.compute_friction_velocity.(
+    u_ave, θ, flux, Δz, z_0, a, Ψ_m_tol, tol_abs, iter_max)
+plot(u_star, Δz, title = "Friction velocity vs dz", xlabel = "Friction velocity", ylabel = "dz")
```
simonbyrne: should this plot correspond to a figure from a paper?

docs/src/Atmos/SurfaceFluxes.md
charleskawczynski: I can confirm that all but `z_0` are reasonable. Looking into `z_0`.

charleskawczynski: To produce Fig 4, I think we can use several values for `z_0`:
```
 70.0
 17.5
  5.83
  3.5
  0.35
```
Need to confirm this (plots are not showing up in the docs for me right now)..

simonbyrne: that's usually a sign of an error (unfortunately it doesn't give a stacktrace). Easiest option is to just load up the `docs` project in a REPL and copy paste

docs/src/Atmos/SurfaceFluxes.md
charleskawczynski: We inlined it in #64, we can revive these functions if we want to use them to recreate the plots. I'm fine either way. They're relatively small functions, so I don't know that we need to plot them for verification.

docs/src/Atmos/SurfaceFluxes.md
charleskawczynski: No, I think I was just seeing if the results made sense.

PR #143, opened 2019-04-01
Breaking out standalone dycore components

jkozdon:
This breaks the ODEsolver, mesh types, and MPI-aware arrays off of the
AtmosDycore in the hopes that they will be reusable components


src/ODESolvers/ODESolvers.jl
```diff
@@ -0,0 +1,80 @@
...
+A series of optional callback functions can be specified using the tuple
+`callbacks`; see [`GenericCallbacks`](@ref).
+
+!!! todo
+
+    Currently `stopaftertimeend` is not used. The idea behind it was that a user
+    might want to stop either one step before or after the final time `timeend`.
+    This should either be removed or used.
+"""
+function solve!(Q, solver::AbstractODESolver; timeend::Real=Inf,
```
glwagner: A couple thoughts (which you may ignore):

- In the context of differential equations, I often think of 'solving' as solving an elliptic equation, whereas time-integration is 'integrating', 'running', or 'stepping' --- though ultimately that's a matter of style.

- If `numberofsteps` is made `Inf` by default, it won't get triggered when `timeend` is set, which obviates the need for `stopaftertimeend`?

- `endtime` or `stoptime` instead of `timeend` --- just like `numberofsteps` is spoken-word-y?

- what's a callback?

jkozdon: FYI this pull request is closed in favor of #146 

Thanks for the feedback. Overall I think these are fair points, though maybe not good for this pull request. I'd like to keep this focused on a pretty straight movement of stuff around, new features and API changes should be separate issues and pull requests I think.

> In the context of differential equations, I often think of 'solving' as solving an elliptic equation, whereas time-integration is 'integrating', 'running', or 'stepping' --- though ultimately that's a matter of style.

We borrowed this term from the [differential equations package](http://docs.juliadiffeq.org/stable/basics/overview.html), I'm fine changing this. Maybe open up an issue and we can discuss there.

> If `numberofsteps` is made `Inf` by default, it won't get triggered when `timeend` is set, which obviates the need for `stopaftertimeend`?

Interface certainly needs cleaning up, I'd welcome a pull request to try to do so.

> `endtime` or `stoptime` instead of `timeend` --- just like `numberofsteps` is spoken-word-y?

This might be a breaking change for some of the drivers, so perhaps once again I'd welcome a separate pull request to improve/change this.

> what's a callback?

The idea is its a way to inject code into the ODE solver after a time step (we may add other callbacks for injecting code after/during the update as well).

Motivation for this is to support displaying run time info, saving output, etc. without the solver having to know a ton about what is being solved.

Take a look at the [rising thermal bubble driver](https://github.com/climate-machine/CLIMA/blob/1086bdb2af3862724af759e9d918b76bb5279005/src/ClimaAtmos/Dycore/drivers/rising_thermal_bubble.jl#L118) as well as the the code that follows in the [`solve!`](https://github.com/climate-machine/CLIMA/blob/1086bdb2af3862724af759e9d918b76bb5279005/src/ClimaAtmos/Dycore/src/CLIMAAtmosDycore.jl#L58).

PR #146, opened 2019-04-01
Breaking out standalone dycore components

jkozdon:
This breaks the ODEsolver, mesh types, and MPI-aware arrays off of the AtmosDycore in the hopes that they will be reusable components


src/Arrays/MPIStateArrays.jl
```diff
@@ -1,13 +1,13 @@
...
 using MPI
 
-export AtmosStateArray
+export MPIStateArray
 
 """
-    AtmosStateArray{S <: Tuple, T, DeviceArray, N,
-                    DATN<:AbstractArray{T,N}, Nm1, DAI1} <: AbstractArray{T, N}
+    MPIStateArray{S <: Tuple, T, DeviceArray, N,
+                  DATN<:AbstractArray{T,N}, Nm1, DAI1} <: AbstractArray{T, N}
```
simonbyrne: Happy to keep this for the time being, but I was wondering if we could come up with a better name for this. Maybe `MPIGhostArray` or something like that?

jkozdon: Not sure what the word `Ghost` means in this context. I tend to think of `Ghost` being data that an mpi rank depends another ran owns.

jkozdon:
> Most of this looks like re-naming / restructuring, so this is fine by me.

Yes that's all it really is. Trying to pull stuff out of the AtmosDycore that is not AtmosDycore specific

src/Arrays/MPIStateArrays.jl
jkozdon: But FYI, I'm not wed to this name... just had to name it something other than `AtmosStateArray`. @lcw might have thoughts.

simonbyrne: Let's leave it as is for now.

simonbyrne:
Looks good! Just need to link the new markdown files into `docs/make.jl`

Would it make sense to move the `Dycore` directory to the top-level? (I realise this could also be done as part of your later work)

jkozdon:
> Looks good! Just need to link the new markdown files into `docs/make.jl`

Great. Updated the commits to include this.

> Would it make sense to move the `Dycore` directory to the top-level? (I realise this could also be done as part of your later work)

Might be best to leave for now so as to minimize trouble for Akshay and Simone later this week?

Were you just thinking that it should live in something like `src/AtmosDycore` eventually?

simonbyrne:
> Were you just thinking that it should live in something like `src/AtmosDycore` eventually?

Or just `src/Dycore` (since the hope is that we can share at least some of it with the ocean)

PR #138, opened 2019-03-28
Moist thermo

asraero:
1) Update VanillaAtmosDiscretizations_kernels.jl to include MoistThermodynamics functions (including saturation adjustment)
2) Update MoistThermodynamics.jl saturation_adjustment.jl to include check for q_t < q_v* 
3) Update drivers/rising_thermal_bubble.jl and drivers/tracers_test.jl to include moist quantities

PR #147, opened 2019-04-01
[MoistThermo] Add ThermodynamicState struct

charleskawczynski:
This is the gist of the idea I mentioned today during the meeting.

Do you guys (@simonbyrne , @jkozdon, @lcw) see any potential performance/memory issues?

Moving in this direction can shift all of the responsibility, of ensuring a consistent thermodynamic state, to `MoistThermodynamics.jl`. Furthermore, if the Thermodynamic formulation changes, fewer places in the implementation would need to be touched (only the initialization of `ThermodynamicState`, I think).

So far, I've only implemented two constructors, one for assuming equilibrium:

`InternalEnergy_Shum_eq{DT} <: ThermodynamicState{DT}`

and another one assuming non-equilibrium:

`InternalEnergy_Shum_neq{DT} <: ThermodynamicState{DT}`

More tests are needed. On the implementation side, one first needs to instantiate a `ThermodynamicState`:

`ts = InternalEnergy_Shum_eq(e_int, q_t, rho)`

Then, only that object is needed to get any other properties:

```
T = temperature(ts)
p = pressure(ts)
```
etc.


simonbyrne:
I like this idea. Would we want separate structs for prognostic only / prognostic + computed?

charleskawczynski:
I think we should stick to a single struct. `ThermodynamicState` should not be concerned with what we happen to solve for/diagnose in the dycore/parameterizations.

And what members we include in each `ThermodynamicState` is a trade-off between memory/computation. Minimally, they only need to include whatever two independent thermodynamic properties are passed in.

simonbyrne:
Sounds good. Would be good to see an example of it in use.

charleskawczynski:
One small example is in the `runtests.jl` of this PR, but the benefits are less clear from this example.

Once #138 is merged, I can submit a PR where I implement this interface in place of the existing MoistThermo function calls.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -37,6 +37,51 @@ export liquid_fraction, phase_partitioning_eq!, saturation_adjustment, phase_par
...
+# Thermodynamic states
+export ThermodynamicState, InternalEnergy_Shum_eq, InternalEnergy_Shum_neq
+
+abstract type ThermodynamicState{DT} end
+
+"""
+    InternalEnergy_Shum_eq{DT} <: ThermodynamicState
+
+A thermodynamic state initialized by internal energy
+`e_int`, specific humidity `q_t`,  and density `ρ`,
```
tapios: @asraero just changed `q_t` and other moisture variables to the more verbose `q_tot`.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -477,6 +802,17 @@ and ice specific humidity `q_i`.
 """
 dry_pottemp(T, p, q_t=0, q_l=0, q_i=0) = T / exner(p, q_t, q_l, q_i)
 
+"""
+    dry_pottemp(T, p, q_t=0, q_l=0, q_i=0)
```
tapios: Want to update doc line here, right?

tapios:
I like the approach. It's very much like thermodynamics--pick your state variables, and everything else can be calculated from them. Only difference is that what is called state here is a redundant description of the state, but that seems ok unless it causes performance issues.

charleskawczynski:
> what is called state here is a redundant description of the state, but that seems ok unless it causes performance issues

Agreed.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,100 +32,258 @@ export saturation_excess
 export liquid_fraction, saturation_adjustment, phase_partitioning_eq
 
 # Auxiliary functions, e.g., for diagnostic purposes
-export liquid_ice_pottemp, dry_pottemp, exner
+export liquid_ice_pottemp, liquid_pottemp, dry_pottemp, density_pottemp, exner
+export mix_ratio_con, mix_ratio_vap
 
+# Thermodynamic states
+export ThermodynamicState, InternalEnergy_Shum_eq, InternalEnergy_Shum_neq
+export LiqPottemp_Shum_eq
```
simonbyrne: I'm not sure on the mixed case. How about
```
InternalEnergySHumEquil
InternalEnergySHumNonEquil
LiquidPotTempSHumEquil
```

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,100 +32,258 @@ export saturation_excess
...
+adjustment is needed).
+"""
+struct InternalEnergy_Shum_eq{DT} <: ThermodynamicState{DT}
+    e_int::DT
+    q_tot::DT
+    ρ::DT
+    T::DT
+    function InternalEnergy_Shum_eq(e_int, q_tot, ρ)
+        return new{typeof(ρ)}(e_int, q_tot, ρ, saturation_adjustment(e_int, ρ, q_tot))
+    end
```
simonbyrne: These could just be outer constructors

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,102 +32,264 @@ export saturation_excess
 export liquid_fraction, saturation_adjustment, phase_partitioning_eq
 
 # Auxiliary functions, e.g., for diagnostic purposes
-export liquid_ice_pottemp, dry_pottemp, exner
+export liquid_ice_pottemp, liquid_pottemp, dry_pottemp, virt_pottemp, density_pottemp, exner
+export mix_ratio_con, mix_ratio_vap
 
+# Thermodynamic states
+export ThermodynamicState, InternalEnergySHumEquil, InternalEnergySHumNonEquil
+export LiquidPotTempSHumEquil
```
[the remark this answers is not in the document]
charleskawczynski: We can if you'd like, but I think that the file size will stagnate, so I think it may be okay as is.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
tapios: I'd replace `SHumEquil` by `PhaseEquil` or something similar, and same for non-equilibrium (because thermally there still is equilibrium...)

tapios: And I'm also not sure we need to support liquid (or, better, liquid-ice) potential temperature as a state variable. It's not generally a good choice (for starters, it is not additive). Seems like overloading the code with unnecessary options to me.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,102 +32,264 @@ export saturation_excess
...
+ - `q_ice` ice specific humidity
+
+Without the specific humidity arguments, it the results
+are that of dry air.
 """
-function gas_constant_air(q_tot=0, q_liq=0, q_ice=0)
+function gas_constant_air(q_tot::DT=0.0, q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT
 
-    return R_d * ( 1 +  (molmass_ratio - 1)*q_tot - molmass_ratio*(q_liq + q_ice) )
+    return DT(R_d) * ( 1 +  (DT(molmass_ratio) - 1)*q_tot - DT(molmass_ratio)*(q_liq + q_ice) )
```
tapios: Would be good to replace `molmass_ratio` here (and in PlanetParameters and SurfaceFluxes) by `\epsilon_dv` as in atmosphere design doc.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,102 +32,264 @@ export saturation_excess
...
 
 end
 
 
 """
+    moist_gas_constants(ts::ThermodynamicState)
     moist_gas_constants([q_tot=0, q_liq=0, q_ice=0])
 
-Wrapper to return R_m, cv_m, cp_m, and gamma_m all at once
+Wrapper to return R_m, cv_m, cp_m, and gamma_m all
```
tapios: Would unicode `gamma` be kosher here?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -255,15 +514,17 @@ struct Liquid <: Phase end
...
 temperature `T`.
 
-    `saturation_vapor_pressure(T, Ice())`
+    saturation_vapor_pressure(T, Ice())
+
 Return the saturation vapor pressure over a plane ice surface at
 temperature `T`.
 
-    `saturation_vapor_pressure(T, LH_0, cp_diff)`
+    saturation_vapor_pressure(T, LH_0, cp_diff)
```
tapios: Maybe we can have unicode `\Delta cp` in place of `cp_diff` throughout?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -407,50 +713,66 @@ end
...
-      return air_temperature(e_int, q_tot)   
+      return air_temperature(e_int, q_tot)
    else
-    tol_abs = 1e-3*cv_d
+    tol_abs = DT(1e-3)*DT(cv_d)
     iter_max = 10
     args = (ρ, q_tot, e_int)
-    T0 = max(T_min, air_temperature(e_int, q_tot, 0.0, 0.0))
-    T1 = air_temperature(e_int, q_tot, 0.0, q_tot)
+    T0 = max(T_min, air_temperature(e_int, q_tot, DT(0), DT(0)))
```
tapios: Add a FIXME here: need to revisit bounds for saturation adjustment to guarantee bracketing of zero. (I'll do it later.)

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+function liquid_ice_pottemp(ts::ThermodynamicState)
+
+    return liquid_ice_pottemp(air_temperature(ts),
+                              air_pressure(ts),
+                              ts.q_tot,
+                              phase_partitioning_eq(ts)...)
+
+end
+
+"""
```
tapios: Please remove the mixing ratios, unless there is a strong reason why we need them. We should stay within one set of moisture variables to avoid unnecessary confusion, and I right now I see no reason why we would need the mixing ratios.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
     return air_temperature(e_int, q_tot, q_liq, q_ice)
   end
 end
+function saturation_adjustment_q_t_θ_l(θ_liq::DT, q_tot::DT, ρ::DT, p::DT) where DT
+  T_1 = θ_liq*(p/DT(MSLP))^(DT(R_d)/DT(cp_d))
```
tapios: This is not consistent with our definition of dry potential temperature (which uses the moist-air gas constants. Replace by a function that inverts the liquid-ice potential temperature. Also, use liquid-ice potential temperature consistently, if at all. No need for a separate liquid potential temperature, I think.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+  qv_star = saturation_shum(T_1, ρ)
+  if (q_tot <= qv_star) # If not saturated
+    return T_1
+  else  # If not saturated, iterate
+    T_2 = T_1 + (q_tot - qv_star) * latent_heat_vapor(T_1) /((1-q_tot)*DT(cp_d) + qv_star * DT(cp_v))
```
tapios: Use saturation excess function and partitioning into liquid and ice, then the correct gas constants for moist air.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+    function eos_root(T, args)
+      qv_star = saturation_shum(T, ρ)
+      θ = T*(DT(MSLP) / p)^(DT(R_d)/DT(cp_d))
```
tapios: Not consistent with our definition of dry potential temperature and liquid-ice potential temperature. Use those functions to make it consistent, rather than redefining potential temperature.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+    return dry_pottemp(air_temperature(ts),
+                       air_pressure(ts),
+                       ts.q_tot,
+                       phase_partitioning_eq(ts)...)
 
 end
 
 """
-    dry_pottemp(T, p, q_tot=0, q_liq=0, q_ice=0)
+    virt_pottemp(ts::ThermodynamicState)
```
tapios: I'd prefer to write out `virtual_pottemp`.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+Return the virtual potential temperature,
+given a thermodynamic state `ts` or
+
+ - `T` temperature
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
+ """
+virt_pottemp(T::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT = gas_constant_air(q_tot, q_liq, q_ice)/R_d*T
```
tapios: This looks like virtual temperature rather than virtual potential temperature (which would replace `T` by `dry_pottemp`).

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
 
+ - `T` temperature
+ - `p` pressure
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 """
-    exner(p, q_tot=0, q_liq=0, q_ice=0)
+function liquid_pottemp(T::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT
```
tapios: I think what you want here is not a liquid potential temperature, but a *saturation* liquid-ice potential temperature. You can get this from liquid-ice potential temperature by computing saturation excess and phase partitioning.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -462,49 +784,212 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+given a thermodynamic state `ts` or
+
+ - `T` temperature
+ - `p` pressure
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
+"""
+function density_pottemp(T::DT, p::DT, q_tot::DT, q_liq::DT, q_ice::DT) where DT
```
tapios: Remove `density_pottemp`. This is the same as the virtual potential temperature.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,102 +32,264 @@ export saturation_excess
...
-    _R_m = gas_constant_air(q_tot, q_liq, q_ice)
-    return sqrt(_γ * _R_m * T)
+"""
+    InternalEnergySHumEquil{DT} <: ThermodynamicState
+
+A thermodynamic state initialized by
+
+ - `e_int` internal energy
+ - `q_tot` total specific humidity
+ - `ρ` density
```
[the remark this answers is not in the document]
charleskawczynski: Should it be the rule to use this?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -139,97 +298,197 @@ function moist_gas_constants(q_tot=0, q_liq=0, q_ice=0)
...
+end
 
 """
+    air_temperature(ts::ThermodynamicState)
     air_temperature(e_int[, q_tot=0, q_liq=0, q_ice=0])
 
-Return the air temperature, given the internal energy `e_int` per unit mass,
-and, optionally, the total specific humidity `q_tot`, the liquid specific humidity
-`q_liq`, and the ice specific humidity `q_ice`.
+Return the air temperature, given a thermodynamic state `ts` or
```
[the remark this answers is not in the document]
charleskawczynski: I tried to be consistent by using a single docstring for each function, should I split all of them then?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
charleskawczynski: I like this idea, but maybe this can be a separate PR? (I think this is used in the, and I’d like to keep changes local)

charleskawczynski: I like this idea, but maybe this can be a separate PR? (I think this is used in the, and I’d like to keep changes local)

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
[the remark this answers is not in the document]
simonbyrne: It's a matter of convenience: not always necessary for small structs, but handy for larger ones.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
tapios: Sure, a separate PR is fine.

simonbyrne:
@charleskawczynski I cleaned up the main things I found. I'll leave you to address the remainder of Tapio's comments.

One thing that did occur to me is that we could have a
```julia
struct PhasePartition{DT}
    q_tot::DT
    q_liq::DT
    q_ice::DT
end
```
which might simplify some of the code, but that can be a separate PR.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -463,48 +739,244 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+
+Compute the temperature that is consistent with
+
+ - `p` pressure
+ - `q_tot` total specific humidity
+ - `θ_liq` liquid potential temperature
+
+See also [`saturation_adjustment`](@ref).
+"""
+function saturation_adjustment_q_t_θ_l(θ_liq::DT, q_tot::DT, ρ::DT, p::DT) where DT
```
tapios: I think a preferable way of writing this function is to make it parallel to the saturation adjustment for internal energy, just replacing the roots equation and the initial temperature bracket.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
tapios: So, parallel to the internal energy adjustment, this would be something like `roots_equation ... = liquid_ice_pottemp_sat(...) - liquid_ice_pottemp`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -463,48 +739,244 @@ function saturation_adjustment(e_int, ρ, q_tot, T_init = T_triple)
...
+The liquid potential temperature where
+
+ - `T` temperature
+ - `p` pressure
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
+"""
+function liquid_pottemp(T::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where {DT}
```
tapios: Replace this by something like `liquid_ice_pottemp_sat`, and use `saturation_excess`, `phase_partitioning_eq` and `liquid_ice_pottemp` to compute it.

charleskawczynski:
> cleaned up the main things I found. I'll leave you to address the remainder of Tapio's comments.

Thanks! I'm going to speak with Tapio today.

Also, I like the `PhasePartition` sub-structure, maybe we can do this in a separate PR.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,102 +32,264 @@ export saturation_excess
...
-function soundspeed_air(T, q_tot=0, q_liq=0, q_ice=0)
+abstract type ThermodynamicState{DT} end
 
-    _γ   = cp_m(q_tot, q_liq, q_ice)/cv_m(q_tot, q_liq, q_ice)
-    _R_m = gas_constant_air(q_tot, q_liq, q_ice)
-    return sqrt(_γ * _R_m * T)
+"""
+    InternalEnergySHumEquil{DT} <: ThermodynamicState
+
+A thermodynamic state initialized by
```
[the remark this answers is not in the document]
simonbyrne: I've changed these a bit, but could do with slightly better descriptions.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
charleskawczynski: I think I was able to clear this up, but I think there's still one loose end.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
charleskawczynski: They'd be convenient. I can work around omitting them out for now, but I'll probably write the `tke` buoyancy source in terms of mixing ratios in the short-term EDMF implementation.

tapios: Let's talk about the buoyancy source terms. They should be written in terms of the specific humidities we have, not mixing ratios.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
charleskawczynski: > I'd replace `SHumEquil` by `PhaseEquil` or something similar, and same for non-equilibrium (because thermally there still is equilibrium...)

This looks like the right abstraction. I've changed `InternalEnergySHumEquil` to `PhaseEquil`, and similar for non-equilibrium.

charleskawczynski: > And I'm also not sure we need to support liquid (or, better, liquid-ice) potential temperature as a state variable. It's not generally a good choice (for starters, it is not additive). Seems like overloading the code with unnecessary options to me.

I think you'll like our new solution. We're accepting these variables as input, but they are immediately converted into a `PhaseEquil` object, using the functions we discussed. So, there are currently only two structs: `PhaseEquil` and `PhaseNonEquil`.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
-`q_liq`, and the ice specific humidity `q_ice`.
+The air temperature, where
+
+ - `e_int` internal energy per unit mass
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
+
+Without the specific humidity arguments, it the results
```
tapios: remove it in `arguments, the results`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
-        ( internal_energy - (q_tot - q_liq) * e_int_v0 + q_ice * (e_int_v0 + e_int_i0) )/
-            cv_m(q_tot, q_liq, q_ice)
+ - `θ` potential temperature
+ - `p` pressure
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 
+Without the specific humidity arguments, it the results
```
tapios: remove `it`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
+are that of dry air.
+"""
+function air_temperature_from_pottemp(θ::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT
```
tapios: I think you want `air_temperature_from_liquid_ice_pottemp` here, and solve the (linearized) liquid-ice potential temperature for `T`. This is what is needed for bracketing in saturation adjustment.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 
+Without the specific humidity arguments, the results
+are that of dry air.
+"""
+function soundspeed_air(T::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where {DT<:Real}
+  γ   = cp_m(q_tot, q_liq, q_ice) / cv_m(q_tot, q_liq, q_ice)
+  R_m = gas_constant_air(q_tot, q_liq, q_ice)
+  return sqrt(γ*R_m* T)
```
tapios: insert spaces around multiplication operator, or remove in front of `T`?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
+latent_heat_generic(T::Real, LH_0::Real, Δcp::Real) =
+  LH_0 + Δcp * (T - T_0)
 
-    return LH_0 + cp_diff * (T - T_0)
 
-end
+"""
+    Phase
 
+A generic gas mixture phase, to dispatch
```
tapios: Replace `generic gas mixture phase` by `condensate phase`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
+over [`saturation_vapor_pressure`](@ref) and
+[`saturation_shum_generic`](@ref).
+"""
 abstract type Phase end
+
+"""
+    Liquid <: Phase
+
+A liquid gas mixture phase, to dispatch
```
tapios: remove `gas mixture`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -35,247 +34,496 @@ export saturation_excess
...
+over [`saturation_vapor_pressure`](@ref) and
+[`saturation_shum_generic`](@ref).
+"""
 struct Liquid <: Phase end
+
+"""
+    Ice <: Phase
+
+An ice gas mixture phase, to dispatch
```
tapios: remove `gas mixture`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -284,26 +532,34 @@ relation to obtain the saturation vapor pressure `p_vs` as a function of
...
-condensate, given the temperature `T` and the (moist-)air density `ρ`.
+Compute the saturation specific humidity over a plane surface of condensate, given
 
-The optional argument `phase` can be ``Liquid()`` or ``"ice"`` and indicates
-the condensed phase.
+ - `T` temperature
+ - `ρ` (moist-)air density
+and, optionally,
+ - `Liquid()` phase indicating liquid
+ - `Ice()` phase indicating ice
```
tapios: replace [`Liquid()/Ice()`] `phase indicating` by `indicating condensate is`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -314,25 +570,31 @@ end
...
+
+ - `T` temperature
+ - `ρ` (moist-)air density
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 
-If the optional liquid, and ice specific humdities `q_tot` and `q_liq` are given,
+If the optional liquid, and ice specific humidities `q_tot` and `q_liq` are given,
```
tapios: remove comma after liquid

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 
 If the optional input arguments `q_liq` and `q_ice` are not given or are zero, the
 fraction of liquid is a function that is 1 above `T_freeze` and goes to zero below
 `T_freeze`. If `q_liq` or `q_ice` are nonzero, the liquid fraction is computed from
 them.
 """
-function liquid_fraction(T, q_liq=0, q_ice=0)
+function liquid_fraction(T::DT, q_liq::DT=DT(0), q_ice::DT=DT(0)) where {DT<:Real}
```
tapios: It would be good to have two functions `liquid_fraction_equil` and `liquid_fraction_nonequil`. `liquid_fraction_equil` should always be the Heaviside function. `liquid_fraction_nonequil` for now can be equal to `liquid_fraction_equil` but eventually can be a more general function.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
     iter_max = 10
     args = (ρ, q_tot, e_int)
-    T0 = max(T_min, air_temperature(e_int, q_tot, 0.0, 0.0))
-    T1 = air_temperature(e_int, q_tot, 0.0, q_tot)
-    roots_equation(x, ρ, q_tot, e_int) = internal_energy_sat(x, ρ, q_tot) - e_int
-    T, converged = find_zero(roots_equation,
-                             T0, T1,
-                             args,
+    # FIXME here: need to revisit bounds for saturation adjustment to guarantee bracketing of zero.
+    T_1 = max(T_min, air_temperature(e_int, q_tot, DT(0), DT(0)))
```
tapios: You can construct this parallel to the saturation adjustment for potential temperature. Add:
` qv_sat = saturation_shum(T_1, ρ)
	  if (q_tot <= qv_sat) # If not saturated
	    return T_1
	  else  # If saturated, iterate`

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
+
+See also [`saturation_adjustment`](@ref).
+"""
+function saturation_adjustment_q_tot_θ_liq_ice(θ_liq_ice::DT, q_tot::DT, ρ::DT, p::DT) where DT
+  T_1 = air_temperature_from_pottemp(θ_liq_ice, p) # Assume all vapor
+  qv_star = saturation_shum(T_1, ρ)
+  if (q_tot <= qv_star) # If not saturated
+    return T_1
+  else  # If saturated, iterate
+    T_2 = air_temperature_from_pottemp(θ_liq_ice, p, q_tot, DT(0), q_tot) # Assume all ice
```
tapios: Same as above. Need to update the `air_temperature_from...` function so that this will work.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
+"""
+function liquid_ice_pottemp(T::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT
     # isobaric specific heat of moist air
     _cp_m   = cp_m(q_tot, q_liq, q_ice)
 
     # liquid-ice potential temperature, approximating latent heats
     # of phase transitions as constants
-    return dry_pottemp(T, p, q_tot, q_liq, q_ice) * exp(-(LH_v0*q_liq + LH_s0*q_ice)/(_cp_m*T))
-
+    return dry_pottemp(T, p, q_tot, q_liq, q_ice) * exp(-(DT(LH_v0)*q_liq + DT(LH_s0)*q_ice)/(_cp_m*T))
```
tapios: Replace by linearized form, so it can easily be inverted for temperature:
`dry_pottemp(T, p, q_tot, q_liq, q_ice) * (1 - (DT(LH_v0)*q_liq + DT(LH_s0)*q_ice)/(_cp_m*T))`

Put the inverse `air_temperature_from...` function near here, and make sure it inverts this expression.

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
+The dry potential temperature, given a thermodynamic state `ts`.
 """
-function exner(p, q_tot=0, q_liq=0, q_ice=0)
+dry_pottemp(ts::ThermodynamicState) =
+  dry_pottemp(air_temperature(ts), air_pressure(ts), ts.q_tot,
+              phase_partitioning_eq(ts)...)
+
 
+"""
+    virtual_temp(T, p, q_tot, q_liq, q_ice)
```
tapios: Replace `virtual_temp` by `virtual_pottemp` (because this is what is now computing)

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -341,170 +603,333 @@ function saturation_shum(T, ρ, q_liq=0, q_ice=0)
...
+ - `T` temperature
+ - `p` pressure
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
+"""
+function liquid_ice_pottemp_sat(T::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where {DT}
+    ρ = air_density(T, p, q_tot, q_liq, q_ice)
+    q_vs = saturation_shum(T, ρ, q_liq, q_ice)
```
tapios: We should use a uniform variable for saturation specific humidity, e.g., 'q_v_sat' would be consistent with our other notation. (Above the same variable is `q_vstar`.)

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -7,504 +7,951 @@ saturation specific humidities
...
-function saturation_vapor_pressure(T, LH_0, cp_diff)
+end
+saturation_vapor_pressure(T::DT, ::Ice) where DT = saturation_vapor_pressure(T, DT(LH_s0), DT(cp_v) - DT(cp_i))
+saturation_vapor_pressure(ts::ThermodynamicState, ::Ice) = saturation_vapor_pressure(air_temperature(ts), LH_s0, cp_v - cp_i)
+
+function saturation_vapor_pressure(T::DT, LH_0::DT, Δcp::DT) where DT
 
-    return press_triple * (T/T_triple)^(cp_diff/R_v) *
-        exp( (LH_0 - cp_diff*T_0)/R_v * (1 / T_triple - 1 / T) )
+    return DT(press_triple) * (T/DT(T_triple))^(Δcp/DT(R_v)) *
```
simonbyrne: It seems like `T/DT(T_triple)` is negative?

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
```diff
@@ -7,504 +7,951 @@ saturation specific humidities
...
+and, optionally,
+ - `q_tot` total specific humidity
+ - `q_liq` liquid specific humidity
+ - `q_ice` ice specific humidity
 
+Without the specific humidity arguments, the results
+are that of dry air.
+"""
+function air_temperature_from_liquid_ice_pottemp(θ::DT, p::DT, q_tot::DT=DT(0), q_liq::DT=DT(0), q_ice::DT=DT(0)) where DT
+    return θ*exner(p, q_tot, q_liq, q_ice)
```
tapios: Here you need to return this: `θ*exner(p, q_tot, q_liq, q_ice) + (DT(LH_v0)*q_liq + DT(LH_s0)*q_ice)/_cp_m` with the usual definition for `_cp_m`. I would also replace `θ` by `θ_liqice` or something like it to make clear it is the liquid ice potential temperature.

charleskawczynski: Thank you!

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
charleskawczynski: Yes, temperatures returned from saturation adjustment were as low as ~`0.001`. I suspect that there is an error in the isentropic vortex test.

The test is breaking [here](https://github.com/climate-machine/CLIMA/blob/master/src/DGmethods/test/Euler/isentropic_vortex.jl#L55), but I'm suspicious of [this](https://github.com/climate-machine/CLIMA/blob/master/src/DGmethods/test/Euler/isentropic_vortex.jl#L54) equation. I don't see why internal energy at zero Kelvin is added to the RHS, this doesn't seem to match the design docs. @jkozdon, I tried adjusting this line, but it didn't seem to fix the test (sorry for roping you into `MoistThermodynamics.jl`!).

charleskawczynski: For now, I've removed `isentropic_vortex` from the test set. We can add it back in once we figure out what's wrong.

tapios: @jkozdon Yes, the initialization of energy in this [line](https://github.com/climate-machine/CLIMA/blob/5a2b2eee73da88629b3472b0b7c11c6857542873/src/DGmethods/test/Euler/isentropic_vortex.jl#L113) is not consistent with the thermodynamics. That line in isentropic_vortex.jl needs to be replaced by a call to `total_energy`, with temperature from `air_temperature`. The `internal_energy(0)` in [this](https://github.com/climate-machine/CLIMA/blob/5a2b2eee73da88629b3472b0b7c11c6857542873/src/DGmethods/test/Euler/isentropic_vortex.jl#L54) then needs to be removed.

charleskawczynski:
Closes #119

src/Utilities/MoistThermodynamics/src/MoistThermodynamics.jl
jkozdon: Sorry for missing this conversation.

I based this based on what was in the current Dycore code for [`isentropic_vortex.jl`](https://github.com/climate-machine/CLIMA/blob/d13e455e315cea579c69263af02f5ca562a6af79/src/Atmos/Dycore/drivers/isentropic_vortex.jl#L68) which came in with Moist thermo pr #138. 

The movement of `internal_energy(0)` out of the initial condition was done to be more `Float32` friendly.

For now I am deleting the test from the balance law complete. When the new framework is in place proper integration tests with thermodynamics can be added.

PR #151, opened 2019-04-04
rename CLIMA.CLIMAAtmos => CLIMA.Atmos

simonbyrne:
rename the directory and module `CLIMAAtmos` to just `Atmos` since they already are subdirectories/submodules of CLIMA


charleskawczynski:
Need to update `ClimaAtmos` in the includes, e.g., here https://github.com/climate-machine/CLIMA/blob/master/src/CLIMA.jl#L7. Unfortunately, this still won't pass the gitlab CI. Maybe we can discuss this

PR #158, opened 2019-04-08
Start of new balance law solver

jkozdon:
This is the start of the new balance law solver.

This is not functionality equivalent to the old atmosdycore yet as it only handles periodic BCs and hyperbolic conservation laws (no sources or gradient terms)


jkozdon:
> What is a balance law solver? Is this a solver for general PDEs?

The idea is to come up with a DG implementation that works for balance laws needed for the atmospheric model. So not all PDES, but sufficient for compressible Naiver-Stokes and microphysics.

glwagner:
Okay, nice! Is a balance law a PDE then? Or something different?

jkozdon:
> Is a 'balance law' a partial differential equation then (for example, the advection-diffusion equation)? 

Yes. Essentially  something that can be written in terms of a conservation law plus a source.

If you can access it, see Equation (3.1.1) of [Dafermos C.M. (2010) Hyperbolic Systems of Balance Laws. In: Hyperbolic Conservation Laws in Continuum Physics. Grundlehren der mathematischen Wissenschaften (A Series of Comprehensive Studies in Mathematics), vol 325. Springer, Berlin, Heidelberg
](https://doi.org/10.1007/978-3-642-04048-1)

Right now what we're building won't work out of the box for incompressible because we won't be handling constraint equations.

glwagner:
Awesome, thank you. I just wanted to make sure I understood. I'll check that reference when I get on a university network!

> Right now what we're building won't work out of the box for incompressible because we won't be handling constraint equations.

While I see what you're saying that it won't work in an automated sense, we may still be able to solve incompressible systems by writing a special time-stepper that includes a pressure correction step (plus elliptic solver) to enforce incompressibility. That's a non-trivial step, but I think we're probably well more than half way there if we can write `∂ϕ/∂t = R` for generic `R`. While the atmosphere people may know what equations they want to implement pretty well already, we probably will need to do a bit of experimenting on the ocean side. 

In addition to that, a generic PDE solver that can be used to solve 1D equations will be super useful in prototyping new 1D column models --- where we don't need to worry about incompressibility.

src/Arrays/MPIStateArrays.jl
```diff
@@ -288,57 +290,50 @@ function knl_L2norm(::Val{Np}, Q, weights, elems) where {Np}
...
-  cpubackend = DeviceArray == Array
-  if vgeo == nothing
-    vgeo = cpubackend ? config.vgeo : Array(config.vgeo)
-  end
-  if Q == nothing
-    Q = cpubackend ? state.Q : Array(state.Q)
-  end
-  if time == nothing
-    time = state.time[1]
+function euclideandist(A::MPIStateArray, B::MPIStateArray)
```
tapios: Can we use camel case or underscores to make this and other function names more easily readable, e.g., `EuclideanDist`?

jkozdon: I would suggest `euclidean_distance` over `EuclideanDist` since that would be more in line with the Julia style for functions (CamelCase tends to be used more for types/structs).

jkozdon: I've updated a handful of other functions you might have been referring to. If there are others please let me know (or feel free to register an issue with a checklist or pop in a pull-request). For the most part I'm not wed to any names or naming conventions.

(I agree that `euclideandist` was a bad name, `euclidiandistance` would have been better at a minimum.)

simonbyrne:
I realise this is a piece of a larger issue (#160), but in future I think we should try to have the documentation in a better state before merging.

jkozdon:
Sorry. My bad on that.

PR #166, opened 2019-04-11
Turn current Atmos/Dycore drivers into tests

jkozdon:
Turns the current atmos drivers into a simple set of tests in order to ensure that future changes (e.g., backend stuff for the new balance law solver) does not break existing functionality.


src/Atmos/Dycore/test/runtests.jl
```diff
@@ -0,0 +1,28 @@
...
+  JL_LOG_NONE = 0
+  JL_LOG_USER = 1
+  JL_LOG_ALL = 2
+  coverage_opts = Dict{Int, String}(JL_LOG_NONE => "none",
+                                    JL_LOG_USER => "user",
+                                    JL_LOG_ALL => "all")
+  coverage_opt = coverage_opts[Base.JLOptions().code_coverage]
+  testdir = dirname(@__FILE__)
+
+  for (n, f) in [(1, "../drivers/isentropic_vortex.jl")
```
simonbyrne: should we move them into the `test` directory?

jkozdon: I'm fine with that. Will do so.

jkozdon: resolved by a91dbbc

src/Atmos/Dycore/test/runtests.jl
```diff
@@ -0,0 +1,28 @@
...
+  for (n, f) in [(1, "isentropic_vortex.jl")
+                 (3, "isentropic_vortex.jl")
+                 (1, "tracers_test.jl")
+                 (3, "tracers_test.jl")
+                 (1, "rising_thermal_bubble.jl")
+                 (3, "rising_thermal_bubble.jl")
+                ]
+    cmd =  `mpiexec -n $n $(Base.julia_cmd()) --startup-file=no --project=$(Base.active_project()) --code-coverage=$coverage_opt $(joinpath(testdir, f))`
+    @info "Running MPI test..." n f cmd
+    @test (run(cmd); true)
```
simonbyrne:
```suggestion
    run(cmd)
```
since an error will give a test failure anyway.

src/Atmos/Dycore/test/rising_thermal_bubble.jl
```diff
@@ -155,8 +155,11 @@ function main(mpicomm, DFloat, ArrayType, brickrange, nmoist, ntrace, N,
   @printf(io, "||Q||₂ ( final ) =  %.16e\n", engf)
   @printf(io, "||Q||₂ (initial) / ||Q||₂ ( final ) = %+.16e\n", engf / eng0)
   @printf(io, "||Q||₂ ( final ) - ||Q||₂ (initial) = %+.16e\n", eng0 - engf)
```
simonbyrne: We probably don't want to be printing diagnostics on every test run, so maybe add an option?
Ideally, could we turn these into tests, e.g
```
@test engf ≈ eng0
```

src/Atmos/Dycore/test/runtests.jl
jkozdon: Sounds good. Will add.

src/Atmos/Dycore/test/rising_thermal_bubble.jl
jkozdon: > We probably don't want to be printing diagnostics on every test run, so maybe add an option?

I can suppress the output.

> ```
> @test engf ≈ eng0
> ```

The methods we're using do not guarantee this, we will generally lose some energy though we could in principle gain some too

jkozdon: Note there is still a lot of chatter on the tests because of the balance law solver. I'll turn this off on the next balance law solver pull request.

jkozdon:
Going to merge this with a larger balance law solver pull request shortly.

PR #169, opened 2019-04-12
split travis tests into two sets

jkozdon:
travis has been timing out on some test, so I thought it might be good to split the into multiple test sets


jkozdon:
closing and merging with #173

Issue #172, opened 2019-04-12
Remove MPI dependency


jkozdon:
Basic idea to remove MPI

```julia
abstract type AbstractComm end

# for MPI
struct MPIComm <: AbstractComm
  comm::MPI.Comm
end

# for without MPI
struct SelfComm <: AbstractComm
  # might be needed if we ever want to support multiple communicators on the MPI side.
  comm::Int
end
```

Doing this will require moving [`Canary.jl`](https://github.com/climate-machine/Canary.jl) into CLIMA

See also how [`GPUArrays.jl`](https://github.com/JuliaGPU/GPUArrays.jl) handles the issue.

jkozdon:
With this issue we may want to consider how to do some sort of communicator based extension of the Julia logger.

tapios:
Is this still an issue?

jkozdon:
Yes, in the sense that `MPI.jl` is still required to run the code

jkozdon:
I'm fine closing this if we always want to have a hard dependency on `MPI.jl`

Issue #134, opened 2019-03-26
Logging framework


ali-ramadhan:
Maybe more of a question for @charleskawczynski @simonbyrne but any suggestions for a logging library? It's probably a good software practice to move away from `print`/`@printf` statements to a logging framework that provides detailed logs for debugging (and for users, with a lower severity).

I quite like [Memento.jl](https://github.com/invenia/Memento.jl) but maybe it's better to rely on Julia's built-in logging? The default format isn't very good for detailed logs but I'm sure we can set a custom format with timestamps, line numbers, etc.

simonbyrne:
I think starting with the built-in logging makes sense. It works pretty well, though there are some rough edges when dealing with multiple processes.

ali-ramadhan:
Thanks, will look into customizing it!

simonbyrne:
Re: distributed logging issues
- https://github.com/JuliaLang/julia/issues/25929
- https://github.com/JuliaLang/julia/issues/31483

simonbyrne:
Let's keep this issue open until we've figure it out.

ali-ramadhan:
Ah sure, sounds good. I just closed it as you answered my question.

jkozdon:
Right now our hack in the dycore is to redirect stdout to `dev/null`; see [here](https://github.com/climate-machine/CLIMA/blob/0c69e8973fa8b3a59b2f1ca0579e83f1a67796cf/src/ClimaAtmos/Dycore/drivers/rising_thermal_bubble.jl#L112)

Would be better to have something like `@info` and `@rootinfo`, where `@info` is that standard logging with a little bit of extra info like which MPI rank is dumping the data and `@rootinfo` just has one MPI rank dump the data; we'd also obviously want all the other functionality of the logging package too available via `@logmsg`.

Could be useful to also have the info statement "registered" in some way so that we know what is dumping the message (atmosdycore, oceandycore, landmodel, etc.). Something along the lines of

```
# assuming mpirank == 0
julia> @info "message"
[ info, AtmosDycore, 3: message

julia> @rootinfo "message"
[ info, AtmosDycore, --: message

julia>
```

```
# assuming mpirank == 3
julia> @info "message"
[ info, AtmosDycore, 3: message

julia> @inforoot "message"

julia>
```

I remember thinking a while ago that this should all be possible with a fairly thin layer on top of `Logging`, I just never quite fully tried to do it.

c42f:
[...]

PR #173, opened 2019-04-12
Balance Law: add docs, refactor some of the interface, add AtmosDycore tests

jkozdon:


src/Atmos/Dycore/test/isentropic_vortex.jl
```diff
@@ -19,6 +19,8 @@ using CLIMA.PlanetParameters: R_d, cp_d, grav, cv_d
 
 const halfperiod = 5
 
+const print_diagnostics = length(ARGS) == 0 || parse(Bool, ARGS[1])
```
simonbyrne: we should use Julia's logging infrastructure for this `@debug`, `@info` for this: want me to have a go?

jkozdon: Sure. I was just trying to suppress the existing output in the tests

jkozdon: We need it to play well with MPI too (so best would be also some `@root_info` as well in which case only 1 ranks does the output not everyone)

jkozdon: You can see how I used this down on line 127 to define an `io` variable.

simonbyrne: I think we should just be able to set a different log level on each rank.

jkozdon: really this is related to larger issue #134

simonbyrne: theoretically you could set up an MPILogger or something like that which would handle everything automatically, but this will work for now.

jkozdon: Sounds good. Are you adding it to this pull request or will you do a separate one?

simonbyrne: I can just add it to this one.

jkozdon: ok

simonbyrne: Let me know what you think of ce11cbdd510f368cc6733f465c3d2d8e39226917: we could move some of this into Utilities proper.

simonbyrne: Okay, I've updated all of them, hopefully that should work nicely.

jkozdon:
Forced a push to update to current master and try to fix the windows test

jkozdon:
Overall I'm good with this for now (assuming the tests pass)

jkozdon:
My fix didn't work. Reverting.

The problem is that `-env` is not a standard option for `mpiexec` (my mac which uses openmpi 3.0.0 doesn't recognize it either).

simonbyrne:
Since the tests are all run locally, it might suffice to set the ENV from
within Julia before calling MPI?

jkozdon:
Done. I just set this for all the tests in `test/runtest.jl` as this seemed most reasonable to me (?). Uniform behaviour across all the tests with any logging output.

jkozdon:
I added the test splitting idea from #169 into this pull request. I know that you weren't a huge fan of it @simonbyrne so I can remove it if you want (only used right now on travis).

I think the test set idea might be useful in the future, since the tests I'm running right now are pretty short compared to what will be coming once we test the full system. I think we'll likely have a short and long test set. This allows folks to also split tests into integration tests, unit test, moist test, etc.

I also added another environment variable "JULIA_CLIMA_TESTING" to run separate tests when in testing versus running tests as stand alone. Doing this allows me to run one fewer resolution in CI.

simonbyrne:
@jkozdon I have a small fix: should I just push it to this branch?

jkozdon:
> @jkozdon I have a small fix: should I just push it to this branch?

Will merge commit from #181 shortly.

jkozdon:
@simonbyrne Just an FYI, we still probably want some sort of MPI-aware logging. Subtle bug that took me a bit to take down was an MPI call inside an `@sinfo` which of course only the root called. Really we want all the ranks to execute everything in the logging calls but only the root rank to do the print.

jkozdon:
Adding logging to #172

jkozdon:
@simonbyrne or @lcw anything else you want addressed here before this is merged?

jkozdon:
> @simonbyrne or @lcw anything else you want addressed here before this is merged?

I know it's turned into a bit of a beast of a PR, hopefully future ones will be more bitesized!

lcw:
I love the docs!  Looks good to me.

PR #180, opened 2019-04-15
[TOML] Add Distributions.jl package to CLIMA

charleskawczynski:
Added Distributions.jl to CLIMA and its docs

Distributions will be needed to solve the EDMF equations.


jkozdon:
gpu environment?

PR #181, opened 2019-04-15
remaining print_diagnostics

simonbyrne:
NOTE: this is a PR to the #173 branch. I didn't want to break @jkozdon git rebasing.


jkozdon:
Thanks! Found a small error that makes sense to fold into #173 as well, so will merge this commit in there shortly.

Issue #155, opened 2019-04-07
Add a module to store hyperparameters


charleskawczynski:
Parameters that will be learned by DA/ML are currently stored in the dycore drivers and parameterization initialization functions. It's probably a good idea that we make a module for initializing/storing these parameters so that collecting them later does not lead to issues.

I think there was discussion of these parameters living in a `Dict` or `struct`, which I like. I'll add on that and suggest that we build initialize a `HyperParameterSet` using a `ValidRange`:

```julia
struct ValidRange{T}
  value_typical::T
  value_min::T
  value_max::T
end

struct HyperParameterSet{T}
params::T
end

function HyperParameterSet()
  params = Dict()
  params[:Cs] = ValidRange(0.0, 12/100, 10.0)  # Smagorinsky constant
  params[:N_subdomains] = ValidRange(2, 2, 10) # subdomain decomposition
  # ...
  params[:c_ε] = ValidRange(0.0, 0.12, 10.0)   # entr-detr factors
  params[:f_c] = ValidRange(0.0, 0.12, 1.0)    # buoyancy factors

  return HyperParameterSet(params)
end
```

I think that this struct will need certain interface functions, like 

 - a nice interface for converting a subset of parameters to a `NamedTuple` so that they can be assumed to be constant when passed in as an argument to the dycore/parameterizations
 - update routine for the DA/ML to update them between separate runs

I'd really like to hear what others think about this (the DA/ML team @agarbuno) @lcw @jkozdon  @simonbyrne @peterahrens @asraero

charleskawczynski:
Closing this. Work on free parameters is continuing at [FreeParameters.jl](https://github.com/climate-machine/FreeParameters.jl)

PR #186, opened 2019-04-17
[FreeParameters] Add FreeParameters module

charleskawczynski:
Adds a `FreeParameters` module for #155. This module is where the free/adjustable-parameters (that CLIMA will learn) will live.


src/Utilities/HyperParameters/src/HyperParameters.jl
```diff
@@ -0,0 +1,75 @@
...
+
+  return HyperParameterSet(value_min, value_typical, value_max)
+end
+
+"""
+    AtmosTurbConv
+
+Hyper-parameters in the [`TurbulenceConvection`](@ref) module.
+"""
+function AtmosTurbConv()
```
jkozdon: Should these take in a type and then use full precision numbers below?
```julia
function AtmosTurbConv(T)
  params = Dict()
  params[:N_subdomains] = ValidRange(2, 2, 10) # Number of sub-domains
  params[:c_ε] = ValidRange(T(0), T(12//100), T(10))   # entr-detr factors
  params[:f_c] = ValidRange(T(0), T(12//100), T(1))    # buoyancy factors
  return params
end
```

src/Utilities/HyperParameters/src/HyperParameters.jl
```diff
@@ -0,0 +1,75 @@
...
+
+  return params
+end
+
+"""
+    AtmosTurbConv
+
+Hyper-parameters in the Sub-Grid-Scale Physics.
+"""
+function AtmosSGS()
```
jkozdon: same

jkozdon:
Would `@kwdef struct` make more sense for parameters than a `Dict`?

```
help?> Base.@kwdef
  @kwdef typedef

  This is a helper macro that automatically defines a keyword-based
  constructor for the type declared in the expression typedef, which must be a
  struct or mutable struct expression. The default argument is supplied by
  declaring fields of the form field::T = default or field = default. If no
  default is provided then the keyword argument becomes a required keyword
  argument in the resulting type constructor.

  Inner constructors can still be defined, but at least one should accept
  arguments in the same form as the default inner constructor (i.e. one
  positional argument per field) in order to function correctly with the
  keyword outer constructor.

  │ Julia 1.1
  │
  │  Base.@kwdef for parametric structs, and structs with supertypes
  │  requires at least Julia 1.1.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> Base.@kwdef struct Foo
             a::Int = 1         # specified default
             b::String          # required keyword
         end
  Foo

  julia> Foo(b="hi")
  Foo(1, "hi")

  julia> Foo()
  ERROR: UndefKeywordError: keyword argument b not assigned
  Stacktrace:
  [...]
```

src/Utilities/HyperParameters/src/HyperParameters.jl
charleskawczynski: Yes, good point

charleskawczynski:
> Would @kwdef struct make more sense for parameters than a Dict?

I like this. Can you show an example of how we can merge an array of them together?

jkozdon:
> I like this. Can you show an example of how we can merge an array of them together?

Not sure I've seen this. Sorry.

charleskawczynski:
@jkozdon what advantages does the `@kwdef struct` have over `Dict`? I figured it would be nice to separate lists of parameters into functions for the ocean/land/atmos modules. But if this module isn't going to grow super large then I suppose we could all just share one big function and return a `@kwdef struct`. @simonbyrne any thoughts?

jkozdon:
Default values is what I was thinking. I think that you can splat a `Dict` or `NamedTuple` into a `@kwdef struct` constructor.

The advantages I see are:
- all the possible parameters are defined in one place
- you get checking to make sure that a user has set the right parameter as opposed to something random (or a typo)
- you get docs for the parameters if you add docs to the structs
- Consistent typing can be ensured if that's required

Just something to consider as we build things out. Might not be the right thing now, but something to keep in mind.

charleskawczynski:
Great points. I'll start converting. Should we make this struct mutable so that the DA/ML team can easily update it?

tapios:
Looks good. But this is about parameters, not hyperparameters (which are usually understood to be, e.g., parameters in a prior distribution, rather than model parameters as in our case). So I suggest to call it, e.g., the Parameters module.

charleskawczynski:
@tapios it seems that there are many different types of parameters, can we call these `LearnedParameters`? This way

 - `PlanetParameters` ∪ `LearnedParameters` ∈ `ModelParameters`

and perhaps, if there are more constant parameters,

 - `PlanetParameters` ∈ `ConstantParameters`

Then, `HyperParameters` can be saved for the DA/ML modules.

tapios:
> @tapios it seems that there are many different types of parameters, can we call these `LearnedParameters`? This way
> 
> * `PlanetParameters` ∪ `LearnedParameters` ∈ `ModelParameters`
> 
> and perhaps, if there are more constant parameters,
> 
> * `PlanetParameters` ∈ `ConstantParameters`
> 
> Then, `HyperParameters` can be saved for the DA/ML modules.

`LearnedParameters` is strange (they may not all be learned). If it needs a modifier, call it, e.g., `AdjustableParameters` or `FreeParameters`.

glwagner:
+1 for `FreeParameters`.

I like `PlanetConstants` too!

glwagner:
Just my two cents, after talking with @charleskawczynski and @simonbyrne --- an idea:

I like the idea of letting free parameters self-organize in an appropriate way within the various components of clima. In other words, if a `ClimateModel` is something like

```julia
struct ClimateModel{A, O, I, L}
    atmos :: A
    ocean :: O
    ice   :: I
    land  :: L
end
```
then the total free parameters of `ClimateModel` are embedded within each component that they pertain to. 

One reason this makes sense is because the free parameters of the climate model may actually depend on the type of the sub-component (e.g. `typeof(atmos) = A = AnelasticModel` versus `typeof(atmos) = FullyCompressibleModel`. Even better, the parameterizations themselves may be type parameters of a component, such that

```julia
struct AtmosModel{Turb}
   turbulence_parameterization :: Turb
end
```
If `typeof(turbulence_parameterization) = EDMF`, this implies a particular set of free parameters for the atmospheric component. But if `typeof(turbulence_parameterization) = Siebesma2007` (or something...), this implies another set.

An advantage of this approach is that it is modular and democratic. The names and number of free parameters are determined bottom-up, rather than top-down.

If `FreeParameters` is implemented as an abstract type, and if developers of model components are required to implement their free parameters as subtypes of `FreeParameters`, we do not lose the ability to identify or abstract the parameters associated with a particular instance of `ClimateModel` (for example, if we wish to optimize over all `ClimateModel` parameters. For example:

```julia
abstract type FreeParameters end 

function add_free_parameters!(params, obj)
    for name in propertynames(obj)
        fld = getproperty(obj, name)
        if typeof(fld) <: FreeParameters
            for param in propertynames(fld)
                push!(params, getproperty(fld, param))
            end 
        else
            add_free_parameters!(params, fld)
        end 
    end 
    return nothing
end

function free_parameters(model)
    params = []
    add_free_parameters!(params, model)
    return params
end

struct SomeParams <: FreeParameters
    a :: Float64
    b :: Float64
end

struct MoreParams <: FreeParameters
    c   
    d   
end

struct MetaParams
    s :: SomeParams
end

struct Model
    more :: MoreParams
    meta :: MetaParams
end

s = SomeParams(1, 2)
m = MoreParams("hi", "bye")
model = Model(m, MetaParams(s))

@show params = free_parameters(model)
```

which gives me

```julia
julia> include("free_parameters_idea.jl")
params = free_parameters(model) = Any["hi", "bye", 1.0, 2.0]
4-element Array{Any,1}:
  "hi" 
  "bye"
 1.0   
 2.0   
```

In other words, we can write functions that extract all values (and their names, if we want) inside `structs` that subtype `FreeParameters`.

We may be able to also invert this procedure to assign updated/perturbed parameters if we, for example, require implementations of `FreeParameters` to define a keyword-argument constructor (or something like that --- perhaps another way is to imbue (mutable) subtypes of `FreeParameters` with a dictionary-like interface... ?) 

This is just one implementation of what I imagine are many possible solutions. The main point is that high-level functionality is easy to design even within this modular design in which `FreeParameters` structs are arbitrarily distributed among the sub-components of the climate model.

Julia gurus may do better than what I have sketched. In particular, the functions should probably be modified to return a dict of (name, value) pairs (we could even implement a convention system within the function to label parameters by the module they belong to) rather than just an array of unnamed values.

charleskawczynski:
> free parameters of `ClimateModel` are embedded within each component

I like this idea. It supports encapsulation. All we'd have to do is fuse these parameter lists from different components into something that has vector-like behavior. I don't think that there are any major design obstacles here.

It would be nice if we can provide a nice framework to support all of the desired behaviors and interface in `Utilities/FreeParameters`, and the implementation/initialization can live in each component.

glwagner:
It seems like we might want something similar to what's implemented in https://github.com/mauro3/Parameters.jl.

Constructors for `FreeParameters` might then build vector-like `structs` of `FreeParameter` that provide a keyword argument for each parameter, and perhaps `assert` statements to enforce the range of the parameters.

Then we can imagine having syntax like

```julia
@free_parameters struct TurbulentConvection{T}
    a :: T = 1.0, (0.5, 1.5)
    n :: Int = 1, (0, 100)
end
```

that generates both the struct and a constructor, and perhaps also `property` methods that allow us to simply write

```julia
tc = TurbulentConvection()
tc.a # returns 1.0
```

something like
```julia
getproperty(p::FreeParameters, fld) = getproperty(getproperty(p, fld), value)
```

charleskawczynski:
Closing, based on recent discussion. Further development of this work was moved to [FreeParameters.jl](https://github.com/climate-machine/FreeParameters.jl)

PR #187, opened 2019-04-17
Adding boundary conditions to balance law solver

jkozdon:


src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -134,12 +139,35 @@ where
...
 
+If `grid.topology` has a boundary then the function
+`inviscid_numerical_boundary_flux!` must be specified. This function is called
+with the data from the neighbouring DOF as
+```
+inviscid_numerical_boundary_flux!(F, nM, QM, auxM, QP, auxP, bctype, t)
+```
+where
+- `F` is an `MArray` of size `(dim, length_state_vector)` to be filled with the
+  numerical flux across the face (note that this is uninitialized so user must
```
lcw: "so user" -> "so the user"

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -134,12 +139,35 @@ where
...
+  set to zero is this desired)
```
lcw: "set to zero is this desired" -> "set to zero if this is desired"

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -134,12 +139,35 @@ where
...
+- `nM` is the unit outward normal to the face with respect to the minus side
+  (`MVector` of length `3`)
+- `QM` and `QP` are the minus and plus side states (`MArray`)
+- `auxM` and `auxP` are the auxiliary states (`MArray`)
+- `bctype` is the boundary condition flag for the connected face and element of
+   `grid.elemtobndy`
+- `t` is the current simulation time
+Note: `QP` and `auxP` are filled with values reflectes in `grid.vmapP`, and
```
lcw: Not sure what "values reflectes" is trying to say

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -134,12 +139,35 @@ where
...
+these values may be modified by the calling function.
+
+Warning: Modifications to `nM`, `QM`, or `auxM` side effects and should not be
```
lcw: Should "to `nM`, `QM`, or `auxM` side effects" be "to `nM`, `QM`, or `auxM` cause side effects"?

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -166,14 +194,18 @@ stored in the auxiliary state.
 
 !!! todo
 
-    - Add support for boundary conditions
     - support viscous fluxes (`gradstates` is in the argument list as part of
       this future interface)
+    - Revisit how to handle plus side state in
+      `inviscid_numerical_boundary_flux!` after gradient state is handled (how
+      or not propagate changes without adverse side effects).
```
lcw: "or not propagate" -> "to or not propagate"?

src/DGmethods/NumericalFluxes.jl
```diff
@@ -66,4 +72,38 @@ function rusanov!(F::MArray{Tuple{nstate}}, nM,
   end
 end
 
+"""
+    rusanov_boundary_flux!(F::MArray{Tuple{nstate}}, nM, QM, auxM, QP, auxP,
+                           bctype, t, flux!, bcstate!, wavespeed,
+                           preflux = (_...) -> (), correctQ! = nothing
+                           ) where {nstate}
+
+The function `bcstate!` is used to calculate the plus side state for the boundary condition `bctype`. The calling convention is:
```
lcw: wrap line?

src/DGmethods/NumericalFluxes.jl
```diff
@@ -66,4 +72,38 @@ function rusanov!(F::MArray{Tuple{nstate}}, nM,
...
+```
+PP = bcstate!(QP, auxP, QM, auxM, bctype, t, preflux(QM, auxM, t)...)
+```
+where `QP` and `auxP` are the plus side state and auxiliary state to be filled
+from the given data; other arguments should not be modified.
+
+The function `bcstate!` should return either `preflux(QP, auxP, t)` or
+`nothing`. If `nothing` is returned then `preflux(QP, auxP, t)` will be called
+otherwise after the function returns. The reason for this behaviour is to allow
```
lcw: "will be called otherwise" is awkward

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: Is this more clear?
```
Note: `QP` and `auxP` are filled with values based on degrees of freedom
referenced in `grid.vmapP`; `QP` and `auxP` may be modified by the calling
function.```

File: src/DGmethods/DGBalanceLawDiscretizations.jl (hunk 4 above)
Comment (jkozdon) [reply src/DGmethods/DGBalanceLawDiscretizations.jl]: yes

File: src/DGmethods/DGBalanceLawDiscretizations.jl (hunk 1 above)
Comment (jkozdon) [reply src/DGmethods/DGBalanceLawDiscretizations.jl]: done

File: src/DGmethods/DGBalanceLawDiscretizations.jl (hunk 2 above)
Comment (jkozdon) [reply src/DGmethods/DGBalanceLawDiscretizations.jl]: done

File: src/DGmethods/DGBalanceLawDiscretizations.jl (hunk 5 above)
Comment (jkozdon) [reply src/DGmethods/DGBalanceLawDiscretizations.jl]: done

File: src/DGmethods/NumericalFluxes.jl (hunk 6 above)
Comment (jkozdon) [reply src/DGmethods/NumericalFluxes.jl]: thanks

File: src/DGmethods/NumericalFluxes.jl (hunk 7 above)
Comment (jkozdon) [reply src/DGmethods/NumericalFluxes.jl]: Better?
```
The function `bcstate!` should return either `preflux(QP, auxP, t)` or
`nothing`; if `nothing` is returned then `preflux(QP, auxP, t)` is called by
`rusanov_boundary_flux!`. The reason for this behaviour is to allow the user to
avoid redoing expensive calculations.
```

lcw:
I think I pre approved your changes (not sure how to take it back).  Once the docs are updated as you suggested then I approve!

Issue #176, opened 2019-04-12
Move components from Canary


simonbyrne:
Tracking issue for things which need to be moved from [Canary.jl](https://github.com/climate-machine/Canary.jl):
- [x] `Canary.brickmesh`
- [x] `Canary.partition`
- [x] `Canary.connectmesh`
- [x] `Canary.linearpartition`
- [x] `Canary.lglpoints`
- [x] `Canary.spectralderivative`
- [x] `Canary.creategrid!`
- [x] `Canary.computemetric!`

PR #191, opened 2019-04-19
Start of tutorial style balance law examples

jkozdon:
I've started a tutorial style example framework for the new balance law solver. The hope is that the functionality is best learned by doing through self documented examples. The first shows how to do periodic advection in order to introduce the concepts.

Since these take a bit of effort to produce I wanted to gauge interest and value before diving deeper.

Right now the only example is constant coefficient advection in a periodic box. Other planned examples are:

- Using callback functions for the ODE solver
- Adding auxiliary data to degrees of freedom
- Handling boundary conditions
- Adding source terms
- Warping a mesh
- Using the spherical shell mesh
- Adding viscous terms (functionality not in solver yet)

I'm also planning/hoping to produce a document on the math behind the method (at least at a high level).

You can either build the tutorial yourself:
```
julia --project=docs -e 'using Pkg; Pkg.instantiate()'
julia --project=docs -e 'include("docs/make.jl")'
open docs/build/examples/DGmethods/generated/ex_001_periodic_advection.jl
```
or can view then [here](http://faculty.nps.edu/jekozdon/CLIMA_site/examples/DGmethods/generated/ex_001_periodic_advection.html)

If you have comments or correction feel free to make them directly to the pull request. If it's a minor typo fix this can also be done by the pull request files (https://github.com/climate-machine/CLIMA/pull/191/files), selecting the line you want to correct, and pressing the "insert a suggestion".

cc: @lcw


jkozdon:
I should note that is inspired by [Deal.ii](https://dealii.org) [Tutorials](https://dealii.org/developer/doxygen/deal.II/Tutorial.html) and [JuAFEM.jl](https://github.com/KristofferC/JuAFEM.jl) [Examples](http://kristofferc.github.io/JuAFEM.jl/dev/examples/generated/heat_equation/)

src/DGmethods/examples/ex_001_periodic_advection.jl
```diff
@@ -0,0 +1,438 @@
...
+# This is done in the following function
+function upwindflux!(fs, nM, stateM, auxM, stateP, auxP, t)
+  DFloat = eltype(fs)
+  @inbounds begin
+    ## determine the advection speed and direction
+    un = dot(nM, DFloat.(uvec))
+    qM = stateM[1]
+    qP = stateP[1]
+    ## Determine which state is "upwind" of the minus side
+    fs[1] = un ≥ 0 ? un * qM : un * qP
```
ali-ramadhan: I wonder if `fs[1] = elseif(un ≥ 0, un * qM, un * qP)` would be faster on a GPU.

ali-ramadhan:
I'm finding this extremely useful, especially as the example is explained section by section!

src/DGmethods/examples/ex_001_periodic_advection.jl
jkozdon: I would be surprised.

On the CPU they seem to generate the llvm same code (though I do not know what `spec.select` is):
 
```
julia> function max_ternary(a,b)
       a > b ? a : b
       end
max_ternary (generic function with 1 method)

julia> @code_llvm max_ternary(1,2)

;  @ REPL[30]:2 within `max_ternary'
define i64 @julia_max_ternary_12352(i64, i64) {
top:
; ┌ @ operators.jl:286 within `>'
; │┌ @ int.jl:49 within `<'
    %2 = icmp slt i64 %1, %0
; └└
  %spec.select = select i1 %2, i64 %0, i64 %1
  ret i64 %spec.select
}

julia> function max_ifelse(a,b)
       ifelse(a > b, a, b)
       end
max_ifelse (generic function with 1 method)

julia> @code_llvm max_ifelse(1,2)

;  @ REPL[32]:2 within `max_ifelse'
define i64 @julia_max_ifelse_12354(i64, i64) {
top:
; ┌ @ operators.jl:286 within `>'
; │┌ @ int.jl:49 within `<'
    %2 = icmp slt i64 %1, %0
; └└
  %3 = select i1 %2, i64 %0, i64 %1
  ret i64 %3
}
```

ali-ramadhan: Ah interesting. I think the way it was explained to me was that `ifelse` eagerly evaluates both branches so the GPU doesn't have to branch which might speed things up if the expressions is evaluated many times. But I must confess that I've never really benchmarked the difference.

```
help?> ifelse
search: ifelse

  ifelse(condition::Bool, x, y)

  Return x if condition is true, otherwise return y. This differs from ? or if in that it is an ordinary function, so all the arguments are evaluated first. In some cases, using ifelse instead of
  an if statement can eliminate the branch in generated code and provide higher performance in tight loops.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> ifelse(1 > 2, 1, 2)
  2
```

jkozdon:
> Handling boundary conditions and adding source terms is something I've been wondering about how to do with DG, so definitely looking forward to those examples.

though there is not an example on it, you can see this in the test functions for [boundary conditions](https://github.com/climate-machine/CLIMA/blob/76355268e25cdab65c5366e70bff870048d708ce/src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl#L78-L84) ([kernel implementation](https://github.com/climate-machine/CLIMA/blob/76355268e25cdab65c5366e70bff870048d708ce/src/DGmethods/DGBalanceLawDiscretizations_kernels.jl#L189-L196)) and [source functions](https://github.com/climate-machine/CLIMA/blob/76355268e25cdab65c5366e70bff870048d708ce/src/DGmethods/test/Euler/isentropic_vortex_standalone_source.jl#L69-L80) ([kernel implementation](https://github.com/climate-machine/CLIMA/blob/76355268e25cdab65c5366e70bff870048d708ce/src/DGmethods/DGBalanceLawDiscretizations_kernels.jl#L70-L71))

jkozdon:
added examples of ODE solver callbacks

src/DGmethods/examples/ex_001_periodic_advection.jl
```diff
@@ -0,0 +1,438 @@
...
+
+  # (end of function)
+end
+#md nothing # hide
+
+#------------------------------------------------------------------------------
+
+# ### Initializing and run the DG method
+# Note: This whole code chunk is in a `let` block
+let
```
ali-ramadhan: Out of curiosity, why run the DG method in a `let` block?

jkozdon: I just like to protect my repl-scope

src/DGmethods/examples/ex_001_periodic_advection.jl
```diff
@@ -0,0 +1,563 @@
...
+
+  # The topology only has element connectivity and corner information, thus we
+  # still need to create a grid (or mesh) of degrees of freedom. In CLIMA the
+  # so-called discontinuous spectral element grid is used (aka tensor-product
+  # quadrilateral and hexahedral elements with Legendre-Gauss-Lobatto
+  # interpolation and quadrature weights). Given a topology and polynomial order
+  # we can create the grid of degrees of free using
+  grid = DiscontinuousSpectralElementGrid(topology; polynomialorder =
+                                          polynomialorder, FloatType = DFloat,
+  DeviceArray = ArrayType,)
```
charleskawczynski:
```suggestion
                                          DeviceArray = ArrayType,)
```

src/DGmethods/examples/ex_001_periodic_advection.jl
jkozdon: Not sure if there is a performance impact too with function closure?

src/DGmethods/examples/ex_001_periodic_advection.jl
jkozdon: Interesting. I tried to dig into the julia source to see how there were implemented, but didn't get very far. Might be a question for @vchuravy

src/DGmethods/examples/ex_001_periodic_advection.jl
```diff
@@ -0,0 +1,563 @@
...
+# and `Ne` x `Ne` x `Ne` elements when `dim == 3`. The floating point type of
+# the computation is `DFloat` and whether the CPU or GPU is used is determined
+# by `ArrayType`; `ArrayType === Array` is for the CPU and `ArrayType ===
+# CuArray` is for NVIDIA GPUs.
+#
+# Note: This whole code chunk is in a function block
+function setupDG(mpicomm, dim, Ne, polynomialorder, DFloat=Float64,
+                 ArrayType=Array)
+
+  @assert ArrayType === Array
```
ali-ramadhan: Hit this line when trying to run
```julia
spatialdiscretization = setupDG(mpicomm, dim, Ne, polynomialorder, Float64, CuArray)
```
so I'm guessing you can't run this example on a GPU just yet.

ali-ramadhan:
Was able to run the example and get the exact same output. Does MPI do anything if there's only a single rank?

src/DGmethods/examples/ex_001_periodic_advection.jl
lcw: That is right.  GPU support is coming.

lcw:
> Was able to run the example and get the exact same output :) Does MPI do anything if there's only a single rank?

Right now we need MPI even in the single node case.  Once #176 is done we can think about removing the hard dependence on MPI #172.

tapios:
Thanks, @jkozdon. This is very helpful. I realize that producing such detailed documentation takes time, but I think we all benefit. I'd encourage to add the other examples you had in mind too.

jkozdon:
I've added a second example on using the auxiliary state and boundary conditions

jkozdon:
cc @lcw

lcw:
@simonbyrne Is this okay to merge if it passes CI?

jkozdon:
> This looks good. I think that the examples should also be run as tests so that they are run on multiple platforms.

@lcw & @simonbyrne: I agree. If we want to ensure our docs remain relevant, I like having multi-platform testing of the examples we publish. Now we have the problem with travis timing out again though. Thoughts? We could explore the split travis build.

jkozdon:
Do not merge until fixed with current master following #204

jkozdon:
@lcw

PR #195, opened 2019-04-22
Update README.md

aramirezreyes:
Updated location of test cases from drivers to test in the README


jkozdon:
Thanks! Only other suggestion would be to remove the GPU references (since this does not currently work with the code).

I can do this if you want.

jkozdon:
Thanks for contributing @aramirezreyes! I've removed the last bit of GPU code from the README.md, and once the CI passes it should be merged.

PR #196, opened 2019-04-23
[MoistThermo] use PhasePartition object

simonbyrne:
As discussed in #147.

### DIFF
--- a/docs/src/Utilities/MoistThermodynamics.md
+++ b/docs/src/Utilities/MoistThermodynamics.md
@@ -28,47 +28,46 @@ There are several types of functions:
 7. Auxiliary functions for diagnostic purposes, e.g., other thermodynamic quantities
     * `liquid_ice_pottemp` (liquid-ice potential temperature)
 
-A moist dynamical core that assumes equilibrium thermodynamics can be obtained from a dry dynamical core with total energy as a prognostic variable by including a tracer for the total specific humidity `q_tot`, using the functions, e.g., for the energies in the module, and computing the temperature `T` and the liquid and ice specific humidities `q_liq` and `q_ice` from the internal energy `e_int` by saturation adjustment:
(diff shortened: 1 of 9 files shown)


docs/src/Utilities/MoistThermodynamics.md
```diff
@@ -28,47 +28,46 @@ There are several types of functions:
...
+geopotential = grav * z
+q_tot          = ...
+ρ            = ...
 
-    # initialize
-    geopotential = grav * z
-    q_tot          = ...
-    ρ            = ...
+(u, v, w)    = ...
+e_kin           = 0.5 * (u.^2 .+ v.^2 .+ w.^2)
```
charleskawczynski:
```suggestion
e_kin           = 0.5 * (u^2 + v^2 + w^2)
```
I think this is typically done element-wise, should we remove the `.`?

docs/src/Utilities/MoistThermodynamics.md
```diff
@@ -28,47 +28,46 @@ There are several types of functions:
...
-    do timestep   # timestepping loop
+  # advance dynamical variables by a timestep (temperature typically
+  # appears in terms on the rhs, such as radiative transfer)
+  advance(u, v, w, ρ, e_tot, q_tot)
 
-      # advance dynamical variables by a timestep (temperature typically
-      # appears in terms on the rhs, such as radiative transfer)
-      advance(u, v, w, ρ, e_tot, q_tot)
+  # compute internal energy from dynamic variables
+  e_int = e_tot - 0.5 * (u.^2 .+ v.^2 .+ w.^2) - geopotential
```
charleskawczynski:
```suggestion
  e_int = e_tot - 0.5 * (u^2 + v^2 + w^2) - geopotential
```
I think this is typically done element-wise, should we remove the `.`?

simonbyrne:
> Still looking through this, but I think this makes sense. It protects against passing in `q_tot`, `q_liq` and `q_ice` in the wrong order, which is really nice. My only immediate suggestion might be to change the name of the `q` constructor to avoid TitleCase and `_`'s:
> 
> `PhasePartition_equil` -> `PhasePartitionEquil`

Yeah, it's not ideal, I agree. I'm not that keen `PhasePartitionEquil` as it doesn't return an object of that name.

charleskawczynski:
> I'm not that keen PhasePartitionEquil as it doesn't return an object of that name.

True. I suppose that you could make it an outer constructor of `PhasePartition`, and use keyword args, but then the user needs to remember the default behavior. I'm fine with whatever you prefer.

PR #199, opened 2019-04-24
Allowing the user to wait on the MPI Isends as well as Irecvs

jkozdon:
Hoping that this will address issues seen in https://github.com/climate-machine/CLIMA/issues/194#issuecomment-486006248.

@simonbyrne Can you give this a try?

@lcw You think this is reasonable?


simonbyrne:
That did seem to fix it.

jkozdon:
> That did seem to fix it.

Great. Assuming the CI passes, I'll merge the change.

PR #203, opened 2019-04-29
[BL] Adding viscous/gradient terms

jkozdon:
Adds ability to have gradient based terms in the balance law solver. 

Before I open this up for merge I need to add one more test of the boundary conditions, but wanted people to have some sense for what the interface will be in case there are suggestions or requests.

cc: @lcw


jkozdon:
@lcw do you have time to look at this today? @fxgiraldo wants to use this stuff.

PR #204, opened 2019-04-29
[BL] Refactor of inviscid in prep for viscous

jkozdon:
Several things happening here:
- inviscid_X -> X (since function will be inviscid + some viscous). All
  these function now get passed the viscous state as well
- Qgrad -> QV (for Q viscous flux state)
- adding some more callbacks and sizes needed for viscous handling
  (nviscstate, nviscfluxstate, viscous_transform!, viscous_flux!,
  viscous_numerical_flux!, viscous_boundary_numerical_flux!)


jkozdon:
cc: @lcw

jkozdon:
spinning this commit off from #203 so it can be merged in first (since it impacts #191

PR #206, opened 2019-04-30
Add test for conservation

jkozdon:
Main purpose of this is to test conservation of the code. To do this we do advection on a sphere with a random velocity field and initial conditions.

Also added a `weightedsum` routine for the `MPIStateArray`. For accuracy I used a `BigFloat` to accumulate, but this should be revisited. It is really a map-reduce operation over several iterators, so maybe something can be done there when we think about the GPU.

Also contains a  bug fix for the `CubedShellTopology` when multiple MPI ranks are used (connected `elementocoord` was not being used, but only the local version).

#204 should be sorted out / merged before this PR

cc @lcw


jkozdon:
Switched over to using `DoubleFloats` in the `weightedsum` accumulation

cc: @lcw

src/Arrays/MPIStateArrays.jl
```diff
@@ -355,7 +356,48 @@ function knl_L2dist(::Val{Np}, A, B, weights, elems) where {Np}
...
+end
+
+function knl_weightedsum(::Val{Np}, A, weights, elems, states) where {Np}
+  DFloat = eltype(A)
+
+  wsum = zero(DoubleFloat{DFloat})
+
+  @inbounds for e = elems
+    for q = states, i = 1:Np
+      wsum += weights[i, e] * A[i, q, e]
```
simonbyrne: should we do the multiplication in extended precision as well?

PR #209, opened 2019-05-03
[BL] Minor tweaks/cleanups

jkozdon:
A few minor tweaks to the Balance law framework based on conversations with @lcw and @fxgiraldo


jkozdon:
bors try

jkozdon:
@lcw updated this to only run single rank jobs on travis. Not ideal, but travis just keeps timing out. Thoughts?

jkozdon:
@lcw: last week we talked about what is the bottleneck in testing, is it the solver time or the julia initialization time. I ran on my make a time test:
```
@time include("isentropic_vortex_standalone.jl")
```
in a fresh julia session. The first time through I get:
```
51.729335 seconds (91.74 M allocations: 4.427 GiB, 4.53% gc time)
```
and the second time through I get
```
2.403413 seconds (5.50 M allocations: 281.082 MiB, 4.57% gc time)
```
So the bottleneck is not the solver time itself (though this has not been optimized yet), but the julia compile time.

PR #219, opened 2019-05-09
use devnull and not NullLogger for non-root MPI logging

jkozdon:
When `NullLogger` is used for non-root logging, MPI calls in log records are optimized away resulting in deadlock. Using `devnull` avoids this problem.

This should fix some of the recent test timeouts


jkozdon:
bors try

jkozdon:
bors try

Issue #160, opened 2019-04-09
Balance Law solver checklist


jkozdon:
- [x] source terms
- [x] constant auxiliary state
- [x] documentation
- [x] boundary conditions
- [x] gradient terms (pr #203)
- [x] GPUify
- [x] Tutorial style docs (like done with [JuAFEM.jl] (http://kristofferc.github.io/JuAFEM.jl/dev/examples/generated/heat_equation/) examples) (pr #191)
- [ ] compare performance with current dycore
- [ ] Better docs for tests
- [ ] Add `Float32` tests
- [ ] pinned memory and multiple GPU queues
- [ ] update interface to allow for separation of tracer/advected state and nonlinear terms
- [ ] IMEX interface

Tabled items:
- switch to skew formulation (as opposed to weak form)
  - This is something we will likely want to think about in the future, but not high priority right now
- dynamic auxiliary state
  - if needed, this would be better stored at the end of the solution vector and updated during the ODE update.
  - To do this will requiring modifying the `MPIStateArray` so that the norm is only performed over "real" states

simonbyrne:
- [ ] documentation

PR #220, opened 2019-05-09
GPUify the balance law solver

jkozdon:
@lcw and I have worked to GPUify the whole balance law solver. This should be a non-breaking update to anyone using the balance law solver on the CPU only.

Major hat tip to @vchuravy and @maleadt for all their help debugging and handling issues related to the Julia GPU interface!


src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
```diff
@@ -252,24 +268,28 @@ let
...
-          err[l] = run(mpicomm, dim, Ne, polynomialorder, timeend, DFloat, dt)
-          @test err[l] ≈ DFloat(expected_error[dim-1, l])
-        end
-        @info begin
-          msg = ""
-          for l = 1:lvls-1
-            rate = log2(err[l]) - log2(err[l+1])
-            msg *= @sprintf("\n  rate for level %d = %e\n", l, rate)
+    for ArrayType in ArrayTypes
+      for DFloat in (Float64,) #Float32)
```
charleskawczynski: Does the `Float32` case work?

src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
```diff
@@ -281,17 +301,20 @@ let
...
-        engf_eng0 = run(mpicomm, dim, numelem[1:dim], polynomialorder, timeend,
-                        DFloat, dt)
-        @test check_engf_eng0[dim, MPI.Comm_size(mpicomm), DFloat] ≈ engf_eng0
+    check_engf_eng0[2, 1, Float64] = 9.9999920444514001e-01
+    check_engf_eng0[3, 1, Float64] = 9.9999530983481677e-01
+    check_engf_eng0[2, 3, Float64] = 9.9999839879889008e-01
+    check_engf_eng0[3, 3, Float64] = 9.9999507253572062e-01
+
+    for ArrayType in ArrayTypes
+      for DFloat in (Float64,) #Float32)
```
charleskawczynski: Does the `Float32` case work?

src/DGmethods/test/Euler/isentropic_vortex_standalone_source.jl
```diff
@@ -263,24 +277,28 @@ let
...
-          err[l] = run(mpicomm, dim, Ne, polynomialorder, timeend, DFloat, dt)
-          @test err[l] ≈ DFloat(expected_error[dim-1, l])
-        end
-        @info begin
-          msg = ""
-          for l = 1:lvls-1
-            rate = log2(err[l]) - log2(err[l+1])
-            msg *= @sprintf("\n  rate for level %d = %e\n", l, rate)
+    for ArrayType in ArrayTypes
+      for DFloat in (Float64,) #Float32)
```
charleskawczynski: Does the `Float32` case work?

charleskawczynski:
I'm so pumped about this! Can GPUifyLoops be isolated from within the balance law framework?

src/DGmethods/DGBalanceLawDiscretizations_kernels.jl
```diff
@@ -1,6 +1,20 @@
+using Requires
+@init @require CUDAnative = "be33ccc6-a3ff-5ff2-a52e-74243cff1e17" begin
+  using .CUDAnative
```
vchuravy: Is this still required?

src/DGmethods/test/Euler/isentropic_vortex_standalone.jl
```diff
@@ -217,6 +223,9 @@ let
   ll == "ERROR" ? Logging.Error : Logging.Info
   logger_stream = MPI.Comm_rank(mpicomm) == 0 ? stderr : devnull
   global_logger(ConsoleLogger(logger_stream, loglevel))
+  @static if Base.find_package("CUDAnative") !== nothing
+    device!(MPI.Comm_rank(mpicomm) % length(devices()))
```
vchuravy: We should probably allow this to come from the scheduler.

src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
jkozdon: In theory yes, but we had some problems early on. We plan to revisit this soon.

jkozdon: Problems being large background state on a test we no longer have.

src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
jkozdon: ditto.

src/DGmethods/test/Euler/isentropic_vortex_standalone_source.jl
jkozdon: ditto

jkozdon:
> I'm so pumped about this! Can GPUifyLoops be isolated from within the balance law framework?

What do you mean?

src/DGmethods/test/Euler/isentropic_vortex_standalone.jl
jkozdon: This certainly isn't a long term solution.

In my experience the scheduler should set the devices you can see on a node so you are at least protected there. Not sure the scheduler sets the node each MPI rank can see, and what about machines that you run on that don't have schedulers?

This also has the problem that it assumes that all nodes have the same numbers of GPUs, that the number of ranks on a node is the same as the number of GPUs, that mpiranks are consecutive across GPUS, etc..

In the past we've used an all gather on host name then assigned GPUs that way with the assumption that number of GPUs on a node is the same as the number of ranks.

Better solutions are certainly welcome, but this should work in the near-term I think(?).

@lcw probably has thoughts too.

src/DGmethods/DGBalanceLawDiscretizations_kernels.jl
jkozdon: Yes it is because otherwise `threadIdx` is undefined.

vchuravy: Ah! Makes sense

src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
charleskawczynski: Ok, sounds good.

jkozdon:
> Great to see this! I assume we need a tag for GPUifyLoops, let me know when I should do that.

Right now we just depend on the current master, not sure what other functionality is needed once the moisture stuff is in.

Also, https://github.com/JuliaGPU/CUDAnative.jl/commit/4a079b69468b07ddb9d4f7ea145128d44d073bd2 bit me today! @lcw reminded me to always check master on the GPU branches when stuff is breaking weirdly. Had fun playing with [Cthulhu](https://github.com/JuliaDebug/Cthulhu.jl) though before jumping on CUDAnative master.

charleskawczynski:
> What do you mean?

I suppose, concretely, will GPUifyLoops be needed in the drivers? e.g., will a user need to use @launch statements into functions supplied to the balance law solver?

src/DGmethods/test/Euler/isentropic_vortex_standalone_bc.jl
jkozdon: Features / requests are being tracked in the BL issue #160

jkozdon:
> I suppose, concretely, will GPUifyLoops be needed in the drivers? e.g., will a user need to use @launch statements into functions supplied to the balance law solver?

Not most of the time. If a user has a fancy initialization they want to they can use it, but for the most part this probably won't be needed. None of the tests currently include GPUifyLoops explicitly

jkozdon:
Also, I should add that though I am the commiter on all the commits in this pull request none of this would have been possible without the tireless efforts over the past weeks of @lcw finding bugs and gotchas. @lcw along with @vchuravy and @maleadt have handled a ton of issues that we ran into, and certainly we'd be months away from this without them.

All three of these guys deserve the lionshare of the credit for making this happen.

jkozdon:
@lcw found the problem with the square root. It was a bug in the test. Not sure why it only hit on the mac with 1.1 (reading a garbage value in the source term)

PR #221, opened 2019-05-09
Remove allow_failure on nightly with gitlab

jkozdon:
Since we currently pass on the gitlab nightly CI, I recommend removing the allow_failures

cc: @lcw @simonbyrne @charleskawczynski @vchuravy

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,13 +33,11 @@ gpu:test:dev:
   extends: .gputest
   variables:
     CI_VERSION_TAG: 'dev'
-  allow_failure: true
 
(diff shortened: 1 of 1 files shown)


lcw:
I think we should merge this change because #220 only runs the GPUified code on nightly because we require changes in 1.2.

charleskawczynski:
I definitely like this, but how stable are the nightly developments? If nobody thinks it will be an issue, then I'm all for it.

charleskawczynski:
> I think we should merge this change because #220 only runs the GPUified code on nightly because we require changes in 1.2.

Makes sense.

PR #222, opened 2019-05-10
[BL] Allowing user to also write out auxstate to vtk

jkozdon:


jkozdon:
@fxgiraldo This should be what you need.

PR #223, opened 2019-05-10
... -> .. ?

trontrytel:
I was trying to load CLIMA.jl from within tests in DG methods from within Julia repl. I was getting errors about Julia not finding certain modules. It helped to change the paths in the using statement like this. Is that correct?


trontrytel:
bors try

PR #229, opened 2019-05-15
[BL] Simple pvtu handling

jkozdon:


simonbyrne:
Should this perhaps go somewhere else?

jkozdon:
> Should this perhaps go somewhere else?

Maybe the existing `Mesh/vtk.jl` file? Right now this file is just included in `DGBalanceLawDiscretization.jl`. Probably not the best approach overall, though it works...

simonbyrne:
Yes, cross-including files seems like a bad idea. Can we move the vtk/pvtu stuff to its own module?

jkozdon:
That's fine with me. I'd also thought perhaps having the VTK functionality lives as functions of the Grid module. Given a grid and an MPIArray we can write VTK files. I just haven't had a chance to do this yet. Maybe for now I will add it to `vtk.jl`, and we can move stuff around later? Just thinking that anyone running in parallel will want pvtu sooner rather than later(?).

test/DGmethods/Euler/isentropic_vortex_standalone.jl
```diff
@@ -175,7 +175,13 @@ function main(mpicomm, DFloat, topl::AbstractTopology{dim}, N, timeend,
     outprefix = @sprintf("vtk/isentropicvortex_%dD_mpirank%04d_step%04d",
                          dim, MPI.Comm_rank(mpicomm), step[1])
     @debug "doing VTK output" outprefix
-    writevtk(outprefix, Q, spacedisc, statenames)
+    DGBalanceLawDiscretizations.writevtk(outprefix, Q, spacedisc, statenames)
```
charleskawczynski: Can we remove `DGBalanceLawDiscretizations.`?

simonbyrne: yes, that should be removed

PR #230, opened 2019-05-16
rearrange directories

simonbyrne:
- move tests to tests
- get rid of intermediate src directories


.gitlab-ci.yml
```diff
@@ -33,6 +33,7 @@ gpu:test:dev:
   extends: .gputest
   variables:
     CI_VERSION_TAG: 'dev'
+  allow_failure: true
```
jkozdon: Just an FYI that this means we are effectively turning off GPU CI for the balance law solver.

charleskawczynski: @simonbyrne why was this necessary?

simonbyrne: See #232

PR #233, opened 2019-05-17
Introduce a DGModel type

vchuravy:
@jkozdon @lcw @simonbyrne and I had a brief discussion how we could tighten up the interface to the BalanceLaw solver, very rough draft of today.

```julia
module DGModels
# abstract type DGModel{Dim, NStates, ...} end
abstract type DGModel{Dim} end

numerical_flux!(::DGModel, args...) = error("Need a numerical flux")
numerical_boundary_flux!(::DGModel, args...) = error("Need a numerical boundary flux")

# ...

# Optional callback
viscous_transform(::DGModel, args...) = nothing

end # module

# define user models
# only 3-dim supported
@model MyModel <: DGModel{3}

# @model MyModel{Dim} <: DGModel{Dim}

# should expand to
import DGModels: numerical_flux!, viscous_transform, ...
struct MyModel <: DGModel{3} end

## 
# Users can then define
##

numerical_flux!(::MyModel, args...) = ...
viscous_transform(::MyModel, args...) = ...

##
# We can define a library of functionality
##

best_flux(args...) = ...
numerical_flux(::MyModel, args...) = best_flux(args...)

# Benefits:
# - reuse
# - improved debugability
# - the solver can extend with more callbacks
# - the model can carry the type parameters, reducing the need for `Val` arguments
# - hierarchy of models MyModel <: IncompressibleModel <: DGModel
```


simonbyrne:
After some discussion with @charleskawczynski, @asraero and @trontrytel, we decided that it would be very useful to be able to declare internal model variables that could be accessed via `getfield` overloading instead of e.g. `ρ == Q[_ρ]`.

Maybe something like:
```
@variables MyModel begin
    @constant begin
        # time-constant variables, e.g. coordinates, reference values, etc.
        # currently called "aux" 
        x,y,z
    end

    @diagnostic begin
         # currently computed via "preflux"
         ql
     end

     @prognostic begin
         # state variables
         U,V,W,ρ,Q,E
      end
end
```

It would also be useful to be able to create compound data structures (e.g. structs or static arrays) from combinations of these.

vchuravy:
Can you give a desugared example? The macro should be implementable, but I am worried about GPU compatibility. Are these static variables for the entire Grid, for each element, or for each nodal point? 

What we might be able to do, is to encapsulate the local view into the global data and have the BalanceLaw solver create a local model instance.

jkozdon:
Let's move this discussion to issue #245 since it's a bit orthogonal to this PR.

src/DGmethods/DGModels.jl
```diff
@@ -0,0 +1,48 @@
...
+        :hasbctype,
+        :nstate,
+        :nviscstate,
+        :ngradstate,
+        :nauxstate,
+        :states_grad)
+    imports = (Expr(:., m) for m in methods)
+
+    Expr(:toplevel,
+        Expr(:import, Expr(Symbol(":"), Expr(:., fullname(@__MODULE__)...), imports...)),
```
simonbyrne: I think we should keep imports out of the macro for now. We can add them in later if need be.

src/DGmethods/DGModels.jl
```diff
@@ -0,0 +1,48 @@
...
+        def)
+end
+
+flux!(::DGModel, args...) = error("")
+numerical_flux!(::DGModel, args...) = error("")
+numerical_boundary_flux!(::DGModel, args...) = error("")
+gradient_transform!(::DGModel, args...) = nothing
+viscous_transform!(::DGModel, args...) = nothing
```
simonbyrne: @tapios suggested diffusive flux

simonbyrne:
One other thing that I was wondering: would it be possible to separate the underlying "mathematical" components (`flux!`, `gradient_transform!`, `viscous_transform!`) from the "numerical" ones (`numerical_flux!`, `numerical_boundary_flux!`, `viscous_penalty!`, `viscous_boundary_penalty!`)?

simonbyrne:
Along those lines: would it make sense to have something like
```julia
abstract type NumericalFlux
end

struct RusanovFlux <: NumericalFlux
 ...
end
```
and then define `numerical_flux!` and `numerical_boundary_flux!` for those? 

We might be able to do something similar for the `viscous_penalty!`/`viscous_boundary_penalty!` (though I am even less clear on how they work.

charleskawczynski:
This appears to have been addressed by #266

PR #236, opened 2019-05-19
Fxg/sphere

fxgiraldo:
I have added an SSP time-integrator and tested it on the Advection on the Sphere test case.  This is the first step towards getting double tableau RK methods.

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -4,7 +4,7 @@ using DoubleFloats
 
 using MPI
 
-export MPIStateArray, euclidean_distance, weightedsum
+export MPIStateArray, euclidean_distance, weightedsum, global_max, global_min, global_extrema_diff
 
 """
     MPIStateArray{S <: Tuple, T, DeviceArray, N,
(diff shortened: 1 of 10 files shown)


jkozdon:
Couple things I see off the bat we should look at before this is merged:

- [ ] Make sure it runs on the GPU
- [ ] Add a test of the time stepping (test exists but does not run)
- [ ] Maybe merge `advection_sphere.jl`, `advection_sphere_lsrk`, and `advection_sphere_ssp`
- [ ] Remove `acoustic_wave` since it's not really part of this PR
- [ ] fix line length and indentation to match CLIMA style guide?

Frank I can help you with all this later today if you want and will also take a closer look at the code.

fxgiraldo:
@jkozdon : the code runs on the GPU. As discussed, have kept each of the advection_sphere cases with time-integrators separate to be able to deploy each time-integrator separately. should probably move these to the examples folder at a later date.  The code runs too slow on the GPU. I have pushed my changes to my branch FXG/SSP.  Maybe I should take back this pull request and open a new one?

jkozdon:
> Maybe I should take back this pull request and open a new one?

No need to. We can just keep updating this request.

Issue #178, opened 2019-04-13
Add more tests to MoistThermodynamics


charleskawczynski:
There are several tests in MoistThermodynamics, but the interface functions are currently only tested for type-stability. It would be nice if we add more tests to the interface.

Also, perhaps we can reproduce some plots of phase diagrams in the docs.

simonbyrne:
Are there published values or plots we can test against?

charleskawczynski:
I'm sure there's something, I'll take a look

trontrytel:
The single_eddy test I'm working on is a good test for many of the MoistThermodynamics components (saturation_adjustment, PhasePartition, etc) Should be fast enough to run as test

charleskawczynski:
Some tests were added in #355 to test the case for `PhaseEquil(q_tot -> 0)` against `PhaseDry`.

charleskawczynski:
Some interface consistency tests were added in #506.

charleskawczynski:
I think Thermodynamics is pretty heavily tested now, so I think we can close or we should better refine this issue. Regarding the plots, @trontrytel are there plots we can compare [these profiles](https://clima.github.io/ClimateMachine.jl/latest/HowToGuides/Common/Thermodynamics/#Moist-Phase-in-thermodynamic-equilibrium) with?

trontrytel:
I agree about closing this issue. 

Nothing comes to my mind about the plots. Do you want to check if they are consistent with some assumptions? Or just have another profile to show that the ones you test for are representative?

charleskawczynski:
We have some test all of the thermo functions with these profiles, so I suppose they're tested for consistency in that way. I think having the doc plots is useful enough just for monitoring. I'll close for now.

PR #237, opened 2019-05-21
Single eddy test

trontrytel:
@lcw Here is my draft pull.

First of all I have no idea why I have changes in ``DGBalanceLaw`` or ``DGmethods`` files. I know I was cherry-picking something from @jkozdon last week. But I don't think this is it. Please let me know if these are not needed and I'll change them to the upstream versions.

Both ``single_eddy_`` tests are set to run just 2 timesteps so that one can profile on the GPU. Right now they dont test anything, they just run:
- The ``single_eddy_test_sat_adj`` is working. 
- The ``single_eddy_test_Kessler`` works on CPU but crashes on the GPU with the error I sent before

My TODO list for now:
- work on the GPU bug (all hints are appreciated)
- add the ``@inline``decorator (?) to see if it helps with my GPU profiling output.
- turn those into actual tests
- add more docs


vchuravy:
There should be no need for the `@inline` annotations for GPU code anymore. What GPU error did you see?

trontrytel:
```ERROR: LoadError: InvalidIRError: compiling #12(Val{2}, Val{4}, Val{6}, Val{0}, Val{2}, typeof(eulerflux!), typeof(source!), CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CuDeviceArray{Float64,3,CUDAnative.AS.Global}, Float64, CuDeviceArray{Float64,2,CUDAnative.AS.Global}, UnitRange{Int64}) resulted in invalid LLVM IR
Reason: unsupported call to the Julia runtime (call to jl_f__apply)
Stacktrace:
 [1] eulerflux! at /central/home/ajaruga/clones/CLIMA/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl:149
 [2] volumerhs! at /central/home/ajaruga/clones/CLIMA/src/DGmethods/DGBalanceLawDiscretizations_kernels.jl:84
 [3] #12 at /home/ajaruga/.julia/packages/GPUifyLoops/fn7x4/src/context.jl:169
Stacktrace:
 [1] check_ir(::CUDAnative.CompilerJob, ::LLVM.Module) at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/compiler/validation.jl:114
 [2] macro expansion at /home/ajaruga/.julia/packages/TimerOutputs/7zSea/src/TimerOutput.jl:216 [inlined]
 [3] #codegen#119(::Bool, ::Bool, ::Bool, ::Bool, ::Bool, ::typeof(CUDAnative.codegen), ::Symbol, ::CUDAnative.CompilerJob) at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/compiler/driver.jl:186
 [4] #codegen at ./none:0 [inlined]
 [5] #compile#118(::Bool, ::Bool, ::Bool, ::Bool, ::Bool, ::typeof(CUDAnative.compile), ::Symbol, ::CUDAnative.CompilerJob) at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/compiler/driver.jl:47
 [6] #compile at ./none:0 [inlined]
 [7] #compile#117 at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/compiler/driver.jl:28 [inlined]
 [8] compile at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/compiler/driver.jl:28 [inlined] (repeats 2 times)
 [9] macro expansion at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/execution.jl:378 [inlined]
 [10] #cufunction#159(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(cufunction), ::getfield(GPUifyLoops, Symbol("##12#13")){typeof(CLIMA.DGBalanceLawDiscretizations.volumerhs!)}, ::Type{Tuple{Val{2},Val{4},Val{6},Val{0},Val{2},typeof(eulerflux!),typeof(source!),CuDeviceArray{Float64,3,CUDAnative.AS.Global},CuDeviceArray{Float64,3,CUDAnative.AS.Global},CuDeviceArray{Float64,3,CUDAnative.AS.Global},CuDeviceArray{Float64,3,CUDAnative.AS.Global},CuDeviceArray{Float64,3,CUDAnative.AS.Global},Float64,CuDeviceArray{Float64,2,CUDAnative.AS.Global},UnitRange{Int64}}}) at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/execution.jl:347
 [11] cufunction(::Function, ::Type) at /home/ajaruga/.julia/packages/CUDAnative/bpN1D/src/execution.jl:347
 [12] macro expansion at /home/ajaruga/.julia/packages/GPUifyLoops/fn7x4/src/GPUifyLoops.jl:113 [inlined]
 [13] macro expansion at ./gcutils.jl:87 [inlined]
 [14] #launch#46(::Base.Iterators.Pairs{Symbol,Any,Tuple{Symbol,Symbol},NamedTuple{(:threads, :blocks),Tuple{Tuple{Int64,Int64,Int64},Int64}}}, ::typeof(GPUifyLoops.launch), ::GPUifyLoops.CUDA, ::typeof(CLIMA.DGBalanceLawDiscretizations.volumerhs!), ::Val{2}, ::Vararg{Any,N} where N) at /home/ajaruga/.julia/packages/GPUifyLoops/fn7x4/src/GPUifyLoops.jl:110
 [15] (::getfield(GPUifyLoops, Symbol("#kw##launch")))(::NamedTuple{(:threads, :blocks),Tuple{Tuple{Int64,Int64,Int64},Int64}}, ::typeof(GPUifyLoops.launch), ::GPUifyLoops.CUDA, ::typeof(CLIMA.DGBalanceLawDiscretizations.volumerhs!), ::Val{2}, ::Vararg{Any,N} where N) at ./none:0
 [16] macro expansion at /home/ajaruga/.julia/packages/GPUifyLoops/fn7x4/src/GPUifyLoops.jl:54 [inlined]
 [17] odefun!(::DGBalanceLaw, ::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::Float64) at /central/home/ajaruga/clones/CLIMA/src/DGmethods/DGBalanceLawDiscretizations.jl:598
 [18] (::getfield(CLIMA.LowStorageRungeKuttaMethod, Symbol("##12#13")){DGBalanceLaw})(::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::Vararg{Any,N} where N) at /central/home/ajaruga/clones/CLIMA/src/ODESolvers/LowStorageRungeKuttaMethod.jl:91
 [19] dostep!(::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::LowStorageRungeKutta{Float64,MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}},5}, ::Int64, ::Bool) at /central/home/ajaruga/clones/CLIMA/src/ODESolvers/LowStorageRungeKuttaMethod.jl:113
 [20] #solve!#1(::Int64, ::Bool, ::Int64, ::Tuple{CLIMA.GenericCallbacks.EveryXWallTimeSeconds,CLIMA.GenericCallbacks.EveryXSimulationSteps}, ::typeof(solve!), ::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::LowStorageRungeKutta{Float64,MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}},5}) at /central/home/ajaruga/clones/CLIMA/src/ODESolvers/ODESolvers.jl:44
 [21] (::getfield(CLIMA.ODESolvers, Symbol("#kw##solve!")))(::NamedTuple{(:timeend, :callbacks),Tuple{Int64,Tuple{CLIMA.GenericCallbacks.EveryXWallTimeSeconds,CLIMA.GenericCallbacks.EveryXSimulationSteps}}}, ::typeof(solve!), ::MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}}, ::LowStorageRungeKutta{Float64,MPIStateArray{Tuple{25,6},Float64,CuArray,3,CuArray{Float64,3},2,CuArray{Int64,1}},5}) at ./none:0
 [22] main(::MPI.Comm, ::Type, ::BrickTopology{2,Float64}, ::Int64, ::Int64, ::Type, ::Float64) at /central/home/ajaruga/clones/CLIMA/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl:314
 [23] run(::Int64, ::Tuple{Int64,Int64}, ::Int64, ::Int64, ::Type) at /central/home/ajaruga/clones/CLIMA/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl:338
 [24] top-level scope at /central/home/ajaruga/clones/CLIMA/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl:352
 [25] include at ./boot.jl:328 [inlined]
 [26] include_relative(::Module, ::String) at ./loading.jl:1094
 [27] include(::Module, ::String) at ./Base.jl:31
 [28] include(::String) at ./client.jl:431
 [29] top-level scope at REPL[1]:1
in expression starting at /central/home/ajaruga/clones/CLIMA/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl:343
```

test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl
```diff
@@ -0,0 +1,358 @@
...
+    #if qt < 0
+    #    @show(qt)
+    #end
+
+  end
+end
+
+
+# physical flux function
+eulerflux!(F, Q, QV, aux, t) = eulerflux!(F, Q, QV, aux, t, preflux(Q)...)
```
vchuravy: I haven't had a chance to look deeply into this, but the error you are seeing (if the error location is correct), means that either `eulerflux!` or `preflux` is causing an issue.
Maybe try removing the `_....` from preflux?

src/Utilities/Microphysics/src/Microphysics.jl
```diff
@@ -0,0 +1,103 @@
...
+"""
+    terminal_velocity(qt, qr, ρ, ρ_ground)
+
+Return rain terminal velocity.
+TODO - add citation
+"""
+function terminal_velocity(qt::DT, qr::DT, ρ::DT,
+                           ρ_ground::DT) where {DT<:Real}
+    rr  = q2r(qr, qt)
+    vel = 0
```
jkozdon: I think this is causing some type instability on the return (since sometimes the return is an `Int64` and other times a `DT`. This worked for me:

```suggestion
    vel::DT = 0
```

glwagner: Or perhaps `zero(DT)`? Or better yet `-zero(DT)` 😏

lcw:
> There should be no need for the `@inline` annotations for GPU code anymore. What GPU error did you see?

We are seeing local loads and stores when we profile the kernels.  I forgot that GPUifyLoops is inlining everything that is not marked `@noinline`.  We will have to look somewhere else to see what is causing the poor performance.

src/Utilities/Microphysics/src/Microphysics.jl
trontrytel: That was it. Thank you @jkozdon.

trontrytel:
> We will have to look somewhere else to see what is causing the poor performance.

I will track the function that is responsible for it today. It will be anonymous no longer

trontrytel:
I tracked down Mr anonymous12_3 who is spilling registers for me:
https://github.com/climate-machine/CLIMA/blob/899e2825dc99044475c6e3ba16684fcc8b1d442a/src/DGmethods/DGBalanceLawDiscretizations.jl#L710-L713

You can see it by running ```nvprof --profile-from-start off --metrics local_load_transactions,local_store_transactions julia --project=env/gpu test/Atmos/Parameterizations/Microphysics/single_eddy_test_sat_adj.jl```

A fun fact for @vchuravy is that if you go to the knl_dof_iteration function herself: 
https://github.com/climate-machine/CLIMA/blob/899e2825dc99044475c6e3ba16684fcc8b1d442a/src/DGmethods/DGBalanceLawDiscretizations_kernels.jl#L568-L574
and uncomment line 574 the kernel will disappear from the profiler (instead of crashing).

I think this is one of the functions I cherry-picked last week(?) @jkozdon let me know if it's already done in a different way somewhere? If not, any suggestions how I should fix this?

jkozdon:
The `knl_dof_iteration!` is the function we added to fill the custom VTK array. It even if it spills, it's impact should be minimal on the performance unless it's performance is atrocious.

Now why the `@profile` doesn't see it... that I do not understand

jkozdon:
Oh, I misread what you said. The uncommenting the line  574 causes it to disappear if you run in the profiler.

This line is an `@show` which is not supported in `CUDAnative`. Don't know why it does funky things with `@profile`, though...

jkozdon:
Seems that my use of the `do` syntax to create the anonymous function might be the culprit. If I spin off the work as a standalone function it seems better (though the profiler is still a bit odd and running for other functions too).

trontrytel:
> The `knl_dof_iteration!` is the function we added to fill the custom VTK array. It even if it spills, it's impact should be minimal on the performance unless it's performance is atrocious.
> 
ok - thanks @jkozdon. I'll leave this function for now. 


In the profiler I was able to identify all ```anonymous_12_``` kernels. But I can't find ```anonymous_19_1``` kernel. It's not any of the ```@launch``` from the DGmethods and its not the kernel in the ODEs. It's called once at the beginning of the simulation, 3 times before the time loop begins and once at the end of the simulation. 

I'm trying to find it because its dram_read is in MBs and dram_write is in GBs. So I want to understand what it does.  



> Seems that my use of the `do` syntax to create the anonymous function might be the culprit. If I spin off the work as a standalone function it seems better (though the profiler is still a bit odd and running for other functions too).

The ```nvprof --profile-from-start off``` that you suggested was working well for me.

jkozdon:
> The `nvprof --profile-from-start off` that you suggested was working well for me.

When I removed the do syntax I suddenly got 4 kernels being profiled... so not sure what was going on there.

jkozdon:
> In the profiler I was able to identify all `anonymous_12_` kernels. But I can't find `anonymous_19_1` kernel. It's not any of the `@launch` from the DGmethods and its not the kernel in the ODEs. It's called once at the beginning of the simulation, 3 times before the time loop begins and once at the end of the simulation.

Weird let me look into it...

jkozdon:
> But I can't find `anonymous_19_1` kernel

This is coming from the `fill!` in `MPIStateArrays`. Probably best not to do this, so I'll put in a fix for it shortly.

src/Utilities/Microphysics/src/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
+"""
+    Microphysics
+
+Microphysics parameterization functions (for example condensation,
+evaporation, autoconversion rates for cloud water and rain water)
+TODO: Snow, ice, and 2-moment formulation
+"""
+module Microphysics
+
+using CLIMA.MoistThermodynamics
```
charleskawczynski:
```suggestion
using ..MoistThermodynamics
```
?

charleskawczynski: @simonbyrne

trontrytel:
I have updated my draft pull request. 

The Microphysics module and the sources handling https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl#L149-L157 needs review from @tapios 

I have 2 examples:

- The ```test/Atmos/Parameterizations/Microphysics/single_eddy_test_sat_adj.jl``` uses the saturation adjustment only. It works. With help from @simonbyrne, @jkozdon  I can turn it into a test. The test should be that the final ql field is close to the initial one. (This should help with issue #178 )

- The ```test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl``` doesn't work. My guess is that it's because the rain fall speed has to be treated better than what I currently do. If you run it the way it is now it will crash after a couple of timesteps. You can see the instability developing at the top of the domain (looking at the terminal_velocity or qr fields is best for that). 
This is where the rain fall speed is calculated: https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/src/Utilities/Microphysics/Microphysics.jl#L77-L78 If you comment out line 77 and uncomment line 78 the simulation should run. Thats why I'm thinking it has to do with advection of rain. @lcw, @jkozdon @fxgiraldo @smarras79 - let me know what you think? Possible problems I can think of:
      - the (u, w) field is divergence free but the (u, w + rain_w) is not
      - the boundary condition for rain advected with w + rain_w should be coded differently?
      - as my rain velocity is changing in the simulation I'm not sure if the model substeps are updated accordingly? 

Other issues:
- In the Kessler test my qr quickly starts having small negative values. For now I ignore them in my Microphysics source terms calculations. But it should be avoided or handled better.  https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/src/Utilities/Microphysics/Microphysics.jl#L76 
- In both tests I have named the model variables following @jkozdon suggestion to improve readability. So instead of U we now have \rhou. https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl#L47 @asraero, @simonbyrne @tapios : Does that look better? https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl#L178-L181
- You will see that there are a lot of similarities between the two tests. Comparing them also with @asraero  Navier-Stokes tests should help define what we can hide from the future users?
- What should be our strategy for handling errors. For example when the saturation adjustment scheme does not converge.https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/src/Utilities/MoistThermodynamics/MoistThermodynamics.jl#L681-L687

I think I have mentioned as many people as I could in this. I'm grateful to any suggestions how to improve this in the DG fashion.  I'm happy to discuss the issues in writing here or during the office hours.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -253,7 +253,7 @@ internal_energy(ts::PhaseEquil) = ts.e_int
 internal_energy(ts::PhaseNonEquil) = ts.e_int
 
 """
-    internal_energy_sat(T, ρ, q_tot)
+    internal_energy_sat(T, ρ, q_tot) #TODO - rename or change documentation (this is misleading)
```
tapios: What is misleading about it?

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+"""
+module Microphysics
+
+using CLIMA.MoistThermodynamics
+using CLIMA.PlanetParameters
+
+using Printf
+
+# rates of conversion between microphysics categories
+export qv2qli, ql2qr, qr2qv
```
tapios: It would be good to stay with the naming conventions we had established, and improve readability. E.g., ``q_v_to_q_liq``, ``q_liq_to_q_rain`` etc.

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+
+# rain fall speed
+export terminal_velocity
+
+# conversion from specific humidity to mixing ratio (TODO - temporary)
+export q2r
```
tapios: Why do we need it even temporarily? It would seem better to me to formulate things with consistent physics from the outset.

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+
+"""
+    qv2qli(PhasePartition, T, ρ[, timescale=1])
+
+Return the ql and qi source terms due to condesation/evaporation (for ql)
```
tapios: We had ``q_liq``, ``q_ice`` etc. elsewhere.

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+and sublimation/resublimation (for qi)
+The source terms are obtained assuming a relaxation to equilibrum with a
+constant timescale. The timescale is an optional parameter.
+"""
+function qv2qli(q_sa::PhasePartition, q::PhasePartition,
+                timescale::DT=DT(1)) where {DT<:Real}
+
+    dqldt = (q_sa.liq - q.liq) / timescale
```
tapios: ``dq_liq_dt`` etc. would be easier to read.

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+Return rain terminal velocity.
+TODO - add citation
+"""
+function terminal_velocity(qt::DT, qr::DT, ρ::DT,
+                           ρ_ground::DT) where {DT<:Real}
+    rr  = q2r(qr, qt)
+    vel::DT = 0
+
+    if (qr > DT(1e-8)) # TODO - assert positive definite elswhere
+      vel = DT(14.34) * ρ_ground^DT(0.5) * ρ^-DT(0.3654) * rr^DT(0.1346)
```
tapios: Please no constants anywhere except nondimensional numbers that actually are constant. Terminal velocity in the presence of buoyancy is ``sqrt(8 g r/C_d_hydrometeor (\rho_liquid/\rho_air - 1)``, where ``r`` is droplet radius and ``C_d_hydrometeor`` is the drag coefficient of the hydrometeor. ``C_d_hydrometeor`` should be a parameter in some parameter file. The radius can be, e.g., a separately defined effective radius. Probably the above relation put these pieces empirically together, but it thereby hides distributional assumptions made for the radius. We want to expose them, so we can carry them through consistently.

src/Utilities/Microphysics/Microphysics.jl
```diff
@@ -0,0 +1,115 @@
...
+
+    qv_sat = saturation_shum(T, ρ, q)
+    qv = q.tot - q.liq - q.ice
+
+    rr = q2r(qr, q.tot)
+    rv = q2r(qv, q.tot)
+    rv_sat = q2r(qv_sat, q.tot)
+
+    # ventilation factor
+    C = DT(1.6) + DT(124.9) * (DT(1e-3) * ρ * rr)^DT(0.2046)
```
tapios: Same as above. Please formulate them physically, and collect the adjustable parameters that remain in a central place.

smarras79:
> * The `test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl` doesn't work. My guess is that it's because the rain fall speed has to be treated better than what I currently do. If you run it the way it is now it will crash after a couple of timesteps. You can see the instability developing at the top of the domain (looking at the terminal_velocity or qr fields is best for that).
>   This is where the rain fall speed is calculated: https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/src/Utilities/Microphysics/Microphysics.jl#L77-L78
>    If you comment out line 77 and uncomment line 78 the simulation should run. Thats why I'm thinking it has to do with advection of rain. @lcw, @jkozdon @fxgiraldo @smarras79 - let me know what you think? Possible problems I can think of:
>   - the (u, w) field is divergence free but the (u, w + rain_w) is not
>   - the boundary condition for rain advected with w + rain_w should be coded differently?
>   - as my rain velocity is changing in the simulation I'm not sure if the model substeps are updated accordingly?

Hi @trontrytel  before moving to a DG implementation of Kessler, did you try to implement the classical FD based precipitation? I would suggest starting from it because we know what to expect from that implementation which relies on an upwinded solution of the transport equation along each column. 
The reason why I am suggesting this is to have a working kessler first and then move that to DG. 
When I worked at the SEM/CG implementation of precipitation in NUMA the issues were the ones that you described. This was expected because CG is non-positivity preserving and limiters can't be really built for CG. The same is expected from DG unless a slope limiter is implemented. 
Furthermore, the upwinding in the Kessler implementation that everyone uses in WRF and other codes is very dissipative, and that explains why it is so stable at the end of the day. In the case of DG, the inherent dissipation coming from the discretization of the advection equation may be too small. 
We may have to decide how to proceed about it in terms of limiter choice.

src/Utilities/Microphysics/Microphysics.jl
tapios: Or better yet, and what other models use, ``d_dt_q_liq``. But because this is not entirely accurate with nonzero mass sources from precipitation, something like ``source_q_liq`` may be even better. It would be good for us to come up with a consistent naming convention for sources like this.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -681,6 +681,9 @@ function saturation_adjustment(e_int::DT, ρ::DT, q_tot::DT) where DT
     T, converged = find_zero(
       T -> internal_energy_sat(T, ρ, q_tot) - e_int,
       T_1, T_2, SecantMethod(); xatol=DT(1e-3), maxiters=10)
+      if !converged
+        @show "SA did not converge" #TODO - error/warning handling
```
tapios: Instead of ``SA``, spell out ``saturation_adjustment.`` (Better yet: we can assure convergence with proper initial brackets and regula falsi.)

tapios:
> * The `test/Atmos/Parameterizations/Microphysics/single_eddy_test_sat_adj.jl` uses the saturation adjustment only. It works. With help from @simonbyrne, @jkozdon  I can turn it into a test. The test should be that the final ql field is close to the initial one. (This should help with issue #178 )
> *

More tests of moist thermodynamics would indeed be good. But I suggest we only have tests that do not depend on specific choices of ``PlanetParameters.`` As it is, this test does, and hence it would fail if parameters are changed (e.g., for different planetary configurations). It would need to be made parameter-agnostic.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: My only hesitation with using Regula Falsi method is that it has trouble running on the GPU (because there are several `if` statements, which creates branches, and GPUs are not good with handling many branches). Maybe @simonbyrne and @vchuravy can weigh in on this?

It may also be nice if we can map out the 3D space for which we believe `saturation_adjustment` should yield valid results. If the inputs are not physically realistic, do we still want to return a value without any warning/error? This is something we can do once and then specify valid ranges in the docs.

tapios: Good point about GPUs. Probably even with the secant method we should get convergence in reasonable cases as long as the initial bracket is chosen appropriately.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
trontrytel: It's just internal energy and not internal energy at saturation. Or in other words it will only return internal energy at saturation if the input temperature provided will be at saturation. Which is misleading to me.

src/Utilities/Microphysics/Microphysics.jl
trontrytel: Can we discuss those conventions at some point?  The shum is not a standard way to mark specific humidities. Also the function names and variable names will continue to get longer if we stay this explicit.

src/Utilities/Microphysics/Microphysics.jl
trontrytel: The microphysics parameterisation I coded was formulated in mixing ratio and not specific humidity. I can ignore it and just code the same parameters or I have to convert  between the mixing ratio and specific humidity.

src/Utilities/Microphysics/Microphysics.jl
trontrytel: Yes - lets set a convention for all the microphysics species. And I'll set the names accordingly.

src/Utilities/Microphysics/Microphysics.jl
trontrytel: I'm not sure how to formulate this parameter physically. I'll have to search the literature and get back to it.

src/Utilities/Microphysics/Microphysics.jl
trontrytel: Could be settle on q_l and q_i?

trontrytel: This way it's 3 symbols and not 5

src/Utilities/Microphysics/Microphysics.jl
trontrytel: OK - I'll rewrite this and test.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
trontrytel: Yes. My question was more about what we want to do in unreasonable cases. For example when the simulation becomes unstable because of the instability in rain. I would expect the saturation adjustment scheme to throw some error or warning when it doesn't converge.

trontrytel:
> > * The `test/Atmos/Parameterizations/Microphysics/single_eddy_test_sat_adj.jl` uses the saturation adjustment only. It works. With help from @simonbyrne, @jkozdon  I can turn it into a test. The test should be that the final ql field is close to the initial one. (This should help with issue #178 )
> 
> More tests of moist thermodynamics would indeed be good. But I suggest we only have tests that do not depend on specific choices of `PlanetParameters.` As it is, this test does, and hence it would fail if parameters are changed (e.g., for different planetary configurations). It would need to be made parameter-agnostic.

I'm not sure. If PlanetParameters represent Earth we should still expect to get the top part of the domain saturated in the initial condition. And then the test would only test if the final solution is similar to the initial condition. If the initial condition is not saturated then with the prescribed velocity we should not create saturation later in the run. So again the final solution should be close to the initial condition

trontrytel:
> > * The `test/Atmos/Parameterizations/Microphysics/single_eddy_test_Kessler.jl` doesn't work. My guess is that it's because the rain fall speed has to be treated better than what I currently do. If you run it the way it is now it will crash after a couple of timesteps. You can see the instability developing at the top of the domain (looking at the terminal_velocity or qr fields is best for that).
> >   This is where the rain fall speed is calculated: https://github.com/climate-machine/CLIMA/blob/e84b123e8b9a8bc489e8828f855a0d84285072ae/src/Utilities/Microphysics/Microphysics.jl#L77-L78
> >   
> >   If you comment out line 77 and uncomment line 78 the simulation should run. Thats why I'm thinking it has to do with advection of rain. @lcw, @jkozdon @fxgiraldo @smarras79 - let me know what you think? Possible problems I can think of:
> >   
> >   * the (u, w) field is divergence free but the (u, w + rain_w) is not
> >   * the boundary condition for rain advected with w + rain_w should be coded differently?
> >   * as my rain velocity is changing in the simulation I'm not sure if the model substeps are updated accordingly?
> 
> Hi @trontrytel before moving to a DG implementation of Kessler, did you try to implement the classical FD based precipitation? I would suggest starting from it because we know what to expect from that implementation which relies on an upwinded solution of the transport equation along each column.
> The reason why I am suggesting this is to have a working kessler first and then move that to DG.
> When I worked at the SEM/CG implementation of precipitation in NUMA the issues were the ones that you described. This was expected because CG is non-positivity preserving and limiters can't be really built for CG. The same is expected from DG unless a slope limiter is implemented.
> Furthermore, the upwinding in the Kessler implementation that everyone uses in WRF and other codes is very dissipative, and that explains why it is so stable at the end of the day. In the case of DG, the inherent dissipation coming from the discretization of the advection equation may be too small.
> We may have to decide how to proceed about it in terms of limiter choice.

@smarras79 Yes - I can do that. But I was told not to do it at first and show the problem to @lcw and @jkozdon. If thats the solution we want to have I can do this next

src/CLIMA.jl
```diff
@@ -4,6 +4,7 @@ include("Utilities/ParametersType/ParametersType.jl")
 include("Utilities/PlanetParameters/PlanetParameters.jl")
 include("Utilities/RootSolvers/RootSolvers.jl")
 include("Utilities/MoistThermodynamics/MoistThermodynamics.jl")
+include("Utilities/Microphysics/Microphysics.jl")
```
simonbyrne: should this go under `Atmos`? (same for `MoistThermodynamics`?)

cc @charleskawczynski

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
tapios: The temperature, density, and q_tot define the equilibrium point in the phase diagram. Seems standard and ok to me. (You need 3 variables to define a thermodynamic equilibrium state.)

trontrytel: I agree with this. But if as a user I don't make sure that my 3 input variables are at equilibrium then this function will not return internal energy at saturation?

src/Utilities/Microphysics/Microphysics.jl
tapios: Longer is good if it improves readability. We do not need to use ``shum`` but can use ``q`` as in other parts of the code and the design documents. Basically, the closer we stay to the standard notation in the design docs, the easier it will be for users.

src/Utilities/Microphysics/Microphysics.jl
tapios: Fine with me. Please discuss with @charleskawczynski . He was in favor of more verbosity before, which is why we changed to ``q_ice`` etc. in the thermodynamics.

src/Utilities/Microphysics/Microphysics.jl
tapios: Yes, and it's not just for microphysics. This will come up in many places, for many other sources (gravity waves, radiation etc.)

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
tapios: Yes, error handling is good and important. We should just make sure that this isn't triggered in usual circumstances.

tapios:
> I'm not sure. If PlanetParameters represent Earth we should still expect to get the top part of the domain saturated in the initial condition. And then the test would only test if the final solution is similar to the initial condition. If the initial condition is not saturated then with the prescribed velocity we should not create saturation later in the run. So again the final solution should be close to the initial condition

If you change thermodynamic parameters, e.g., to methane on Titan (which we may want to do one day), this may not work. It would be annoying to track down such inconsistencies in those situations. (I have spent a lot of time doing it in the past and want to avoid building in specific assumptions for parameters in this model. It'll increase usability down the line tremendously: other planets, unusual situations, running a non-dimensional dycore, exoplanets etc. etc.)

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
tapios: Any 3 input variables **define** an equilibrium point.

trontrytel: Sorry. I was thinking in terms of the equilibrium I want to reach when doing saturation adjustment. I'll remove the todo.

trontrytel:
@tapios - I have changed the terminal velocity and accretion formulations to get rid of the empirical relations. All the parameters are now in src/Utilities/Microphysics/MicrophysicsParameters.jl`. I set the parameter values to something that works but it definitely requires a revisit. 

I have described the derivation and plotted the comparisons agains the Smolarkiewicz and Grabowski 1996 paper in the CLIMA documentation. Could you take a look at the documentation and check if I'm on the correct path? The remaining thing is rain evaporation and ventilation coefficient. I'll do that next.

docs/src/Utilities/Microphysics.md
```diff
@@ -0,0 +1,229 @@
...
+where
+ - ``D`` is the drop diameter
+ - ``n_{0_{MP}}`` and ``\lambda_{MP}`` are the Marshall-Palmer distribution parameters.
+
+In the remaining part of the Microphysics module derivation this size distribution is expressed as a function of drop radius ``r``
+```math
+\begin{equation}
+n(D) = 2 n_{0_{MP}} exp\left(- 2 \lambda_{MP} r \right)
+\end{equation}
+```
```
tapios: My preference would be just to write the equation in terms of radius, and absorb the 2's into the constants n_0 and \lambda. No need to state the equation with diameter if we do not use it (it just confuses).

docs/src/Utilities/Microphysics.md
```diff
@@ -0,0 +1,229 @@
...
+q_rain_range = range(0, stop=5e-3, length=100)
+ρ_air, q_tot, ρ_air_ground = 1.2, 20 * 1e-3, 1.22
+
+plot(q_rain_range * 1e3,  [terminal_velocity(q_rai, ρ_air) for q_rai in q_rain_range], xlabel="q_rain [g/kg]", ylabel="velocity [m/s]", title="Average terminal velocity of rain", label="CLIMA")
+plot!(q_rain_range * 1e3, [terminal_velocity_empirical(q_rai, q_tot, ρ_air, ρ_air_ground) for q_rai in q_rain_range], label="Empirical")
+savefig("rain_terminal_velocity.svg") # hide
+nothing # hide
+```
+![](rain_terminal_velocity.svg)
+
```
tapios: This is really great use of examples embedded in documentation!

src/CLIMA.jl
tapios: Yes, `Microphysics` should go under `Atmos/Parameterizations`. Moist thermodynamics is more general (same functions should be used, e.g., for sea ice, snow etc.), so this is better under `Utilities`.

simonbyrne:
bors try

examples/Microphysics/ex_2_Kessler.jl
```diff
@@ -0,0 +1,421 @@
...
+                                                Q[_ρq_tot], Q[_ρq_liq],
+                                                Q[_ρq_rai], Q[_ρe_tot]
+    u, w, q_tot, q_liq, q_rai, e_tot = ρu / ρ, ρw / ρ,
+                                       ρq_tot / ρ, ρq_liq / ρ,
+                                       ρq_rai / ρ, ρe_tot / ρ
+
+    DF = eltype(ρ)
+    if(q_rai >= DF(0))
+      # compute rain fall speed
+      rain_w = terminal_velocity(q_rai, ρ)
```
lcw: As we discussed over slack today.  You need to make the following change because you are zeroing out `rain_w` in the `eulerflux!` function.

```suggestion
      rain_w = DF(0)
```

trontrytel: Yes. Just added. Testing now

trontrytel: Thank you!

trontrytel:
bors try

trontrytel:
@tapios @lcw @simonbyrne - I think I'm ready with this pull. 

I opened a separate issue #276 to deal with all the stability issues in the Kessler example.

I'll add some simple tests to the Microphysics module so that there is some code coverage on that. But the ultimate test for me would be to get the Kessler example actually working

trontrytel:
bors try

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
+# Microphysics Module
+
+The `Microphysics` module provides 1-moment bulk parameterization of cloud microphysical processes. The module describes the warm rain (no ice and snow) formation and is based on the ideas of Kessler 1995.
```
tapios: ... provides *a* 1-moment ...

... describes ** warm rain ....

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+n(r) = n_{0_{MP}} exp\left(- \lambda_{MP} \; r \right)
+\end{equation}
+```
+where:
+ - ``r`` is the drop radius,
+ - ``n_{0_{MP}}`` and ``\lambda_{MP}`` are the Marshall-Palmer distribution parameters (twice the values used in the Marshall Palmer 1948, because we use drop radius and not diameter).
+
+## Terminal velocity
+
+The terminal velocity of an individual rain drop is defined by the balance between the gravity (taking into account the density difference between water and air) and the drag force:
```
tapios: gravity -> gravitational acceleration

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+ - ``\rho`` is the density of air.
+
+The mass weighted terminal velocity ``v_t`` is defined following Ogura and Takahashi 1971
+```math
+\begin{equation}
+v_t = \frac{F_{rain}}{RWC}
+\end{equation}
+```
+where:
+ - ``F_{rain} = \int_0^\infty n(r) \; m(r) \; v_{drop}(r) dr`` is the vertical flux of rain drops,
```
tapios: dr -> \, dr

(and I would change \; to \, ...:)

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+where:
+ - ``r`` is the drop radius,
+ - ``n_{0_{MP}}`` and ``\lambda_{MP}`` are the Marshall-Palmer distribution parameters (twice the values used in the Marshall Palmer 1948, because we use drop radius and not diameter).
+
+## Terminal velocity
+
+The terminal velocity of an individual rain drop is defined by the balance between the gravity (taking into account the density difference between water and air) and the drag force:
+```math
+\begin{equation}
+v_{drop} = \left(\frac{8}{3 \; C_{drag}} \left( \frac{\rho_{water}}{\rho} -1 \right) \right)^{1/2} g^{1/2} \; r^{1/2} = v_c(\rho) \; g^{1/2} \; r^{1/2}
```
tapios: would be a bit clearer to write this as (g r)^{1/2}

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+\end{equation}
+```
+```math
+F_{rain} = \Gamma \left(\frac{9}{2} \right) \frac{4}{3} n_{0_{MP}} \; \pi \; \rho_{water} v_c(\rho) \; g^{1/2} (\lambda_{MP})^{-9/2}
+```
+Eliminating ``\lambda_{MP}`` between the two equations and dividing by ``RWC`` results in
+```math
+\begin{equation}
+v_t = \Gamma \left( \frac{9}{2} \right) \; \frac{v_c(\rho)}{6} \; (8 \pi)^{-1/8} \left( \frac{\rho}{\rho_{water}} \right)^{1/8} n_{0_{MP}}^{-1/8} g^{1/2} q_{rai}^{1/8} = v_{t_{c}}(\rho) \; n_{0_{MP}}^{-1/8} g^{1/2} q_{rai}^{1/8}
+\end{equation}
```
tapios: I'd find it clearer to write this as 

v_t = \frac{4}{3}\pi \, \Gamma \left( \frac{9}{2} \right) v_c(\rho) \left( \frac{g}{\lambda_{MP}} \right)^{1/2} 

with \lambda_{MP} = \left( \frac{1}{8\pi n_{0_{MP}}} \frac{\rho}{\rho_{water}} q_{rai} \right)^{1/4}

This makes it clearer where length scales come from, and that you are using q_{rai} to get at \lambda. (Please check details.)

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+plot(q_rain_range * 1e3,  [terminal_velocity(q_rai, ρ_air) for q_rai in q_rain_range], xlabel="q_rain [g/kg]", ylabel="velocity [m/s]", title="Average terminal velocity of rain", label="CLIMA")
+plot!(q_rain_range * 1e3, [terminal_velocity_empirical(q_rai, q_tot, ρ_air, ρ_air_ground) for q_rai in q_rain_range], label="Empirical")
+savefig("rain_terminal_velocity.svg") # hide
+nothing # hide
+```
+![](rain_terminal_velocity.svg)
+
+## Cloud condensation/evaporation
+
+Condensation and evaporation of cloud water is parameterized as a relaxation to equilibrium. The equilibrium state is calculated using saturation adjustment defined in the `MoistThermodynamics` module.
```
tapios: If this is time-stepped explicitly, we do not need saturation adjustment, just the equilibrium value at the current time step.

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+ - ``q_{liq\_threshold}`` - autoconversion (parameter in `MicrophysicsParameters` module).
+
+The default values of ``\tau_{acnv}`` and ``q_{liq\_threshold}`` are based on Smolarkiewicz and Grabowski 1996.
+
+## Accretion
+
+Accretion defines the rate of conversion from cloud to rain water resulting from collisions between cloud droplets and rain drops. It is parameterized following Kessler 1995:
+```math
+\begin{equation}
+\left. \frac{d \; q_{rai}}{dt} \right|_{accr} = \int_0^\infty n(r) \; \pi r^2 \; v_c(\rho) \; g^{1/2} \; r^{1/2} E_{col} q_{liq} dr
```
tapios: Why not writing v_{drop} here in place of v_c(\rho) \; g^{1/2} \; r^{1/2} ? Seems clearer to me.

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+                  \left(\frac{q_{vap}}{q_{vap}^{sat}} - 1 \right)
+                  \left(
+                    \frac{L}{KT} \left(\frac{L}{R_v T} - 1 \right) +
+                    \frac{R_v T}{p_{vap}^{sat} D}
+                  \right)^{-1}
+                = \frac{1}{\rho_{water}} S(q_{vap}, q_{vap}^{sat}) \; G(T)
+\end{equation}
+```
+where:
+ - ``q_{vap}{sat}`` is the saturation vapor specific humidity,
```
tapios: -> _{cap}*^*{sat}

docs/src/Atmos/Microphysics.md
```diff
@@ -0,0 +1,369 @@
...
+\begin{equation}
+
+r \frac{dr}{dt} = \frac{1}{\rho_{water}}
+                  \left(\frac{q_{vap}}{q_{vap}^{sat}} - 1 \right)
+                  \left(
+                    \frac{L}{KT} \left(\frac{L}{R_v T} - 1 \right) +
+                    \frac{R_v T}{p_{vap}^{sat} D}
+                  \right)^{-1}
+                = \frac{1}{\rho_{water}} S(q_{vap}, q_{vap}^{sat}) \; G(T)
+\end{equation}
```
tapios: Define S and G explicitly.

src/Atmos/Parameterizations/CloudPhysics/MicrophysicsParameters.jl
```diff
@@ -0,0 +1,25 @@
+"""
+    MicrophysicsParameters
+
+Module containing 1-moment bulk microphysics parameters.
+"""
+module MicrophysicsParameters
+using CLIMA.ParametersType
+
+@exportparameter MP_n_0          8e6 * 2         "Marshal-Palmer distribution n_0 coeff [1/m4]"
```
tapios: -> Marshal*l*-Palmer

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,139 @@
...
+  ref_term_vel_big   = [11.75, 12.47, 13.11, 13.90, 14.74, 15.85]
+
+  for idx in range(Int(1), stop=Int(6))
+    vc = terminal_velocity_single_drop_coeff(ρ_range[idx])
+
+    term_vel_small = vc .* sqrt(r_small .* grav)
+    term_vel_big   = vc .* sqrt(r_big   .* grav)
+
+    @test term_vel_small ≈ ref_term_vel_small[idx] atol = 0.01
+    @test term_vel_big   ≈ ref_term_vel_big[idx]   atol = 0.01
```
tapios: These tests are fine, but seem brittle: if we change formulation of the terminal velocity (e.g., by changing size distribution of raindrops), they will fall. Wouldn't it be better to test universal limits? (e.g., density liquid = density air, density liquid >> density air, 4xg leads to 2x terminal velocity etc.)

examples/Microphysics/ex_1_saturation_adjustment.jl
```diff
@@ -0,0 +1,335 @@
...
+  cbvtk = GenericCallbacks.EveryXSimulationSteps(60) do (init=false)
+    DGBalanceLawDiscretizations.dof_iteration!(postprocessarray, spacedisc,
+                                               Q) do R, Q, QV, aux
+      @inbounds begin
+        u, w, ρ, q_tot, e_tot = preflux(Q)
+        z = aux[_c_z]
+        p = aux[_c_p]
+
+        e_int = e_tot - 1//2 * (u^2 + w^2) - grav * z
+        ts = PhaseEquil(e_int, q_tot, ρ)  # saturation adjustment happens here
```
tapios: Why do we need it for explicit time stepping, if q_liq is part of the state?

tapios:
Looks great overall. Some minor comments in the file. The one more significant comment concerns saturation adjustment: If liquid water is prognostic, why do we do it at all? The source term with explicit time stepping would be evaluated with the current q_liq etc., which is sufficient to compute the saturation vapor pressure and the super/sub-saturation appearing in the relaxation term.

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: I like it better too. I was trying to always keep the q_rai out in the open. But this way is clearer. I also changed the code implementation to reflect the documentation structure.

src/Atmos/Parameterizations/CloudPhysics/MicrophysicsParameters.jl
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

docs/src/Atmos/Microphysics.md
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: Those would be good tests. But I don't know how to overwrite the dens_liquid and grav, because they are planet parameters. I'll add an issue to improve on the tests for Microphysics

trontrytel: issue #280

examples/Microphysics/ex_1_saturation_adjustment.jl
trontrytel: In the Kessler example I change it to q_eq = PhasePartition_equil(T, ρ, q_tot). In the example 1 for saturation adjustment I keep it

docs/src/Atmos/Microphysics.md
trontrytel: correct - changed!

trontrytel:
bors try

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+  # terminal_vel_of_individual_drop = v_drop_coeff * drop_radius^(1/2)
+  v_c::DT = terminal_velocity_single_drop_coeff(ρ)
+
+  #gamma(7/2)
+  gamma_7_2::DT = DT(3.32)
+
+  accr_coeff::DT = gamma_7_2 * DT(8)^DT(-7/8) * π^DT(1/8) * v_c * E_col *
+                   (ρ / dens_liquid)^DT(7/8)
+
+  src_q_rai::DT = accr_coeff * MP_n_0^DT(1/8) * sqrt(grav) *
```
simonbyrne:
```suggestion
  return accr_coeff * DT(MP_n_0)^DT(1/8) * sqrt(DT(grav)) *
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+
+where:
+- `q_liq` - is the liquid water specific humidity
+
+Returns the q_rai tendency due to collisions between cloud droplets
+(autoconversion) parametrized following Kessler 1995.
+"""
+function conv_q_liq_to_q_rai_acnv(q_liq::DT) where {DT<:Real}
+
+  src_q_rai::DT = max(DT(0), q_liq - q_liq_threshold) / τ_acnv
```
simonbyrne:
```suggestion
  return max(DT(0), q_liq - q_liq_threshold) / τ_acnv
```
and can get rid of extra empty lines

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+
+where:
+- `q_sat` - PhasePartition at equilibrium
+- `q`     - current PhasePartition
+
+Returns the q_liq tendency due to condensation/evaporation.
+The tendency is obtained assuming a relaxation to equilibrium with
+constant timescale.
+"""
+function conv_q_vap_to_q_liq(q_sat::PhasePartition, q::PhasePartition)
```
simonbyrne:
```suggestion
function conv_q_vap_to_q_liq(q_sat::PhasePartition{DT}, q::PhasePartition{DT}) where {DT}
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+
+  DT = eltype(q.tot)
```
simonbyrne: this can then be removed

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+where:
+  - `ρ` is the density of air
+
+Returns the proportionality coefficient between terminal velocity of an
+individual water drop and the square root of its radius * g.
+"""
+function terminal_velocity_single_drop_coeff(ρ::DT) where {DT<:Real}
+
+    # terminal_vel_of_individual_drop = v_drop_coeff * (g * drop_radius)^(1/2)
+    v_c::DT = sqrt(DT(8/3) / C_drag * (dens_liquid / ρ - DT(1)))
```
simonbyrne:
```suggestion
    return sqrt(DT(8/3) / C_drag * (dens_liquid / ρ - DT(1)))
```
and get rid of empty whitespace

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+  S::DT = q_v/qv_sat - DT(1)
+
+  L::DT = latent_heat_vapor(T)
+  p_vs::DT = saturation_vapor_pressure(T, Liquid())
+  G::DT = DT(1) / (
+            L / K_therm / T * (L / R_v / T - DT(1)) + R_v * T / D_vapor / p_vs
+          )
+
+  # gamma(11/4)
+  gamma_11_4::DT = DT(1.61)
```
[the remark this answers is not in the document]
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+
+Returns the mass weighted average rain terminal velocity assuming
+Marshall Palmer 1948 distribution of rain drops.
+"""
+function terminal_velocity(q_rai::DT, ρ::DT) where {DT<:Real}
+
+    v_c = terminal_velocity_single_drop_coeff(ρ)
+
+    # gamma(9/2)
+    gamma_9_2::DT = DT(11.63)
```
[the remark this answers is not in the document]
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+Returns the q_rai tendency due to collisions between cloud droplets
+and rain drops (accretion) parametrized following Kessler 1995.
+"""
+function conv_q_liq_to_q_rai_accr(q_liq::DT, q_rai::DT, ρ::DT) where {DT<:Real}
+
+  # terminal_vel_of_individual_drop = v_drop_coeff * drop_radius^(1/2)
+  v_c::DT = terminal_velocity_single_drop_coeff(ρ)
+
+  #gamma(7/2)
+  gamma_7_2::DT = DT(3.32)
```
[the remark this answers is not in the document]
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+  v_c::DT = terminal_velocity_single_drop_coeff(ρ)
+
+  av = sqrt(2 * π) * a_vent * sqrt(ρ / dens_liquid)
+  bv = DT(2)^DT(7/16) * gamma_11_4 * π^DT(5/16) * b_vent * (N_Sc)^DT(1/3) *
+       sqrt(v_c) * (ρ / dens_liquid)^DT(11/16)
+
+  F::DT = av * sqrt(qr) +
+          bv * grav^DT(1/4) / (MP_n_0)^DT(3/16) / sqrt(ν_air) * qr^DT(11/16)
+
+  ret = S * F * G * sqrt(MP_n_0) / ρ
```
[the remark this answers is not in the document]
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,187 @@
...
+constant timescale.
+"""
+function conv_q_vap_to_q_liq(q_sat::PhasePartition, q::PhasePartition)
+
+  DT = eltype(q.tot)
+
+  src_q_liq::DT = (q_sat.liq - q.liq) / τ_cond_evap
+
+  if q_sat.ice != DT(0)
+    @show("1-moment bulk microphysics is not defined for snow/ice")
```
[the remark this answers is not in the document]
trontrytel: done

trontrytel:
bors try

trontrytel:
@simonbyrne  - the build failed but it doesn't look like it was because of microphysics?

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -681,6 +681,9 @@ function saturation_adjustment(e_int::DT, ρ::DT, q_tot::DT) where DT
     T, converged = find_zero(
       T -> internal_energy_sat(T, ρ, q_tot) - e_int,
       T_1, T_2, SecantMethod(); xatol=DT(1e-3), maxiters=10)
+      if !converged
```
charleskawczynski: @trontrytel Codecov may be failing if this segment of the code is not being tested.

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+ - ρ - air density
+
+Returns the q_rai tendency due to rain evaporation. Parameterized following
+Smolarkiewicz and Grabowski 1996.
+"""
+function conv_q_rai_to_q_vap(qr::DT, q::PhasePartition{DT},
+                             T::DT, p::DT, ρ::DT) where {DT<:Real}
+
+  qv_sat = saturation_shum(T, ρ, q)
+  q_v::DT = q.tot - q.liq - q.ice
```
charleskawczynski:
```suggestion
  q_v = q.tot - q.liq - q.ice
```
The type is already inherited from the passed in argument

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+  S::DT = q_v/qv_sat - DT(1)
```
charleskawczynski:
```suggestion
  S = q_v/qv_sat - 1
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+
+  L::DT = latent_heat_vapor(T)
```
charleskawczynski:
```suggestion
  L = latent_heat_vapor(T)
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+  p_vs::DT = saturation_vapor_pressure(T, Liquid())
```
charleskawczynski:
```suggestion
  p_vs = saturation_vapor_pressure(T, Liquid())
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+  G::DT = DT(1) / (
```
charleskawczynski:
```suggestion
  G = 1 / (
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -0,0 +1,172 @@
...
+            L / K_therm / T * (L / R_v / T - DT(1)) + R_v * T / D_vapor / p_vs
```
charleskawczynski:
```suggestion
            L / K_therm / T * (L / R_v / T - 1) + R_v * T / D_vapor / p_vs
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+  end
+end
+
+@testset "RainFallSpeed" begin
+
+  # eq. 5d in Smolarkiewicz and Grabowski 1996
+  # https://doi.org/10.1175/1520-0493(1996)124<0487:TTLSLM>2.0.CO;2
+  function terminal_velocity_empir(q_rai::DT, q_tot::DT, ρ::DT,
+                                      ρ_air_ground::DT) where {DT<:Real}
+      rr  = q_rai / (DT(1) - q_tot)
```
charleskawczynski:
```suggestion
      rr  = q_rai / (1 - q_tot)
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+      vel = DT(14.34) * ρ_air_ground^DT(0.5) * ρ^-DT(0.3654) * rr^DT(0.1346)
+      return vel
+  end
+
+  # some example values
+  q_rain_range = range(1e-8, stop=5e-3, length=10)
+  ρ_air, q_tot, ρ_air_ground = 1.2, 20 * 1e-3, 1.22
```
charleskawczynski:
```suggestion
  ρ_air, q_tot, ρ_air_ground = DT(1.2), DT(20 * 1e-3), DT(1.22)
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+
+  for q_rai in q_rain_range
+    @test terminal_velocity(q_rai, ρ_air) ≈
+          terminal_velocity_empir(q_rai, q_tot, ρ_air, ρ_air_ground) atol =
+            0.2 * terminal_velocity_empir(q_rai, q_tot, ρ_air, ρ_air_ground)
```
charleskawczynski:
```suggestion
            DT(0.2) * terminal_velocity_empir(q_rai, q_tot, ρ_air, ρ_air_ground)
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+  function rain_evap_empir(q_rai::DT, q::PhasePartition,
+                           T::DT, p::DT, ρ::DT) where {DT<:Real}
+
+      q_sat  = saturation_shum(T, ρ, q)
+      q_vap  = q.tot - q.liq
+      rr     = q_rai / (DT(1) - q.tot)
+      rv_sat = q_sat / (DT(1) - q.tot)
+      S      = q_vap/q_sat - DT(1)
+
+      ag, bg = 5.4 * 1e2, 2.55 * 1e5
```
charleskawczynski:
```suggestion
      ag, bg = DT(5.4 * 1e2), DT(2.55 * 1e5)
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+  # eq. 5c in Smolarkiewicz and Grabowski 1996
+  # https://doi.org/10.1175/1520-0493(1996)124<0487:TTLSLM>2.0.CO;2
+  function rain_evap_empir(q_rai::DT, q::PhasePartition,
+                           T::DT, p::DT, ρ::DT) where {DT<:Real}
+
+      q_sat  = saturation_shum(T, ρ, q)
+      q_vap  = q.tot - q.liq
+      rr     = q_rai / (DT(1) - q.tot)
+      rv_sat = q_sat / (DT(1) - q.tot)
+      S      = q_vap/q_sat - DT(1)
```
charleskawczynski:
```suggestion
      S      = q_vap/q_sat - 1
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+@testset "RainEvaporation" begin
+
+  # eq. 5c in Smolarkiewicz and Grabowski 1996
+  # https://doi.org/10.1175/1520-0493(1996)124<0487:TTLSLM>2.0.CO;2
+  function rain_evap_empir(q_rai::DT, q::PhasePartition,
+                           T::DT, p::DT, ρ::DT) where {DT<:Real}
+
+      q_sat  = saturation_shum(T, ρ, q)
+      q_vap  = q.tot - q.liq
+      rr     = q_rai / (DT(1) - q.tot)
```
charleskawczynski:
```suggestion
      rr     = q_rai / (1 - q.tot)
```

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -0,0 +1,140 @@
...
+      rv_sat = q_sat / (DT(1) - q.tot)
```
charleskawczynski:
```suggestion
      rv_sat = q_sat / (1 - q.tot)
```

examples/Microphysics/ex_1_saturation_adjustment.jl
```diff
@@ -0,0 +1,335 @@
...
+
+using LinearAlgebra
+using StaticArrays
+using Printf
+
+using CLIMA.PlanetParameters
+using CLIMA.MoistThermodynamics
+using CLIMA.Microphysics
+
+@static if Base.find_package("CuArrays") !== nothing
```
charleskawczynski: Let's switch this code snippet to
```
@static if haspkg("CuArrays")
```
in the `src/misc.jl`.

src/Utilities/MoistThermodynamics/MoistThermodynamics.jl
trontrytel: I think it's still better to have it

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: done

test/Atmos/Parameterizations/Microphysics/runtests.jl
trontrytel: done

trontrytel:
bors try

examples/Microphysics/ex_1_saturation_adjustment.jl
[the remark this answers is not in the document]
trontrytel: @charleskawczynski what additional package should I include? I'm getting 

```ERROR: LoadError: LoadError: UndefVarError: haspkg not defined```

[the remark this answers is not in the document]
trontrytel: Works if I add ```using CLIMA```. Does it include everything?

trontrytel:
bors try

trontrytel:
bors try

trontrytel:
bors try

examples/Microphysics/ex_1_saturation_adjustment.jl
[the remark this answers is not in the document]
charleskawczynski: Yeah, you need `using CLIMA` before this line.

trontrytel:
bors try

trontrytel:
bors try

trontrytel:
bors try

PR #238, opened 2019-05-21
Define haspkg function

simonbyrne:
Currently we use `find_package` to determine if a package is available. This can cause conflicts in cases where the package is installed, but not actually available in the current environment.

This seems to be the simplest option for now.


simonbyrne:
included as part of #246

PR #239, opened 2019-05-21
Make CLIMA work with Revise

simonbyrne:
There was a weird issue with using 2 `@init` statements, so I had to merge them.


jkozdon:
I had a message on slack last week from @vchuravy about Revise and that we shouldn't have the `@eval` in the `@parameter` macro. Haven't had a chance to explore this, but related to https://github.com/timholy/Revise.jl/issues/309.

simonbyrne:
I'm looking at it at the moment: removing `eval` doesn't seem to fix it.

simonbyrne:
This was fixed upstream. Everything now works!

PR #240, opened 2019-05-21
Fix VTK for old atmos dycore

jkozdon:
Bug fix for vtk in old dycore introduced by #230

### DIFF
--- a/src/InputOutput/Vtk/writevtk.jl
+++ b/src/InputOutput/Vtk/writevtk.jl
@@ -96,12 +96,13 @@ end
 function writevtk(prefix, Q::MPIStateArray, disc::VanillaAtmosDiscretization)
   vgeo = disc.grid.vgeo
   host_array = Array ∈ typeof(Q).parameters
-  (h_vgeo, h_Q) = host_array ? (vgeo, Q.Q) : (Array(vgeo), Array(Q))
-  writevtk(prefix, h_vgeo, h_Q, disc.grid)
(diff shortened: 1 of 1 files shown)


src/InputOutput/Vtk/writevtk.jl
```diff
@@ -96,12 +96,13 @@ end
...
   host_array = Array ∈ typeof(Q).parameters
-  (h_vgeo, h_Q) = host_array ? (vgeo, Q.Q) : (Array(vgeo), Array(Q))
-  writevtk(prefix, h_vgeo, h_Q, disc.grid)
+  (vgeo, Q) = host_array ? (vgeo, Q.Q) : (Array(vgeo), Array(Q))
+  writevtk_VanillaAtmosDiscretization(prefix, vgeo, Q, disc.grid)
 end
 
-function writevtk(prefix, vgeo::Array, Q::Array,
-                  G::Grids.AbstractGrid{T, dim, N}) where {T, dim, N}
+function writevtk_VanillaAtmosDiscretization(prefix, vgeo::Array, Q::Array,
```
simonbyrne: would it be easier just to combine the functions?

jkozdon: We could. I was just doing a minimal fix since this code will likely go away very soon once @asraero and @smarras79 are done with the old dycore and we dump all the Vanilla atmos code.

PR #242, opened 2019-05-22
remove unnecessary use of Documenter

simonbyrne:


simonbyrne:
included as part of #246

PR #244, opened 2019-05-23
Removing default zeroing of MPIStateArray

jkozdon:
As discussed in https://github.com/climate-machine/CLIMA/pull/237#issuecomment-495392923 we do not need to be filling MPIStateArrays with zeros by default.


jkozdon:
@fxgiraldo  and @mwarusz this might effect the ODE solver stuff you're working on if you assume that your similared arrays are zero'd out (which isn't a safe assumption anyway in Julia).

jkozdon:
bors try

PR #252, opened 2019-05-30
Rework tests to add @testset and remove random

jkozdon:
Realized after discussions with @mwarusz that we should be using `@testset` in the tests and that the first level integration test is probably short enough for our current CI set up (so we can get rid of the random testing)

### DIFF
--- a/test/DGmethods/Euler/isentropic_vortex_standalone.jl
+++ b/test/DGmethods/Euler/isentropic_vortex_standalone.jl
@@ -41,16 +41,16 @@ else
   const ArrayTypes = (Array,)
 end
 
+if !@isdefined integration_testing
+  const integration_testing =
+    parse(Bool, lowercase(get(ENV,"JULIA_CLIMA_INTEGRATION_TESTING","false")))
+end
+
 const _nstate = 5
 const _ρ, _U, _V, _W, _E = 1:_nstate
 const stateid = (ρid = _ρ, Uid = _U, Vid = _V, Wid = _W, Eid = _E)
 const statenames = ("ρ", "U", "V", "W", "E")
 const γ_exact = 7 // 5
-if !@isdefined integration_testing
-  const integration_testing =
-    parse(Bool, lowercase(get(ENV,"JULIA_CLIMA_INTEGRATION_TESTING","false")))
-  using Random
-end
 
 # preflux computation
 @inline function preflux(Q, _...)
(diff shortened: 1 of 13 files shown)


jkozdon:
bors try

jkozdon:
@mwarusz this might impact some of what you're doing. Let me know if you have suggestions/thoughts

test/DGmethods/Euler/isentropic_vortex_standalone.jl
```diff
@@ -41,16 +41,16 @@ else
   const ArrayTypes = (Array,)
 end
 
+if !@isdefined integration_testing
+  const integration_testing =
+    parse(Bool, lowercase(get(ENV,"JULIA_CLIMA_INTEGRATION_TESTING","false")))
```
simonbyrne: As an aside: what is this for?

jkozdon: Long versus short testing. If `integration_test == false` only one mesh level is run. When this breaks we probably want to longer set of data to know if stuffs broken, or just changed.

jkozdon: This removes the need to have my random test hashing stuff to get short tests for CI. Most of the time for the tests is actually Julia initialization time (not running time) for the smallest mesh.

jkozdon: Also when `integration_test == true` we compute a rate of convergence of the scheme (this is the only rigorous way I know to make sure things are working as they should) as opposed to just working in the eyeball norm.

jkozdon:
Main thing to check before this is merged is that the CI time has not significantly changed.

test/DGmethods/Euler/isentropic_vortex_standalone.jl
simonbyrne: okay, can we give it a better name? Or perhaps make it a command line argument?

Also, should we do this as part of our tests, now we're not so limited in terms of running time? My experience is that things which aren't explicitly tested will break as the code evolves.

jkozdon: > okay, can we give it a better name? Or perhaps make it a command line argument?

I'm fine with another name if you have a suggestion. If you think that these should be command line arguments instead of environment variables that's fine.

> Also, should we do this as part of our tests, now we're not so limited in terms of running time?

I'm fine running the fully suite all the time, but this probably won't work for travis or appveyor runs. I just don't want to get back into the CI bottleneck state so I was avoiding this.

As a time point, running `JULIA_CLIMA_INTEGRATION_TESTING=true julia --project=@. isentropic_vortex_standalone.jl` on a single CPU (with my AWS instance) takes a little over 8 minutes and this is probably shortest test.

> My experience is that things which aren't explicitly tested will break as the code evolves.

I agree. One thing though is that nothing is being hidden behind a switch now (so even if we cannot afford to run the full suite the whole time this is better than before since the actual driver is being tested not some random hash of it).

jkozdon:
bors try

test/DGmethods/Euler/isentropic_vortex_standalone.jl
jkozdon: > > My experience is that things which aren't explicitly tested will break as the code evolves.
> 
> I agree. One thing though is that nothing is being hidden behind a switch now (so even if we cannot afford to run the full suite the whole time this is better than before since the actual driver is being tested not some random hash of it).

On this note, there was some code hidden behind the switch in `dycoms.jl` and `rtb_visc.jl` that [popped](https://gist.github.com/climabot/a8ba3c533c1814b8a394deda58db0c06)

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

PR #254, opened 2019-05-30
[BL] Adding vertical integration kernels

jkozdon:
Adds functionality to compute indefinite line integrals for stacked meshes along the stacked dimension though the function `DGBalanceLawDiscretizations.indefinite_stack_integral!`.

The function `preodefun!` is also introduced which allows the user to "inject" code right before the ODE function is evaluated.

Few other bug/feature fixes:
- Read geo indices from Grids and unify names
- Removing unused functions in MPIStateArrays
- Remove some structure assumptions in MPIStateArrays
- Fix bug Topology for stacksize and getproperty
- Fix `grad_test_sphere.jl` to actually use a sphere


jkozdon:
cc: @lcw @smarras79 @asraero

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -659,6 +688,64 @@ function grad_auxiliary_state!(disc::DGBalanceLaw, id, (idx, idy, idz))
                            Dmat, topology.elems, id, idx, idy, idz))
 end
 
+
+"""
+    indefinite_stack_integral!(disc, f, Q, out_states, [P=disc.auxstate])
+
+Computes the indefinite integral along an element stack using state `Q`
+```math
+∫_{ζ_{0}}^{ζ} f(q; aux, t)
```
simonbyrne: Is this the integral from the base upwards? I think in some cases they need the integral in the other direction

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -659,6 +688,64 @@ function grad_auxiliary_state!(disc::DGBalanceLaw, id, (idx, idy, idz))
                            Dmat, topology.elems, id, idx, idy, idz))
 end
 
+
+"""
+    indefinite_stack_integral!(disc, f, Q, out_states, [P=disc.auxstate])
```
simonbyrne: should be added to the relevant `docs/src/` file.

src/DGmethods/DGBalanceLawDiscretizations.jl
```diff
@@ -659,6 +688,64 @@ function grad_auxiliary_state!(disc::DGBalanceLaw, id, (idx, idy, idz))
...
+where `F` is an `MVector` of length `length(out_states)`, `Q` and `aux` are
+the `MVectors` for the state and auxiliary state at a single degree of freedom.
+The function is responsible for filling `F`.
+
+Requires the `isstacked(disc.grid.topology) == true`
+
+!!! note
+
+    When the topology has one two dimensions the integral is computed along the
+    `η` direction.
```
simonbyrne: would it be simplest to say "trailing dimension"?

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: Yes it is. Given the top value you can extract the other one... that said we do not have convenient way to access this at the user level.

jkozdon: Could change this to:
```julia
indefinite_stack_integral!(disc::DGBalanceLaw, f, Q, out_states,
                                    P=disc.auxstate; direction=:ascending)
```
or something like that.

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: sure.

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: done (as well as fixed other docs issues)

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: How would one `accumulate` backward in julia? This would really be the analog in summation case.

jkozdon: One option would be to have a reverse a precomputed indefinite integral function if up is the standard way, but sometimes backward is also needed.

jkozdon:
bump.

@asraero and @smarras79 do you guys need anything else in this functionality?

Also, rebased this on master (since there were going to be conflicts)

src/DGmethods/DGBalanceLawDiscretizations.jl
jkozdon: I've added a `reverse_indefinite_stack_integral!` function

jkozdon:
I've added a reverse function for the integral. Does this work for you @asraero and @smarras79?

Issue #245, opened 2019-05-24
accessing elements of degree of freedom arrays in balance law solver


jkozdon:
After some discussion with @charleskawczynski, @asraero and @trontrytel, we decided that it would be very useful to be able to declare internal model variables that could be accessed via `getfield` overloading instead of e.g. `ρ == Q[_ρ]`.

Maybe something like:
```
@variables MyModel begin
    @constant begin
        # time-constant variables, e.g. coordinates, reference values, etc.
        # currently called "aux" 
        x,y,z
    end

    @diagnostic begin
         # currently computed via "preflux"
         ql
     end

     @prognostic begin
         # state variables
         U,V,W,ρ,Q,E
      end
end
```

It would also be useful to be able to create compound data structures (e.g. structs or static arrays) from combinations of these. An example would be getting a static array of velocities.

_Originally posted by @simonbyrne in https://github.com/climate-machine/CLIMA/pull/233#issuecomment-495404115_

jkozdon:
Can you give a desugared example? The macro should be implementable, but I am worried about GPU compatibility. Are these static variables for the entire Grid, for each element, or for each nodal point? 

What we might be able to do, is to encapsulate the local view into the global data and have the BalanceLaw solver create a local model instance.

I explicitly excluded models having state in the first draft, since I don't quite know how that fits into what the solver needs to do. 

_Originally posted by @vchuravy  in https://github.com/climate-machine/CLIMA/pull/233#issuecomment-495411957

jkozdon:
Something to keep in mind if someone tries to do this:

- Right now we grab all of the states, aux state, and viscous state to pass to the code injection functions. I imagine we'll likely have to move away from this for performance reason and only grab values out of the global arrays the kernel actually needs.
- When we have O(100) tracers, there will be some sub-grouping that has to occur.

Whatever framework is chosen, will need to be flexible enough to support this. The introduction of the `DGModel` type should help here though.

Also, if this were done the above sub-selection would be seen more cleanly by the user possibly.

Finally, we may want to support functions as well. For instance, `Q.U` gives the momentum (`ρu`) and `Q.u` gives the velocity (`U/ρ`). You could even imagine support `Q.u⃗` giving the vector `(u, v, w)`

simonbyrne:
Basically the idea is that the macros would generate the necessary `getproperty`/`setproperty!` overloads, so that any such accesses will be inlined and resolved to regular array indexing at compile time. 

[...]

PR #256, opened 2019-05-31
Draft get/setproperty overloading

simonbyrne:
Based on ideas from #245.

Which of the tests are a good benchmark?

### DIFF
--- a/test/DGmethods/compressible_Navier_Stokes/mms_bc.jl
+++ b/test/DGmethods/compressible_Navier_Stokes/mms_bc.jl
@@ -57,34 +57,66 @@ end
   @inbounds abs(n[1] * u + n[2] * v + n[3] * w) + sqrt(ρinv * γ * P)
 end
 
+struct State{T}
+  arr::T
+end
+struct Flux{T}
+  arr::T
+end
+
+function Base.getproperty(state::State, sym::Symbol)
+  arr = getfield(state,:arr)
+  if sym == :ρ
+    @inbounds arr[_ρ]
+  elseif sym == :ρ𝐮
+    @inbounds SVector(arr[_U], arr[_V], arr[_W])
+  elseif sym == :𝐮
+    @inbounds SVector(arr[_U], arr[_V], arr[_W]) ./ arr[_ρ]
+  elseif sym == :ρe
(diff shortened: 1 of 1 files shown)


test/DGmethods/compressible_Navier_Stokes/mms_bc.jl
```diff
@@ -57,34 +57,66 @@ end
...
+
+function Base.getproperty(state::State, sym::Symbol)
+  arr = getfield(state,:arr)
+  if sym == :ρ
+    @inbounds arr[_ρ]
+  elseif sym == :ρ𝐮
+    @inbounds SVector(arr[_U], arr[_V], arr[_W])
+  elseif sym == :𝐮
+    @inbounds SVector(arr[_U], arr[_V], arr[_W]) ./ arr[_ρ]
+  elseif sym == :E
```
tapios: I suggest we use ``\rho e`` for ``E`` for clarity and consistency with the rest. Similarly for ``\rho q`` for humidities per unit volume etc.

jkozdon:
In terms of performance, `mms_bc` is not great due to the heavy use of `sin` and `cos` in the manufactured solution. That said, functionality wise it has much of the complication including the fact that the `velocities!` only receives a partial state instead of the full state. Unfortunately, it gets the first 4 values. This could be made more complicated by changing
```julia
const _ρ, _U, _V, _W, _E = 1:_nstate
```
to
```julia
const _E, _ρ, _U, _V, _W = 1:_nstate
```

In general, for performance reason, in the future will likely need a number of the functions to get only the partial set of components from `Q`, `QV`, and `aux` that they need instead of the full set. Right now only the `gradient_transform!` callback has this functionality.

The reason this might be needed is that loading data out of the global arrays we don't need will degrade performance (unless we can do it in a way that the on the compiler to optimize the erroneous loads away?)

jkozdon:
Some of the stuff @trontrytel @asraero and @smarras79 probably have more complication as well.

`isentropic_vortex_standalone` might be our best performing kernel since it's the simplest right now. Though I haven't done a thorough profiling of the code.

simonbyrne:
Fixed by #266.

PR #259, opened 2019-06-04
exclude upgraded GPU nodes

simonbyrne:


jkozdon:
Is this what's been causing the recent CI problems? (just noticed a bunch of failures recently).

simonbyrne:
> Is this what's been causing the recent CI problems? (just noticed a bunch of failures recently).

Yes, the problem is that some of the nodes have different versions of the CUDA driver installed, which causes problems when you Pkg.build and run on different machines (https://github.com/JuliaGPU/CUDAdrv.jl/issues/143). 

I may need to change how I handle this.

PR #264, opened 2019-06-10
ODESolvers tests & fixes

mwarusz:
This pull request:
- enables testing of time integrators without any spatial discretization, by using simple ODE problems
- adds tests of time integrators based on the above idea
- fixes small bugs that the new tests exposed and updates the reference results accordingly


jkozdon:
bors try

PR #282, opened 2019-06-24
Reduce GPU tests

simonbyrne:
Only test what is actually GPU-enabled. Should help avoid timeout issues.


simonbyrne:
okay, this seems to have worked. Just now waiting on the Travis/Appveyor.

PR #285, opened 2019-06-26
User input infrastructure

smarras79:
PR to incorporate the user input interface by means of a configuration file. The configuration file content is read and passed to the driver. 

- The following new directory was created:
* ```CLIMA/src/InputOutput/UserInput```
which now contains:
1) ReadConfigurationFile.jl
2) configuration.ini

- The testbed for this PR is the driver ```test/DGmethods/compressible_Navier_Stokes/dc_Smago.jl```
where all geometrical properties of the problem are read from the input file via this infrastructure.

- Next PR in continuation to this one:
* A second PR will include a template JL file that should allow the user to simply define an initial condition.
* All benchmark tests will have their own pre-defined file.


tapios:
The configuration setup looks clean and clear.

- Why not putting it into directory simply called Config?

- Can we separate the configuration aspects from all the rest in this PR? Hard to review it all in one go.

smarras79:
@tapios 

> The configuration setup looks clean and clear. Thanks!
> 
> * Why not putting it into directory simply called Config?
> * Can we separate the configuration aspects from all the rest in this PR? Hard to review it all in one go.

I will do that. I will probably need to cancel this PR and create a new one with only those changes; it was my mis-use of git and my master branch.

src/DGmethods/test/compressible_Navier_Stokes/soundings/create_sounding.m
```diff
@@ -0,0 +1,134 @@
+clear all
```
charleskawczynski: I think that we should convert this into Julia and skip adding the `.dat` files into the repo.

src/DGmethods/DGBalanceLawDiscretizations_kernels.jl
```diff
@@ -706,6 +716,7 @@ function knl_indefinite_stack_integral!(::Val{dim}, ::Val{N}, ::Val{nstate},
               l_int[ind_out, k, i, j] = l_int[ind_out, Nq, i, j]
             end
           end
+        #FIXME
```
charleskawczynski: ?

charleskawczynski:
```suggestion
```

src/DGmethods/test/Sphere/advection_sphere.jl
```diff
@@ -0,0 +1,366 @@
...
+using CLIMA.MPIStateArrays
+using CLIMA.LowStorageRungeKuttaMethod
+using CLIMA.ODESolvers
+using CLIMA.GenericCallbacks
+using LinearAlgebra
+using StaticArrays
+using Logging, Printf, Dates
+using Random
+
+@static if Base.find_package("CuArrays") !== nothing
```
charleskawczynski: I would suggest using the same new code snippet here:
```
@static if haspkg("CuArrays")
  using CUDAdrv
  using CUDAnative
  using CuArrays
  CuArrays.allowscalar(false)
  const DeviceArrayType = CuArray
else
  const DeviceArrayType = Array
end
```

jkozdon:
> I really like the idea of adding an input/config file. However, I think the input file should be hierarchical, rather than flat

Agreed. To add to this:

- why not make the config file an actual julia file. Then the config file is a program itself which allows you to do some cool stuff (like defining the initial condition as a function rather than through a string like "dycom"); this is essentially something that we attempted to start working towards in our [last attempt at an atmos dycore](https://github.com/climate-machine/CLIMA/blob/4cf3a005e4ce7ab29b4643e5017fed0abe28d257/test/Atmos/Dycore/rising_thermal_bubble.jl#L63-L89)
- additional benefit is that multiple places in the code do not need to be modified to add a parameter
- If you used a julia struct or named tuple or something like this, the user would also be able to easily see/query default values

Of course take anything I say with a grain of salt, we all know what interface I'm responsible for...

smarras79:
> > I really like the idea of adding an input/config file. However, I think the input file should be hierarchical, rather than flat
> 
> Agreed. To add to this:
> 
> * why not make the config file an actual julia file. Then the config file is a program itself which allows you to do some cool stuff (like defining the initial condition as a function rather than through a string like "dycom"); this is essentially something that we attempted to start working towards in our [last attempt at an atmos dycore](https://github.com/climate-machine/CLIMA/blob/4cf3a005e4ce7ab29b4643e5017fed0abe28d257/test/Atmos/Dycore/rising_thermal_bubble.jl#L63-L89)
> * additional benefit is that multiple places in the code do not need to be modified to add a parameter
> * If you used a julia struct or named tuple or something like this, the user would also be able to easily see/query default values
> 
> Of course take anything I say with a grain of salt, we all know what interface I'm responsible for...

Yep, @charleskawczynski has shown me some good ideas on how to do it in Julia but yet, keeping it almost unknown to the user that the configuration file is an actual part of the code (i.e. written in Julia).
I am working with him to let him know what type of parameters we expect and their hierarchy so that he can think of a non-flat structure from the algorithmic point of view.

smarras79:
I am closing this pull request because I am working with @charleskawczynski on an improvement of this idea.

PR #289, opened 2019-06-29
Topography reader for NOAA topography files

smarras79:
Topography reader complete for box geometries.

* The reader consists of a new module: `CLIMA/src/Mesh/Topography.jl`
As of now, its functions are called from a test driver `test/DGmethods/compressible_Navier_Stokes/test_external_topography.jl `

To be added to enhance this tool:
* The ability to read DEM and NCDF files. Only NOAA text files can be read right now. 
* Map the topography onto the sphere. This should be quick since these topographies are already provided with respect to their lon/lat coordinates.

Notice: a discussion with @simonbyrne / @charleskawczynski can define how to best interface the reader to a user (i.e. where the calls to the reader should be made: driver vs should they be completely hidden to the user?)

During this implementation a bug was found in the brick construction. See issue reported in https://github.com/climate-machine/CLIMA/issues/290#issue-462267550


jkozdon:
Should we be using git LFS for the binary files? If these change much, they are going to blow up the storage costs of our git repo (which is going to suck whenever the repo is pulled for the first time)

src/Mesh/Grids.jl
```diff
@@ -1,5 +1,6 @@
 module Grids
 using ..Topologies
+using Printf
```
jkozdon: No needed

src/Mesh/Grids.jl
```diff
@@ -228,7 +229,7 @@ function computegeometry(topology::AbstractTopology{dim}, D, ξ, ω, meshwarp,
   Canary.creategrid!(X..., topology.elemtocoord, ξ)
 
   @inbounds for j = 1:length(x)
-    (x[j], y[j], z[j]) = meshwarp(x[j], y[j], z[j])
+      (x[j], y[j], z[j]) = meshwarp(x[j], y[j], z[j])
```
jkozdon: No need for changing spacing (`src/Mesh/Grids.jl` can just be reverted to master, no changes needed for this pull request)

src/Mesh/Topography.jl
```diff
@@ -0,0 +1,159 @@
...
+
+# {{{ READTOPOtxt_header
+"""
+   READTOPOtxt_header(txt_inputfile,nlon, nlat, deltaLon, deltaLat)
+
+   reads the header file containing the details of the topography from a NOAA text file
+"""
+function ReadExternalHeader(header_file_in)
+    
+    @info @sprintf """ Topography header file %s ... DONE""" header_file_in
```
jkozdon: Some of these `@info` lines seem more like `@debug`

smarras79: No, the user needs to know where the code is in terms of reading his/her grid file. The file that the user passes to the code may be broken and hence the would be hanging without knowing where it is hanging.

src/Mesh/Grids.jl
smarras79: This may be a matter of different editors.

jkozdon: My main point was that unnecessary changes should not be included in a commit/pull request to keep things cleaner. The two changes to this file have nothing to do with the PR. Only included the changes you need to in PR makes it much easier to review.

Best to avoid using `git commit -a` as you will pick up unneeded file additions like this. Use instead `git add` or even better `git add -p`.

Also, technically the [CLIMA coding convention](https://climate-machine.github.io/CLIMA/latest/CodingConventions/) stipulates 2 spaces (not the julia 4 spaces) so you may want to adjust your editor settings accordingly.

smarras79: Should I use ```git add -p ONLY_THE_FILE_TO_COMMIT``` instead of ```git commit -am 'message'``` ?

vchuravy:
I have had bad experience with git lfs. Generally I would say big binary files should not be included as part of Git, but rather be downloaded as part of a build script. E.g. `deps/build.jl`

simonbyrne:
I don't think we should be adding binary data files to git (I'm still not particularly keen about the other datafiles already in there, but they're pretty small).

Something like [DataDeps.jl](https://github.com/oxinabox/DataDeps.jl) is probably the best option.

smarras79:
> I don't think we should be adding binary data files to git (I'm still not particularly keen about the other datafiles already in there, but they're pretty small).
> 
> Something like [DataDeps.jl](https://github.com/oxinabox/DataDeps.jl) is probably the best option.

Yes, I agree that we should have data files as part of the repository. 
This said, we should think ahead and explore options (maybe a centralized repository specifically for data like the one @simonbyrne suggests); this software will heavily rely on topographical files and thermodynamic sounding files which we should provide the users, so that handling this properly will be necessary.

christophernhill:
@smarras79 etc... I definitely agree that we should keep large files out of the repo if we can. Anyone who has downstream codes that are doing Pkg.add + https://github.com/climate-machine (like me for example) a lot might be sad if the code repo gets too much larger. 

I have a quick question about #290 too. It can be nice to keep PRs clean, narrow and on a single sub-sub-topic. Would it be better to have a separate PR for #290 code? I know its related to #289, so maybe its a judgement call? Reading the #290 thread it looks like in the end it is maybe adding a new example, and maybe not a bug fix? Perhaps it could be its own PR, as a new example?

smarras79:
I will create a new clean one.

PR #291, opened 2019-06-29
Remove Canary.jl dependency

simonbyrne:
This copies the necessary files from Canary.jl into the main CLIMA repo, which should make it easier to maintain.


simonbyrne:
@lcw I'm just using the GaussQuadrature package, instead of copying the `GaussQuadrature` directory from Canary. Is that okay?

lcw:
Simon Byrne <notifications@github.com> writes:

> @lcw I'm just using the GaussQuadrature package, instead of copying
> the `GaussQuadrature` directory from Canary. Is that okay?

That should be fine.

simonbyrne:
Great. I just need to copy over the tests as well.

PR #293, opened 2019-07-01
Remove kwargs from find_zero

lcw:
Key word arguments can prevent functions from being inlined when run in a
GPU kernel.  For this reason, we remove the keyword arguments from
`find_zero`.


charleskawczynski:
That's odd, should we always avoid keyword args then?

lcw:
Yes, for everything that is going to run on the GPU.  @vchuravy helped me trace it and it was unclear why the compiler decided not to inline the function.

charleskawczynski:
Interesting. It would be nice to figure out how to allow keyword args, but I'm fine with this for now if it helps you guys improve performance.

PR #297, opened 2019-07-03
Names cleanup

trontrytel:
To make naming more consistent among different modules I suggest to:
-  use \rho_water and \rho_ocean for water density. (We already use \rho for air density.)
- use q_vap_saturation instead of saturation_shum for saturation specific humidity. (We use q_ to denote specific humidities of our tracers)

### DIFF
--- a/docs/src/Atmos/Microphysics.md
+++ b/docs/src/Atmos/Microphysics.md
@@ -270,7 +270,7 @@ using Plots
 # https://doi.org/10.1175/1520-0493(1996)124<0487:TTLSLM>2.0.CO;2
 function rain_evap_empirical(q_rai::DT, q::PhasePartition, T::DT, p::DT, ρ::DT) where {DT<:Real}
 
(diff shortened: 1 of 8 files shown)


trontrytel:
bors try

trontrytel:
@charleskawczynski - is that the density you wanted changed?

trontrytel:
bors try

tapios:
The change to ‘q_vap...’ is fine. There was a reason why I had put ‘..._liquid’ for the density of the liquid condensate: here as elsewhere, I wanted to keep formulations and variable names general, so that the model can easily be used for other planetary atmospheres, where the condensing species may not be water (e.g., methane on Titan). So my preference would be to keep ‘liquid’ (and similarly use other variable names that are generic). It has the added advantage of being unambiguous about the water phase we are referring to.

trontrytel:
@tapios  - changed water to liquid

trontrytel:
bors try

PR #298, opened 2019-07-03
Improve surface fluxes

charleskawczynski:
This PR:
 - Removed the Byun functions, since we'll want to use a discretization-independent model (as is done in the Nishizawa paper)
 - Combines the functions into a single function call
 - Now offers to pass the potential temperature flux as an optional argument. If not given, the surface fluxes are computed as is done in the Nishizawa paper. Otherwise, the friction velocity and temperature scale are computed in a decoupled system of equations.
 - Adjusted the docs accordingly

More tests would be great to add, but I'm not sure how to do that at the moment.


charleskawczynski:
bors try

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -45,146 +42,165 @@ module SurfaceFluxes
...
+                       ) where DT<:AbstractFloat
+
+Surface conditions given
+ - `u_ave` volume-averaged horizontal wind speed
+ - `θ_ave` volume-averaged potential temperature
+ - `θ_bar` basic potential temperature
+ - `θ_s` surface potential temperature
+ - `Δz` layer thickness (not spatial discretization)
+ - `z_0_m` roughness length for momentum
+ - `z_0_h` roughness length for heat
```
tapios: Add a roughness length for moisture and other tracers?

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -45,146 +42,165 @@ module SurfaceFluxes
...
+ - `z` coordinate axis
+ - `F_m` momentum flux at the top
+ - `F_h` heat flux at the top
+ - `a` free model parameter with prescribed value of 4.7
+ - `Pr` Prantl number at neutral stratification
```
tapios: It’s probably good to make clear that this is a turbulent Prandtl number (and make sure the same is used in other parameterizations, e.g., Smagorinsky-Lilly and convection schemes).

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -45,146 +42,165 @@ module SurfaceFluxes
...
+                            pottemp_flux::Union{Nothing, DT}=nothing
+                            ) where DT<:AbstractFloat
+
+  L_MO = 1.0
+  if pottemp_flux!=nothing
+
+    @assert 0 <= z <= z_0_m
+    @assert 0 <= z <= z_0_h
+
+    xatol, maxiters = DT(1e-3), 100
```
tapios: Do we really need so high a default ‘maxiters’?

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -45,146 +42,165 @@ module SurfaceFluxes
...
-    stable = z_1/compute_MO_len(ustar_0, flux) >= 0
-    function compute_ψ_m(u)
-      L_MO = compute_MO_len(u, flux)
-      ζ   = z_1/L_MO
-      ζ_0 = z_0/L_MO
-      return stable ? ψ_m_stable(ζ, ζ_0, β_m) : ψ_m_unstable(ζ, ζ_0, γ_m)
-    end
-    function compute_u_ave_over_ustar(u)
-      return (log(z_1 / z_0) - compute_ψ_m(u)) / k_Karman # Eq. 10 in Ref. Byun1990
+    initial_x = [100, 15.0, 350.0] # TODO: improve initial guess
```
tapios: Use values from previous time step by default?

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -45,146 +42,165 @@ module SurfaceFluxes
...
+Buoyancy flux given
+ - `heat_flux_sensible` sensible heat flux,
+ - `heat_flux_latent` latent heat flux,
+ - `T_b` surface boundary temperature,
+ - `alpha0_0` specific volume,
+ - `q_pt` `PhasePartition`,
+"""
+function compute_buoyancy_flux(heat_flux_sensible, heat_flux_latent, T_b, q_pt, alpha0_0)
+  cp_ = cp_m(q_pt)
+  l_vap = latent_heat_vapor(T_b)
```
tapios: I’d use ‘L_vap’ here to be consistent with names in MoistThermodynamics.

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -231,66 +247,102 @@ end
...
 """
-function compute_exchange_coefficients(z, F_m, F_h, a, u_star, θ, flux, Pr)
+function compute_exchange_coefficients(z, F_m, F_h, a, u_star, θ_star, θ_bar, Pr, pottemp_flux)
 
-  L_mo = compute_MO_len(u_star, θ, flux)
-  ψ_m = compute_ψ_m(z/L_mo, L_mo, a)
-  ψ_h = compute_ψ_h(z/L_mo, L_mo, a, Pr)
+  L_MO = compute_MO_len(u_star, θ_bar, pottemp_flux)
+  ψ_m = compute_psi_m(z/L_MO, L_MO, a)
+  ψ_h = compute_ψ_h(z/L_MO, L_MO, a, Pr)
```
tapios: Why Unicode psi here and not in line above?

tapios:
Looks good to me. But how about moisture and other tracer fluxes? I’m assuming they’ll be treated like heat, but it would be good to make that explicit.

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
charleskawczynski: This will change things quite a bit. Which input parameters need their own version for its corresponding moisture/tracer variables? Also perhaps I should nix the case when `pottemp_flux` is given? (this is what was done in SCAMPy, but it's not what is done in the Nishizawa paper)

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
charleskawczynski: No, I don't think so, I can adjust these if we don't end up removing this case (when `pottemp_flux` is passed as an input argument).

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
charleskawczynski: Agreed, I'll pass this as an input argument.

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
charleskawczynski: I changed these because it is very difficult to distinguish the upper and lowercase `psi`. I forgot to replace all of them, I'll spell them all out (none should be exposed to the user anyway since the user should only have to call `surface_conditions`).

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
tapios: We will need moisture fluxes (evaporation) and passive tracer fluxes (for the `n` passive tracers we'll have in the end, where `n` is a variable). It would be good for moisture to have a separate roughness length (by default set equal to heat), and for all other tracers to have one roughness length in common.

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
tapios: Having `pottemp_flux` as optional input would be good for now. We'll need it for many experiments.

charleskawczynski: Ok, I managed to allow both cases using the same code base, but we'll need to re-work some things to get this working on the GPU (also, I don't know if NLSolve is GPU friendly).

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
Reopened in #556

PR #301, opened 2019-07-04
[BL] Remove states_for_gradient_transform

jkozdon:
Based on how I have seen things being used, I think that removing the `states_for_gradient_transform` from the balance law solver will simplify things (and unify the interface).

cc: @lcw @asraero @simonbyrne @smarras79


jkozdon:
bors try

simonbyrne:
Fine by me. I think we will need a more general solution to handle kernel splitting but we can review that then.

PR #266, opened 2019-06-11
initial sketch of DGmodel

simonbyrne:

PR #302, opened 2019-07-04
[BL] Remove preflux from flux computations

jkozdon:
Based on how I have seen things being used, I think that removing the `preflux` from the balance law solver will simplify things (and unify the interface).

cc: @lcw @asraero @simonbyrne @trontrytel @smarras79

### DIFF
--- a/examples/Microphysics/ex_1_saturation_adjustment.jl
+++ b/examples/Microphysics/ex_1_saturation_adjustment.jl
@@ -56,7 +56,7 @@ const _c_z, _c_x, _c_p = 1:_nauxcstate
 
 
 # preflux computation for wavespeed function
-@inline function preflux(Q, _...)
+@inline function preflux(Q)
   @inbounds begin
     # unpack all the state variables
     ρ, ρu, ρw, ρe_tot, ρq_tot = Q[_ρ], Q[_ρu], Q[_ρw], Q[_ρe_tot], Q[_ρq_tot]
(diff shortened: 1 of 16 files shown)


jkozdon:
bors try

jkozdon:
cc @smarras79

charleskawczynski:
Will we still be able to retain the functionality that the `preflux` function offered?

jkozdon:
bors try

jkozdon:
> Will we still be able to retain the functionality that the `preflux` function offered?

What functionality are you worried about losing? The downside to removing it is that whatever is computed in the preflux will be computed twice if it's needed in both the physical flux function and the wave speed calculation if Rusanov flux is used.

I think that folks thought that some magical caching was happening of these values so that they wouldn't be recalculated between kernels calls, but this is not the case. The user still needs to manually cache the values in aux if this is the behavior they desire.

That said, I'm not opposed to keep this functionality if people weigh in to support it (@lcw @smarras79 @trontrytel @asraero?)

jkozdon:
If we do keep it, I realized that we need to correct the description text in `NumericalFluxes.jl` [here](https://github.com/climate-machine/CLIMA/blob/09c6bf5930380a1b2094dec5e1f7a50e50ac7623/src/DGmethods/NumericalFluxes.jl#L12-L19) and [here](https://github.com/climate-machine/CLIMA/blob/09c6bf5930380a1b2094dec5e1f7a50e50ac7623/src/DGmethods/NumericalFluxes.jl#L74-L77)

jkozdon:
bors try

smarras79:
I would suggest the least invasive solution at least until we have very specific test cases up and running.

simonbyrne:
I was planning on getting rid of it in #266 anyway, so lets do it.

simonbyrne:
r+

PR #303, opened 2019-07-04
[BL] Unify aux treatment in init of MPIStateArray

jkozdon:
Sometimes there was an aux array passed to the initial condition function and sometimes not. Since all realistic use case have aux state, we might as well always pass it.


jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

PR #304, opened 2019-07-04
Remove unneeded MPI finalization

jkozdon:


jkozdon:
@lcw and @simonbyrne Please verify that this is correct

jkozdon:
The problem here seems to be that you cannot call `run` on `mpiexec` within an initalized MPI state:
```julia
julia> using MPI

julia> run(`mpirun -n 2 echo "hi"`)
hi
hi
Process(`mpirun -n 2 echo hi`, ProcessExited(0))

julia> MPI.Init()

julia> run(`mpirun -n 2 echo "hi"`)
ERROR: failed process: Process(`mpirun -n 2 echo hi`, ProcessExited(1)) [1]

Stacktrace:
 [1] pipeline_error at ./process.jl:813 [inlined]
 [2] #run#536(::Bool, ::typeof(run), ::Cmd) at ./process.jl:728
 [3] run(::Cmd) at ./process.jl:726
 [4] top-level scope at REPL[4]:1

julia> MPI.Finalize()

julia> run(`mpirun -n 2 echo "hi"`)
hi
hi
Process(`mpirun -n 2 echo hi`, ProcessExited(0))
```

This only pops up right now in `test/Mesh` where some of the tests using MPI directly and other tests use `runmpi` which tries to run `mpiexec`.

Since you cannot reinitialize MPI once it's been finalized we'll have to be careful with our tests.

I'll put in a work around for now, but something we might hit again.

jkozdon:
bors try

jkozdon:
bors try

PR #308, opened 2019-07-08
dycoms3d driver for reference

smarras79:
This PR is meant to incorporate the driver 

`dycoms3d-reference.jl`

This driver was tested on multiple GPUs to reproduce the dycoms test of June 19.

Important suggestion for future testing:
When a new important addition is added to the main code, this test should be run and make sure that it does not break. 
It runs in approx 10 minutes to timeend = 100 s using 1 GPU so that it is a good test that users should always compare against their changes.

@lcw @jkozdon @asraero @simonbyrne


src/Mesh/Grids.jl
```diff
@@ -5,6 +5,7 @@ export DiscontinuousSpectralElementGrid, AbstractGrid
 export ExponentialFilter, CutoffFilter, AbstractFilter
 export dofs_per_element, arraytype, dimensionality, polynomialorder
 export referencepoints
+export grid_stretching
```
jkozdon: I do not think that this should be part of the PR. It's not used in the driver.

jkozdon: I don't think that adding `grid_stretching` should be part of this PR since it's not used in the driver at all.

src/Mesh/Grids.jl
```diff
@@ -433,4 +434,52 @@ end
 
 # }}}
 
+
+"""
+    grid_stretching(DFloat,
```
jkozdon: I do not understand the purpose of this function. How is it any different than passing different ranges to the brick code from the start?

jkozdon:
I like the idea of having a unified dycoms we all work from. Thanks for putting this together.

jkozdon:
bors try

test/DGmethods/compressible_Navier_Stokes/dycoms3d-reference.jl
```diff
@@ -0,0 +1,862 @@
...
+    # User defined timestep estimate
+    # User defined simulation end time
+    # User defined polynomial order 
+    numelem = (Nex,Ney,Nez)
+    dt = 0.005
+    timeend = 100 #14000
+    polynomialorder = Npoly
+    DFloat = Float64
+    dim = numdims
+    if MPI.Comm_rank(mpicomm) == 0
```
jkozdon: this if statement is not needed

src/Mesh/Grids.jl
smarras79: fixed

smarras79:
These files no longer exist so that I don't know why they appear in the upstream. 
Please advise on how to solve it.

smarras79:
<Conflicting files
test/DGmethods/compressible_Navier_Stokes/dycoms.jl
test/DGmethods/compressible_Navier_Stokes/rtb_visc.jl
>

These files no longer exist so that I don't know why they appear in the upstream.
Please advise on how to solve it.

test/DGmethods/compressible_Navier_Stokes/dycoms3d-reference.jl
```diff
@@ -0,0 +1,805 @@
...
+    F_rad = ρ * radiation(aux)
+
+    SijSij = VF[_SijSij]
+
+    #Dynamic eddy viscosity from Smagorinsky:
+    ν_e = sqrt(2SijSij) * C_smag^2 * DFloat(Δsqr)
+    D_e = ν_e / Prandtl_t
+
+    # Multiply stress tensor by viscosity coefficient:
+    τ11, τ22, τ33 = VF[_τ11] * ν_e, VF[_τ22]* ν_e, VF[_τ33] * ν_e
```
simonbyrne: should be multiplied by rho

test/DGmethods/compressible_Navier_Stokes/dycoms3d-reference.jl
```diff
@@ -0,0 +1,805 @@
...
+# Compute the velocity from the state
+gradient_vars!(vel, Q, aux, t, _...) = gradient_vars!(vel, Q, aux, t, preflux(Q,~,aux)...)
+@inline function gradient_vars!(vel, Q, aux, t, u, v, w)
+  @inbounds begin
+    T = aux[_a_T]
+    E, QT = Q[_E], Q[_QT]
+
+    # ordering should match states_for_gradient_transform
+    vel[1], vel[2], vel[3] = u, v, w
+    vel[4], vel[5], vel[6] = E, QT, T
```
simonbyrne: Should this be the gradient of q_t instead of rho*q_t?

smarras79:
Should we close this PR since @simonbyrne  is currently working a new user-code infrastructure which will drastically change this driver?

asraero:
#380 (merged) moves components of this into the `AtmosModel` framework.

asraero:
#380 moves this into the new framework so this PR is now outdated. The older code is still maintained in feature branches by @smarras79 for reference.

PR #310, opened 2019-07-09
\rho_liquid -> \rho_cloud_liq, add \rho_cloud_ice

trontrytel:
@tapios is that what you had in mind?


tapios:
> @tapios is that what you had in mind?

Yes, this looks great.

trontrytel:
bors try

PR #316, opened 2019-07-10
Select GPU devices for MPI runs

lcw:
This pull request adds GPU device selection to the examples and removes an assumption of MPI rank ordering in the tests.


lcw:
> :) But yes this is what the big ML frameworks seem to use.

I used to roll something a long time ago based on `MPI_Get_processor_name` but this is simpler :).

simonbyrne:
Is there a way we could avoid adding more boiler plate to the driver files?

PR #317, opened 2019-07-10
Added shared code/folder to driver tests

charleskawczynski:
This PR
 - includes to `using` statements at the beginning of the drivers
 - includes the logger prep
 - Adds a `print_norm!` function


charleskawczynski:
bors try

charleskawczynski:
bors try

jkozdon:
This seems like a bit clunky of an approach in my opinion, since now were saying all the drivers have the same dependencies. And we're magically defining constants outside of the file.

That said these are just tests.... so maybe we do not care so much?

For the logging, could we add some sort of clima init logging function? (We might be able to do something similar for the GPU and MPI initialization?)

jkozdon:
But I agree, we have a lot of boilerplate code that applies to many (or all) of our drivers...

charleskawczynski:
Yeah, it's clunky. I figure when #266  folds in, we'll be able to write a shared function library and do things more nicely. I think it'll just be nice to reduce the boilerplate for now.

And yeah, I like the idea of an init logging function :)

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
This has gotten stale

PR #323, opened 2019-07-13
First IMEX test

jkozdon:
Adds IMEX tests for
- isentropic vortex
- rising thermal bubble in 2D (without moist thermodynamics routines)

Still todo for IMEX:
- Expand testing to 3D problems
- Test/use the moist thermodynamics routines
- Add/test linearization with velocity in reference states
- preconditioning/initial guess improvements
- explore 1D IMEX
- consider stabilization (Smagorinsky)


jkozdon:
Still todo for IMEX:
- Expand testing to 3D problems
- Test/use the moist thermodynamics routines
- Add/test linearization with velocity in reference states
- preconditioning/initial guess improvements
- explore 1D IMEX

jkozdon:
bors try

jkozdon:
bors try

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try

jkozdon:
@simonbyrne Maybe I just don't understand the current CI framework, but why did this come back as failed?

It seems to have passed [climabot](https://gist.github.com/climabot/bff6678c5cd0b619981c593e4993116b) and [Azure](https://dev.azure.com/spjbyrne/CLIMA/_build/results?buildId=210)

charleskawczynski:
@jkozdon if you click on the red x next to `...bors Try #323:`, you'll see several showing that it's failing
https://dev.azure.com/spjbyrne/CLIMA/_build/results?buildId=212&view=logs&jobId=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e

Not sure why this doesn't match what's on this page, though.

jkozdon:
> if you click on the red x next to `...bors Try #323:`

Thanks @charleskawczynski. Guess I  just got so used to seeing the output in the bors message that I didn't even see the red x.

jkozdon:
bors try

mwarusz:
Seems like this works, but the Linux job timeouts on Azure. If I read

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml&viewFallbackFrom=vsts#timeouts

correctly, we can increase out time limits by setting `timeoutInMinutes` for each job.
Alternatively, we can try to make some DGmethods tests shorter as they seem to be the biggest offender.

@jkozdon @simonbyrne

jkozdon:
> we can try to make some DGmethods tests shorter as they seem to be the biggest offender

I know when I timed things a while ago the bulk of the time we were seeing in our test was Julia initialization time (since each of the tests is a separate Julia session). We might be able to redesign things to batch them together.

That said, it must be the IMEX test that is taking the longest. We could shorten this for CI purposes on azure I suppose (or just not run it through azure since the pieces are tested elsewhere).

jkozdon:
@mwarusz I wonder if we switched to`ARK2GiraldoKellyConstantinescu` if things would run a bit faster? I can check on Monday that results still look OK when this is done.

mwarusz:
@jkozdon Switching to `ARK2GiraldoKellyConstantinescu` should help. So should fixing the performance bug in the update kernel that I discovered. However, my worry is that we will still end up on the verge of timeout. Then I might run into the same issue after adding GMRES tests, for example.

jkozdon:
This has been a continual problem for our testing. Probably trying to batch some of the DG tests together would lead to a significant reduction in CI time (since we avoid tens of Julia startup costs).

PR #324, opened 2019-07-17
[BL] Fix num of aux state in integral_test_sphere

jkozdon:


jkozdon:
The cause of some of the recent CI failures...

PR #329, opened 2019-07-24
Read topography

smarras79:
This PR is a replacement to "Topography reader for NOAA topography files" (#289) which was closed yesterday for being out of date wrt the current DG code by @simonbyrne 

The current PR is meant to add the external topography reader to CLIMA using the current CLIMA structure.

* test/DG/MWE_external_topography.jl

Features: 
* The external topographic files are not incorporated into the CLIMA repository. 
They are automatically downloaded via a call to `wget` from the `Julia` driver into the user's working directory `$CLIMA\_HOME/USERS/TEST/DIRECTORY`. As of now, the available topography files are hosted here `https://web.njit.edu/~smarras/TopographyFiles` (notice: this directory is not directly accessible from the browser).

TO DOs:
* Map the topography grid onto the sphere


test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
```diff
@@ -0,0 +1,267 @@
...
+    topo_files_str       = string(TEST_DIR, "/TopographyFiles/")
+    if !isfile(topo_files_str)
+        #create directory IF directory not created yet:
+        mkpath(topo_files_str)    
+    end
+    
+    #NOAA HEADER FILE
+    header_file_name_and_path = string(topo_files_str, region_file_name_hdr);
+    if !isfile(header_file_name_and_path)
+        url_path = string("https://web.njit.edu/~smarras/TopographyFiles/NOAA/", region_file_name_hdr);
```
jkozdon: Can we do something besides your personal website here?

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
```diff
@@ -0,0 +1,267 @@
...
+        cd(topo_files_str)
+        Base.run(`wget $url_path`)
+        cd(TEST_DIR)
+    end
+    
+    #NOAA COORDINATES FILE
+    body_file_name_and_path = string(topo_files_str, region_file_name_xyz);
+    if !isfile(body_file_name_and_path)
+        url_path = string("https://web.njit.edu/~smarras/TopographyFiles/NOAA/", region_file_name_xyz);
```
jkozdon: Can we do something besides your personal website here?

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
```diff
@@ -0,0 +1,267 @@
...
+                         periodicity = (false, false, false))
+    dt = 0.00001
+    warpfun = warp_external_topography
+    
+    timeend = dt
+    nsteps = ceil(Int64, timeend / dt)
+    
+    run(mpicomm, ArrayType, dim, topl, warpfun,
+        polynomialorder, timeend, DFloat, dt)
+
```
jkozdon: Can we actually check something (as opposed to having the test just be that the code runs?)

src/Mesh/Topography.jl
```diff
@@ -0,0 +1,173 @@
...
+    end
+  
+    #
+    # Store 1D X array in the x direction: size -> nnodes_lat
+    # and order in increasing order
+    l = 0
+    interval = nnodes_lon
+    for k = 1:interval:nnodes
+        l = l + 1
+        yTopo1d[l] = topo_body[k,2]
```
jkozdon:
```julia
yTopo1d = topo_body[1:interval:nnodes,2]
```

src/Mesh/Topography.jl
```diff
@@ -0,0 +1,173 @@
...
+            k = k + 1
+            zTopo2d[i,j] = topo_body[k,3]
+        end
+    end
+    npoin = k
+    
+    #
+    # Store 1D X array in the x direction: size -> nnodes_lat
+    # and order in increasing order
+    for i = 1:nnodes_lon
```
jkozdon: This for loop code be replaced with just a reshape
```julia
xTopo1d = topo_body[1:nnodes_lon,1]
```

src/Mesh/Topography.jl
```diff
@@ -0,0 +1,173 @@
...
+    # Create array
+    #
+    nnodes_lon, nnodes_lat, nnodes = nlon, nlat, nlon*nlat #Linear grid
+    zTopo2d = Array{DFloat, 2}(undef, nnodes_lon, nnodes_lat)
+    
+    xTopo1d = Array{DFloat, 1}(undef, nnodes_lon)
+    yTopo1d = Array{DFloat, 1}(undef, nnodes_lat)
+    
+    k = 0
+    for j = 1:nnodes_lat
```
jkozdon: This for loop code be replaced with just a reshape
```julia
zTopo2d = reshape(topo_body[:,3], nnodes_lon, nnodes_lat);
```

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
smarras79: Any server would work as long as we know it won't be closed.

smarras79:
> Overall this seems reasonable.
> 
> My big comment/request would be if this is the standard file format can we generate a very small file and add a check that the data is read in as expect.
> 
> Using a full PDE solves seems a bit overkill for the functionality that is actually added (namely reading the topography file).

I agree and I tried that but had issues in getting things to work when removing certain parts the driver.  Given the current way drivers are setup now, it was difficult to reduce it to a minimum. With a working example that does not go through the solver I can re-write the test accordingly.

src/Mesh/Topography.jl
smarras79: will change this.

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
smarras79: other than visualizing the grid, I don't know how to check. unless we save a grid file and compare all the next runs to it.

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
jkozdon: I guess this is largely a bigger project question for how to deal with "data" files. @simonbyrne might have thoughts.

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
jkozdon: You could compute the norm of the solution or something like this. We have done this [elsewhere](https://github.com/climate-machine/CLIMA/blob/70414b906c2e962a0ca40eba94d14afc47d50fb0/test/DGmethods/Euler/RTB_IMEX.jl#L457-L464) as a test. At least that way if something changes we have some sort of warning that things broke / have changed.

Since you seem to be using an mms solution, you should also be able to [compute error and convergence rates](https://github.com/climate-machine/CLIMA/blob/70414b906c2e962a0ca40eba94d14afc47d50fb0/test/DGmethods/compressible_Navier_Stokes/mms_bc.jl).

smarras79: I will add this check but it may take me a while because I am leaving in two days and I am wrapping other things up. I will get to this though.

test/DGmethods/compressible_Navier_Stokes/MWE_external_topography.jl
simonbyrne: Can we download them directly from somewhere else? Or are they being preprocessed?

A decent permalink option might be figshare.com? If you have a tarball, I can upload it and set up DataDeps.jl (which handles local caching).

smarras79: The central DB in NOAA is here: https://www.ngdc.noaa.gov/mgg/global/global.html
however, the specific subregions need to be extracted from a global file. This can be done either manually and then download only the region of interest (as they do in WRF) or --which I haven't implemented yet-- automatically extract the region from the global file. The global file is 377MB large.

@simonbyrne

simonbyrne:
From the call: can you open a new issue linking the data source, and what sort of preprocessing we need to do.

akshaysridhar:
See #807

PR #330, opened 2019-07-24
Add GMRES linear solver

mwarusz:
Aside from GMRES another notable change is that iterative solvers now use relative tolerance
for the convergence criterion.


mwarusz:
bors try

simonbyrne:
Out of curiosity, any reason we can't use IterativeSolvers.jl?

mwarusz:
I have considered IterativeSolvers.jl but using it would require a lot of work to adapt our data structures and also require some changes inside IterativeSolvers.jl. For example their GMRES
is not generic with respect to the Krylov vector type since it stores them in a `Matrix{T}`

https://github.com/JuliaMath/IterativeSolvers.jl/blob/17ef2613dff1d80724886797160ee01bd1bd2c71/src/gmres.jl#L7

We would need to mimic the action of our linear operator as a matrix vector multiply,
implement views of `MPIStateArray`s and some BLAS operations on them. Reusing the
same solver on different time steps also did not seem straightforward.

simonbyrne:
Good to know, thanks.

mwarusz:
bors try

jkozdon:
bors try

PR #332, opened 2019-07-29
[MoistThermo] Added constructors

charleskawczynski:
This PR adds new thermodynamic constructors in `states.jl`:
 - one that does not require density
 - one that does not require density, and accepts phase partitions to return `PhaseEquil`
 - one that does not require density, and accepts phase partitions to return `PhaseNonEquil`


charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/MoistThermodynamics/states.jl
```diff
@@ -88,6 +88,25 @@ function LiquidIcePotTempSHumEquil(θ_liq_ice, q_tot, ρ, p)
...
+"""
+    LiquidIcePotTempSHumEquil_no_ρ(θ_liq_ice, q_tot, p)
+
+Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
+
+ - `θ_liq_ice` - liquid-ice potential temperature
+ - `q_tot` - total specific humidity
+ - `p` - pressure
+"""
+function LiquidIcePotTempSHumEquil_no_ρ(θ_liq_ice, q_tot, p)
```
simonbyrne: can we just make it a 3-arg form of `LiquidIcePotTempSHumEquil`?

charleskawczynski: Potentially, I need to check to see if the results/physics are the same for this case.

charleskawczynski: Yeah, there's not a clear way to recover the density (corresponding to the saturation-adjusted temperature), given liquid potential temperature, q_tot and pressure.

charleskawczynski:
Ok, maybe we just keep MoistThermo in Utilities for now. I'll update the PR to just include the constructor, and add docs.

simonbyrne:
I just want to move it out of `utilities`. Maybe create a `common` directory?

My main complaint here is that utilities should be what the name suggests (maybe we can move PlanetParameters as well), as opposed to model components that happen to be shared.

tapios:
> I just want to move it out of `utilities`. Maybe create a `common` directory?
> 
> My main complaint here is that utilities should be what the name suggests (maybe we can move PlanetParameters as well), as opposed to model components that happen to be shared.

Some directory for shared components would be good.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #333, opened 2019-07-30
Moist thermo wrapper

charleskawczynski:
This PR
 - [x] Makes the `mms_bc_atmos.jl` test implement MoistThermodynamics/PlanetParameters
 - [x] adds a wrapper to get a `ThermodynamicState` for different moisture models
 - [x] Simplify other thermo function wrappers (e.g. `soundspeed`)

Now, different moister models will implement their own
`TS = thermo_state(m::MoistureModel, ...)`


charleskawczynski:
Ok, I wasn't able to make the mms case use MoistThermo, but I've annotated the functions to be mms-specific (since MoistThermo will not work with unrealistic pressures/temperatures). Maybe we need a numerical test-friendly functionality from MoistThermo that supports working with pressure/temperature/energy near 1?

charleskawczynski:
bors try

jkozdon:
> Ok, I wasn't able to make the mms case use MoistThermo

So it didn't work to modify the mms solution?

charleskawczynski:
No, unfortunately. The symbolic differentiation did not like the MoistThermo struct :/

charleskawczynski:
I didn't dig into things much more once I saw it break, maybe it's possible, and perhaps more likely if we made a dry thermodynamic state, since its behavior would be far simpler than the moist one.

charleskawczynski:
bors try

jkozdon:
> No, unfortunately. The symbolic differentiation did not like the MoistThermo struct :/

Bummer but makes sense. I think that for the dry case things should work though, since there we can just analytically write down the pressure equation.

Since the forced solution is fairly smooth, I wonder if you could use something like `approxfun` to get a very high order representation? Not saying you should, would just be another option.

charleskawczynski:
Yeah, for now I've just moved some of the mms-specific functionality out. I think we should be able to move them all in eventually, but we may just need more examples to see how to best do it.

charleskawczynski:
bors try

charleskawczynski:
@simonbyrne I ran `julia --project test\DGmethods\compressible_Navier_Stokes\mms_bc_atmos.jl` locally, and the test passed. Also, I'm confused by the difference in results between Azure and SlurmCI.

src/Atmos/Model/moisture.jl
```diff
@@ -15,27 +15,40 @@ end
 function gradvariables!(::MoistureModel, transform::Vars, state::Vars, aux::Vars, t::Real)
 end
 
-# TODO: rewrite to use MoistThermo
-struct DryModel <: Atmos.MoistureModel
+function internal_energy(m::MoistureModel, state::Vars, aux::Vars)
```
tapios: Wouldn't `MoistThermodynamics` be a better place for this function?

src/Atmos/Model/moisture.jl
```diff
@@ -15,27 +15,40 @@ end
...
+  T = eltype(state)
+  q_pt = get_phase_partition(m, state)
+  ρinv = 1 / state.ρ
+  ρe_kin = ρinv*sum(abs2, state.ρu)/2
+  ρe_pot = state.ρ * grav * aux.coord.z
+  ρe_int = state.ρe - ρe_kin
+  # ρe_int = state.ρe - ρe_kin - ρe_pot # FIXME: Should we always include/exclude ρe_pot?
```
tapios: Specific internal energy should always be `ρe_int = state.ρe - ρe_kin - ρe_pot`.

src/Atmos/Model/moisture.jl
```diff
@@ -76,25 +78,24 @@ function gradvariables!(::EquilMoist, transform::Vars, state::Vars, aux::Vars, t
...
   diag_ν = ν isa Real ? ν : diag(ν) # either a scalar or vector
   D_q_vap = D_q_liq = D_q_ice = D_q_tot = diag_ν / Prandtl_t # either a scalar or vector
-  
+
   # diffusive flux of q_tot
-  # FIXME 
+  # FIXME
   ρd_q_vap = state.ρ * (-D_q_vap) .* ∇transform.moisture.q_vap # a vector
   ρd_q_liq = state.ρ * (-D_q_liq) .* ∇transform.moisture.q_liq # a vector
   ρd_q_ice = state.ρ * (-D_q_ice) .* ∇transform.moisture.q_ice # a vector
```
tapios: We don't need the separate vapor, liquid, ice fluxes. Just compute total flux from gradient of total specific humidity. This will be more stable (less cancelation, e.g., between vapor and condensate at cloud boundaries).

src/Atmos/Model/moisture.jl
```diff
@@ -76,25 +78,24 @@ function gradvariables!(::EquilMoist, transform::Vars, state::Vars, aux::Vars, t
...
   diffusive.moisture.ρd_q_tot = ρd_q_vap + ρd_q_liq + ρd_q_ice
 
   D_T = diag_ν / Prandtl_t
-  phase = PhaseEquil(aux.moisture.e_int, q_tot, ρ, aux.moisture.temperature)
-  q = PhasePartition(phase)
+  phase = thermo_state(m, state, aux)
 
   # J is the conductive or SGS turbulent flux of sensible heat per unit mass
-  ρJ = state.ρ * cv_m(q) * D_T * ∇transform.moisture.temperature
+  ρJ = state.ρ * cv_m(phase) * D_T * ∇transform.moisture.temperature
```
tapios: This isn't right. I am currently updating docs. The way to do this is to compute sum `ρJ + ρD` by diffusing specific total energy `e_tot`. We don't need the fluxes separately.

src/Atmos/Model/moisture.jl
```diff
@@ -109,25 +110,15 @@ function diffusive!(::EquilMoist, diffusive::Vars, ∇transform::Grad, state::Va
   e_tot_vap = e_kin + e_pot + I_vap
   e_tot_liq = e_kin + e_pot + I_liq
   e_tot_ice = e_kin + e_pot + I_ice
-  ρD = state.ρ * ((e_tot_vap + R_v*T)*diffusive.moisture.ρd_q_vap + e_tot_liq*diffusive.moisture.ρd_q_liq + e_tot_ice*diffusive.moisture.ρd_q_ice) 
+  ρD = state.ρ * ((e_tot_vap + R_v*T)*diffusive.moisture.ρd_q_vap + e_tot_liq*diffusive.moisture.ρd_q_liq + e_tot_ice*diffusive.moisture.ρd_q_ice)
 
   diffusive.moisture.ρJ_ρD = ρJ + ρD
```
tapios: Just compute this summed flux from gradient of `e_tot`.

src/Atmos/Model/moisture.jl
charleskawczynski: This is just a wrapper function to help facilitate computing internal energy with a simple interface (since MoistThermo has total energy, and does not have kinetic and potential energy).

src/Atmos/Model/moisture.jl
charleskawczynski: I'll update this

src/Atmos/Model/moisture.jl
```diff
@@ -45,89 +57,44 @@ Assumes the moisture components are computed via thermodynamic equilibrium.
...
-  e_kin = 0.5 * sum(abs2,u)
-  e_pot = grav * aux.coord.z
-  e_tot_vap = e_kin + e_pot + I_vap
-  e_tot_liq = e_kin + e_pot + I_liq
-  e_tot_ice = e_kin + e_pot + I_ice
-  ρD = state.ρ * ((e_tot_vap + R_v*T)*diffusive.moisture.ρd_q_vap + e_tot_liq*diffusive.moisture.ρd_q_liq + e_tot_ice*diffusive.moisture.ρd_q_ice) 
-
-  diffusive.moisture.ρJ_ρD = ρJ + ρD
+  # diffusive flux of total energy
+  diffusive.moisture.ρJ_ρD = state.ρ * (-D_T) .* ∇transform.transform.moisture.e_x
```
simonbyrne: @tapios is this what you had in mind?

src/Atmos/Model/moisture.jl
```diff
@@ -45,89 +57,44 @@ Assumes the moisture components are computed via thermodynamic equilibrium.
...
+  R_m = gas_constant_air(q)
+  T = aux.moisture.temperature
+  e_tot = state.ρe * invρ
 
-  transform.moisture.q_vap = q.tot - q.liq - q.ice
-  transform.moisture.q_liq = q.liq
-  transform.moisture.q_ice = q.ice
-  transform.moisture.temperature = aux.moisture.temperature
+  # a better name?
+  transform.moisture.e_x = e_tot + R_m*T
```
simonbyrne: @tapios is this what you had in mind?

src/Atmos/Model/moisture.jl
tapios: Yes, looks good. I'd just give `\rho J_\rho D' a more intuitive name, e.g., `\rho_SGS_energyflux` or something like it. I also do not find the `e_x` very clear (subscript `x` is confusing--for us it usually means the zonal coordinate)

src/Atmos/Model/moisture.jl
tapios: Yes, just a better name. `e_x` without the kinetic energy we call moist static energy. So maybe this should be `moist dynamical energy`, or `dyn_e_tot` or something like it.

tapios: Or, perhaps better: "total enthalpy" (a slight abuse of terminology, because total enthalpy does not usually contain potential energy). It's strange this quantity does not have an established name...

charleskawczynski:
@simonbyrne I'm a bit confused why commit 7911add seemed to pass, and now saturation adjustment is not converging.

charleskawczynski:
bors try

PR #336, opened 2019-08-01
CPU optimizations

mwarusz:
The main changes are

- faster vectorized reductions on the CPU
- faster `volumerhs!` kernel due to reordering some loops
- kernel for computing linear combinations in linear solvers


mwarusz:
bors try

mwarusz:
bors try

azure-pipelines.yml
```diff
@@ -9,6 +9,8 @@ trigger:
 jobs:
 - job: Linux
 
+  timeoutInMinutes: 0
+
```
simonbyrne: what does this do?

lcw: I believe it allows the azure jobs to run the maximum time (more than the default of 60 minutes).

src/LinearSolvers/LinearSolvers_kernels.jl
```diff
@@ -0,0 +1,20 @@
+using Requires
+@init @require CUDAnative = "be33ccc6-a3ff-5ff2-a52e-74243cff1e17" begin
+  using .CUDAnative
+end
+
+function linearcombination!(Q, cs, Xs, increment::Bool)
```
simonbyrne: This looks like a mat-vec multiply: is doing this via vectors of tuples really more efficient?

azure-pipelines.yml
simonbyrne: That doesn't appear to be documented. Maybe just set it to the maximum (360 minutes)?

mwarusz: Quoting from https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml&viewFallbackFrom=vsts#timeouts

"The timeoutInMinutes allows a limit to be set for the job execution time. When not specified, the default is 60 minutes. When 0 is specified, the maximum limit is used (described above)."

simonbyrne: Ah, I missed that. Thanks.

src/LinearSolvers/LinearSolvers_kernels.jl
mwarusz: It is a mat-vec multiply, but it is special in a way. The size of the matrix is m x length(state_vector) where m <= 10. Not sure if viewing this operation as a mat-vec
multiply is the best way to reason about its performance. Anyway, this kernel is probably
not yet optimal, but it is faster that what was in there before (a deep broadcasted expression).

test/LinearSolvers/poisson.jl
```diff
@@ -77,7 +77,7 @@ end
   @inbounds Q[1] = prod(sol1d.(xs[1:dim]))
 end
 
-function run(mpicomm, ArrayType, DFloat, dim, polynomialorder, brickrange, periodicity, linmethod)
+function run(mpicomm, ArrayType, DFloat, dim, polynomialorder, brickrange, periodicity, linmethod, tol)
```
charleskawczynski:
```suggestion
function run(mpicomm, ArrayType, DFloat, dim, polynomialorder, brickrange, periodicity, linmethod, tol=1e9)
```

test/LinearSolvers/poisson.jl
```diff
@@ -131,28 +130,29 @@ let
 
   polynomialorder = 4
   base_num_elem = 4
+  tol = 1e-9
```
charleskawczynski: Delete this in favor of default input?

mwarusz: Cannot do that because `tol` is also used outside of `run` in calculating `isapprox` relative tolerance. Thanks for taking a look at this though, I think I found a better way to deal with `tol`.

PR #338, opened 2019-08-05
access first interior at boundary nodes

asraero:
1) kernel + function to loop over first element to extract and store first-interior values (currently accessible only via boundary nodes in driver) 
2) stratocumulus ex: bulk formulae source terms 
@smarras79 @lcw @jkozdon

### DIFF
--- a/src/DGmethods/DGBalanceLawDiscretizations.jl
+++ b/src/DGmethods/DGBalanceLawDiscretizations.jl
@@ -735,6 +735,66 @@ function indefinite_stack_integral!(disc::DGBalanceLaw, f, Q, out_states,
                                          1:nhorzelem, Val(out_states)))
 end
 
+function aux_firstnode_values!(disc::DGBalanceLaw, Q, out_states, in_states,
+                                    P=disc.auxstate)
+  grid = disc.grid
+  topology = grid.topology
+  @assert isstacked(topology)
+
+  dim = dimensionality(grid)
+  N = polynomialorder(grid)
+
+  auxstate = disc.auxstate
+  nauxstate = size(auxstate, 2)
+  nstate = size(Q, 2)
+
+  vgeo = grid.vgeo
+  device = typeof(Q.Q) <: Array ? CPU() : CUDA()
+
+  nelem = length(topology.elems)
+  Nq = N + 1
+  Nqk = dim == 2 ? 1 : Nq
+
+  nvertelem = topology.stacksize
+  nhorzelem = div(nelem, nvertelem)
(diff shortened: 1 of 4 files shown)


lcw:
@jkozdon and I took a look and we would suggest changing
`numerical_boundary_flux!(F, nM, QM, VM, auxM, QP, VP, auxP, bctype, t)`
to `numerical_boundary_flux!(F, nM, QM, VM, auxM, QM1, VM1, auxM1, QP, VP, auxP, bctype, t)` where `QM1, VM1, auxM1` are the values at the first node inside the element.  Would that work for you?  If so I can try to implement the change.

jkozdon:
Should we move these extra arguments to the end so they can be more easily ignored with an `_...`?

In particular this interface:
```julia
numerical_boundary_flux!(F, nM, QM, VM, auxM, QP, VP, auxP, bctype, t, QM1, VM1, auxM1)
```

asraero:
@lcw the `bcstate` solution sounds good : however for now if we can access these quantities in the source terms as well it would be useful. @jkozdon, I agree we could add the first-interior as extra at the end so they can be ignored in simpler cases.

lcw:
Why do you need these in the source terms as well?

Storing the surface terms in the volume aux fields is going to decrease the performance of all of the kernels because right now all aux states are loaded in most of the kernels.

simonbyrne:
Out of curiousity: why would it be better to do this via the numerical flux function rather than a boundary condition function?

asraero:
@lcw Having reviewed the b.c. requirements ; the first-nodal values aren't needed in sources - I'll follow up once I've tested the code, I think that we can use the `bcstate` / `boundary_flux` functions for what is needed.

lcw:
@simonbyrne We currently impose boundary conditions weakly through the `numerical_boundary_flux!` function.  Do you have an example of a boundary condition function?

simonbyrne:
What I meant was: what is the effective difference between specifying a `bcstate` function vs a `numerical_boundary_flux!`, and when should one choose one over the other?

jkozdon:
> What I meant was: what is the effective difference between specifying a `bcstate` function vs a `numerical_boundary_flux!`, and when should one choose one over the other?

The Rusanov numerical flux uses the `bcstate` to define the `numerical_boundary_flux!`. So the `bcstate` function will be similarly modified to pass this through to the user level.

asraero:
This can be closed in favor of #380. The first-interior information is limited to the `numerical_boundary_flux` (at the boundary faces only, can be ignored in existing `bcstate` usage via `_...` ) rather than being passed around in `auxstate`

PR #345, opened 2019-08-08
Add type restrictions to MoistThermo

charleskawczynski:
- Removed some type-unstable possibilities (e.g., `T{Float32}, q::PhasePartition{Float64}` when `q::PhasePartition` is given)
 - Added `<:Real` restriction


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -46,9 +46,9 @@ include("states.jl")
 The specific gas constant of moist air given
  - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
 """
-gas_constant_air(q::PhasePartition{DT}) where {DT} =
+gas_constant_air(q::PhasePartition{DT}) where {DT<:Real} =
   DT(R_d) * ( 1 +  (DT(molmass_ratio) - 1)*q.tot - DT(molmass_ratio)*(q.liq + q.ice) )
-gas_constant_air() = gas_constant_air(PhasePartition(0.0))
+gas_constant_air(::Type{DT}) where {DT<:Real} = gas_constant_air(PhasePartition{DT}(zeros(DT, 3)...))
```
simonbyrne: I think this might be the issue: the compiler isn't so good at handling splatting of arrays, since it doesn't know their length (as opposed to tuples which have known length at compile time).

charleskawczynski: Ok, should I update this with `PhasePartition{DT}(DT(0),DT(0),DT(0))`? Or just use the outer constructor like before `PhasePartition(DT(0))`?

simonbyrne: `PhasePartition{DT}(DT(0),DT(0),DT(0))` would probably be simplest.

charleskawczynski:
bors try

PR #346, opened 2019-08-09
Made DryModel use PhaseDry

charleskawczynski:
- Made the `DryModel` use `PhaseDry`
 - Removed unecessary function call to `get_phase_partition`
 - Removed `get_phase_partition` wrapper for `PhaseDry`


charleskawczynski:
bors try

PR #356, opened 2019-08-12
Run integration tests in slurm testing

### LINKED ISSUE #337: Integration Testing
Right now the integration tests are not being run automatically. Should we turn on integration testing via `JULIA_CLIMA_INTEGRATION_TESTING` for the slurm runs?

cc: @mwarusz

jkozdon:
To close #337

### DIFF
--- a/.slurmci/cpu.sh
+++ b/.slurmci/cpu.sh
@@ -14,3 +14,6 @@ export OPENBLAS_NUM_THREADS=1
 module load cmake/3.10.2 openmpi/4.0.1
 
 mpiexec julia --color=no --project $1
+if grep -q JULIA_CLIMA_INTEGRATION_TESTING $1; then
+  JULIA_CLIMA_INTEGRATION_TESTING=true mpiexec julia --color=no --project $1
+fi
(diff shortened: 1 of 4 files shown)


jkozdon:
bors try

lcw:
If the tests pass and don't take too long, looks good to me!

jkozdon:
@simonbyrne: Do you know if this Is this still running on the caltech cluster? If so, probably guessing that one of the tests is taking a long time with full integration testing turned on...

jkozdon:
bors try-

jkozdon:
bors try

simonbyrne:
Yes, it does look like some are timing out, but there are also some failures.

You can try to extend the timeout?

jkozdon:
> Yes, it does look like some are timing out, but there are also some failures.

For some reason my initial `bors try` failed to launch. Hence my question this morning. After posting the message to you I decided to `bors try` again and it went through.

jkozdon:
This seems like it'll be a bit of mess to get working. Since most of these tests are getting phased out, it seems like it would be a waste of time to fix all the issues.

@lcw wondering if we should instead just make sure that the tests we move over to the new framework run the full suite when running on slurm.

lcw:
Jeremy E Kozdon <notifications@github.com> writes:

> @lcw wondering if we should instead just make sure that the tests we
> move over to the new framework run the full suite when running on
> slurm.

This seems reasonable.

PR #367, opened 2019-08-15
Helds-Suarez test

mwarusz:


examples/DGmethods/held_suarez_forcing.jl
```diff
@@ -0,0 +1,1017 @@
...
+    temp0 :: DFloat = 315
+    temp_min :: DFloat = 200
+    hd :: DFloat = 7000 #from Smolarkiewicz JAS 2001 paper    
+    σ_b :: DFloat = 7 / 10
+    
+    #Compute Rayleigh Damping terms from Held-Suarez 1994 paper
+    (r, λ, φ) = cartesian_to_spherical(x, y, z)
+    h = r - DFloat(planet_radius) # height above the planet surface
+    σ = exp(-h/hd) #both approx of sigma behave similarly
+#    σ = P/p0        #both approx of sigma behave similarly
```
tapios: This second (commented out version) is closer to the definition of `sigma` (it takes temperature variations in the atmosphere into account')

examples/DGmethods/held_suarez_forcing.jl
```diff
@@ -0,0 +1,1017 @@
...
+    c = max(0, Δσ)
+    T_eq = ( temp0 - ΔT_y*sin(φ)^2 - ΔT_z*log(σ)*cos(φ)^2 )*π
+    T_eq = max(temp_min, T_eq)
+    kt = ka + (ks-ka)*c*cos(φ)^4
+    kv = kf*c
+    kt = kt*(1 - T_eq/T)
+
+    if ramp_up_forcings
+      t_ramp :: DFloat = 86400
+      ramp_factor = (1 - exp(-2t / t_ramp)) / (1 + exp(-2t / t_ramp))
```
tapios: Why do we need this ramp?

examples/DGmethods/held_suarez_forcing.jl
```diff
@@ -0,0 +1,1017 @@
...
+    P = air_pressure(T, ρ)
+    
+    #Create Held-Suarez forcing
+    (kv,kt,T_eq)=held_suarez_forcing(x,y,z,T,P,t) 
+    
+    #Apply forcing
+    if use_held_suarez_forcings
+      S[_ρu ]  -= kv*ρu
+      S[_ρv ]  -= kv*ρv
+      S[_ρw ]  -= kv*ρw
```
tapios: Same here. No w terms for hydrostatic model configuration.

examples/DGmethods/held_suarez_forcing.jl
```diff
@@ -0,0 +1,1017 @@
...
+    σ = exp(-h/hd) #both approx of sigma behave similarly
+#    σ = P/p0        #both approx of sigma behave similarly
+    Δσ = (σ - σ_b)/(1 - σ_b)
+    π = σ^(R_d/cp_d)
+    c = max(0, Δσ)
+    T_eq = ( temp0 - ΔT_y*sin(φ)^2 - ΔT_z*log(σ)*cos(φ)^2 )*π
+    T_eq = max(temp_min, T_eq)
+    kt = ka + (ks-ka)*c*cos(φ)^4
+    kv = kf*c
+    kt = kt*(1 - T_eq/T)
```
tapios: It's quite confusing to have the part of the source that involves the state variable (or aux variable) here, and the part involving the reference state in another part of the code. Makes it difficult to piece together what is going on.

examples/DGmethods/held_suarez_forcing.jl
```diff
@@ -0,0 +1,1017 @@
...
+    
+    #Create Held-Suarez forcing
+    (kv,kt,T_eq)=held_suarez_forcing(x,y,z,T,P,t) 
+    
+    #Apply forcing
+    if use_held_suarez_forcings
+      S[_ρu ]  -= kv*ρu
+      S[_ρv ]  -= kv*ρv
+      S[_ρw ]  -= kv*ρw
+      S[_dρe ] -= ( kt*ρ*cv_d*T + kv*(ρu*ρu + ρv*ρv + ρw*ρw)/ρ )
```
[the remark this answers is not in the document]
lcw: Note, here w is the cartesian component of the velocity, not the radial velocity

examples/DGmethods/held_suarez_forcing.jl
lcw: When we were trying to figure out why the code was unstable, we decided to try to smoothly increase the forcing.  In the end, I don't think it was actually needed to stabilize the code.

examples/DGmethods/held_suarez_forcing.jl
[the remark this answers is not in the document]
tapios: Ah, thanks, I had missed that. It remains true, however, that the vertical velocity component (however you get it here) should not appear in the kinetic energy for a consistent hydrostatic configuration.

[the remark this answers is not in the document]
tapios: And if w is not vertical velocity, then why is the w component of the Coriolis acceleration set to zero? Only the vertical component is set to zero in the thin-shell approximation (which you do not need to make).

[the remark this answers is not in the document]
mwarusz: Note that the `PDE_level_hydrostatic_balance` flag means that the reference state
in an exact hydrostatic balance and therefore some terms cancel out in the governing PDEs.
It does not mean that we are solving hydrostatic equations. Perhaps the name is a bit unfortunate.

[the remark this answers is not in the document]
tapios: Ok, thanks. And is the w Coriolis component then in fact the vertical?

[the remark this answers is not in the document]
mwarusz: The `w` Coriolis component is not in the vertical. We are solving equations for Cartesian velocity components. If Earth's axis of rotation is assumed to be along `z` then there is no Coriolis force in this direction. This is why this component is zero.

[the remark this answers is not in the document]
tapios: Ok, makes sense. It’s pretty confusing to use the conventional u, v, w (usually eastward, northward, upward) for velocities in these Cartesian coordinates. We should change names to avoid that confusion.

examples/DGmethods/held_suarez_balanced.jl
```diff
@@ -0,0 +1,567 @@
...
+    r = hypot(x, y, z)
+    λ = atan(y, x)
+    φ = asin(z / r)
+    (r, λ, φ)
+end
+#md nothing # hide
+
+# FXG: reference state
+# Setup the reference state based on a N=0.0158725 uniformly stratified atmosphere with θ0=315K
+function auxiliary_state_initialization!(T0, aux, x, y, z)
```
tapios: This looks much more complicated than HS stipulates. HS initial state is just constant temperature T_r. Initial density then is rho = rho_0 exp(-z//H), where H = R_d T_r/g and rho_0 = p_MSLP/R_d T_r. Add some noise to rho to break symmetry.

examples/DGmethods/held_suarez_balanced.jl
```diff
@@ -0,0 +1,567 @@
...
+    
+    ## Convert to Spherical coordinates
+    (r, λ, φ) = cartesian_to_spherical(DFloat, x, y, z)
+
+    ## Calculate the geopotential ϕ
+    h = r - DFloat(planet_radius) # height above the planet surface
+    ϕ = gravity * h
+
+    ## Reference Potential Temperature from constant Brunt-Vaisala definition
+    θ_ref = θ0 * exp( N_bv^2*h/gravity )
```
charleskawczynski: Was this `θ0` on purpose? If so, why not `θ_ref = DFloat(0)`?

mwarusz: Not sure I understand, `θ0` is defined as`θ0 = DFloat(315)` above .

charleskawczynski: Oops, haha, I got confused by the syntax highlighting and forgot that variable names are not commutative (`θ0 != 0θ`)!

PR #373, opened 2019-08-17
1D IMEX Demo

lcw:
Here is a demonstration in the old BalanceLaw framework how to do 1D IMEX.

To make it performant we may want to replace the iterative solver with a direct solver.  We may want to combine the 1D volume and face kernels.  And of course, we need to move this to the new framework.

Co-authored-by: Jeremy E Kozdon <jekozdon@nps.edu>


jkozdon:
Thanks for pushing it over the line tonight @lcw!

@fxgiraldo thanks for helping us with the formulation!

test/DGmethods_old/Euler/RTB_1DIMEX.jl
```diff
@@ -0,0 +1,427 @@
...
+  S .= 0
+  lin_source_geopotential!(S, Q, aux, t)
+end
+@inline function lin_source_geopotential!(S, Q, aux, t)
+  @inbounds begin
+    δρ = Q[_δρ]
+    ∇ϕ = SVector(aux[_a_ϕ_x], aux[_a_ϕ_y], aux[_a_ϕ_z])
+    S[_ρu] -= δρ * ∇ϕ[1]
+    S[_ρv] -= δρ * ∇ϕ[2]
+    S[_ρw] -= δρ * ∇ϕ[3]
```
tapios: Does it make a substantial difference performance-wise whether this buoyancy term is included in the linear solve or not? (I think it makes sense to include it here for a GCM configuration, I’m just wondering how much, if any, overhead it creates.)

test/DGmethods_old/Euler/RTB_1DIMEX.jl
```diff
@@ -0,0 +1,427 @@
...
+  @inbounds begin
+    P = pressure(Q, aux)
+    δρ, δρe = Q[_δρ], Q[_δρe]
+    ρu⃗ = SVector(Q[_ρu], Q[_ρv], Q[_ρw])
+    ρ0, ρe0 = aux[_a_ρ0], aux[_a_ρe0]
+    ρ = ρ0 + δρ
+    ρe = ρe0 + δρe
+    ρinv = 1 / ρ
+    u⃗ = ρinv * ρu⃗
+    abs(n⃗' * u⃗) + sqrt(ρinv * γ * P)
```
tapios: The speed of sound here and elsewhere is just an approximation, valid in the unstratified case. In the presence of stratification, there are (usually small) correction terms that depend, e.g., on the buoyancy frequency squared. (They are responsible for noise, e.g., from freeways being refracted at a stable inversion in winter. People in Pasadena experience this...) How important is it that the sound speed is accurate here?

tapios:
Thanks for putting this together so quickly!

test/DGmethods_old/Euler/RTB_1DIMEX.jl
lcw: Good question.

If we are using a direct solver then I think the performance penalty is only in the initialization.

If we are using an iterative method to solve the system, there has to be some performance penalty because we are loading more data but we would have to measure to find out how much (I am guessing not much).  Also I don't know the impact it has on the number of iterations of the iterative solver, maybe @fxgiraldo has some experience.

test/DGmethods_old/Euler/RTB_1DIMEX.jl
lcw: This is an approximate Riemann solver, like the ones used in finite volume methods.  Usually, we just want the maximum speed.  That could be locally or globally.  If it is too large it can add more dissipation than is needed.  But I don't think it needs to be extremely accurate.

fxgiraldo: I agree with @lcw .  We just need an upper bound on the wave speed.

tapios: Well. the Laplace wavespeed we use here is not the upper bound. It's probably good to within a few percent. Maybe that's good enough based on what @lcw says?

test/DGmethods_old/Euler/RTB_1DIMEX.jl
```diff
@@ -0,0 +1,427 @@
...
+    e0 = ρinv0 * ρe0
+
+    P0 = (γ-1)*(ρe0 - ρ0 * ϕ)
+    δP = (γ-1)*(δρe - δρ * ϕ)
+
+    p0 = ρinv0 * P0
+
+    F[:, _δρ ] = ρu⃗
+    F[:, _ρu⃗ ] = δP * I + @SMatrix zeros(3,3)
+    F[:, _δρe] = (e0 + p0) * ρu⃗
```
tapios: This looks odd. Shouldn't there be `(e0 + p0/\rho0)` here?

jkozdon: The division `P0 / \rho0` happens in line 231 above. i'll grant that the notation I choose for the proof of concept implementation was not the best! But `P` was pressure and `p` was `P / \rho`.

tapios: Ok, thanks.

PR #374, opened 2019-08-17
Encapsulate RootSolvers results

### LINKED ISSUE #365: Return more information from RootSolvers
It might be nice if we return a struct from RootSolvers containing, e.g., `x::T, f(x)::T, converged::Bool, iter_used::Int`. Just so that we can diagnose stuff easier from outside the function call. Even better, if we can dispatch between returning debug info to simply `x::T, converged::Bool`.

charleskawczynski:
Added

`CompactSolution <: SolutionType`
`VerboseSolution <: SolutionType`

to RootSolvers. Now,
`sol = find_zero(..., CompactSolution(),...)`
yields
`sol.root`
`sol.converged`

But if you want more information, you can use:
`sol = find_zero(..., VerboseSolution(),...)`
to get
`sol.root`
`sol.converged`
`sol.err`
`sol.iter_performed`

The latter may be useful for debugging / making a convergence map for our moist thermo state.

Also, added `Float32` tests (including type-stability) for CPU/GPU.

Closes #365


charleskawczynski:
bors try

charleskawczynski:
bors try

tapios:
Looks good to me. @simonbyrne?

charleskawczynski:
Re-opened, after git-ocalypse, in #548

PR #375, opened 2019-08-19
Added TurbConv to new framework

charleskawczynski:
This PR
 - Adds the sub-domains for EDMF model
 - Adds aux functions for computing several source terms and aux variables
   - [x] EDMF buoyancy
   - [x] EDMF entrainment-detrainment
   - [x] EDMF mixing length
   - [x] EDMF moisture auxiliary functions
 - Tests that an `AtmosModel` can be constructed with the `EDMF` sub-model

Design questions:

 - Should any of the EDMF fields be included by default (e.g. velocity/energy?) Right now, they are composed to allow for flexibility, but accessing velocities is a bit verbose (`atmosmodel.turbconv.momentum.ρu`)

Additional needed functionality:

 - [ ] Need to accommodate e.g., `vars_state(    ::AreaFractionModel, T, N) = @vars()` having two arguments here is not currently supported [example here](https://github.com/climate-machine/CLIMA/blob/master/src/DGmethods/balancelaw.jl#L37)
 - [ ] Surface fluxes / BCs
 - [ ] EDMF Momentum sources
 - [ ] EDMF Momentum fluxes
 - [ ] EDMF Energy sources
 - [ ] EDMF Energy fluxes
 - [ ] EDMF moisture sources
 - [ ] EDMF moisture fluxes


charleskawczynski:
Reopened in #555

PR #376, opened 2019-08-19
Add hydrostaticstate to AtmosModel

charleskawczynski:
This PR:

 - Adds a reference state to `AtmosModel`
 - Adds kernels to compute the reference state
 - Adds a new MoistThermo constructor used in the EDMF reference state computations


src/Atmos/Model/refstate.jl
```diff
@@ -0,0 +1,74 @@
...
+
+Column-wise integral kernel to compute reference state fields:
+  - ρ density
+  - p presure
+  - T temperature
+
+For accuracy, the logarithm of the expression is in the integrand, which must be exponentiated
+after computing the integral.
+"""
+function column_integration_kernel!(m::ReferenceState, am::AtmosModel, state::Vars, aux::Vars)
```
charleskawczynski: @simonbyrne Here's an example of a kernel that will need to be called at initialization. This answers @tapios request to add reference states as laid out in 5.4 of the design docs.

I'll add a `NoReferenceState` to use as default so that we can adapt to the other existing cases, but we'll need a test for when a non-`NoReferenceState` is used. I'll add such a test once this kernel call is supported.

jkozdon:
@charleskawczynski If you want to see how the reference state will be needed in the IMEX solvers, you can see that in [`RTB_IMEX.jl`](https://github.com/climate-machine/CLIMA/blob/52d173ce5eeb0917906990128529fd144f82ec65/test/DGmethods_old/Euler/RTB_IMEX.jl); search for `_a_ρ0` and `_a_ρe0`

src/Atmos/Model/refstate.jl
```diff
@@ -0,0 +1,74 @@
...
+struct ReferenceState{T} <: AbstractReferenceState{T}
+  T_ground::T
+  P_ground::T
+  q_tot_ground::T
+end
+
+vars_state(m::AbstractReferenceState    , DT) = @vars()
+vars_gradient(m::AbstractReferenceState , DT) = @vars()
+vars_diffusive(m::AbstractReferenceState, DT) = @vars()
+vars_aux(m::AbstractReferenceState      , DT) = @vars(ρ::DT, p::DT, T::DT)
```
jkozdon: Should `p` be in the reference state of `ρe`? Seems that it would make more sense to match that name here to whats in the [`AtmosModel.jl`](https://github.com/climate-machine/CLIMA/blob/52d173ce5eeb0917906990128529fd144f82ec65/src/Atmos/Model/AtmosModel.jl#L38-L46)

jkozdon: My thinking here is when the full state is needed the reference state can just be added.

charleskawczynski: Good point. I'm not sure how to answer that. In the EDMF model, we've typically only used the reference `ρ, p, T` (maybe this is due to available measurements?). We can certainly add these fields if it makes sense to mirror `AtmosModel`.

tapios:
@charleskawczynski I cleaned up the definition of the hydrostatic state, with the family of temperature profiles we discussed, in the design doc. I'd be grateful if you implement this. (You'll probably need to distinguish zero lapse rate (isothermal atmosphere) from nonzero lapse rate, to avoid divisions by zero.)

tapios:
I also suggest you call this `hydrostatic_state` or something like it, because this is what it is. I think it is clearer than reference state, and we will use it for several purposes, not just for the reference state for linearizations.

src/Atmos/Model/refstate.jl
tapios: Adding `ρe` makes sense--we'll need it.

tapios: You'd also want to add a `q_tot` that corresponds to a fixed and prescribed relative humidity RH, as described in the docs. We'll needs this for spinning up moist models (otherwise it takes years of simulated time to moisten the stratosphere).

charleskawczynski: Ok, I'll make sure to add `ρe` when updating according to the new docs. And I'll take a look at the RTB case.

PR #377, opened 2019-08-20
added LocalGeometry object to query length scale and coordinates

simonbyrne:
Thoughts welcome.


simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
I gave it the name before putting `coord` in there, so i realise it isn't really "local" any more. Any suggestions for a better name?

We might also be able to use this to capture the gravity potential as well (though then the name is even less appropriate)

charleskawczynski:
> I gave it the name before putting `coord` in there, so i realise it isn't really "local" any more. Any suggestions for a better name?

How is it not local?

simonbyrne:
it's still local in the sense that it pertains to the particular node, but not in the differential geometry sense (because it stores the global coordinates).

charleskawczynski:
Ahh, we need a naming convention for our coordinate systems to distinguish these differences.

charleskawczynski:
How about

`GlobalCoord` for `x1,x2,x3` and
`ElemCoord` for `ξ1, ξ2, ξ3`

simonbyrne:
`CartesianCoord`?

simonbyrne:
bors try

PR #351, opened 2019-08-12
Dry rayleigh-benard test

asraero:
1) Working dry rayleigh-benard test in old framework
2) Porting dry test to new AtmosModel

PR #378, opened 2019-08-20
Communicate aux states, zero out source terms

simonbyrne:
This includes 2 of the fixes from #351.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -194,6 +194,7 @@ Computes flux `S(Y)` in:
 ```
 """
 function source!(m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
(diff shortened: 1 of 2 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -194,6 +194,7 @@ Computes flux `S(Y)` in:
 ```
 """
 function source!(m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
+  fill!(getfield(source, :array), zero(eltype(source)))
```
jkozdon: I know that I'm a bit late to the game here since the PR is closed, but why not zero the source in `DGmodel_kernels.jl`?

charleskawczynski: I like this idea. It removes the responsibility from the user.

PR #380, opened 2019-08-20
DYCOMS: move LES driver to AtmosModel

asraero:
- [x] Initial Condition + driver w/ domain description
- [x] Flux BCs
- [x] Access to first-interior through boundary flux  -> note faceid is hardcoded for bottom walls based on topology.
- [x] Sponge - Linear Rayleigh type damping. 
- [x] VTK state output
- [x] Radiation (requires version of `dof_iteration` or `preodefun` in new framework)
- [x] Sources
Status: CI Testing


jkozdon:
> Add this to the test suite!

It would also be good if the test was some statistic or metric that will change if the code is wrong (not that it just runs)

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  DF         = eltype(state)
+  xvert::DF  = z
+
+  # Constants as defined in Stevens et. al (2005) 
+  R_d::DF     = 287.0
+  cp_d::DF    = 1015.0
+  cp_v::DF    = 1859.0
+  cp_l::DF    = 4181.0
+  Lv::DF      = 2.47e6
+  epsdv::DF   = 1.61
```
tapios: Since (from what I understand) these constants are not overriding the `PlanetParameters` used in our`MoistThermodyamics`, I suggest we stick with our constants. Otherwise, we have an initialization that is not consistent with our thermodynamics. (One day we will need an easy way to adjust `PlanetParameters` that does not involve changing the source code, but using our constants is fine for now.)

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  g::DF       = grav
+  p0::DF      = 1.0178e5
+  ρ0::DF      = 1.22
+  r_tot_sfc::DF=8.1e-3
+  Rm_sfc          = R_d * (1.0 + (epsdv - 1.0)*r_tot_sfc)
```
tapios: Same for these constants and functions: use the values from PlanetParameters and MoistThermodynamics (except for `p0` and `rho0`)

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  zi         = 840.0    #initial cloud top
+  dz_cloud   = zi - zb
+  q_liq_peak = 0.00045 #cloud mixing ratio at z_i    
+  if xvert > zb && xvert <= zi        
+    q_liq = (xvert - zb)*q_liq_peak/dz_cloud
+  end
+
+  if ( xvert <= zi)
+    θ_liq  = 289.0
+    r_tot      = 8.1e-3                  #kg/kg  specific humidity --> approx. to mixing ratio is ok
```
tapios: Why don't you use `q_tot` in place of `r_tot`, to avoid confusion?

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  end
+
+  if xvert <= 200.0
+      θ_liq += θ_liq 
+      q_tot += q_tot
+  end
+  
+  Rm       = R_d * (1 + (epsdv - 1)*q_tot - epsdv*q_liq);
+  cpm     = cp_d + (cp_v - cp_d)*q_tot + (cp_l - cp_v)*q_liq;
+
```
tapios: Get `Rm` and `cpm` from MoistThermodynamics.

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  Lv::DF      = 2.47e6
+  epsdv::DF   = 1.61
+  g::DF       = grav
+  p0::DF      = 1.0178e5
+  ρ0::DF      = 1.22
+  r_tot_sfc::DF=8.1e-3
+  Rm_sfc          = R_d * (1.0 + (epsdv - 1.0)*r_tot_sfc)
+  ρ_sfc::DF   = 1.22
+  P_sfc           = 1.0178e5
+  T_0::DF     = 285.0
```
tapios: Maybe good to use `T_ref` or `T_BL` (for boundary layer) to avoid confusion with the reference temperature `T_0` that we use in internal energy definitions.

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  #Pressure
+  H = Rm_sfc * T_0 / g;
+  P = P_sfc * exp(-xvert/H);
+  
+  #Exner
+  exner = (P/P_sfc)^(R_d/cp_d);
+  
+  #T, Tv 
+  T     = exner*θ_liq + Lv*q_liq/(cpm*exner);
+  Tv    = T*(1 + (epsdv - 1)*q_tot - epsdv*q_liq);
```
tapios: Why do we need `Tv`? It seems it can be omitted.

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  
+  #Density
+  ρ  = P/(Rm*T);
+  
+  #θ, θv
+  θ      = T/exner;
+  θv     = θ*(1 + (epsdv - 1)*q_tot - epsdv*q_liq);
```
tapios: If we need `theta_v` (do we?), get it from MoistThermodynamics.

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  Lz = zmax - ymin
+  # User defines the grid size:
+  ratiox = (Lx/Δx - 1)/polynomialorder
+  ratioy = (Ly/Δy - 1)/polynomialorder
+  ratioz = (Lz/Δz - 1)/polynomialorder
+  Nex = ceil(Int64, ratiox)
+  Ney = ceil(Int64, ratioy)
+  Nez = ceil(Int64, ratioz)
+  Ne = (Nex, Ney, Nez)
+  # User defined domain parameters
```
tapios: It would be nice if the `Lx`, `ratiox`, `Nex` etc. do not need to be defined here but are calculated in some lower layer from `Delta x` etc.

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,259 @@
...
+  brickrange = (range(DF(xmin), length=Ne[1]+1, DF(xmax)),
+                range(DF(ymin), length=Ne[2]+1, DF(ymax)),
+                range(DF(zmin), length=Ne[3]+1, DF(zmax)))
```
tapios: Same here: I suggest this be pushed to a lower layer, so that the setup becomes slimmer and more transparent.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -229,12 +228,95 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+                        WM - 2 * nM[3] * UnM)
+
+    # Assign scalar values at the boundaries 
+    stateP.ρ = ρM
+    stateP.moisture.ρq_tot = QTM
+    # Assign diffusive fluxes at boundaries
+    diffP = diffM
+    xvert = auxM.coord[3]
+    
+    if xvert < 0.00001
```
tapios: Why do we need this conditional?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -229,12 +228,95 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+      # Bottom boundary quantities 
+      # --------------------------
+      zM          = auxM.coord[3] 
+      q_totM      = QTM/ρM
+      windspeed   = sqrt(uM^2 + vM^2 + wM^2)
+      e_intM      = EM/ρM - 0.5*windspeed^2 - grav*zM
+      TSM         = PhaseEquil(e_intM, q_totM, ρM) 
+      q_vapM      = q_totM - PhasePartition(TSM).liq
+      TM          = air_temperature(TSM)
+      # ----------------------------------------------------------
```
tapios: It looks like more quantities are defined here than we really need.

tapios:
Great to see this coming together!

src/Atmos/Model/AtmosModel.jl
charleskawczynski: We do, unfortunately, for now. We need an elegant way to determine when we're at the ground element.

src/Atmos/Model/AtmosModel.jl
smarras79: Yes, but as long as @asraero has this code working for me as it is, I can then take care of all the cleanup. The reference driver I passed him has a lot of things that are there for testing purposes only but will disappear after I have made sure that the dycoms in the new code works correctly.

src/Atmos/Model/AtmosModel.jl
asraero: If we use the `boundary` flag in the topology setup then we can assign specific indices to each `boundary` wall - this information is contained in `bctype` : default values for `bctype` are 1 (non-periodic) or 0 (periodic)

lcw: @asraero is correct.  `bctype` is designed to be used to determine which boundary is the ground.  You should not need that conditional.

charleskawczynski: Awesome

src/Atmos/Model/AtmosModel.jl
```diff
@@ -230,12 +229,95 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+      # (<var>_FN) First node values (First interior node from bottom wall)
+      # ------------------------------------------------------------------------
+      z_FN             = aux1.coord[3]
+      ρ_FN             = state1.ρ
+      U_FN, V_FN, W_FN = state1.ρu
+      E_FN             = state1.ρe
+      u_FN, v_FN, w_FN = U_FN/ρ_FN, V_FN/ρ_FN, W_FN/ρ_FN
+      windspeed_FN     = sqrt(u_FN^2 + v_FN^2 + w_FN^2)
+      q_tot_FN         = state1.moisture.ρq_tot / ρ_FN
+      e_int_FN         = E_FN/ρ_FN - 0.5*windspeed_FN^2 - grav*z_FN
```
lcw:
```suggestion
      e_int_FN         = E_FN/ρ_FN - windspeed_FN^2/2 - grav*z_FN
```

For type stability we should not use `Float64` constants.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -230,12 +229,95 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+      TS_FN            = PhaseEquil(e_int_FN, q_tot_FN, ρ_FN) 
+      T_FN             = air_temperature(TS_FN)
+      q_vap_FN         = q_tot_FN - PhasePartition(TS_FN).liq
+      # --------------------------
+      # Bottom boundary quantities 
+      # --------------------------
+      zM          = auxM.coord[3] 
+      q_totM      = QTM/ρM
+      windspeed   = sqrt(uM^2 + vM^2 + wM^2)
+      e_intM      = EM/ρM - 0.5*windspeed^2 - grav*zM
```
lcw:
```suggestion
      e_intM      = EM/ρM - windspeed^2/2 - grav*zM
```

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,244 @@
...
+  #θ, θv
+  θ      = T/exner_dry;
+  θv     = virtual_pottemp(T, P, PhPartObj)
+
+  # energy definitions
+  u, v, w     = 7, -5.5, 0.0 
+  U           = ρ * u
+  V           = ρ * v
+  W           = ρ * w
+  e_kin       = 0.5 * (u^2 + v^2 + w^2)
```
charleskawczynski:
```suggestion
  e_kin       = DF(1/2) * (u^2 + v^2 + w^2)
```

jkozdon: Why not just `(u^2 + v^2 + w^2) / 2`

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,246 @@
...
+  exner_dry = exner(P, PhasePartition(DF(0)))
+  
+  #T 
+  T     = exner_dry*θ_liq + LH_v0*q_liq/(cpm*exner_dry);
+  
+  #Density
+  ρ  = P/(Rm*T);
+  
+  #θ, θv
+  θ      = T/exner_dry;
```
charleskawczynski: This appears to never be used

charleskawczynski: Neither does \theta_v

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,246 @@
...
+  end
+  
+  #=
+  if xvert <= 200.0
+      θ_liq += θ_liq 
+      q_tot += q_tot
+  end
+  =# 
+
+  PhPartObj = PhasePartition(q_tot, q_liq, DF(0))
```
charleskawczynski:
```suggestion
  PhPartObj = PhasePartition(q_tot, q_liq, q_ice)
```

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,246 @@
...
+  Rm    = gas_constant_air(PhPartObj)
+  cpm   = cp_m(PhPartObj)
+
+  #Pressure
```
charleskawczynski:
```
  #Pressure
  H = Rm_sfc * T_BL / grav;
  P = P_sfc * exp(-xvert/H);

  # Thermodynamic state
  q_pt = PhasePartition(q_tot, q_liq, q_ice)
  ts = LiquidIcePotTempSHumEquil_no_ρ(θ_liq, q_pt, P)

  #Density
  ρ  = air_density(ts)

  # energy definitions
  u, v, w     = DF(7), DF(-5.5), DF(0)
  U           = ρ * u
  V           = ρ * v
  W           = ρ * w
  e_kin       = (u^2 + v^2 + w^2)/2
  e_pot       = grav * xvert
  E           = ρ * total_energy(e_kin, e_pot, ts)
```

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,246 @@
...
+                     Dates.format(convert(Dates.DateTime,
+                                          Dates.now()-starttime[]),
+                                  Dates.dateformat"HH:MM:SS"),
+                     energy)
+    end
+  end
+
+  step = [0]
+    cbvtk = GenericCallbacks.EveryXSimulationSteps(1000) do (init=false)
+    mkpath("./vtk/")
```
charleskawczynski:
```suggestion
    mkpath("./vtk-dycoms/")
```

src/Atmos/Model/AtmosModel.jl
smarras79: if `bctype` indicates what physical boundary we are on (e.g. bottom wall, left wall, etc.) and not the `type` of boundary condition that is being applied, I suggest renaming`bctype` to something like `bdyposition`.

It would also help define it with a descriptive string rather than a number: 'BOTTOM_WALL', 'TOP_WALL', etc.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -244,23 +244,34 @@ struct DYCOMS_BC <: BoundaryCondition
...
     ρM = stateM.ρ 
     UM, VM, WM = stateM.ρu
     EM = stateM.ρe
     QTM = stateM.moisture.ρq_tot
     uM, vM, wM  = UM/ρM, VM/ρM, WM/ρM
     q_totM = QTM/ρM
     UnM = nM[1] * UM + nM[2] * VM + nM[3] * WM
+    
+    # Assign reflection wall boundaries (top wall)
     stateP.ρu = SVector(UM - 2 * nM[1] * UnM,
```
smarras79: Is it possible to hide the way the no-flux (and b.c. in general) from the person imposing the boundary conditions?
A user who sets up a problem should only need to worry about the type of b.c. and not how it is implemented.

For example, hide this implementation behind something like stateP.\rhou = bc(NO_FLUX)

src/Atmos/Model/AtmosModel.jl
lcw: `bctype` is an integer (which exists for each face) designed to be used to indicate the type of the boundary condition that is to be applied.

As @asraero indicated, you can set these numbers when defining your topology, for example see [[1]].  Currently, we support setting different numbers for the faces of the brick mesh or the top and bottom of the cubed sphere.  

I don't think we can use a string on the GPU.  I would suggest using an `Enum` to define your boundary condition types.  I am guessing this should go into the `AtmosModel`

[1]: https://github.com/climate-machine/CLIMA/blob/master/src/Mesh/Topologies.jl#L228-L230

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,233 @@
...
+using CLIMA.Atmos
+using CLIMA.VariableTemplates
+using CLIMA.MoistThermodynamics
+using CLIMA.PlanetParameters
+using LinearAlgebra
+using StaticArrays
+using Logging, Printf, Dates
+using CLIMA.VTK
+
+@static if haspkg("CuArrays")
```
smarras79: This should be hidden

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -0,0 +1,233 @@
...
+if !@isdefined integration_testing
+  const integration_testing =
+    parse(Bool, lowercase(get(ENV,"JULIA_CLIMA_INTEGRATION_TESTING","false")))
+end
+
+"""
+  Initial Condition for DYCOMS_RF01 LES
+"""
+function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
+
```
smarras79: Based on my conversation with @asraero yesterday, this initialization returns a q_l field with max(q_l)=0.37 instead of the 0.45 requested by Steven's paper. 
To achieve that, the initialization should follow this:

https://github.com/smarras79/CLIMA/blob/02c6dc286b080d71d56fbadd7bd20e1e4dc74237/test/DGmethods/compressible_Navier_Stokes/dycoms-analyticinit.jl#L749

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -37,6 +37,102 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+                      WM - 2 * nM[3] * UnM)
+
+  # Assign scalar values at the boundaries 
+  stateP.ρ = ρM
+  stateP.moisture.ρq_tot = QTM
+  # Assign diffusive fluxes at boundaries
+  diffP = diffM
+  xvert = auxM.coord[3]
+  
+  if bctype == 5 # bctype identifies bottom wall
```
smarras79: using numbers to identify a boundary may lead to errors. Can we wrap those numbers by explicit naming such as "front, back, east, west, top, bottom" or something along these lines?

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -37,6 +37,102 @@ Set the value at the boundary to match the `init_state!` function. This is mainl
...
+  ρM = stateM.ρ 
+  UM, VM, WM = stateM.ρu
+  EM = stateM.ρe
+  QTM = stateM.moisture.ρq_tot
+  uM, vM, wM  = UM/ρM, VM/ρM, WM/ρM
+  q_totM = QTM/ρM
+  UnM = nM[1] * UM + nM[2] * VM + nM[3] * WM
+  
+  # Assign reflection wall boundaries (top wall)
+  stateP.ρu = SVector(UM - 2 * nM[1] * UnM,
```
smarras79: See comment on the same function posted to PR #397

asraero:
bors try

asraero:
bors try

src/Atmos/Model/AtmosModel.jl
asraero: @smarras79 There is a separate PR https://github.com/climate-machine/CLIMA/pull/412 that moves towards this

asraero:
bors try

asraero:
bors try

simonbyrne:
bors try

PR #382, opened 2019-08-20
add gravitational potential

simonbyrne:
Ideally we would make use of the topology structure here, but I can't think of a good way to do that in the current configuration. This seems like the easiest option for now.

### DIFF
--- a/src/Atmos/Model/orientation.jl
+++ b/src/Atmos/Model/orientation.jl
@@ -1,0 +1,40 @@
+# TODO: add Coriolis vectors
+import ..PlanetParameters: grav, planet_radius
+
+abstract type Orientation
+end
+
+function vars_aux(m::Orientation, T)
+  @vars begin
+    Φ::T # gravitational potential
+    ∇Φ::SVector{3,T}
+  end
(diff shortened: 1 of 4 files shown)


simonbyrne:
@charleskawczynski and @asraero pointed out that `RectangularSpace` was not the most informative name.

Basically this determines the orientation of the global forces (gravity / Coriolis) so how about `SphericalOrientation`/`FlatOrientation`?

src/Atmos/Model/orientation.jl
```diff
@@ -0,0 +1,39 @@
...
+struct SphericalOrientation <: Orientation
+end
+function init_aux!(m::SphericalOrientation, aux::Vars, x)
+  normcoord = norm(aux.coord)
+  aux.orientation.Φ = grav * (normcoord - planet_radius)
+  aux.orientation.∇Φ = grav / normcoord .* aux.coord
+end
+
+"""
+    SphericalOrientation <: Orientation
```
lcw:
```suggestion
    FlatOrientation <: Orientation
```

PR #383, opened 2019-08-21
Zero source in DGmodel_kernels (not in AtmosModel)

jkozdon:


jkozdon:
bors try

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -112,6 +112,7 @@ function volumerhs!(bl::BalanceLaw, ::Val{dim}, ::Val{polyorder},
           end
 
           # if source! !== nothing
+          fill!(l_S, -zero(eltype(l_S)))
```
simonbyrne: 😄

charleskawczynski:
@simonbyrne @jkozdon I only just realized, should we add this same behavior for the fluxes?

simonbyrne:
I think to make it consistent we should make `flux!`/`source!` functions operate in an incrementing fashion (i.e. `flux.X += z` instead of `flux.X = z`), and the kernel functions can then either zero or not depending on whether the `increment=true/false` is passed from the ODE function.

jkozdon:
Increment from the ODE function is different than incrementing the source.

The ode function is about whether: `dq = f(t, q)` or `dq += f(t, q)` where `f(t, q)` is the ode function from `dq/dt = f(t, q)`

The source and flux arrays always need to be set somewhere, I do think that it is reasonable to zero any arrays that callback functions are in-charge of setting (flux, source, numerical flux, etc.) lest they have `undef` values when used later.

PR #384, opened 2019-08-21
Vtk should be VTK

jkozdon:
Total yak shaving, but VTK is an acronym for Visualization Toolkit [1]
and should be all caps

[1] https://vtk.org/


charleskawczynski:
Should we also change `writevtk` to `writeVTK`? No, right?

simonbyrne:
happy to merge for now. I'll leave it up to @jkozdon if he wants to change `writevtk`

jkozdon:
> Should we also change `writevtk` to `writeVTK`? No, right?

I could go either way. The filename has a lowercase `.vtk` so maybe not?

PR #397, opened 2019-08-26
SGS (Smagorinsky-Lilly) : AtmosModel version

asraero:
1) Buoyancy-adjusted Smagorinsky-Lilly
**TODO:** 
1) SymmetricSMatrix instead of SVector for \rho\tau
[x] Port over SGS parameter list from old framework
[x] CI Test problems


src/Atmos/Model/moisture.jl
```diff
@@ -83,23 +85,21 @@ function gradvariables!(m::EquilMoist, transform::Vars, state::Vars, aux::Vars,
   transform.moisture.h_tot = e_tot + R_m*T
 end
 
-
-function diffusive!(m::EquilMoist, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real, ρν::Union{Real,AbstractMatrix})
+function diffusive!(m::EquilMoist, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real, ρν::Union{Real,AbstractMatrix}, inv_Pr_turb::Real)
```
smarras79: Can we replace `∇transform` with `∇grad` and add an additional operator called `∇div` as well?

charleskawczynski: I think that the these spatial operators need to be handled in the back end in order to hide DG details from the user. From my understanding, the purpose of this function is to compute `∇transform.f = grad(f)`. That way, we can add `flux.var += ∇transform.f` in `flux!`  for terms that appear in the form `div(grad(f))`.

charleskawczynski: Maybe there is a case for splitting up the name convention this way, but it's a bit unclear to me.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -223,8 +228,12 @@ struct NoFluxBC <: BoundaryCondition
 end
 function boundarycondition!(m::AtmosModel{O,T,M,R,S,BC,IS}, stateP::Vars, diffP::Vars, auxP::Vars,
     nM, stateM::Vars, diffM::Vars, auxM::Vars, bctype, t) where {O,T,M,R,S,BC <: NoFluxBC,IS}
-
-  stateP.ρu -= 2 * dot(stateM.ρu, nM) * nM
+    DF = eltype(stateM)
+    UM, VM, WM = stateM.ρu
+    stateP.ρ = stateM.ρ
+    stateP.ρu -= 2 * dot(stateM.ρu, nM) * collect(nM)
```
smarras79: I think that this layer should be invisible for the user who is not familiar with the mathematics of b.c.
He should only see something like "no-flux", "heat flux", "periodic", etc. Based on this, the b.c. are defined in the background.
What makes it even more difficult for the user not familiar with DG  is the fact that we assign a `Plus` state as a function of the `Minus` state. A user should never have to wonder what this means.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model.jl
```diff
@@ -0,0 +1,202 @@
...
+using CLIMA.VariableTemplates
+using CLIMA.MoistThermodynamics
+using CLIMA.PlanetParameters
+using LinearAlgebra
+using StaticArrays
+using Logging, Printf, Dates
+using CLIMA.VTK
+using Random
+
+@static if haspkg("CuArrays")
```
smarras79: This should be hidden.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model.jl
```diff
@@ -0,0 +1,202 @@
...
+const zmax      = 1000
+const Ne        = (10,2,10)
+const polynomialorder = 4
+const dim       = 3
+const dt        = 0.01
+const timeend   = 10dt
+# ------------- Initial condition function ----------- # 
+function initialise_rising_bubble!(state::Vars, aux::Vars, (x1,x2,x3), t)
+  DF            = eltype(state)
+  R_gas::DF     = R_d
```
smarras79: I think it is repetitive to define new constants that are simply equal to the global constants.

src/Atmos/Model/moisture.jl
simonbyrne: Note that `∇transform` isn't an operator, it just a variable name that happens to contain the spatial derivative of the variable `transform` from the `gradtransform` function.

src/Atmos/Model/AtmosModel.jl
simonbyrne: That's the plan, we're slowly in the process of hiding more of this.

PR #405, opened 2019-08-28
Add squal line

charleskawczynski:
This PR adds the Squal Line test (by @charleskawczynski  and @trontrytel)
 - [x] Initial conditions
 - [x] Boundary conditions
 - [ ] Non-diffusive flux terms
 - [ ] Diffusive flux terms 
 - [ ] Source terms

We're seeing non-deterministic behavior, even at the initialization level. This PR has a test to first ensure that separate runs produce the same result, which fails at large `n_runs`.


charleskawczynski:
We're seeing non-deterministic behavior on a single cpu (I've been testing this only locally for now).

#403 seemed to fix some issues, but we're still unsure about the remaining ones.

jkozdon:
If I wanted to run the code how should I do this?

I tried running `test/DGmethods/compressible_Navier_Stokes/squall-line-model.jl` but some sounding files seem to be missing. 

```
ERROR: LoadError: SystemError: opening file "/Users/jekozdon/codes/CLIMA/test/DGmethods/compressible_Navier_Stokes/../soundings/sounding_gabersek.dat": No such file or directory
```

charleskawczynski:
> I tried running `test/DGmethods/compressible_Navier_Stokes/squall-line-model.jl` but some sounding files seem to be missing.

Hm, I did `git reset`, maybe those didn't land on top of master somehow. I'll make sure to add these files.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -182,15 +225,136 @@ end
...
+function init_sponge!(m::AtmosModel, aux::Vars, geom::LocalGeometry)
+  DT = eltype(aux)
+  #Sponge
+  csleft  = DT(0)
+  csright = DT(0)
+  csfront = DT(0)
+  csback  = DT(0)
+  ctop    = DT(0)
+
+  cs_left_right = DT(0.0)
```
tapios: Would be good to extract all of these as parameters in a central place.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -182,15 +225,136 @@ end
...
+      #Vertical sponge:
+      if z >= top_sponge
+          ctop = ct * (sinpi(1/2 * (z - top_sponge)/(domain_top - top_sponge)))^4
+      end
+
+  elseif sponge_type == 2
+
+
+      alpha_coe = DT(0.5)
+      bc_zscale = DT(7500.0)
```
tapios: Same for these: better extracted as parameters with other SGS parameters.

smarras79:
> If I wanted to run the code how should I do this?
> 
> I tried running `test/DGmethods/compressible_Navier_Stokes/squall-line-model.jl` but some sounding files seem to be missing.
> 
> ```
> ERROR: LoadError: SystemError: opening file "/Users/jekozdon/codes/CLIMA/test/DGmethods/compressible_Navier_Stokes/../soundings/sounding_gabersek.dat": No such file or directory
> ```

all the soundings were pushed to a previous version of the code. However, I see that the sounding directory no longer exists anywhere in the master. I will add a PR to fix this now.

src/Atmos/Model/AtmosModel.jl
asraero: There is a `RayleighSponge` subtype of `Source` https://github.com/climate-machine/CLIMA/blob/296648553346f2e7630152113ad3e04520e573d0/src/Atmos/Model/source.jl#L40 - perhaps we can use this / modify it as required for the squall line case.

smarras79: To extend  @tapios comment on this:
because the sponge parameters depend on the specific problem that is being solved, it would be nice to let the user decide what sponge type (notice that the sponge used for the squall line is different from that used in dycoms) and the parameters such as the sponge intensity.

asraero: > There is a `RayleighSponge` subtype of `Source`
> 
> https://github.com/climate-machine/CLIMA/blob/296648553346f2e7630152113ad3e04520e573d0/src/Atmos/Model/source.jl#L40
> 
> - perhaps we can use this / modify it as required for the squall line case.

Noticed the docs typo here - will fix it with #426

charleskawczynski:
Re-opened, after git-ocalypse, in #549

PR #406, opened 2019-08-28
Add normpervars to MPIStateArray

charleskawczynski:
This PR adds `npv = normpervars(Q::MPIStateArray, ...)`, which computes norms for each of the states. i.e., `length(npv) = numstates(Q)`.

This may help debugging since it's difficult to see how `ρq_tot~O(1)` varies while `ρe ~O(10^4)`.

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -10,6 +10,7 @@ using MPI
 using Base.Broadcast: Broadcasted, BroadcastStyle, ArrayStyle
 
 export MPIStateArray, euclidean_distance, weightedsum
+export normpervar
 
 """
     MPIStateArray{S <: Tuple, T, DeviceArray, N,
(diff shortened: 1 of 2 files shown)


src/Arrays/MPIStateArrays.jl
```diff
@@ -367,13 +368,33 @@ function LinearAlgebra.norm(Q::MPIStateArray, p::Real=2, weighted::Bool=true)
   else
     locnorm = norm_impl(Q.realQ, Val(p))
   end
+  @show "norm", MPI.Comm_rank(Q.mpicomm), locnorm
```
simonbyrne: remove `@show`

src/Arrays/MPIStateArrays.jl
```diff
@@ -367,13 +368,33 @@ function LinearAlgebra.norm(Q::MPIStateArray, p::Real=2, weighted::Bool=true)
...
 
   mpiop = isfinite(p) ? MPI.SUM : MPI.MAX
   r = MPI.Allreduce([locnorm], mpiop, Q.mpicomm)[1]
+  @show "norm", MPI.Comm_rank(Q.mpicomm), r
   isfinite(p) ? r ^ (1 // p) : r
 end
 
+function normpervar(Q::MPIStateArray, p::Real=2, weighted::Bool=true)
```
simonbyrne: It might be possible to combine this with the above function by using the same `dims=nothing` trick.

charleskawczynski: Wouldn't this lead to a type instability? (since they return different ranks)

charleskawczynski:
Closing this since #409 did this more nicely

PR #411, opened 2019-08-30
Allow AtmosModel run with DryModel moisture

lcw:
I am not sure if this is the fix we want but something along these lines is needed to run the RTB model with DryModel moisture.

### DIFF
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -27,7 +27,7 @@ function atmos_boundarycondition!(bc::NoFluxBC, m::AtmosModel, stateP::Vars, dif
     stateP.ρ = stateM.ρ
     stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
     diffP.ρτ = SVector(DF(0), DF(0), DF(0), DF(0), DF(0), DF(0))
-    diffP.moisture.ρd_h_tot = SVector(DF(0), DF(0), DF(0))
+    typeof(diffP.moisture) <: EquilMoist && (diffP.moisture.ρd_h_tot = SVector(DF(0), DF(0), DF(0)))
 end
 
 """


simonbyrne:
I think this relates to something @charleskawczynski and I were discussing yesterday: does it makes sense to specify different boundary conditions on different components, and if so, how should we do it?

src/Atmos/Model/moisture.jl
```diff
@@ -103,3 +106,8 @@ function flux_diffusive!(m::EquilMoist, flux::Grad, state::Vars, diffusive::Vars
   flux.ρe += diffusive.moisture.ρd_h_tot
   flux.moisture.ρq_tot = diffusive.moisture.ρd_q_tot
 end
+
+function atmos_boundarycondition!(bc::NoFluxBC, m::EquilMoist, stateP::Vars, diffP::Vars, auxP::Vars, nM, stateM::Vars, diffM::Vars, auxM::Vars, bctype, t)
+  diffP.moisture.ρd_q_tot = zero(eltype(diffP.moisture.ρd_q_tot))
+  diffP.moisture.ρd_h_tot = zero(eltype(diffP.moisture.ρd_h_tot))
```
simonbyrne: @asraero should we set both diffusive terms to be zero?

lcw:
Simon Byrne <notifications@github.com> writes:

> I think this relates to something @charleskawczynski and I were
> discussing yesterday: does it makes sense to specify different
> boundary conditions on different components, and if so, how should we
> do it?

Good question, as @mwarusz says "boundary conditions are hard".

I guess this leads to more overarching issues on the design of the
boundary conditions interface. Like, how do we want to specify Neumann,
Dirichlet, ...

charleskawczynski:
I'm working on something now that can hopefully help with this. But I think I'll need help with loop unrolling

simonbyrne:
I think this should get fixed by #413

charleskawczynski:
Ah, yes, [`q_pt`](https://github.com/climate-machine/CLIMA/pull/413/files#diff-6f6f3c257f7970bd031f46109bd72cceL22) is not used, and it doesn't make sense to call `PhasePartition` for a dry model. So this should fix `internal_energy`/`thermo_state` calls for `DryModel`.

charleskawczynski:
Can we close this @lcw ?

lcw:
@charleskawczynski yep!

PR #412, opened 2019-08-30
Add `Component`

charleskawczynski:
This is a work-in-progress, but the basic idea is to add a `Component` type for each component that extends the state vector. This way, we can write kernels that can be shared by each of the AtmosModel components (e.g., `Mass`, `Momentum`, `Energy`, etc.).

 - Some parts are a bit more verbose (e.g., `state.mass.ρ` instead of `state.ρ`)
 - BCs can be cherry-picked from a library that works on a `Component`

 - [ ] Need to apply the same pattern to momentum, energy, moisture
 - [ ] Need to address loop unrolling (search `# TODO: Unroll loop`)
 - [ ] Need to add `dim=3` back to one of the test cases


charleskawczynski:
After discussing this with Simon, it seems there are some design drawbacks to this (e.g. adds too much boiler plate)

PR #413, opened 2019-08-31
Update MoistureModel

charleskawczynski:
This PR:

 - [x] Removes unecessary aux variables from Dry and EquilMoist models
 - [x] Fixes notation regarding rho factor in diffuse function (that @simonbyrne pointed out)
 - [x] Moves h_tot up to AtmosModel


charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -185,6 +194,13 @@ function diffusive!(m::AtmosModel, diffusive::Vars, ∇transform::Grad, state::V
   # momentum flux tensor
   diffusive.ρτ = scaled_momentum_flux_tensor(m.turbulence, ρν, S)
 
+  # turbulent Prandtl number
+  diag_ρν = ρν isa Real ? ρν : diag(ρν) # either a scalar or matrix
+  # Diffusivity D_t = ρν/Prandtl_turb
+  D_T = diag_ρν * inv_Pr_turb
```
simonbyrne: While you're doing this, lets rename this to `ρD_t`.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -185,6 +194,13 @@ function diffusive!(m::AtmosModel, diffusive::Vars, ∇transform::Grad, state::V
...
+  # diffusive flux of total energy
+  diffusive.ρd_h_tot = -D_T .* state.ρ .* ∇transform.h_tot
+
   # diffusivity of moisture components
   diffusive!(m.moisture, diffusive, ∇transform, state, aux, t, ρν)
```
simonbyrne: pass `ρD_t` instead of `ρν` (since we just use it to compute `ρD_t` anyway)

src/Atmos/Model/moisture.jl
```diff
@@ -59,47 +58,41 @@ Assumes the moisture components are computed via thermodynamic equilibrium.
...
-
-  phase = thermo_state(m, state, aux)
-  R_m = gas_constant_air(phase)
-  T = aux.moisture.temperature
-  e_tot = state.ρe * ρinv
-  transform.moisture.h_tot = e_tot + R_m*T
 end
 
-function diffusive!(m::EquilMoist, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real, ρν::Union{Real,AbstractMatrix}, inv_Pr_turb::Real)
+function diffusive!(m::EquilMoist, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real, ρν::Union{Real,AbstractMatrix})
```
simonbyrne: receive `ρD_t` instead of `ρν`, remove unnecessary parts.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -185,6 +194,13 @@ function diffusive!(m::AtmosModel, diffusive::Vars, ∇transform::Grad, state::V
   # momentum flux tensor
   diffusive.ρτ = scaled_momentum_flux_tensor(m.turbulence, ρν, S)
 
+  # turbulent Prandtl number
+  diag_ρν = ρν isa Real ? ρν : diag(ρν) # either a scalar or matrix
+  # Diffusivity D_t = ρν/Prandtl_turb
+  D_T = diag_ρν * inv_Pr_turb
+  # diffusive flux of total energy
+  diffusive.ρd_h_tot = -D_T .* state.ρ .* ∇transform.h_tot
```
simonbyrne: My point was that it already has a factor of `ρ` in there.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Makes sense. I love this naming scheme!

charleskawczynski:
bors try

simonbyrne:
need to update the `mms_bc_atmos.jl` example.

charleskawczynski:
bors try

charleskawczynski:
Do not merge yet. We want [this term](https://github.com/climate-machine/CLIMA/pull/413/files#diff-c3513c2fab9c4791668155909dd1f5e6R202), according to the design docs, but the `mms_bc_atmos.jl` test fails when we include it. Not sure how we want to make this term flexible for both cases.

simonbyrne:
@asraero and I discussed this on Friday. Ideally we would generate a new system of equations which had this term. 
@lcw @jkozdon would that be hard to do?

Otherwise, we could disable that but it is the only case which actually tests the AtmosModel at this point, so we should come up with some more tests.

asraero:
> @asraero and I discussed this on Friday. Ideally we would generate a new system of equations which had this term.
It seems the manufactured solution in https://github.com/climate-machine/CLIMA/blob/master/test/DGmethods/compressible_Navier_Stokes/mms_solution.jl#L15 can be used to re-generate the appropriate source terms by adjusting the gradient and flux expressions to reflect the `AtmosModel` equations.

jkozdon:
> It seems the manufactured solution in https://github.com/climate-machine/CLIMA/blob/master/test/DGmethods/compressible_Navier_Stokes/mms_solution.jl#L15 can be used to re-generate the appropriate source terms by adjusting the gradient and flux expressions to reflect the `AtmosModel` equations.

You should be able to modify the manufactured solution to include any terms you want. Just pop the extra diffusion terms into the [fluxes](https://github.com/climate-machine/CLIMA/blob/2629ce8f5ff7eb53f4b1eea8bfb90de7e3c0a380/test/DGmethods/compressible_Navier_Stokes/mms_solution.jl#L51-L65)

charleskawczynski:
> You should be able to modify the manufactured solution to include any terms you want. Just pop the extra diffusion terms into the [fluxes](https://github.com/climate-machine/CLIMA/blob/2629ce8f5ff7eb53f4b1eea8bfb90de7e3c0a380/test/DGmethods/compressible_Navier_Stokes/mms_solution.jl#L51-L65)

Ok, implementing this now, just want to be sure that I adjust `expected_result` correctly.

jkozdon:
> Ok, implementing this now, just want to be sure that I adjust `expected_result` correctly.

Sounds good. Let me know if you need some help with this.

charleskawczynski:
> Sounds good. Let me know if you need some help with this.

Thanks, I would definitely appreciate a review. Generating the `expected_result` now, it's taking a little while using `integration_testing = true`.

jkozdon:
> it's taking a little while

The solution is a little too complicated, the use of special functions really hurts the performance. At some point we may want to reevaluate how it's implemented and see if we can do anything smarter.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -185,8 +194,16 @@ function diffusive!(m::AtmosModel, diffusive::Vars, ∇transform::Grad, state::V
   # momentum flux tensor
   diffusive.ρτ = scaled_momentum_flux_tensor(m.turbulence, ρν, S)
 
+  # turbulent Prandtl number
+  diag_ρν = ρν isa Real ? ρν : diag(ρν) # either a scalar or matrix
+  # Diffusivity ρD_t = ρν/Prandtl_turb
+  ρD_t = diag_ρν * inv_Pr_turb
+  # diffusive flux of total energy
+  # diffusive.ρd_h_tot = -ρD_t .* ∇transform.h_tot
```
asraero: Uncomment line 202, remove line 203

charleskawczynski:
The integration tests were taking too long (granted I was trying to run it locally). Also, I heard that they're not currently running, so maybe we can add them in later. For now, I've just filled the remaining values as zeros, and put a `TODO`.

@asraero @simonbyrne @jkozdon I think this is ready for review/merge

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Good catch! Thanks!

jkozdon:
> The integration tests were taking too long (granted I was trying to run it locally).

Yes the MMS is fairly inefficient right now due to the use of special functions.

> Also, I heard that they're not currently running, so maybe we can add them in later. For now, I've just filled the remaining values as zeros, and put a `TODO`

The main point of the integration testing is to make sure that when the test was added, the first number you added was a meaningful piece of data (and that you haven't just created a random number generator).

In my opinion, without running the integration tests at least once you don't know that your code is correct.

It would be great if the integration tests always ran, but that probably cannot happen until the MMS is sped up in some way.

jkozdon:
Can you run the code on the cluster? My guess is running over a bunch of CPUs will be better than the GPU (due to the special functions).

charleskawczynski:
Yes, I can do this. Need to reconfigure my duo first..

jkozdon:
#423 effects this PR

test/DGmethods/compressible_Navier_Stokes/mms_bc_atmos.jl
```diff
@@ -198,8 +199,9 @@ let
   polynomialorder = 4
   base_num_elem = 4
 
-  expected_result = [1.5606226382564500e-01 5.3302790086802504e-03 2.2574728860707139e-04;
-                     2.5803100360042141e-02 1.1794776908545315e-03 6.1785354745749247e-05]
+  # TODO: Get solution for integration_testing = true
+  expected_result = [0.36679329773810837 0.3353991836118424 0.33691843904190333;
```
jkozdon: This suggests that things are not converging.

charleskawczynski: > #423 effects this PR

Thanks for the heads up. I'll check this out.

> This suggests that things are not converging.

That makes sense. Any idea what's wrong? Maybe this won't be converged quite as soon as I expected..

charleskawczynski: I made a separate PR to at least remove the unnecessary fields, but we should still figure out how to get this working @jkozdon

simonbyrne:
Merged #453 instead.

PR #414, opened 2019-09-04
Add `linearized_air_pressure` to MT

charleskawczynski:
Adds a function to compute the linearized air pressure, as specified in the design docs.


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -85,6 +86,29 @@ air_pressure(ts::ThermodynamicState) =
   air_pressure(air_temperature(ts), air_density(ts), PhasePartition(ts))
 air_pressure(ts::PhaseDry) = air_density(ts) * gas_constant_air(ts) * air_temperature(ts)
 
+"""
+    linearized_air_pressure(ρ, ρe_tot, Φ[, q::PhasePartition])
+
+The air pressure, linearized around the reference state, from the equation of state
```
tapios: 'linearized around a dry rest state' would be better--the linearization does not actually depend on what the reference state is, as long as it is dry (q=0) and at rest (u=0).

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -85,6 +86,29 @@ air_pressure(ts::ThermodynamicState) =
...
+ - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
+"""
+linearized_air_pressure(ρ::DT, ρe_tot::DT, Φ::DT, q::PhasePartition{DT}=PhasePartition{DT}(DT(0), DT(0), DT(0))) where {DT<:Real} =
+  ρ*DT(R_d)*DT(T_0) + DT(R_d)/DT(cv_d)*(ρe_tot - ρ*Φ - (ρ*q.tot - ρ*q.liq)*DT(e_int_v0) + ρ*q.ice*(DT(e_int_i0) + DT(e_int_v0)))
+
+linearized_air_pressure(ts::ThermodynamicState{DT}, ρe_tot::DT, Φ::DT) where {DT<:Real} =
+  linearized_air_pressure(air_density(ts), ρe_tot, Φ, PhasePartition(ts))
+
+function linearized_air_pressure(ts::PhaseDry{DT}, ρe_tot::DT, Φ::DT) where {DT<:Real}
+  ρ = air_density(ts)
```
tapios: What is the rationale for treating density asymmetrically depending on whether the model is moist or dry? It is a state variable either way, so why is it not passed into this function in the same way as in the moist case?

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -181,6 +187,7 @@ end
       ts_dry = PhaseDry(e_int, ρ)
       ts_eq  = PhaseEquil(e_int, typeof(ρ)(0), ρ)
 
+      @test linearized_air_pressure(ts_eq, ρe_tot, Φ)  ≈ linearized_air_pressure(ts_dry, ρe_tot, Φ)
```
tapios: You might also test that `linearized_air_pressure = air_pressure` for dry air with zero velocity.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Good point, I'll update this.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: This `PhaseDry` function has a lower memory footprint because `PhasePartion` is not needed. Also, I pre-compute `ρ` here since it's used 2 times in the expression, the same function is called in either case (technically it's dispatched to the moist/dry case depending on the particular thermodynamic state).

test/Common/MoistThermodynamics/runtests.jl
charleskawczynski: I'm not sure how to perform this test since `linearized_air_pressure` and `air_pressure` are not a function of velocity. Do you mean for zero gravitational potential energy? In that case, the comparison would be `air_pressure` against `ρ*R_d*(T_0 + e_tot/cv_d)`.

tapios: I mean zero velocity, and said that because velocity enter `e_tot`. If the thermodynamic state does not depend on total energy (only internal energy), you simply do not need to worry about velocity. Then we should generally have `linearized_air_pressure = air_pressure` for dry air.

charleskawczynski: I think I get it, and it looks a lot nicer now (the function signature is the same as `total_energy`), and the total energy is now consistent with the internal energy (I think that's what I was missing before).

charleskawczynski:
@simonbyrne @tapios I think this is ready to merge

PR #417, opened 2019-09-05
Removing stale kernels from the `DGmodel_kernels.jl`

jkozdon:
- an old version of `initstate!` is still in the file
- `knl_apply_filter` has moved to `src/Mesh/Filters_kernels`
- `elem_grad_field!` is not ever called by `DGmodel.jl`


jkozdon:
bors try

lcw:
bors try

PR #420, opened 2019-09-06
Add isentropic vortex test using AtmosModel

mwarusz:
Also adds `NoViscosity` "turbulence closure" to `AtmosModel`.


mwarusz:
bors try

test/DGmethods/Euler/isentropicvortex.jl
```diff
@@ -0,0 +1,278 @@
...
+  e_kin = u' * u / 2
+  state.ρe = ρ * total_energy(e_kin, DFloat(0), T)
+end
+
+function do_output(mpicomm, vtkdir, vtkstep, dg, Q, Qe, testname = "isentropicvortex")
+  ## name of the file that this MPI rank will write
+  filename = @sprintf("%s/%s_mpirank%04d_step%04d",
+                      vtkdir, testname, MPI.Comm_rank(mpicomm), vtkstep)
+
+  statenames = ("ρ", "ρu", "ρv", "ρw", "ρe")
```
charleskawczynski:
```suggestion
  statenames = flattenednames(vars_state(model, eltype(Q)))
```

test/DGmethods/Euler/isentropicvortex.jl
```diff
@@ -0,0 +1,278 @@
...
+  p = p∞ * (T / T∞) ^ (DFloat(1) / kappa_d)
+  ρ = air_density(T, p)
+
+  state.ρ = ρ
+  state.ρu = ρ * u
+  e_kin = u' * u / 2
+  state.ρe = ρ * total_energy(e_kin, DFloat(0), T)
+end
+
+function do_output(mpicomm, vtkdir, vtkstep, dg, Q, Qe, testname = "isentropicvortex")
```
charleskawczynski:
```suggestion
function do_output(mpicomm, vtkdir, vtkstep, dg, Q, Qe, model, testname = "isentropicvortex")
```

test/DGmethods/Euler/isentropicvortex.jl
```diff
@@ -0,0 +1,278 @@
...
+  callbacks = (cbinfo,)
+
+  if output_vtk
+    # create vtk dir
+    vtkdir = "vtk_isentropicvortex" *
+      "_poly$(polynomialorder)_dims$(dims)_$(ArrayType)_$(DFloat)_level$(level)"
+    mkpath(vtkdir)
+
+    # output initial step
+    do_output(mpicomm, vtkdir, vtkstep, dg, Q, Q)
```
charleskawczynski:
```suggestion
    do_output(mpicomm, vtkdir, vtkstep, dg, Q, Q, model)
```

test/DGmethods/Euler/isentropicvortex.jl
```diff
@@ -0,0 +1,278 @@
...
+
+    # setup the output callback
+    outputtime = timeend
+    vtkstep = 0
+    cbvtk = EveryXSimulationSteps(floor(outputtime / dt)) do
+      vtkstep += 1
+      Qe = init_ode_state(dg, param, gettime(lsrk))
+      do_output(mpicomm, vtkdir, vtkstep, dg, Q, Qe)
```
charleskawczynski:
```suggestion
      do_output(mpicomm, vtkdir, vtkstep, dg, Q, Qe, model)
```

src/Atmos/Model/turbulence.jl
```diff
@@ -23,6 +23,41 @@ end
 function gradvariables!(::TurbulenceClosure, transform::Vars, state::Vars, aux::Vars, t::Real)
 end
 
+struct NoViscosity <: TurbulenceClosure end
```
asraero: Thanks for setting this test up! Could we use the current setup with `ConstantViscosityWithDivergence(zero(DFloat))` to get us the same physical result of having no viscosity ? diffusive.\rho\tau is defined at the `AtmosModel` top level

jkozdon: The problem I see with the `ConstantViscosityWithDivergence(zero(DFloat))` approach is that if you want to look at performance you will still call all the viscous kernels, which are not needed in this case. When we start splitting the calculation into substeps, we certainly want to be able to turn these kernel calls off.

mwarusz: Another bonus is that being able to turn viscosity off completely can sometimes help with debugging. If my current changes are too intrusive then there is another approach. You could disable viscosity at the top level of `AtmosModel` by defining methods like

```
flux_diffusive!(m::AtmosModel{O,RS,T}, _...) where {O,RS,T<:NoViscosity} = nothing
```
Then other parts of `AtmosModel`, including the location of `ρτ`, can remain as they were before my changes. Would that  be better ?

@charleskawczynski

charleskawczynski: > The problem I see with the `ConstantViscosityWithDivergence(zero(DFloat))` approach is that if you want to look at performance you will still call all the viscous kernels, which are not needed in this case

That's a great point, I definitely see the motivation for this.

@mwarusz The only issue I have with the example you suggested is that every function signature of this flavor may have to change every time a new component is added to `AtmosModel`, which is kind of a pain. Might we instead just add an argument and use something like 

```
function flux_diffusive!(m::AtmosModel, _...)
  flux_diffusive!(m, ..., m.turbulence)
end
function flux_diffusive!(m::AtmosModel, _..., ::NoViscosity)
...
end
```

But this might lead to very nested calls. @simonbyrne  what do you think?

PR #422, opened 2019-09-06
Fix type in ConstantViscosityWithDivergence

jkozdon:
Remove `Float64` hard coding from `ConstantViscosityWithDivergence`

### DIFF
--- a/src/Atmos/Model/turbulence.jl
+++ b/src/Atmos/Model/turbulence.jl
@@ -28,8 +28,8 @@ end
 
 Turbulence with constant dynamic viscosity (`ρν`). Divergence terms are included in the momentum flux tensor.
 """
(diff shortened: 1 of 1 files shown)


jkozdon:
Even if we do not want to do this, we should at least change https://github.com/climate-machine/CLIMA/blob/e7033ea4f4430dad1f088ee0575f39f9c20d255f/src/Atmos/Model/turbulence.jl#L31-L33

to
```julia
 struct ConstantViscosityWithDivergence{T} <: TurbulenceClosure 
   ρν::T
 end
```

simonbyrne:
I'd be fine with the latter proposal. I'd only do the former if we were sure there was a clear benefit (also we are unlikely to be using this for production, it is mainly for testing).

jkozdon:
> I'd be fine with the latter proposal.

Changed.

PR #425, opened 2019-09-07
Performance improvements

mwarusz:
This pull request:
- brings back `volumerhs!` performance-improving loop reorderings that accidentally got lost when moving to the new framework
- adds further minor changes that should improve the performance


mwarusz:
bors try

PR #431, opened 2019-09-11
Add flexibility to how aux is updated

jkozdon:


charleskawczynski:
What does this do?

jkozdon:
Allows the model to define the logic for updating aux, as opposed to assuming that the aux is always updated the way atmos model needs it updated. Atmos needed integrals and point updates, in the ocean we needed to define some more logic.

jkozdon:
So mainly this adds some flexibility in how aux is updated.

PR #432, opened 2019-09-11
Update boundary conditions interface

jkozdon:
Make the BC interfaces more general

### DIFF
--- a/src/Utilities/VariableTemplates/VariableTemplates.jl
+++ b/src/Utilities/VariableTemplates/VariableTemplates.jl
@@ -117,7 +117,7 @@ end
       offset += N
     elseif T <: StaticArray
       N = length(T)
-      retexpr = :(array[$(offset + 1):$(offset + N)] .= val)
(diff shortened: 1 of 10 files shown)


src/Utilities/VariableTemplates/VariableTemplates.jl
```diff
@@ -117,7 +117,7 @@ end
       offset += N
     elseif T <: StaticArray
       N = length(T)
-      retexpr = :(array[$(offset + 1):$(offset + N)] .= val)
+      retexpr = :(array[$(offset + 1):$(offset + N)] .= val[:])
```
charleskawczynski: Is this to fix shape mismatch with the symmetric matrix?

jkozdon: No, if you have a `SMatrix` in the `Vars`

jkozdon: Everywhere right now we have `SVector`

PR #437, opened 2019-09-13
Allow access to diffusive vars in knl_nodal_update_aux

asraero:
Allows access to diffusive variables in the update_aux function. 

It's closer to this https://github.com/climate-machine/CLIMA/blob/fdff985b7b0416ca50305af000a0043cd6ec1ebe/src/DGmethods/DGmodel.jl#L229 (`knl_node_apply_aux!` no longer exists in DGmodel_kernels.jl). It emulates the functionality of the `dof_iteration!` function in the older DGBalanceLaw framework.

@jkozdon Does this seem reasonable?


src/Atmos/Model/orientation.jl
```diff
@@ -23,7 +23,7 @@ No gravitional force or potential.
 """
 struct NoOrientation <: Orientation
 end
-function vars_aux(m::NoOrientation, T)
+function vars_aux!(::NoOrientation, T)
```
jkozdon:
```suggestion
function vars_aux(::NoOrientation, T)
```

jkozdon:
Yes I changed the `apply_aux!` to `update_aux!`

The problem I see with putting this in aux is that `Q` and `Qvisc` are not consistent when `update_aux!` is called.

src/Atmos/Model/turbulence.jl
```diff
@@ -57,12 +57,19 @@ end
...
 vars_gradient(::SmagorinskyLilly,T) = @vars(θ_v::T)
 vars_diffusive(::SmagorinskyLilly,T) = @vars(∂θ∂Φ::T)
 function atmos_init_aux!(::SmagorinskyLilly, ::AtmosModel, aux::Vars, geom::LocalGeometry)
   aux.turbulence.Δ = lengthscale(geom)
 end
+function atmos_nodal_update_aux!(::SmagorinskyLilly, ::AtmosModel, state::Vars, diffusive::Vars, aux::Vars, t::Real)
+  ρ𝛕 = diffusive.ρτ
+  ρ = state.ρ
+  𝛕 = ρ𝛕 ./ ρ
+  @inbounds τ_w = sqrt(𝛕[1,3]^2 + 𝛕[2,3]^2)
```
jkozdon:
```suggestion
  @inbounds τ_w = hypot(𝛕[1,3], 𝛕[2,3])
```

src/Atmos/Model/turbulence.jl
```diff
@@ -57,12 +57,19 @@ end
...
+  aux.turbulence.u_τ = sqrt(τ_w / ρ)
```
jkozdon: Does this need to be consistent with the state later? Updating in `update_aux!` will not guarantee this.

jkozdon: If this is needed to be consistent, then I would suggest computing when it's needed.

src/DGmethods/DGmodel.jl
```diff
@@ -212,7 +212,7 @@ function init_ode_state(dg::DGModel, param, args...; commtag=888)
   return state
 end
 
-
+#=
```
jkozdon: This chunk of code should just be deleted rather than commented out.

src/Atmos/Model/turbulence.jl
asraero: Noted - I placed this in there as an example of such a value requiring the diffusive state; computing this as necessary seems to be the solution since it needs to be consistent with the state. I believe I can access the diffusive state for visualisation through `param.diff` so this is fine for now.

jkozdon: > I believe I can access the diffusive state for visualisation through `param.diff` so this is fine for now.

Yeah, that's what I'd suggest if you just need it for viz.

jkozdon:
Should this PR be closed?

PR #439, opened 2019-09-16
implement different bufferkinds for CUDA-aware MPI

vchuravy:
We recently discussed what would be needed to implement support for CUDA-aware MPI.
This is an early prototype I was working on.

# TODO:
- [x] Implement device/host buffer support 
  - [x] Blocked on https://github.com/JuliaGPU/CuArrays.jl/issues/409
- [ ] Decide on how to configure
- [ ] Performance evaluation
- [ ] Tests 
- [ ] Docs

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -19,6 +19,9 @@ Base.similar(::Type{A}, ::Type{FT}, dims...) where {A<:Array, FT} = similar(Arra
   Base.similar(::Type{A}, ::Type{FT}, dims...) where {A<:CuArray, FT} = similar(CuArray{FT}, dims...)
 end
 
+include("CMBuffers.jl")
+using .CMBuffers
+
 cpuify(x::AbstractArray) = convert(Array, x)
 cpuify(x::Real) = x
 
@@ -29,7 +32,7 @@ export MPIStateArray, euclidean_distance, weightedsum
                   DAT2<:AbstractArray{FT,2}} <: AbstractArray{FT, 3}
 """
 struct MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
-                     DAT2<:AbstractArray{FT,2}} <: AbstractArray{FT, 3}
+                    Buf<:CMBuffer} <: AbstractArray{FT, 3}
   mpicomm::MPI.Comm
   data::DATN
   realdata::DAV
@@ -43,15 +46,12 @@ struct MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
   sendreq::Array{MPI.Request, 1}
   recvreq::Array{MPI.Request, 1}
 
-  host_send_buffer::Array{FT, 2}
-  host_recv_buffer::Array{FT, 2}
+  send_buffer::Buf
+  recv_buffer::Buf
 
   nabrtorank::Array{Int64, 1}
   nabrtovmaprecv::Array{UnitRange{Int64}, 1}
   nabrtovmapsend::Array{UnitRange{Int64}, 1}
-
-  device_send_buffer::DAT2
-  device_recv_buffer::DAT2
 
   weights::DATN
 
@@ -59,18 +59,21 @@ struct MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
 
   function MPIStateArray{FT}(mpicomm, DA, Np, nstate, numelem, realelems,
                              ghostelems, vmaprecv, vmapsend, nabrtorank,
-                             nabrtovmaprecv, nabrtovmapsend, weights, commtag
+                             nabrtovmaprecv, nabrtovmapsend, weights, commtag;
+                             mpi_knows_cuda=false
                             ) where {FT}
     data = similar(DA, FT, Np, nstate, numelem)
 
-    device_recv_buffer = similar(data, FT, nstate, length(vmaprecv))
-    device_send_buffer = similar(data, FT, nstate, length(vmapsend))
+    if data isa Array || mpi_knows_cuda
+      kind = SingleCMBuffer
+    else
+      kind = DoubleCMBuffer
+    end
+    recv_buffer = CMBuffer{FT}(DA, kind, nstate, length(vmaprecv))
+    send_buffer = CMBuffer{FT}(DA, kind, nstate, length(vmapsend))
 
     realdata = view(data, ntuple(i -> Colon(), ndims(data) - 1)..., realelems)
     DAV = typeof(realdata)
-
-    host_send_buffer = zeros(FT, nstate, length(vmapsend))
-    host_recv_buffer = zeros(FT, nstate, length(vmaprecv))
 
     nnabr = length(nabrtorank)
     sendreq = fill(MPI.REQUEST_NULL, nnabr)
@@ -81,6 +84,7 @@ struct MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
     # anything).
     #
     # Better way than checking the type names?
+    # XXX: Use Adapt.jl vmaprecv = adapt(DA, vmaprecv)
     if typeof(vmaprecv).name != typeof(data).name
       vmaprecv = copyto!(similar(DA, eltype(vmaprecv), size(vmaprecv)),
                          vmaprecv)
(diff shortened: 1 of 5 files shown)


src/Arrays/MPIStateArrays.jl
```diff
@@ -295,7 +327,7 @@ function finish_ghost_recv!(Q::MPIStateArray)
   MPI.Waitall!(Q.recvreq)
 
   # copy data to state vectors
-  transferrecvbuf!(Q.device_recvQ, Q.host_recvQ, Q.Q, Q.vmaprecv)
+  transferrecvbuf!(Q.device_recvQ, Q.recvQ, Q.Q, Q.vmaprecv)
```
vchuravy: I am not quite sure what `device_recvQ` is being used for, but in the case that `recvQ` resides on the device we should be able to collapse them

lcw: The array `device_recvQ` is being used to gather the data that needs to be communicated into contiguous chunks for each rank.

lcw: A long time ago, say ~10 years, I had issues passing a CUDA pinned array into `MPI_Isend`/`MPI_Irecv`.  I guess the MPI installation, which was MVAPICH, was trying to pin the memory region as well and it caused some deadlocking.  Thus, it required double buffering on the host.  I hope this is no longer the case and you can collapse them but haven't tested it.

src/Arrays/buffers.jl
```diff
@@ -0,0 +1,165 @@
+module Buffers
```
charleskawczynski: Should we make this filename `Buffers` (title case)?

src/Arrays/MPIStateArrays.jl
```diff
@@ -20,19 +20,15 @@ end
 
 export MPIStateArray, euclidean_distance, weightedsum
 
-@enum BufferKind begin
-  Host
-  Pinned
-  HostBuffer
-  DeviceBuffer
-end
+include("buffers.jl")
```
charleskawczynski: `include("Buffers.jl")`

src/Arrays/buffers.jl
```diff
@@ -0,0 +1,165 @@
...
+    buffer = new{T, typeof(primary), typeof(secondary)}(primary, secondary)
+
+    return buffer
+  end
+end
+
+register(x) = nothing
+unregister(x) = nothing
+
+@init @require CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
```
charleskawczynski: Is there a reason why we're not using `haspkg` here?

simonbyrne: `haspkg` is going away soon, since it is broken in 1.3

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,123 @@
...
+function get_transfer(buf::Buffer)
+  if buf.transfer === nothing
+    return buf.stage
+  else
+    return buf.transfer
+  end
+end
+
+# XXX: rename and be clearer about direction
+function staged!(buf::Buffer)
```
kpamnany: `prepare_transfer()` (to be followed by `get_transfer()`)

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,123 @@
...
+  if buf.transfer === nothing
+    # nothing to do here
+  else
+    # XXX: async
+    copy!(buf.transfer, buf.stage)
+  end
+end
+
+function transferred!(buf::Buffer)
```
kpamnany: `prepare_stage()`

vchuravy:
bors try

kpamnany:
bors try

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,192 @@
...
+    return buf.transfer
+  end
+end
+
+###
+# By design we intend to run async copies on the default stream
+# to obtain concurrency computation should run on a non-blocking
+# secondary stream. This is due to the fact that CUDA-aware MPI
+# will run operations on the default stream and those operations
+# are implicitly synchronising.
```
lcw: What does this mean for CuArrays based kernels?  Last time I checked it was not possible to use anything but the default queue for CuArrays.

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,192 @@
...
+end
+
+function get_transfer(buf::Buffer)
+  if buf.transfer === nothing
+    return buf.stage
+  else
+    return buf.transfer
+  end
+end
+
```
charleskawczynski: Looks like we could write `get_transfer` as

```
get_transfer(buf::Buffer{T, Arr, Nothing}) where {T, Arr} = buf.stage
get_transfer(buf::Buffer{T, Arr, Buff}) where {T, Arr, Buff} = buf.transfer
```

vchuravy: That is the same (also performance and code-generation wise).

I like this way since it makes the decision clear.

src/Arrays/Buffers.jl
vchuravy: Right, CuArray also operates on the default stream. In general this is something we need to fix working with upstream, but IIUC in the critical loop we shouldn't be doing any CuArray operations that need to happen concurrently to MPI or memcopies?

lcw: I believe you are right that our use of CuArrays is outside the ODE RHS function.  So we should be okay for now.

simonbyrne:
Our of curiousity, is there any reason we can't use MPI Datatypes, and let MPI handle the buffering itself (assuming we have a CUDA-aware MPI)?

vchuravy:
bors try

src/Arrays/Buffers.jl
```diff
@@ -164,6 +165,42 @@ end
...
+          return true
+      else
+          CUDAdrv.throw_api_error(err)
+      end
+  end
+
+  """
+      friendlysynchronize(stream)
+
+  MPI defines a notion of progress which means that MPI operations
```
vchuravy: taken from https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/mpi.html#mpi-progress

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
It looks like we indeed had a deadlock because a stream synchronization inhibited MPI to make progress... If bors is happy with this, I think this should be ready to merge.

I added some detailed reasoning to the commit messages.

vchuravy:
bors try

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,232 @@
...
+        #   stage    = # CuArray(transfer)
+        # ```
+        register(transfer)
+        # XXX: Should this finalizer be attached to the Buffer struct?
+        finalizer(unregister, transfer)
+      end
+    else
+      error("Bufferkind $kind is not implemented yet")
+    end
+
```
lcw: Why is pinnned memory turned off?  As far as I understand it, we need pinned memory to overlap communication and computation.

I am not sure it is correct but here
https://github.com/lcw/Heptapus.jl/blob/0f235861c201cbfdcfc4368629e1dccda3868b57/examples/mpicuda/try.jl#L6
is an example of how I registered pinned memory to achieve overlapping communication and computation.
￼

src/Arrays/Buffers.jl
```diff
@@ -0,0 +1,232 @@
...
+  """
+  function friendlysynchronize(stream)
+    status = false
+    while !status
+      status = queryStream(stream)
+      MPI.Iprobe(MPI.MPI_ANY_SOURCE, MPI.MPI_ANY_TAG, MPI.COMM_WORLD)
+    end
+    return
+  end
+
```
lcw: Interesting.  I have not run into this issue before.  Is it specific to CUDA aware MPI?  Or is it because you are trying to do multiple buffers with your BufferKind type?

For the codes we have written in the past we have used a pattern like the following

https://github.com/lcw/Heptapus.jl/blob/0f235861c201cbfdcfc4368629e1dccda3868b57/examples/mpicuda/try.jl#L66-L83

src/Arrays/Buffers.jl
vchuravy: >  As far as I understand it, we need pinned memory to overlap communication and computation.

IIUC that is not tied to pinned memory, but to using non-blocking streams + async memory transfers.

> Why is pinnned memory turned off?

I was trying to debug the freezes bors is experiencing.

lcw: I see.  I always though I needed pinned memory for async memory transfers.

src/Arrays/Buffers.jl
lcw: Interesting.  Thanks for the link.

src/Arrays/Buffers.jl
vchuravy: > Is it specific to CUDA aware MPI?

Not really, generally any spinlock will block MPI from making progress. I am probably overly cautious here, but I was trying to figure out why bors was freezing.

src/Arrays/Buffers.jl
lcw: FYI, see for example Overlapping Kernel Execution and Data Transfers section of <https://devblogs.nvidia.com/how-overlap-data-transfers-cuda-cc>.

simonbyrne:
bors try

PR #443, opened 2019-09-19
Splitting AtmosModel into nonlinear and linear parts

mwarusz:
This PR:
- enables using `AtmosModel` for IMEX or split-explicit time stepping in a
  very naive way by creating two separate `AmosModel`s for the nonlinear
  and linear parts
- adds a test that that verifies this splitting using IMEX

Marking this as a draft since I am not sure that this is the way to go and there is
a bunch other issues. However, it shows roughly what is needed.

PR #445, opened 2019-09-19
Create linearisation of AtmosModel

simonbyrne:
@mwarusz This is an alternative approach to #443, based on using wrappers around `AtmosModel` to operate on subcomponents. Would this be easier to work with?


tapios:
Looks very nice and clean!

tapios:
We'll primarily need the linear components in the vertical (for 1D IMEX), so only vertical fluxes. An option may be just to implement this, rather than the 3D version here. Or else, if we want both, would it be useful to think of a good abstraction of doing that, rather than defining separate 1d linear, 3d linear models, each with different options for including or excluding gravity waves?

lcw:
If you look at how 1D IMEX worked in the previous framework, https://github.com/climate-machine/CLIMA/pull/373, there is one linear model for 1D and 3D.  We did this by specifying the 3D linearization and zeroing out horizontal derivatives in the 1D computational kernels.

simonbyrne:
bors try

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
```diff
@@ -0,0 +1,220 @@
...
+               Rusanov(),
+               CentralNumericalFluxDiffusive(),
+               CentralGradPenalty())
+
+  linmodel = LinearType(model)
+  lindg = DGModel(linmodel,
+               grid,
+               Rusanov(),
+               CentralNumericalFluxDiffusive(),
+               CentralGradPenalty(); auxstate=dg.auxstate)
```
simonbyrne: Note we need to share the aux state from the main model so that it has access to the reference state. It may make sense to move the reference state to the model instead (in which case it would then have its own aux state).

simonbyrne:
bors try

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,71 @@
...
+init_state!(lm::AtmosLinearModel, state::Vars, aux::Vars, coords, t) = nothing
+
+
+function linear_pressure(moisture, orientation, state::Vars, aux::Vars)
+  ρ = state.ρ
+  invρ = inv(ρ)
+  TS = thermo_state(moisture, orientation, state, aux)
+  # TODO: avoid dividing then multiplying by ρ
+  linearized_air_pressure(ρ, state.ρe * invρ, gravitational_potential(orientation, aux), PhasePartition(TS))
+end
```
jkozdon: Is this actually linear?

simonbyrne: I'm not sure? Maybe we should just remove it for now, since it isn't being used.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
```diff
@@ -0,0 +1,220 @@
...
+end
+
+# -------------- Problem constants ------------------- # 
+const (xmin,xmax)      = (0,1000)
+const (ymin,ymax)      = (0,400)
+const (zmin,zmax)      = (0,1000)
+const Ne        = (10,2,10)
+const polynomialorder = 4
+const dim       = 3
+const dt        = 0.01
```
jkozdon: This seems to be the same `dt` as the current explicit version of the [rising thermal bubble](https://github.com/climate-machine/CLIMA/blob/9594889aa4b0b461122bc08cc1ee73ccd0eba27c/test/DGmethods/compressible_Navier_Stokes/rising_bubble-model.jl#L47). Can this be made any bigger? (Also don't know if the explicit `dt` is as large as possible.)

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,71 @@
...
+vars_diffusive(lm::AtmosLinearModel, T) = @vars()
+vars_aux(lm::AtmosLinearModel, T) = vars_aux(lm.atmos,T)
+vars_integrals(lm::AtmosLinearModel,T) = @vars()
+
+
+update_aux!(dg::DGModel, lm::AtmosLinearModel, Q::MPIStateArray, auxstate::MPIStateArray, t::Real) = nothing
+integrate_aux!(lm::AtmosLinearModel, integ::Vars, state::Vars, aux::Vars) = nothing
+flux_diffusive!(lm::AtmosLinearModel, flux::Grad, state::Vars, diffusive::Vars, aux::Vars, t::Real) = nothing
+function wavespeed(lm::AtmosLinearModel, nM, state::Vars, aux::Vars, t::Real)
+  return soundspeed_air(Float64(T_0))
```
jkozdon: I would expect this to be the reference profile temperature here? @fxgiraldo or @mwarusz

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,71 @@
...
+end
+
+boundary_state!(nf, lm::AtmosLinearModel, x...) = nothing
```
jkozdon: Shouldn't we need boundary conditions for the linear problem? @fxgiraldo or @mwarusz

src/Atmos/Model/linear.jl
simonbyrne: @charleskawczynski what function should I use here to get temperature of reference state?

src/Atmos/Model/linear.jl
jkozdon: Also this is never actually used. Is that deliberate?

jkozdon:
Also need to add this to the slurmci tests

src/Atmos/Model/linear.jl
jkozdon: Don't know that it should be, just wondering.

simonbyrne: 🤦‍♂  that's how we specify the reference state.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
simonbyrne: We need a proper integration test for this, see #462.

src/Atmos/Model/linear.jl
tapios: Let‘s avoid `T_0` for reference state temperature (for dynamical calculations) because we use `T_0` as the thermodynamic reference temperature (a PlanetParameter) that specifies at which temperature reference specific heats are evaluated.

tapios: The reference temperature is a function of height generally, and so would be the speed of sound. Do you want to take the `max` of this here?

charleskawczynski: Yeah, so we'll want something like `return soundspeed_air(aux.ref_state.T)`

simonbyrne: We store the reference temperature, so can just use that.

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,63 @@
...
+  DF = eltype(state)
+  ref = aux.ref_state
+  e_pot = gravitational_potential(lm.atmos.orientation, aux)
+
+  flux.ρ = state.ρu
+  # TODO: use MoistThermodynamics.linearized_air_pressure 
+  # need to avoid dividing then multiplying by ρ
+  pL = state.ρ * DF(R_d) * DF(T_0) + DF(R_d) / DF(cv_d) * (state.ρe - state.ρ * e_pot)
+  flux.ρu += pL*I
+  flux.ρe = ((ref.ρe + ref.p)/ref.ρ - e_pot)*state.ρu
```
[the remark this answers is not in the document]
simonbyrne: this is my interpretation of eqn 2.35 in the overleaf docs: is this correct @tapios?

src/Atmos/Model/linear.jl
tapios: Yes, you need linear boundary conditions. But keep in mind that the wave modes resolved here transport nothing. Hence, the boundary conditions are not the same as for the transport term. Just use linear reflecting boundary conditions.

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,63 @@
...
+end
+function flux_nondiffusive!(lm::AtmosAcousticLinearModel, flux::Grad, state::Vars, aux::Vars, t::Real)
+  DF = eltype(state)
+  ref = aux.ref_state
+  e_pot = gravitational_potential(lm.atmos.orientation, aux)
+
+  flux.ρ = state.ρu
+  # TODO: use MoistThermodynamics.linearized_air_pressure 
+  # need to avoid dividing then multiplying by ρ
+  pL = state.ρ * DF(R_d) * DF(T_0) + DF(R_d) / DF(cv_d) * (state.ρe - state.ρ * e_pot)
```
tapios: @charleskawczynski provided a function `linearized_air_pressure` in `MoistThermodynamics` for this. It seems preferable to me to use that.

src/Atmos/Model/linear.jl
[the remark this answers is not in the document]
tapios: Correct, you want to subtract the potential energy if gravity waves are excluded. To include gravity waves, you leave the potential energy term here and in the momentum equations.

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,63 @@
...
+struct AtmosAcousticGravityLinearModel{M} <: AtmosLinearModel
+  atmos::M
+end
+function flux_nondiffusive!(lm::AtmosAcousticGravityLinearModel, flux::Grad, state::Vars, aux::Vars, t::Real)
+  DF = eltype(state)
+  ref = aux.ref_state
+  e_pot = gravitational_potential(lm.atmos.orientation, aux)
+
+  flux.ρ = state.ρu
+  pL = state.ρ * DF(R_d) * DF(T_0) + DF(R_d) / DF(cv_d) * (state.ρe - state.ρ * e_pot)
```
tapios: As above, use `linearized_air_pressure`?

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,63 @@
...
+vars_aux(lm::AtmosLinearModel, T) = vars_aux(lm.atmos,T)
+vars_integrals(lm::AtmosLinearModel,T) = @vars()
+
+
+update_aux!(dg::DGModel, lm::AtmosLinearModel, Q::MPIStateArray, auxstate::MPIStateArray, t::Real) = nothing
+integrate_aux!(lm::AtmosLinearModel, integ::Vars, state::Vars, aux::Vars) = nothing
+flux_diffusive!(lm::AtmosLinearModel, flux::Grad, state::Vars, diffusive::Vars, aux::Vars, t::Real) = nothing
+function wavespeed(lm::AtmosLinearModel, nM, state::Vars, aux::Vars, t::Real)
+  ref = aux.ref_state
+  return soundspeed_air(ref_state.T)
```
mwarusz:
```suggestion
  return soundspeed_air(ref.T)
```

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
```diff
@@ -0,0 +1,220 @@
...
+  linmodel = LinearType(model)
+  lindg = DGModel(linmodel,
+               grid,
+               Rusanov(),
+               CentralNumericalFluxDiffusive(),
+               CentralGradPenalty(); auxstate=dg.auxstate)
+
+  Q = init_ode_state(dg, DF(0))
+
+  linearsolver = GeneralizedMinimalResidual(10, Q, 1e-10)
```
mwarusz: `1e-10` is too high tolerance for `Float32` . Something like `1e-5` may work in that case.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
mwarusz: After adding boundary conditions for the linear model and fixing small problems mentioned in my recent comments I was able to run this test with much bigger time step (e.g. `dt = 0.4` worked).

jkozdon: Do you know the largest time step running in fully explicit mode?

fxgiraldo: It would be helpful to be able to be able to extract Courant number values (e.g., both with respect to acoustic waves and velocity). It would also be beneficial to subdivide this information into vertically and horizontally propagating waves.

src/Atmos/Model/linear.jl
simonbyrne: Yes, see the comment. There were some issues in that it divides then multiplying by ρ.

test/DGmethods/compressible_Navier_Stokes/rising_bubble-model-imex.jl
simonbyrne: How about?
```suggestion
  linearsolver = GeneralizedMinimalResidual(10, Q, sqrt(eps(DF)))
```

mwarusz: Seems fine. The tolerance in the old IMEX examples was set very high to get reproducible test results. In normal runs it will be much smaller. Whatever works to have a test that will not break on different machines is fine here.

src/Atmos/Model/linear.jl
```diff
@@ -0,0 +1,68 @@
...
+  ref = aux.ref_state
+  return soundspeed_air(ref.T)
+end
+
+function boundary_state!(nf::Rusanov, lm::AtmosLinearModel, x...)
+  atmos_boundary_state!(nf, NoFluxBC(), lm.atmos, x...)
+end
+function boundary_state!(nf::CentralNumericalFluxDiffusive, lm::AtmosLinearModel, x...)
+  nothing
+end
```
simonbyrne: @jkozdon @mwarusz does this seem a reasonable default?

simonbyrne:
bors try

src/Atmos/Model/linear.jl
jkozdon: Seems reasonable to me.

mwarusz: I agree, seems fine.

src/Atmos/Model/linear.jl
charleskawczynski: I'll add a function in MoistThermo that doesn't require the multiply-divide

PR #426, opened 2019-09-09
SGS-Vreman moved to AtmosModel framework

asraero:
This PR moves the Vreman (2004) SGS model into the new AtmosModel framework. 
1) Enables additional testing with the DYCOMS LES test case. 
[x] Tested with Robert (1993) Bubble 
[x] Tested with Straka (1993) Density Current
http://www.vremanresearch.nl/Vreman-PF2004-subgridmodel.pdf 

@smarras79

PR #446, opened 2019-09-20
First pass at updating naming conventions. 

blallen:
Noticed that some branches are modifying these kernels so I wanted to open a PR to start this discussion. The purpose of this was to have consisent naming across functions in the core DG kernels:  DGmodel.jl, DGmodel_kernels.jl, and NumericalFluxes.jl. ShallowWaterModel.jl and GyreDriver.jl show how this would affect the user's end. 

Here's the dictionary for what I did. DG model currently implements equation 8 from [the working doc from the NPS-MIT visit](https://www.overleaf.com/project/5d6ea7b157caba0001d29b5a)

- Q/q = state variables
- α = auxiliary variables
- σ = viscous / diffusive variables
- F = F(q, α) is the non-diffusive flux 
- F' = F'(q, α, σ) is the diffusive flux
- G = G(q, α) to refer to gradient transform / gradvariables
- ∇G to refer to the gradient of the gradient transform (decided it was too clunky to use an entirely new variable)
- S = S(q, α) to refer to the source terms

Additionally I've started a little on updating the topology / grid terms:
- ι = interface / boundary mappings
- Ω = the set of real elements

one proposal for the flux naming convention is to label all fluxes 𝚽 and use subscripts to refer to the order of the flux (and maybe superscripts for the number of integrals). For example, 𝚽 is the non-diffusive flux, 𝚽₁ is the diffusive flux, 𝚽₂ would be used for a third-order term, and 𝚽¹ would refer to a flux that contains a single integral of the state and/or auxiliary variables. I tentatively used this convention in facerhs! to distinguish between the two Q⁺ and α⁺ states needed for the nondiffusive and diffusive fluxes.

The exact symbols / names of objects can be changed, but I think I have broken things down into their logical parts. I think we should try to stick as close to math conventions for names in the DG kernels and then the individual models can use atmospheric and/or oceanographic language where appropriate.


jkozdon:
@blallen Thanks for taking a stab at this! There is certainly a lot of cleaning up that needs to happen across the code, glad to see someone start to take it on.

src/DGmethods/DGmodel.jl
```diff
@@ -8,124 +8,110 @@ end
...
-  Nq = N + 1
-  Nqk = dim == 2 ? 1 : Nq
-  Nfp = Nq * Nqk
-  nrealelem = length(topology.realelems)
-
-  Qvisc = param.diff
-  auxstate = param.aux
+  Ω    = grid.topology.realelems
+  nΩ   = length(Ω)
+  Nᵈ   = dimensionality(grid)
```
jkozdon: `Nd` or just `d` as this might be easier to type regularly (this matches `Nq`, `Np`, etc. below)

src/DGmethods/DGmodel.jl
```diff
@@ -8,124 +8,110 @@ end
...
-  polyorder = polynomialorder(dg.grid)
-
-  Np = dofs_per_element(grid)
-
-  if hasmethod(update_aux!, Tuple{typeof(dg), typeof(bl), typeof(Q),
-                                  typeof(auxstate), typeof(t)})
-    update_aux!(dg, bl, Q, auxstate, t)
+  ω    = grid.ω
+  D    = grid.D
+  ι⁻   = grid.vmapM
```
jkozdon: Personally I like `vmap*`, (especially since it's hard to see what character exactly `ι` is in my browser font)

src/DGmethods/NumericalFluxes.jl
```diff
@@ -31,39 +31,47 @@ function diffusive_boundary_penalty! end
...
+  DFloat = eltype(Q⁻)
 
   @inbounds begin
-    ndim = 3
+    Nᵈ = 3
     ngradstate = num_gradient(bl,DFloat)
-    n_Δdiff = similar(VF, Size(ndim, ngradstate))
-    for j = 1:ngradstate, i = 1:ndim
-      n_Δdiff[i, j] = nM[i] * (diffP[j] - diffM[j]) / 2
+    ∇G = similar(σ, Size(Nᵈ, ngradstate))
```
jkozdon: This is maybe a little confusing since it's not actually the gradient of `G`, but the normal time times `G`

src/DGmethods/DGmodel.jl
```diff
@@ -8,124 +8,110 @@ end
...
-  dim = dimensionality(grid)
-  N = polynomialorder(grid)
-  Nq = N + 1
-  Nqk = dim == 2 ? 1 : Nq
-  Nfp = Nq * Nqk
-  nrealelem = length(topology.realelems)
-
-  Qvisc = param.diff
-  auxstate = param.aux
+  Ω    = grid.topology.realelems
```
jkozdon: I tend to think of `Ω` as the physical domain. Not sure that I have a better suggestion, just this is what comes to mind when I see this. Maybe `E` or `K` or something like this? Perhaps looking a standard finite element text could help (say Ciarlet?)

jkozdon:
I couldn't get to the overleaf document, but another comment would be that `F'` might be easily confused with a derivative.

jkozdon:
Overall I think this cleanup looks very nice. Thanks for taking a stab at it!

cc: @lcw @fxgiraldo @mwarusz @thomasgibson

asraero:
Thanks @blallen ! @charleskawczynski pointed out inconsistent rendering of symbols (sub/superscripts) in #426, so we should watch out for these cases as we rename components

tapios:
Could you make the Overleaf public, or better yet, add to the ocean folder in the Overleaf where all design docs are? it would be good to keep them together. (I can send you a link in pm if you do not have it.)

src/DGmethods/DGmodel.jl
tapios: Better to avoid Omega because we use it for planetary spin vector.

src/DGmethods/DGmodel.jl
```diff
@@ -8,124 +8,110 @@ end
...
   ###################
   # RHS Computation #
   ###################
-  @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
-          volumerhs!(bl, Val(dim), Val(polyorder), dQdt.Q, Q.Q, Qvisc.Q, auxstate.Q,
-                     vgeo, t, lgl_weights_vec, Dmat,
-                     topology.realelems, increment))
+  @launch(device, threads=(Nq, Nq, Nqk), blocks=nΩ,
+          volumerhs!(bl, Val(Nᵈ), Val(N), dQdt.Q, Q.Q, σ.Q, α.Q, vgeo, t, ω, D,
+                     Ω, increment))
```
tapios: I think it's better if we replace rhs by tendency to avoid confusion (a discussion we had a few times in the past...)

tapios:
Great first pass to make this clearer.

But please have a look at the atmosphere model design doc. We chose variable names in there and in the atmosphere model code in an attempt to avoid ambiguities. 

> Q/q = state variables

`Q` is a diabatic source, `q` a specific humidity in other parts of the code. Please avoid them for states. We choose `Y` for state in the design document. In the code, something more verbose would be better (`state` etc., as in the atmosphere model)

> σ = viscous / diffusive variables

See the SGS fluxes @asraero and others implemented. 

> F = F(q, α) is the non-diffusive flux
> F' = F'(q, α, σ) is the diffusive flux

More verbose is better; the prime is ambiguous. E.g., 'flux_diffusive' etc. Same for other variables. @simonbyrne will probably have input.

Basically, it would be good to be able to read the code without having to consult a dictionary.

sandreza:
It seems like there are a few ways we could go about naming things:
a) Consistency across the DG Kernel, Atmospheric Model, Ocean Model, etc. (This may lead to too much bickering about "proper names".)
b) Consistency within the norms of each respective field. (This will lead to inconsistent naming in the code.)
c) Some combination of a) and b). For example, we could agree to keep naming consistent between the Atmosphere and the Ocean, but let the DG kernels themselves have naming conventions consistent with a textbook (let's say the Nodal Discontinuous Galkerin Book) or applied mathematics in general. It is then the job of the atmospheric and ocean code to translate and make it readable for people within the field.

I believe the changes that Brandon was suggesting was strictly for the DG kernels themselves rather than the Ocean or Atmospheric codes.

tapios:
Complete consistency across the stack is probably an impossible goal. That being said, no need to be inconsistent (e.g., about naming fluxes) where consistency can be achieved. That just creates confusion and makes the code hard to read (if the same objects have very different names in different parts of the stack).

src/DGmethods/DGmodel.jl
blallen: we can replace \iota. I used simple roman i in our implementation of the textbook code. main point is that it is a collection of indices for the GL points, so the name/symbol should give some reference to that

src/DGmethods/NumericalFluxes.jl
blallen: ah, when does the actual gradient of G happen then? 

I can change this line to be nG or n\hat G or or n\Delta G or some other variant

src/DGmethods/DGmodel.jl
blallen: I used Ω because this is what was used in the NDG book by Hesthaven and Warburton. Open to renaming it, just not sure what a good choice is

src/DGmethods/DGmodel.jl
blallen: makes sense to me

do we want something like volume_tendency! or volume_tendency_contribution! etc?

src/DGmethods/DGmodel.jl
blallen: yeah for this variable I can see that. I generally prefer using superscript where you would do so in writing an equation, but there are so many of these N variables that it is hard to do it properly

src/DGmethods/DGmodel.jl
fxgiraldo: I think either E or K should work. We tend to call it \Omega_e in DG notation but K also is used in the finite element formalism.

blallen:
I took a look at section 2.1 of the design doc and will base future decisions off that. Just pushed a commit switching the state variables to Y.

For the DG kernels, I would like to keep the notation as close to the math notation as possible, as this code doesn't know anything physics specific and usually won't be touched by end-users. I think a one-to-one correspondence between the symbols used in the equations in Section 2.1 and those used for variable names in the code would greatly ease understanding these kernels.

On that note, @jkozdon, there's a note that you are rewriting this section. Are Y, Σ, D, and G finalized? For the fluxes, we can go with Fⁿ and Fᵈ to refer to the two fluxes for now, but this doesn't really generalize all that well. 

Also, MPIStateArray currently uses Q to refer to the array of the nodal values of state variables. Do we want to change this to Y? Or do we want to use a different symbol since state, auxiliary, and diffusive variables are all MPIStateArrays and avoid things like α.Y? If the latter, I can propose ϕ as an option to refer to the value of the fields.

src/DGmethods/DGmodel.jl
```diff
@@ -6,126 +6,112 @@ struct DGModel{BL,G,NFND,NFD,GNF}
...
+  Ω    = grid.topology.realelems
+  nΩ   = length(Ω)
+  Nᵈ   = dimensionality(grid)
+  N    = polynomialorder(grid)
+  Nq   = N + 1
+  Nqk  = Nᵈ == 2 ? 1 : Nq
+  Nfp  = Nq * Nqk
+  Np   = dofs_per_element(grid)
+
+  σ  = param.diff
```
tapios: Please retain variables with something like `diff` or `diffusive` and `aux` here, so we do not need a dictionary to translate.

src/DGmethods/DGmodel.jl
```diff
@@ -6,126 +6,112 @@ struct DGModel{BL,G,NFND,NFD,GNF}
...
-  if nviscstate > 0
+  MPIStateArrays.start_ghost_exchange!(Y)
+  MPIStateArrays.start_ghost_exchange!(α)
 
-    @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
-            volumeviscterms!(bl, Val(dim), Val(polyorder), Q.Q, Qvisc.Q, auxstate.Q, vgeo, t, Dmat,
-                             topology.realelems))
+  if nσ > 0
+    @launch(device, threads=(Nq, Nq, Nqk), blocks=nΩ,
+            volumeviscterms!(bl, Val(Nᵈ), Val(N), Y.Q, σ.Q, α.Q, vgeo, t, D, Ω))
```
tapios: While you are at it, might as well replace `visc` by `diff` or `diffusive`, which is more generic and captures more broadly what these are about.

src/DGmethods/DGmodel.jl
```diff
@@ -6,126 +6,112 @@ struct DGModel{BL,G,NFND,NFD,GNF}
...
 
   weights = view(h_vgeo, :, grid.Mid, :)
   weights = reshape(weights, size(weights, 1), 1, size(weights, 2))
 
-
-
-
   # TODO: Clean up this MPIStateArray interface...
-  diffstate = MPIStateArray{Tuple{Np, num_diffusive(bl,DFloat)},DFloat, DA}(
+  σ = MPIStateArray{Tuple{Np, num_diffusive(bl,DFloat)},DFloat, DA}(
```
tapios: Please leave it verbose with `diff` in the name.

src/DGmethods/DGmodel.jl
```diff
@@ -138,7 +124,7 @@ function init_ode_param(dg::DGModel)
     weights=weights,
     commtag=111)
 
-  auxstate = MPIStateArray{Tuple{Np, num_aux(bl,DFloat)}, DFloat, DA}(
+  α = MPIStateArray{Tuple{Np, num_aux(bl,DFloat)}, DFloat, DA}(
```
tapios: Same here: please leave verbose, it‘s clearer.

src/DGmethods/DGmodel.jl
```diff
@@ -175,176 +157,149 @@ Initialize the ODE state array.
...
+                   Y::MPIStateArray)
 
 Iterate over each dof to fill `R` using the `dof_fun!`. The syntax of the
 `dof_fun!` is
 ```
-dof_fun!(l_R, l_Q, l_Qvisc, l_aux)
+dof_fun!(l_R, l_Y, l_σ, l_aux)
 ```
-where `l_R`, `l_Q`, `l_Qvisc`, and `l_aux` are of type `MArray` filled initially
+where `l_R`, `l_Y`, `l_σ`, and `l_α` are of type `MArray` filled initially
```
tapios: Same as above: `l_Ydiff` (or similar) and `l_aux` is better. May even consider `state` in variable name.

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -439,7 +445,7 @@ function volumeviscterms!(bl::BalanceLaw, ::Val{dim}, ::Val{polyorder},
           ξ2x1, ξ2x2, ξ2x3 = vgeo[ijk, _ξ2x1, e], vgeo[ijk, _ξ2x2, e], vgeo[ijk, _ξ2x3, e]
           ξ3x1, ξ3x2, ξ3x3 = vgeo[ijk, _ξ3x1, e], vgeo[ijk, _ξ3x2, e], vgeo[ijk, _ξ3x3, e]
 
-          @unroll for s = 1:ngradstate
+          @unroll for s = 1:nG
```
tapios: More verbose here as well. `ngradstate` was better than `nG`. Something that makes the function of the variable clearer.

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -469,107 +477,110 @@ function volumeviscterms!(bl::BalanceLaw, ::Val{dim}, ::Val{polyorder},
...
-
-  if dim == 1
+                        Y, σ, α, vgeo, sgeo, t, ι⁻, ι⁺,
+                        ιᴮ, Ω) where {Nᵈ, N}
+  DFloat = eltype(Y)
+
+  nY = num_state(bl,DFloat)
+  nG = num_gradient(bl,DFloat)
+  nσ = num_diffusive(bl,DFloat)
+  nα = num_aux(bl,DFloat)
```
tapios: Previous variable names were clearer. Stick with something closer to them.

tapios:
> For the DG kernels, I would like to keep the notation as close to the math notation as possible, as this code doesn't know anything physics specific and usually won't be touched by end-users. 

Lots of people will need to interact with this code (atmosphere, ocean, land, sea ice, land ice etc.). We need to make barriers to understanding it as low as possible, and this means making the code function as clear as possible, including variable names that are self-explanatory to the degree they can be (i.e., as understandable as possible without consulting documents as look-up tables). In other words, think of a list of symbols with explanations as included in your document, and name variables as much as possible according to the explanations, not the symbols. (Same is true for all other parts of the code.)

jkozdon:
> On that note, @jkozdon, there's a note that you are rewriting this section. Are Y, Σ, D, and G finalized? For the fluxes, we can go with Fⁿ and Fᵈ to refer to the two fluxes for now, but this doesn't really generalize all that well.

Nothing is finalized yet. I just modified notation to sort of match what was currently going on in the code.

charleskawczynski:
Glad to see the naming convention is being worked on. This will likely need continuous attention until the layers of abstraction are converged.

src/DGmethods/DGmodel.jl
jkozdon: In Hesthaven and Warburton (NDG): Ω is the domain (page 3), Ωh is the union of all the elements (page 19). The individual elements are denoted Dk (page 3). As far as I know, they do not have any notation for the set of elements {Dk}_{k=1}^{K} which is really what we are looking for here.

Ωh is close, but I tend to think of Ωh the domain covered by the set of elements and many different tessellations could have the same Ωh (especially since Ω ≃ Ωh, see page 19).

@fxgiraldo is Ωe usually a single element, as opposed to the collection/set of elements?

Just my two cents on the matter.

src/DGmethods/DGmodel.jl
rafferrari: I am not sure `aux' is self-explanatory. You will always end up needing a dictionary. I think the issue here is to find the right balance between using somewhat self-explanatory variable names without ending up with equations that are too long and verbose. There is a reason why we introduce symbols in math :-)

src/DGmethods/DGmodel.jl
rafferrari: I agree here.

src/DGmethods/DGmodel.jl
rafferrari: Isn't the alpha subscript what is used in the design docs?

src/DGmethods/DGmodel.jl
```diff
@@ -6,126 +6,112 @@ struct DGModel{BL,G,NFND,NFD,GNF}
   gradnumflux::GNF
 end
 
-function (dg::DGModel)(dQdt, Q, param, t; increment=false)
+function (dg::DGModel)(dYdt, Y, param, t; increment=false)
```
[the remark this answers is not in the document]
blallen: Quoting Tapio:

> > Q/q = state variables
> 
> `Q` is a diabatic source, `q` a specific humidity in other parts of the code. Please avoid them for states. We choose `Y` for state in the design document. In the code, something more verbose would be better (`state` etc., as in the atmosphere model)

I think `u` is a bad choice as it is also the component of velocity in the 'x'-direction. Andre and I used it when learning DG and it led to some confusion because of that

src/DGmethods/DGmodel.jl
```diff
@@ -6,126 +6,112 @@ struct DGModel{BL,G,NFND,NFD,GNF}
...
+  nΩ   = length(Ω)
+  Nᵈ   = dimensionality(grid)
+  N    = polynomialorder(grid)
+  Nq   = N + 1
+  Nqk  = Nᵈ == 2 ? 1 : Nq
+  Nfp  = Nq * Nqk
+  Np   = dofs_per_element(grid)
+
+  σ  = param.diff
+  α  = param.aux
```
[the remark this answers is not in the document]
blallen: Yep, I already think we need to move all the outputs of integration to a similar structure to the diffusive variables

src/DGmethods/DGmodel.jl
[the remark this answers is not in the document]
simonbyrne: That seems reasonable.

src/DGmethods/DGmodel.jl
blallen: switched to using `Nd`

src/DGmethods/NumericalFluxes.jl
blallen: After thinking about this some more, I think `∇G` makes sense. You are calculating the normal times the jump in `G`, which is the contribution to the DG gradient coming from the surface. the `diffusive!` function call then adds the contribution from the volume turning this variable into the full DG gradient of `G`.

does this interpretation make sense to you?

src/DGmethods/DGmodel.jl
blallen: I'll go with `E` since it is relatively intuitive. There is some overlap with energy, but hopefully we are never explicitly referring to the set of all elements and the energy explicitly in the same kernels.

src/DGmethods/DGmodel.jl
blallen: okay changes are 

`volumerhs!` --> `volume_tendency!`
`facerhs!` --> `face_tendency!`

src/DGmethods/DGmodel.jl
blallen: went with 

`volumeviscterms!` --> `volume_diffusive_terms!`
`faceviscterms!` --> `face_diffusive_terms!`

I think in the DG kernels we might want to replace diffusive with gradient or something similar, as not all second order terms are diffusive. it will also be more consistent if we do introduce integral variables and need to compute those here are well. for the higher level interfaces of AtmosModel we can still use diffusive though

src/DGmethods/NumericalFluxes.jl
blallen: @jkozdon bump

src/DGmethods/DGmodel.jl
blallen: @jkozdon bump

jkozdon: I not sure what you're asking. vmapM = volume map to the minus side, that seems pretty clear to me.

src/DGmethods/NumericalFluxes.jl
jkozdon: I think that interpretation is mostly fine. What you are calling `∇G` isn't quite that though, but at the end of the day not sure your name is any worse than `n_Δdiff`. If it makes sense to you, I'm fine with it (it's defined right above, so someone can see what it is).

src/DGmethods/DGmodel.jl
jkozdon: Also, if folks want to alias this to `ι⁻` in the kernels that's fine. I'd personally keep a more descriptive name in the type. But even that would be fine to change if it doesn't make sense to you what this is.

src/DGmethods/NumericalFluxes.jl
jkozdon: I think that interpretation is mostly fine. What you are calling ∇G isn't quite that though, but at the end of the day not sure your name is any worse than n_Δdiff. If it makes sense to you, I'm fine with it (it's defined right above, so someone can see what it is).

src/DGmethods/DGmodel.jl
sandreza: oh so that's vmap meant! All summer I thought it meant "vector map" ... I like the symbol with iotas, but using vmap makes it easier to follow for anyone who has read the DG book

jkozdon: `vmap` = volume map
`fmap` = face map (sometimes just `map` in the book codes)

blallen: okay, maybe for the grid struct we can expand out to `volume_map` and `face_map` and use `vmap` and `fmap` in the kernels themselves. I can't think of a good symbol that encompasses both

blallen:
Tried to test out the diagnostics from #508 on my ocean3D branch and realized we have diverged significantly. Caught this branch up to master. @jkozdon can you check the changes in the latest commit for DGmodel.jl and DGmodel_kernels.jl? I think I preserved everything, but might have made some mistakes.

The only remaining issue for this PR is symbols versus abbreviations debate. I think using symbols in the kernel is the proper way to go, especially since there seems to be a general syntax for DG across various different papers.

src/DGmethods/DGmodel.jl
blallen: actually decided to just go with `M` for map here since we are only using the volume maps in these kernels

src/DGmethods/DGmodel.jl
```diff
@@ -21,226 +21,199 @@ function DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux;
           diffstate, direction)
 end
 
-function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
+function (dg::DGModel)(dYdt, Y, param, t; increment=false)
```
jkozdon:
```suggestion
function (dg::DGModel)(dYdt, Y, ::Nothing, t; increment=false)
```

src/DGmethods/DGmodel.jl
```diff
@@ -21,226 +21,199 @@ function DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux;
...
+  E    = grid.topology.realelems
+  nE   = length(E)
+  Nd   = dimensionality(grid)
+  N    = polynomialorder(grid)
+  Nq   = N + 1
+  Nqk  = Nd == 2 ? 1 : Nq
+  Nfp  = Nq * Nqk
+  Np   = dofs_per_element(grid)
+
+  σ  = param.diff
```
jkozdon:
```suggestion
  σ  = dg.diff
```

src/DGmethods/DGmodel.jl
```diff
@@ -21,226 +21,199 @@ function DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux;
...
+  α  = param.aux
```
jkozdon:
```suggestion
  α  = dg.aux
```

jkozdon:
Not sure what you want me to check. There are a lot of changes to the files. Might be best to start clean from master and re-add everything you did?

Does the CI pass, that's really the best way to check.

blallen:
> Not sure what you want me to check. There are a lot of changes to the files. Might be best to start clean from master and re-add everything you did?
> 
> Does the CI pass, that's really the best way to check.

Okay, I'll start testing that way. Already found some fixes when running the ShallowWater model. I'll run the full CI now.

blallen:
bors try

blallen:
bors try

christophernhill:
@blallen how about we use a short prefix to help with ambiguity and give some context. i.e.

Q (currently state) would be DGQ, A (currently aux) would be DGA, G (currently diff) would be DGG, F (currently flux) would be DGF, S (currently source) would be DGS. This would allow things to line up for readability and provide context to help avoid confusion.

johncmarshall54:
This sounds like a good plan that addresses various concerns:
readability
ability to search
disambiguation between different kernel.
Another comment is that this is deep in the innards of the code which few users will be dealing with.

blallen:
bors try

blallen:
bors try

bischtob:
@blallen, is this PR still relevant?

blallen:
> @blallen, is this PR still relevant?

Nope, was keeping it around to scavenge from. But now it is so stale I can't even see the differences easily.

PR #449, opened 2019-09-23
Performance CI (with SLURM)

kpamnany:
Add performance scripts and test set. Also move to OpenMPI 4.0.1 for both CPU and GPU.

This is still in an early stage and needs work, but see https://gist.github.com/climabot/903b740b240bfa52f1e8ca3726f9cfc2 for an idea of what it will do.

Edit to expand: the primary purpose is to show performance changes across commits. Analysis of overall application performance is also intended, rather than just the kernels. Apart from that, please suggest what sort of performance data you'd like to see.


.slurmci/cpu-init.sh
```diff
@@ -7,12 +7,9 @@
 set -euo pipefail
 set -x #echo on
 
-export PATH="${PATH}:${HOME}/julia-1.2/bin"
 export JULIA_DEPOT_PATH="$(pwd)/.slurmdepot/cpu"
 export OPENBLAS_NUM_THREADS=1
 
-module load openmpi/3.1.4
+module load julia/1.2.0 openmpi/4.0.1
```
jkozdon: Just a note that we ran into issues with openmpi/4.0.1, which is why we downgraded to 3.1.4. (see https://github.com/climate-machine/CLIMA/pull/381#issuecomment-523306892)

simonbyrne:
bors try

.slurmci/cpu-init.sh
kpamnany: I saw that, but didn't see any such errors when testing. At some point, using a CUDA-aware MPI is necessary...

jkozdon: Just wanted to make sure you were aware of it. Something we need to keep an eye on.

simonbyrne:
As discussed on the call, we probably don't need all these. From our point of view, dycoms is probably the most useful at the moment.

@lcw @jkozdon are there any others you would like to see tracked?

lcw:
Dycoms sounds fine.  We should also make sure the number of elements and polynomial order is the desired size for the production runs.

I would also like to see kernels get flagged if they have local loads or stores.

jkozdon:
> Dycoms sounds fine. We should also make sure the number of elements and polynomial order is the desired size for the production runs.

agreed. GPU performance makes the most sense when the GPU is fairly loaded up (most of our tests do not do this right now).

kpamnany:
As far as I can tell, getting kernel runtime and _any_ other metrics, will require two invocations of `nvprof`, i.e. running the whole app twice, unless anyone knows better?

Just FYI: `facerhs`, `volumeviscterms`, and `initauxstate` would be flagged for both local loads and local stores. Also the reduce kernels use a number of local stores.

kpamnany:
> Dycoms sounds fine. We should also make sure the number of elements and polynomial order is the desired size for the production runs.

@asraero: can you confirm?

asraero:
> > Dycoms sounds fine. We should also make sure the number of elements and polynomial order is the desired size for the production runs.
> 
> @asraero: can you confirm?

Yes. DYCOMS is a reasonable case for this. I can increase the problem resolution per @lcw 's suggestion.

kpamnany:
The idea is now to remove all the Slurm stuff from this repository (I haven't done this yet). Tests and performance tests will be specified in the newly added `.toml` files for SlurmCI to load and run.

This allows the independent development of the performance testing framework.

Two things also change:
- Dependencies cannot be specified (i.e. test B should not run until test A finishes) because I don't see an easy way to specify this in TOML. Also, I didn't see that feature used.
- There's no distinction between `cpu_gpu` and `gpu`. Until a recent commit, `gpu` was empty. Now, there is a test there, but what does it mean to have a GPU only test?

jkozdon:
> The idea is now to remove all the Slurm stuff from this repository (I haven't done this yet). Tests and performance tests will be specified in the newly added `.toml` files for SlurmCI to load and run.
> 
> This allows the independent development of the performance testing framework.

Is this for all the CI or just for the performance testing?

> Two things also change:
> 
> * Dependencies cannot be specified (i.e. test B should not run until test A finishes) because I don't see an easy way to specify this in TOML. Also, I didn't see that feature used.

Won't this mean that the CI will take a long time since nothing is run in parallel?

> * There's no distinction between `cpu_gpu` and `gpu`. Until a recent commit, `gpu` was empty. Now, there is a test there, but what does it mean to have a GPU only test?

CLIMA is not very efficient on the CPU, running the added test on only three CPU-only ranks would be very slow. Personally I like the idea of specifying tests that are to run on the CPU-only, GPU-only, or on both since we've had problems in the past with tests not getting run on the GPU since people forgot to add them to both lists.

slurmci-test.toml
```diff
@@ -0,0 +1,40 @@
...
+]
+
+gpu = [
+    ["3", "examples/DGmethods_old/ex_001_periodic_advection.jl"],
+    ["3", "examples/DGmethods_old/ex_002_solid_body_rotation.jl"],
+    ["3", "examples/DGmethods_old/ex_003_acoustic_wave.jl"],
+    ["3", "examples/DGmethods_old/ex_004_nonnegative.jl"],
+    ["3", "examples/Microphysics/ex_1_saturation_adjustment.jl"],
+    ["3", "examples/Microphysics/ex_2_Kessler.jl"],
+    ["3", "test/DGmethods/advection_diffusion/pseudo1D_advection_diffusion.jl"],
```
jkozdon: Please pass the "true" command line argument through to this test

kpamnany:
> Is this for all the CI or just for the performance testing?

This is for all of it; SlurmCI is a common framework to run CLIMA stuff on the central cluster with SLURM.

> > * Dependencies cannot be specified (i.e. test B should not run until test A finishes) because I don't see an easy way to specify this in TOML. Also, I didn't see that feature used.
> 
> Won't this mean that the CI will take a long time since nothing is run in parallel?

No, everything runs in parallel. The dependencies are currently used to prevent all tests from running until the package depot is set up, i.e. packages are downloaded, built, and precompiled (the `*-init.sh` scripts). That will still happen, but without it being explicitly specified.

> CLIMA is not very efficient on the CPU, running the added test on only three CPU-only ranks would be very slow. Personally I like the idea of specifying tests that are to run on the CPU-only, GPU-only, or on both since we've had problems in the past with tests not getting run on the GPU since people forgot to add them to both lists.

I'm missing something because I could see no difference between `cpu_gpu` and `gpu` in terms of how they were being run. What difference would you expect to see?

kpamnany:
> I'm missing something because I could see no difference between `cpu_gpu` and `gpu` in terms of how they were being run. What difference would you expect to see?

Thinking about it: perhaps `cpu_gpu` should mean run the test on the CPU _and_ on the GPU, i.e. run it twice? That's not what it currently does though, AFAICS.

kpamnany:
So, see:
https://github.com/climate-machine/CLIMA/blob/35b27da370a485d48a6e0d066cfb8c3023255431/.slurmci/test_set.jl#L1-L7

The 3 tests in `cpu_tests` are all also in `cpu_gpu_tests` with no added or removed parameters. So they are being run twice on the CPU and once on the GPU, surely not intentional?

slurmci-test.toml
kpamnany: That test exists in both `cpu_gpu` with no parameters, and in `gpu` with parameters. This means it will run 3 times: on the CPU with no parameters, on the GPU with no parameters, and on the GPU with the "true" parameter. Is that what is wanted?

jkozdon:
FYI, I'm trying to add more longer tests to the code as well. See PR #468

kpamnany:
> The problem with the test I have in the `gpu_test` section is running the longer test on the cpu takes too long (e.g., hours), so we just want to run it only on the GPU. We also want it on the CPU as a shorter test (e.g., smaller mesh) just to make sure that things are likely OK on the CPU.
> 
> Not sure what the best way to deal with this is, but I would like the test run with `true` at least once on the GPU.

I think we can accomplish this by putting it in `gpu` with the `true` parameter and in `cpu` without the parameter (I assume this is how you make it a smaller mesh?) and removing it from `cpu_gpu` altogether.

jkozdon:
> I think we can accomplish this by putting it in `gpu` with the `true` parameter and in `cpu` without the parameter (I assume this is how you make it a smaller mesh?) and removing it from `cpu_gpu` altogether.

Certainly. I can fix this in PR #468

jkozdon:
From: https://github.com/climate-machine/CLIMA/pull/468#issuecomment-538130916
> Please let me know if any of these should be in the performance tests also, alongside Dycoms, for SlurmCI testing.

@mwarusz based on what you're looking at should rising thermal bubble or density current be added to the performance CI? If so, can you suggest mesh sizes to @kpamnany for this?

mwarusz:
Possibly, but before rising thermal bubble can be used for this #462 needs to be addressed. Right now this test is not set up correctly to handle bigger meshes, for example `dt` is not scaled with grid spacing.

asraero:
> Possibly, but before rising thermal bubble can be used for this #462 needs to be addressed. Right now this test is not set up correctly to handle bigger meshes, for example `dt` is not scaled with grid spacing.

@mwarusz right; I'll set up these `slow` tests for the RTB / density current cases to run per #462 (likewise for DYCOMS with a larger mesh / longer `timeend`).

jkozdon:
> longer `timeend`

Do we need longer `timeend` for the performance metrics?

asraero:
> > longer `timeend`
> 
> Do we need longer `timeend` for the performance metrics?

Not for performance metrics; only for integration tests on gpu

kpamnany:
This now simply adds `slurmci-test.toml` and `slurmci-perf.toml`. The former should be directly equivalent to `.slurmci/test_set.jl`, with the double runs fixed (i.e. nothing should be in `cpu_gpu` _and_ either of the other two sections unless there's a difference in the parameters). This has to be merged before the new SlurmCI can be brought online.

slurmci-test.toml
kpamnany: I've assumed the answer is no, so this test is now in `cpu` with no parameters and in `gpu` with the `true` parameter.

kpamnany: Ah, I see you've fixed this in https://github.com/climate-machine/CLIMA/pull/468 as well as the other point discussed below so I'll mark this resolved.

PR #448, opened 2019-09-20
Adding Multirate time stepping

jkozdon:
Add support for low storage slow time loop with multirate (with substepping) based on LSRK or SSP.

Co-authored-by: Frank Giraldo <fxgirald@nps.edu>

PR #450, opened 2019-09-23
Testing multirate with  isentropic vortex

jkozdon:
Adds a test for isentropic vortex using multirate time stepping (see PR #448).

This is based on the splitting introduced in PR #443, and is dependent on the finalization of the splitting interface (i.e., approach implemented in #443 vs style proposed in #445).

Note that for isentropic vortex, multirate is not needed/beneficial so this code is substantially slower than the no standard approach.


src/Atmos/Model/nonlinear.jl
```diff
@@ -0,0 +1,82 @@
...
+
+  soundspeed_full = soundspeed(m.moisture, m.orientation, state, aux)
+  soundspeed_linear = soundspeed(m.moisture, m.orientation, aux.ref_state, aux)
+
+  if use_linearized_pressure
+    soundspeed_contribution = -zero(eltype(state))
+  else
+    soundspeed_contribution = soundspeed_full - soundspeed_linear
+  end
+  return abs(dot(nM, u)) - abs(dot(nM, u_ref)) + soundspeed_contribution
```
simonbyrne: Why do you subtract the linear soundspeed here?

PR #451, opened 2019-09-24
improve grid stretching interface, add tests

simonbyrne:
As an alternative to #444.

@smarras79 @slifer50 does this help make the code a bit simpler?

### DIFF
--- a/src/Mesh/Topologies.jl
+++ b/src/Mesh/Topologies.jl
@@ -6,7 +6,7 @@ using DocStringExtensions
 export AbstractTopology, BrickTopology, StackedBrickTopology,
     CubedShellTopology, StackedCubedSphereTopology, isstacked
 
-export grid_stretching_1d
+export grid1d, SingleExponentialStretching, InteriorStretching
 
 """
     AbstractTopology{dim}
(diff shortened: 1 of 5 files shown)


src/Mesh/Topologies.jl
```diff
@@ -1010,48 +1010,48 @@ end
 
 
 """    
-    grid_stretching_1d(coord_min, coord_max, Ne, stretching_type)
+    grid1d(a, b[, stretch::AbstractGridStretching]; elemsize, nelem)
```
simonbyrne: I'm still not 100% sold on the name `grid1d`, since we used "Grid" to mean other things.

@lcw @jkozdon any suggestions?

src/Mesh/Topologies.jl
```diff
@@ -1010,48 +1010,48 @@ end
...
-    #build logical space
-    s  = range(DFloat(0), length=Ne[1]+1, DFloat(1))
-
-    stretch_coe = 0.0
-    if (stretching_type == "boundary_stretching")
-        stretch_coe = 2.5
-        range_stretched = (coord_max - coord_min).*(exp.(stretch_coe * s) .- 1.0)./(exp(stretch_coe) - 1.0)
-    elseif (stretching_type == "top_stretching")
-        stretch_coe = 2.5
-        range_stretched = -(coord_max - coord_min).*(exp.(stretch_coe * s) .- 1.0)./(exp(stretch_coe) - 1.0)
```
simonbyrne: @smarras79 I couldn't quite figure out what this was doing. Do you have some references for these?

src/Mesh/Topologies.jl
lcw: I don't have a suggestion on the name but are people aware of the `meshwarp` argument to the constructor of `DiscontinuousSpectralElementGrid`?  That will change the element locations and the grid inside the element.  Here https://github.com/climate-machine/CLIMA/blob/61f69d451aa2fa3f7e905a232794ad74c5fcf7e4/examples/DGmethods/held_suarez_forcing.jl#L756-L795 is an example of where we used it in the past.

simonbyrne: Ah, interesting. I do hope to revamp the warp interface at some point, this is more of a stopgap until then.

simonbyrne:
bors try

simonbyrne:
bors try

src/Mesh/Topologies.jl
jkozdon: `gridrange`?

jkozdon: `stretchrange`? Or can we just go with `range` with some extra arguments?

simonbyrne: it may not return a range (if it has been stretched)

simonbyrne:
bors try

src/Mesh/Topologies.jl
jkozdon: Make sense.

Would the more Julian way be a new range type with an [irregular step](https://github.com/JuliaLang/julia/blob/c44644442949238da70670b547312ca1ae9a9c7d/base/traits.jl#L25-L53)? (Just wondering, I'm personally fine with the array...)

simonbyrne: Perhaps? But these aren't performance or space critical, so probably not worth the effort.

slifer50:
Hi Simon,

This implementation looks simpler than my implementation, it should do the
trick assuming all works as intended when with the stretching.

src/Mesh/Topologies.jl
slifer50: How about `gridspacing1d` for the name

src/Mesh/Topologies.jl
smarras79: @simonbyrne these functions simply move the nodes towards the top or bottom boundaries of the domain.
They are generally referred to as  "grid stretching functions" in the grid generation literature.

src/Mesh/Topologies.jl
smarras79: > it may not return a range (if it has been stretched)

is this because `range` in julia means something very specific? 
The stretching only moves the nodes. It does not change the topological structure of the grid.

smarras79: > I'm still not 100% sold on the name `grid1d`, since we used "Grid" to mean other things.
> 
> @lcw @jkozdon any suggestions?

Maybe `mesh_stretching` ?

fxgiraldo: what about vertical_stretching.

smarras79: `vertical` wouldn't work because these are general stretching functions that we also use horizontally.

fxgiraldo: Not sure I understand what the use case is for stretching horizontally. In the vertical it is clear.

fxgiraldo: I guess I think of the horizontal (2D) and vertical (1D) grid decomposition in this fashion so in the horizontal we have more flexibility (in theory) to construct specific grids (via some refinement process although this doesn't exist in the code at the moment). This is why I suggested the name vertical stretching since it is only a 1D operation.  Not tied to this name but is what, say, NRL calls it in Neptune for example.

smarras79: Horizontal stretching is the closest we have to grid refinement without a grid refinement algorithm. We often use it for problems such as the squall line where the cloud is really happening within a small central portion of the domain. 

Any naming is good, but the reasoning behind having a 1D function was to use the same function whenever needed rather than building two based on the direction.

src/Mesh/Topologies.jl
simonbyrne: Is there a decent reference I can use? I would like to document these.

src/Mesh/Topologies.jl
jkozdon: I would agree with @smarras79 that there is nothing inherently "vertical" about a grid stretching. Our grid generator allows you to stretch in any direction.

One question is should these stretchings really be done inside the element as well. Doing the stretching here you are only stretching the element boundaries, but not the degrees of freedom inside the element.

If you want the later functionality, then this should really be a meshwarping applied elsewhere.

src/Mesh/Topologies.jl
smarras79: "Handbook of Grid Generation"  J. F. Thompson,  B. K. Soni, N. P. Weatherill (Editors) RCR Press 1999

src/Mesh/Topologies.jl
smarras79: > One question is should these stretchings really be done inside the element as well. Doing the stretching here you are only stretching the element boundaries, but not the degrees of freedom inside the element.
> 
> If you want the later functionality, then this should really be a meshwarping applied elsewhere.

I don't think we will ever want to change the grid spacing within an element unless there is a very specific numerical requirement for that.

fxgiraldo: This is what @lcw was getting at with the ExponentialVerticalWarp function. This allows for the element (and the internal DOFs) to be stretched as required to maintain the theoretical accuracy.  We introduced this back in the summer for the Held-Suarez test case to better represent the boundary layer.

simonbyrne: I would eventually like to unify the stretching/warping into one place: perhaps instead of specifying a range in the topology constructors, it should simply be the number of elements in each dimension?

However that would be a much larger change, and I don't really want to spend too much time on it now.

src/Mesh/Topologies.jl
simonbyrne: Do you know where in the book? I wasn't able to find them.

smarras79: I don't have a copy of the book anymore, but if tyou can't find them there, you will certainly find them here:

`Computational Fluid Mechanics And Heat` Pletcher, Tennehill, Anderson`. They are juist standard exponential stretching function used in CFD regardless of the numerical method.

simonbyrne: ah, nevermind.

simonbyrne: I couldn't find a reference for the internal one. If you know of anything, let me know.

simonbyrne:
I'll merge this for now. We can revisit the warping/stretching later.

src/Mesh/Topologies.jl
smarras79: You don't need a reference for these functions. Anyone can come up with any arbitrary function with some specific properties to move the nodes where needed. I wouldn't worry about it at all.

smarras79: Thanks for merging it all together to the new code, btw!

PR #452, opened 2019-09-25
Remove the param structure of the ODE solver

jkozdon:
Removes the `params` from the ODE solver interface and merges previously stored data into the `DGmodel`.

The `params` was complicating the ODE interface when multiple balance laws are needed (e.g., IMEX and multirate). The interface I added should also simplify the sharing of `auxstate` and `diffstate` (which is really just scratch space) between different DGmodels

### DIFF
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -107,56 +107,12 @@ function (dg::DGModel)(dQdt, Q, t; increment=false)
   MPIStateArrays.finish_ghost_send!(Q)
 end
 
-"""
-    init_ode_param(dg::DGModel)
-
-Initialize the ODE parameter object, containing the auxiliary and diffusive states. The extra `args...` are passed through to `init_state!`.
-"""
-function init_ode_param(dg::DGModel)
+function init_ode_state(dg::DGModel, args...; commtag=888)
   bl = dg.balancelaw
   grid = dg.grid
-  auxstate = create_auxstate(bl, grid)
-  diffstate = create_diffstate(bl, grid)
-  return (aux=auxstate, diff=diffstate)
-end
 
+  state = create_state(bl, grid, commtag)
 
-
-
-"""
-    init_ode_state(dg::DGModel, param, args...)
-
-Initialize the ODE state array.
-"""
-function init_ode_state(dg::DGModel, commtag)
-  bl = dg.balancelaw
-  grid = dg.grid
-  topology = grid.topology
-  # FIXME: Remove after updating CUDA
-  h_vgeo = Array(grid.vgeo)
-  DFloat = eltype(h_vgeo)
-  Np = dofs_per_element(grid)
-  DA = arraytype(grid)
-
-  weights = view(h_vgeo, :, grid.Mid, :)
-  weights = reshape(weights, size(weights, 1), 1, size(weights, 2))
-
-  state = MPIStateArray{Tuple{Np, num_state(bl,DFloat)}, DFloat,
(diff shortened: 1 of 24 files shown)


jkozdon:
bors try

jkozdon:
bors try

src/DGmethods/DGmodel.jl
```diff
@@ -1,12 +1,20 @@
...
+end
+function DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux;
+                 auxstate=create_auxstate(balancelaw, grid),
+                 diffstate=create_diffstate(balancelaw, grid))
+  DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux, auxstate,
+          diffstate)
 end
 
-function (dg::DGModel)(dQdt, Q, param, t; increment=false)
+function (dg::DGModel)(dQdt, Q, t; increment=false)
```
simonbyrne: I'm happy to move the functionality, but it might make sense to keep an empty argument here (i.e. `(dg::DGModel)(dQdt, Q, _, t; increment=false)`) for the purposes of compatibility with DifferentialEquations.jl.

jkozdon: Is this what you were thinking: https://github.com/climate-machine/CLIMA/pull/452/commits/f5a9548b9da85da797f58cf376e58039b13ee90a

simonbyrne: That works. I guess the question is whether you would want to change the interface for the ODE solvers as well, but since that's probably not a pressing issue at the moment.

jkozdon: I think that e1fcee5 is what you suggest, but now I get an error
```
all-underscore identifier used as rvalue around /Users/jekozdon/codes/CLIMA/src/DGmethods/DGmodel.jl:18
```
if I use the underscores as opposed to catching the `p` explicitly. Not sure why. Spent a little time trying to track it down, but no luck.

src/DGmethods/DGmodel.jl
```diff
@@ -1,12 +1,20 @@
...
+end
+function DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux;
+                 auxstate=create_auxstate(balancelaw, grid),
+                 diffstate=create_diffstate(balancelaw, grid))
+  DGModel(balancelaw, grid, numfluxnondiff, numfluxdiff, gradnumflux, auxstate,
+          diffstate)
 end
 
-function (dg::DGModel)(dQdt, Q, param, t; increment=false)
+function (dg::DGModel)(dQdt, Q, _, t; increment=false)
```
simonbyrne: Looks like an upstream issue (https://github.com/JuliaLang/julia/issues/32727).

```suggestion
function (dg::DGModel)(dQdt, Q, _unused, t; increment=false) # can't use just _ due to #32727
```

jkozdon: How about that way if some parameters do come through we know something went wrong?

```suggestion
function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
```

jkozdon:
bors try

src/DGmethods/DGmodel.jl
simonbyrne: sounds good!

simonbyrne:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

PR #453, opened 2019-09-25
AtmosModel: move total specific enthalpy to top level

simonbyrne:
Reduces the assumptions on the moisture model, and also avoids the need for updating the mms test case for the time being.

Also changes the order of `diffusive!` calculations so that SmagorinskyLilly usese the most recent gradient. As a result, this affects the DYCOMS tolerance. cc @asraero @smarras79 

Supersedes #413.


simonbyrne:
bors try

charleskawczynski:
Looks good, my only suggestion would be to include integration tests, as @jkozdon suggested, since this is not an easy test to pass.

charleskawczynski:
Oh I guess [this](https://github.com/climate-machine/CLIMA/pull/453/files#diff-2b9e65635b38cc0c463d5206cc073050R48) bypasses that need since the total enthalpy is not needed for that case.

PR #456, opened 2019-09-26
Revise TurbConv framework

charleskawczynski:
Revision of the turbulence convection framework.


charleskawczynski:
Note that a TDMA is in this PR, and I would like to replace it by `Tridiagonal` and `inv` in `LinearAlgebra`, but the current implementation of the EDMF equations for BOMEX needs to be modified in order to do so-- so, I'd prefer we leave it in for now and refactor it away once we can remove the dependence on the current algorithm (which is in-place and has an offset index for the lower-diagonal).

charleskawczynski:
bors try

charleskawczynski:
@simonbyrne I'm seeing a [documentation error](https://dev.azure.com/spjbyrne/CLIMA/_build/results?buildId=1071&view=logs) that appears unrelated to this PR. Not sure what the problem is, but it seems to be coming from [`MPIStateArrays`](https://github.com/climate-machine/CLIMA/blob/master/src/Arrays/MPIStateArrays.jl#L394). Maybe due to differences in default and docs environment?

PR #461, opened 2019-10-01
use FT for floating point type parameters

simonbyrne:
We are inconsistent in the name of the type parameter to use for floating point values. Traditionally Julia code uses `T`, but that conflicts with the use of `T` for temperature. As a result we have had a mix of `T`, `DT`, `DFloat`, and `DF` used throughout the code.

This changes all the `DT`/`DFloat` to use `DF` to represent the type parameter of a floating point type. There are a few more `T` ones that also need to be changed, but I can't do those as easily via `sed`. Once we reach agreement, I can go through and change those as well.

Thoughts? Is `DF` a good symbol? Does this rule also need to apply to the Mesh/DGmethods/ODEsolvers modules where the risk of confusion is less likely?


charleskawczynski:
My two cents:

 - `DF` (Double Float, I believe) rings nicely, but seems to be a misnomer
 - `DT` (Data Type) seems nice, but also a bit generic
 - `WP` (Working Precision) also seems nice but, as @simonbyrne pointed out, `Float64`,`Float32` are data types, _not_ precision.

Just thought of this:
 - `PT` (Precision Type) ?

Thanks, @simonbyrne for unifying things to start with!

simonbyrne:
`FT` float type?

charleskawczynski:
> `FT` float type?

Even better

jkozdon:
🎈though I guess it would get old having to type `\:balloon:` all the time would get annoying.

blallen:
FT works for me! 

I think doing this in at least the DGmethods directory would be useful as well

glwagner:
The name of this PR should be "use DF for...", right?

PR #459, opened 2019-10-01
Add EDMF implementation

charleskawczynski:
Adds stand-alone implementation of EDMF equations for BOMEX case

There are many refactoring todo's, listed in #457, and there are also inconsistencies that need to be resolved with CLIMA's `PlanetParameters` and `MoistThermodynamics`.

Closes #108

PR #463, opened 2019-10-01
Try to recreate Windows failure

simonbyrne:
See #459.


kpamnany:
Old, and presumably dead.

PR #464, opened 2019-10-02
Update b.c. to reflect h_tot at top level (per #453)

asraero:
diffusive.moisture.h_tot -> diffusive.h_tot	
modified:   src/Atmos/Model/boundaryconditions.jl


asraero:
bors try

PR #468, opened 2019-10-03
Turn on more integration testing

jkozdon:


jkozdon:
bors try

jkozdon:
cc @lcw

jkozdon:
bors try

kpamnany:
This is great. Please let me know if any of these should be in the performance tests also, alongside Dycoms, for SlurmCI testing.

.slurmci/test_set.jl
```diff
@@ -29,10 +26,12 @@ cpu_gpu_tests = Set(((3, "examples/DGmethods_old/ex_001_periodic_advection.jl"),
                      (2, "test/DGmethods_old/sphere/advection_sphere_lsrk.jl"),
                      (2, "test/DGmethods_old/sphere/advection_sphere_ssp33.jl"),
                      (2, "test/DGmethods_old/sphere/advection_sphere_ssp34.jl"),
-                     (2, "test/LinearSolvers/poisson.jl"),
                      (4, "examples/DGmethods/ex_001_dycoms.jl"),
                      (1, "test/Ocean/shallow_water/GyreDriver.jl")
```
simonbyrne:
```suggestion
                     (1, "test/Ocean/shallow_water/GyreDriver.jl"),
```

test/DGmethods/integral_test_sphere.jl
```diff
@@ -130,7 +129,7 @@ let
   expected_result = [6.228615762850257e-7
                      9.671308320438864e-9
                      1.5102832678375277e-10
-                     2.359860999112363e-12]
+                     2.3598612210570662e-12]
```
simonbyrne: Slurm doesn't like this change: 
https://gist.github.com/climabot/f8b1957d388a35030d5bcc272c22e0dd#file-out_7303963-L43

simonbyrne: @jkozdon any reason for this change?

simonbyrne:
```suggestion
                     2.359860999112363e-12]
```

simonbyrne:
bors try

test/DGmethods/integral_test_sphere.jl
simonbyrne: Ah, so the issue is that the CPU and GPU give different answers. Is this simply an order-of-operations issue, or something more serious. Can we just loosen the tolerance?

kpamnany: Parallel vs serial reduction perhaps? If so, typical solution is to loosen the tolerance.

PR #469, opened 2019-10-04
nonlinear AtmosModel and isentropic vortex IMEX test

mwarusz:
This adds `AtmosAcousticNonlinearModel` and tests the model splitting using IMEX isentropic vortex test.

One thing I was not sure about is how to initialize the model state when doing the nonlinear-linear splitting. Right now the user has to create a `DGmodel` based on the full model to initialize the state and the aux state. Alternatively, the nonlinear part could handle initialization. Any thoughts on this or other parts of this PR @simonbyrne


mwarusz:
bors try

simonbyrne:
Thanks!

> One thing I was not sure about is how to initialize the model state when doing the nonlinear-linear splitting. Right now the user has to create a `DGmodel` based on the full model to initialize the state and the aux state. Alternatively, the nonlinear part could handle initialization. Any thoughts on this or other parts of this PR @simonbyrne

Yeah, I'm not sure of the best way to do this yet, will need to think a bit more about it.

PR #471, opened 2019-10-04
speed up test suite

simonbyrne:


simonbyrne:
The culprit was the density current test, which for some reason was taking 50 minutes on the Linux machine. Since this is run as part of the Slurm batch, i'm removing it for now.

PR #478, opened 2019-10-08
Multirate Infinitesimal Step methods

simonbyrne:


mwarusz:
Hi @fxgiraldo I just added to this PR some docs with a reference. This is very much a work in progress.

simonbyrne:
@fxgiraldo in section 2.c they transform the system from a nonautonomous system to an autonomous one by adjusting the time. Is there a standard name for this?

fxgiraldo:
@simonbyrne ok thanks.  Not sure but it is the approach used by many in the literature to do this. But this map requires carrying another ODE though so is best to avoid if at all possible.

simonbyrne:
What do you mean about carrying another ODE? How does this differ vs the multirate RK? Are there alternatives we could use?

fxgiraldo:
To go from non-autonomous to autonomous the usual trick is to include an additional ODE dtau/dt=1. I assume this is what you were doing?

mwarusz:
What we are doing is based on that. However, for the methods that we are implementing it is possible to just calculate the resulting coefficients. So there is no need to explicitly integrate additional time variable.

fxgiraldo:
I agree. we just need to compute the correct time value which can then be put directly into the time-integrator.  I believe the LSRK are already setup this way.

src/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -0,0 +1,276 @@
...
+    blocks = div(length(realview(Q)) + threads - 1, threads)
+    @launch(device(Q), threads=threads, blocks=blocks,
+            update!(realview(Q), realview(offset), Val(i), realview(yn), map(realview, ΔYnj[1:i-2]), map(realview, fYnj[1:i-1]), 
+            α[i,:], β[i,:], γ[i,:], d[i], dt))
+
+    fastrhs!.a = time + c̃[i]*dt
+    fastrhs!.b = (c[i] - c̃[i]) / d[i]
+
+    τ = zero(FT)
+    dτ = d[i] * dt / mis.nsubsteps
```
jkozdon: Should number of substeps be hard coded, or should a fast dt be specified and the number of substeps be chosen to match that fast dt?

src/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -0,0 +1,276 @@
...
+    c̃ = α*c
+    
+    new{T, RT, AT, typeof(fastsolver), Nstages, Nstages-1, Nstages-2, Nstages ^ 2}(dt, t0, yn, ΔYnj, fYnj, offset,
+                                         slowrhs!, tsfastrhs!, fastsolver, nsubsteps,
+                                         α, β, γ, d, c, c̃)
+  end
+end
+
+
+# TODO almost identical functions seem to be defined for every ode solver,
```
jkozdon: I'm fine with that. Had noticed the same thing once I started adding the initial multirate stuff.

src/ODESolvers/MultirateInfinitesimalStepMethod.jl
simonbyrne: My understanding is that we will typically specify everything in terms of ratio of steps, and doing it this way avoids the need to worry about floating point rounding, but we could probably support both?

jkozdon: My thought here is that the number of substeps will be dictated by the largest value in the `d` array. If they are not all about the same size, you could end up substepping the stages with smaller `d_i` values using many more steps that needed for stability/accuracy.

For the LSRK, the ratio was about `~75` for the 14 stage LSRK so we'd do substantially more work for the small partial steps than we needed to.

A quick check suggested the situation is not as bad for the methods you all implemented here.

You can see how I handled it [here](https://github.com/climate-machine/CLIMA/blob/72c35b7687601ad0669861ef36b199223851e3f8/src/ODESolvers/MultirateRungeKuttaMethod.jl#L153-L154)

jkozdon: Also, realized that the ratio of steps is the outer to inner full step size, not the step size needed to substep the stage values.

PR #479, opened 2019-10-08
implement Euler timestepping for debugging

blallen:
Using an Euler timestepper allows us to easily compare the results of the code against analytical / by-hand numbers for small numbers of time steps. Should never be used in a production run, but can be very useful for seeing exactly what the DGmethod kernels are computing.


blallen:
bors try

blallen:
bors try

PR #482, opened 2019-10-10
add RemainderModel for general purpose splitting

simonbyrne:


simonbyrne:
TODO: need to benchmark.

mwarusz:
I ran some benchmarks on the rising thermal bubble and it seems that `RemainderModel` is as fast as `AtmosAcousticNonlinearModel` so the latter can be removed :thumbsup:

src/Utilities/VariableTemplates/VariableTemplates.jl
```diff
@@ -143,6 +144,7 @@ Grad{S}(array) where {S} = Grad{S,typeof(array),0}(array)
 
 Base.eltype(g::Grad) = eltype(getfield(g,:array))
 Base.propertynames(::Grad{S}) where {S} = fieldnames(S)
+Base.similar(g::Grad{S,A,offset}) where {S,A,offset} = Grad{S,A,offset}(similar(getfield(g,:array)))
```
charleskawczynski: Is this tested in the tests? Maybe good to add.

simonbyrne: Good point.

PR #483, opened 2019-10-11
allow array initialization on CPU

simonbyrne:


src/DGmethods/DGmodel.jl
```diff
@@ -121,27 +121,39 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
   end
 end
 
-function init_ode_state(dg::DGModel, args...; commtag=888)
+function init_ode_state(dg::DGModel, args...; device=typeof(state.data) <: Array ? CPU() : CUDA(), commtag=888)
+  array_device = typeof(state.data) <: Array ? CPU() : CUDA()
```
smarras79: state.data is not passed to init_ode_state because that's what is being returned at this point in the driver.
How do you recommend modifying it?

PR #485, opened 2019-10-11
Held-Suarez example

simonbyrne:
Visually checked by @mwarusz. We still need some diagnostics, then we can start exploring modifications (imex, turbulence models, etc).

### DIFF
--- a/examples/Atmos/heldsuarez.jl
+++ b/examples/Atmos/heldsuarez.jl
@@ -26,14 +26,9 @@ using MPI, Logging, StaticArrays, LinearAlgebra, Printf, Dates, Test
   using CUDAnative
   using CuArrays
   CuArrays.allowscalar(false)
-  const ArrayTypes = (CuArray,)
+  const ArrayType = CuArray
 else
-  const ArrayTypes = (Array,)
-end
-
-if !@isdefined integration_testing
-  const integration_testing =
-    parse(Bool, lowercase(get(ENV,"JULIA_CLIMA_INTEGRATION_TESTING","false")))
+  const ArrayType = Array
 end
 
 const output_vtk = true
@@ -54,17 +49,15 @@ function main()
   polynomialorder = 5
   numelem_horz = 6
   numelem_vert = 8
-  days = 86400
+  days = 60*60*24
   timeend = 60 # 400days
   outputtime = 2days
   
-  #@testset "$(@__FILE__)" begin
-    for ArrayType in ArrayTypes, FT in (Float64,)
+  for FT in (Float64,)
 
-      run(mpicomm, polynomialorder, numelem_horz, numelem_vert,
-          timeend, outputtime, ArrayType, FT)
-    end
-  #end
+    run(mpicomm, polynomialorder, numelem_horz, numelem_vert,
+        timeend, outputtime, ArrayType, FT)
+  end
 end
 
 function run(mpicomm, polynomialorder, numelem_horz, numelem_vert,
@@ -211,15 +204,16 @@ function held_suarez_forcing!(source, state, aux, t::Real)
   Δθ_z = FT(10)
   T_equator = FT(315)
   T_min = FT(200)
-  scale_height = FT(7000) #from Smolarkiewicz JAS 2001 paper
   σ_b = FT(7 / 10)
   r = norm(coord, 2)
   @inbounds λ = atan(coord[2], coord[1])
   @inbounds φ = asin(coord[3] / r)
   h = r - FT(planet_radius)
+  scale_height = FT(7000) #from Smolarkiewicz JAS 2001 paper
   σ = exp(-h / scale_height)
-  #p = air_pressure(T, ρ)
-#  σ = p/p0
+  # TODO: use
+  #  p = air_pressure(T, ρ)
+  #  σ = p/p0
   exner_p = σ ^ (R_d / cp_d)
   Δσ = (σ - σ_b) / (1 - σ_b)
   height_factor = max(0, Δσ)
--- a/examples/Atmos/heldsuarez.jl
+++ b/examples/Atmos/heldsuarez.jl
@@ -10,7 +10,7 @@ using CLIMA.LowStorageRungeKuttaMethod: LSRK144NiegemannDiehlBusch
 using CLIMA.VTK: writevtk, writepvtu
 using CLIMA.GenericCallbacks: EveryXWallTimeSeconds, EveryXSimulationSteps
 using CLIMA.MPIStateArrays: euclidean_distance
-using CLIMA.PlanetParameters: R_d, grav, MSLP, planet_radius, cp_d, cv_d
+using CLIMA.PlanetParameters: R_d, grav, MSLP, planet_radius, cp_d, cv_d, day
 using CLIMA.MoistThermodynamics: air_density, total_energy, soundspeed_air, internal_energy, air_temperature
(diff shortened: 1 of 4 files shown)


examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+                     setup)
+
+  dg = DGModel(model, grid, Rusanov(),
+               CentralNumericalFluxDiffusive(), CentralGradPenalty())
+
+  # determine the time step
+  element_size = (setup.domain_height / numelem_vert)
+  acoustic_speed = soundspeed_air(FT(315))
+  lucas_magic_factor = 14
+  dt = lucas_magic_factor * element_size / acoustic_speed / polynomialorder ^ 2
```
tapios: Why is this `polynomialorder ^ 2` rather than `polynomialorder` here?

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+  ΔT_y = FT(60)
+  Δθ_z = FT(10)
+  T_equator = FT(315)
+  T_min = FT(200)
+  σ_b = FT(7 / 10)
+  r = norm(coord, 2)
+  @inbounds λ = atan(coord[2], coord[1])
+  @inbounds φ = asin(coord[3] / r)
+  h = r - FT(planet_radius)
+  scale_height = FT(7000) #from Smolarkiewicz JAS 2001 paper
```
tapios: Better to use `scale_height` from setup (more consistent), if you do this to define sigma.

examples/Atmos/heldsuarez.jl
simonbyrne: I think it’s due to the fact that the distance between the closest nodes scales this way. Is that right @lcw?

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+                  "ERROR" => Logging.Error,
+                  "INFO"  => Logging.Info)[ll]
+
+  logger_stream = MPI.Comm_rank(mpicomm) == 0 ? stderr : devnull
+  global_logger(ConsoleLogger(logger_stream, loglevel))
+
+  polynomialorder = 5
+  numelem_horz = 6
+  numelem_vert = 8
+  days = 60*60*24
```
tapios: You might take this from `PlanetParameters` too.

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+  ρu = state.ρu
+  ρe = state.ρe
+  coord = aux.coord
+  Φ = aux.orientation.Φ
+  e = ρe / ρ
+  u = ρu / ρ
+  e_int = e - u' * u / 2 - Φ
+  T = air_temperature(e_int)
+  # Held-Suarez constants
+  k_a = FT(1 / 40 / 86400)
```
tapios: Clearer: `1/ (40 * day)` etc.

examples/Atmos/heldsuarez.jl
mwarusz: I think we want to just use the model pressure directly here (see the commented out part below this line). However, this is the exact setup that ran stably for 400 days so I wanted to start with that.

tapios: Ok. And, yes, the commented part is what you'd want here in the end.

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+  k_s = FT(1 / 4 / 86400)
+  ΔT_y = FT(60)
+  Δθ_z = FT(10)
+  T_equator = FT(315)
+  T_min = FT(200)
+  σ_b = FT(7 / 10)
+  r = norm(coord, 2)
+  @inbounds λ = atan(coord[2], coord[1])
+  @inbounds φ = asin(coord[3] / r)
+  h = r - FT(planet_radius)
```
tapios: This way of determining altitude is fine for now. But it'd be good to make this more general at some point, so that we can have a non-spherical (geoid) planet, and then carry that more general altitude coordinate in the same way as the geopotential.

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,253 @@
...
+  scale_height = FT(7000) #from Smolarkiewicz JAS 2001 paper
+  σ = exp(-h / scale_height)
+  # TODO: use
+  #  p = air_pressure(T, ρ)
+  #  σ = p/p0
+  exner_p = σ ^ (R_d / cp_d)
```
tapios: Maybe better `exner_σ` because this isn't pressure, or simply writing it out in the formula where it appears, without an extra variable.

lcw:
Yep. This is a rough rule of thumb we use to determine stable time step size.

examples/Atmos/heldsuarez.jl
lcw: That is right, it is related to the fact that we need nodes clustered toward the edges of the elements.   The radius of the spectrum of the spatial discretization operator scales roughly like `O(polynomial^2)`.  We could think about implementing  covolume mesh based filtering https://doi.org/10.1137/060672601 to avoid this.  Since we are on structured grids it may not be that bad.

examples/Atmos/heldsuarez.jl
simonbyrne:
```suggestion
  k_a = FT(1 / (40 * day))
```

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,252 @@
...
```
tapios:
```suggestion
  
```

examples/Atmos/heldsuarez.jl
```diff
@@ -0,0 +1,252 @@
...
+  Δσ = (σ - σ_b) / (1 - σ_b)
+  height_factor = max(0, Δσ)
+  T_equil = (T_equator - ΔT_y * sin(φ) ^ 2 - Δθ_z * log(σ) * cos(φ) ^ 2 ) * exner_p
```
tapios:
```suggestion
  T_equil = (T_equator - ΔT_y * sin(φ) ^ 2 - Δθ_z * log(σ) * cos(φ) ^ 2 ) * σ^(R_d / cp_d)
```

examples/Atmos/heldsuarez.jl
tapios: Thanks for the explanation @lcw! 
* It would be good if we provide a function somewhere that gives an effective \Delta x to be used in the Courant number, so it is more transparent how the CFL bounds arise here.
* The covolume meshing looks interesting. It's a question whether it is worth it for us. It can give a factor `polynomialorder` increase in allowable timestep, but at the expense of the mesh projections (and constraints on the timestepping scheme and additional tweaks needed to restore conservation). For us, the order probably will not be larger than 4 or 5. The paper suggests that the damped projections (Eq. 6.1) roughly double the cost per timestep, at the expense of an order reduction. So it may give us a modest speedup. Is that worth exploiting for us?

lcw: I can add a function for calculating an effective \Delta x.  I would say covolume mesh based filtering is not worth perusing right now.  Just wanted to let you know it exists.

tapios: Thanks and sounds good!

PR #492, opened 2019-10-17
add TVDMIS methods

simonbyrne:
TVDMISB works, however beta values for TVDMISA appear to be incorrect. Awaiting input from the author.

### DIFF
--- a/src/ODESolvers/MultirateInfinitesimalStepMethod.jl
+++ b/src/ODESolvers/MultirateInfinitesimalStepMethod.jl
@@ -8,7 +8,7 @@ using GPUifyLoops
 include("MultirateInfinitesimalStepMethod_kernels.jl")
 
 
-export MIS2, MIS3C, MIS4, MIS4a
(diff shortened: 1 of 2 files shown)


simonbyrne:
TVDMISA values have been updated.

bors r+

jkozdon:
> TVDMISB works, however beta values for TVDMISA appear to be incorrect

Was these an error in the paper? If so, could you add a comment for what needs to be corrected? (As someone who has had to track down inconsistencies in coefficients before, I think this would be useful to others to have noted somewhere)

simonbyrne:
> Was these an error in the paper? If so, could you add a comment for what needs to be corrected? (As someone who has had to track down inconsistencies in coefficients before, I think this would be useful to others to have noted somewhere)

No: the error was in the initial code sample provided by the author. The paper didn't have the full coefficients (I believe the rest of them should have been computable from the details given, but I didn't quite understand how they were arrived at).

PR #489, opened 2019-10-14
Remove SlurmCI scripts and migrate test specification to TOML file

kpamnany:
Merge only after https://github.com/climate-machine/slurmci/pull/1 is merged and tested.

PR #495, opened 2019-10-18
Add RB test to slurmci-test.toml (re: #489)

asraero:
@kpamnany 
(Toward removing SLURMCI scripts and moving test specification to TOML file)


asraero:
Addressed by @simonbyrne already.

PR #496, opened 2019-10-20
Fix `ρ_cloud_ice` and update `gas_constant` parameters

ali-ramadhan:
Two small fixes:

1. `ρ_cloud_ice` was missing a decimal place.

2. The SI units were redefined in May so that the Avogadro constant and the Boltzmann constant have exact values, giving the universal `gas_constant` the exact value 8.31446261815324 J/mol/K.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -26,9 +26,9 @@ version = "1.2.1"
 
 [[BandedMatrices]]
 deps = ["FillArrays", "LazyArrays", "LinearAlgebra", "MatrixFactorizations", "Random", "SparseArrays"]
-git-tree-sha1 = "1fd0676c958fe070c0edbffe3197edfdfcced523"
+git-tree-sha1 = "5f200899b4f217545f1b15fdde2f326231e5392c"
 uuid = "aae01518-5342-5314-be14-df237901396f"
-version = "0.12.0"
+version = "0.12.4"
 
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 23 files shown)


tapios:
Thank you for catching the cloud ice problem (those hailstones would have hurt...), and I love the precision on the gas constant!

simonbyrne:
@charleskawczynski this seems to have broken some tests in MoistThermo: can you look into it?

charleskawczynski:
> @charleskawczynski this seems to have broken some tests in MoistThermo: can you look into it?

Yep, I think this is just from the increased precision. I can update the values in the tests.

charleskawczynski:
Alright, @simonbyrne , fixed MoistThermo/SurfaceFluxes. The remaining problem looks strange and related to `MbedTLS`

ali-ramadhan:
Ah sorry I should have realized that changing `gas_constant` would more than likely break some tests.

tapios:
@charleskawczynski @ali-ramadhan Could you guys resolve whatever issues there are and merge this?

charleskawczynski:
As silly as this may sound, this may be good to split into 2 PRs. The `ρ_cloud_ice` is a simple bug fix, but the `gas_constant` requires updating many coefficients in all of the integration tests (which takes some time to run/check).

I'll update this now to only fix `ρ_cloud_ice`, and we can update `gas_constant` in a separate PR.

tapios:
Sounds good. Let’s look over these tests together when you do this. We shouldn’t have tests that fail when physical constants change.

On Nov 6, 2019, at 9:02 AM, Charles Kawczynski <notifications@github.com> wrote:

﻿

As silly as this may sound, this may be good to split into 2 PRs. The ρ_cloud_ice is a simple bug fix, but the gas_constant requires updating many coefficients in all of the integration tests (which takes some time to run/check).

I'll update this now to only fix ρ_cloud_ice, and we can update gas_constant in a separate PR.

PR #499, opened 2019-10-23
Trying 1D-IMEX with GMRES in Held-Suarez

jkozdon:
With 10X bigger time step than the currently set explicit time step, we see `~400` GMRES iterations

With 100X bigger time step, GMRES did not converge (with 30 iterations before restart)


tapios:
@jkozdon Can we close this?

jkozdon:
Done.

PR #501, opened 2019-10-25
Adding upwind flux interface

jkozdon:
- Adds infrastructure for upwind fluxes (primarily for use in linear models)
- Adds Upwind flux for `AtmosAcousticLinearModel`
- Tests upwind flux for `isentropicvortex-imex`

### DIFF
--- a/src/DGmethods/NumericalFluxes.jl
+++ b/src/DGmethods/NumericalFluxes.jl
@@ -1,7 +1,7 @@
 module NumericalFluxes
 
 export Rusanov, CentralGradPenalty, CentralNumericalFluxDiffusive,
-       CentralNumericalFluxNonDiffusive
+       CentralNumericalFluxNonDiffusive, Upwind
 
 using StaticArrays
 using GPUifyLoops: @unroll
@@ -110,6 +110,24 @@ function numerical_boundary_flux_nondiffusive!(nf::NumericalFluxNonDiffusive,
   numerical_flux_nondiffusive!(nf, bl, F, nM, QM, auxM, QP, auxP, t)
 end
 
+"""
+    Upwind <: NumericalFluxNonDiffusive
+
+The Upwind numerical flux.
+
+# Usage
+
+    Upwind()
+
+!!! note
+
+    A balance law must provide a concrete implementation of
+    `numerical_flux_nondiffusive!(::Upwind, bl::BalanceLaw, F, nM,
+                                  QM, auxM, QP, auxP, t)`
+    for this flux to be used
+
+"""
+struct Upwind <: NumericalFluxNonDiffusive end
 
 
 """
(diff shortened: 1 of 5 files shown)


jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

src/DGmethods/NumericalFluxes.jl
```diff
@@ -110,6 +110,24 @@ function numerical_boundary_flux_nondiffusive!(nf::NumericalFluxNonDiffusive,
...
+
+The Upwind numerical flux.
+
+# Usage
+
+    Upwind()
+
+!!! note
+
+    A balance law must provide a concrete implementation of
```
blallen: Is there a reason the implementation of upwind flux is balance-law specific? For a given variable, the implementation should be straight forward. Is the constraint on specifying which variables get upwind fluxes?

jkozdon: In general, upwinding requires solving an eigenvalue problem. You could in theory for linear systems, we could specify the flux as `F = A * q` for some matrix `A` and then use an eigenvalue solver to diagonalize etc, but since this would need to be done at every degree of freedom it would likely not be efficient.

For nonlinear problems and problems with discontinuous coefficients this is even more challenging.

jkozdon: Note also that there isn't just one variable that gets upwinded, but its a combination of the variables because it comes from the left and right eigenvectors, and looking at the signs of the eigenvalues.

jkozdon:
This seems to have trouble outside isentropic vortex, some tests with (perhaps) rising thermal bubble should be added before this is merged.

jkozdon:
bors try

jkozdon:
One of the `Rusanov` tests fails when the `Remainder` model is used in the `rising_bubble-model-imex.jl` test. The newly added `Upwind` tests do not, but I did need to add the `Remainder` model to run upwind.

Also, I don't know how to test if this is really "correct" since all the models seem to blow up when run longer in time.

jkozdon:
bors try

simonbyrne:
bors try

jkozdon:
This should be ready to go.

I ran some longer runs with Upwind and Rusanov using Float64 with the rising thermal bubble, and the two perform similarly in the eyeball norm. (More rigorous testing is done as well with isentropic vortex.)

PR #505 is relevant here, but I don't think that this should effect merging the upwind in (since it's not an upwind issue but a Float32 issue).
 
cc: @lcw

simonbyrne:
bors try

simonbyrne:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
This is pretty stale, so I am closing it.

If we ever try to implement a rotated version of the banded solver (e.g., to use single column on the sphere, we will likely need something like this)

PR #504, opened 2019-10-28
Calculation of metric terms for calculating surface integrals

slifer50:
This uses the existing way of calculating the metric terms in 2D to generate and store the metric terms necessary for accurately computing surface averages in vgeo. This has been named MH in Grids.jl accordingly some additions were made in Metrics.jl


src/Mesh/Grids.jl
```diff
@@ -278,14 +279,19 @@ function computegeometry(topology::AbstractTopology{dim}, D, ξ, ω, meshwarp,
...
-                           ξ1x3, ξ2x3, ξ3x3, sJ, n1, n2, n3, D)
+                           ξ1x3, ξ2x3, ξ3x3, sJ, n1, n2, n3, D, JH)
   end
 
   M = kron(1, ntuple(j->ω, dim)...)
   MJ .= M .* J
   MJI .= 1 ./ MJ
   vMJI .= MJI[vmapM]
 
+  MH = kron(1, ntuple(j->ω, dim)...)
```
smarras79: This quantity is a surface quantity that lives in the element volume. Is this why you use `dim` and not `dim-1` in the Kronecker function? @slifer50 @jkozdon

src/Mesh/Grids.jl
```diff
@@ -30,17 +30,17 @@ Returns the points on the reference element.
...
-const _nvgeo = 15
-const _ξ1x1, _ξ2x1, _ξ3x1, _ξ1x2, _ξ2x2, _ξ3x2, _ξ1x3, _ξ2x3, _ξ3x3, _M, _MI,
+const _nvgeo = 17
+const _ξ1x1, _ξ2x1, _ξ3x1, _ξ1x2, _ξ2x2, _ξ3x2, _ξ1x3, _ξ2x3, _ξ3x3, _M, _MI, _MH, _MHI,
        _x1, _x2, _x3, _JcV = 1:_nvgeo
 const vgeoid = (ξ1x1id = _ξ1x1, ξ2x1id = _ξ2x1, ξ3x1id = _ξ3x1,
                 ξ1x2id = _ξ1x2, ξ2x2id = _ξ2x2, ξ3x2id = _ξ3x2,
                 ξ1x3id = _ξ1x3, ξ2x3id = _ξ2x3, ξ3x3id = _ξ3x3,
-                Mid  = _M , MIid = _MI,
+                Mid  = _M , MIid = _MI, MHid = _MH, MHIid = _MHI,
```
jkozdon: Do you need the inverse, or just the mass matrix?

jkozdon:
What about when `dim = 2`? In this case I assume you want a line integral, so do we need this infrastructure to work in this case?

src/Mesh/Grids.jl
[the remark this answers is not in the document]
slifer50: I think it's needed in order to calculate the average. Let me know if I'm mistaken.

src/Mesh/Metrics.jl
```diff
@@ -291,6 +292,7 @@ function computemetric!(x1, x2, x3, J, ξ1x1, ξ2x1, ξ3x1, ξ1x2, ξ2x2, ξ3x2,
         x3ξ2 += D[j, n] * x3[i, n, k, e]
         x3ξ3 += D[k, n] * x3[i, j, n, e]
       end
+      JH[i, j, k, e] = x1ξ1 * x2ξ2 - x2ξ1 * x1ξ2
```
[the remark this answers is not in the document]
slifer50: I see, so we would need the face normals aswell?

src/Mesh/Grids.jl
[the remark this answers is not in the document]
jkozdon: I don't think so. You want to divide by the slab surface area, which would be computed with `MH` (compute the integral, then divide).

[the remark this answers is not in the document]
slifer50: You're absolutely right. Would there be any possible use for the inverse, any reason to keep it?

src/Mesh/Metrics.jl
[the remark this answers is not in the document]
slifer50: I'm not sure I follow, the surface metrics are naturally calculated completely from the normals. What in the surface metrics should I be trying to imitate in order to be able to capture the warping, I was under the impression the warping was taken into account here : https://github.com/climate-machine/CLIMA/blob/3e72b701ae6192309ef51f2fcc65ecde11663631/src/Mesh/Grids.jl#L271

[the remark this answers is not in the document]
jkozdon: That gives you the points, but to compute integrals you need relationships between the derivatives. To do this you need something like what's done for the [surface Jacobians](https://github.com/climate-machine/CLIMA/blob/c10b76acadc62e89fed459420dbd1bcd5011a9fc/src/Mesh/Metrics.jl#L395-L397)

src/Mesh/Grids.jl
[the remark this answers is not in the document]
jkozdon: I cannot think of anything, unless you're solving PDES on the surfaces or something.

src/Mesh/Metrics.jl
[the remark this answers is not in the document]
slifer50: I'm confused, do you mean having something along the lines of hypot(x1ξ1 * (x2ξ2 * x3ξ3 - x3ξ2 * x2ξ3) ,x2ξ1 * (x3ξ2 * x1ξ3 - x1ξ2 * x3ξ3), x3ξ1 * (x1ξ2 * x2ξ3 - x2ξ2 * x1ξ3))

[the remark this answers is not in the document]
jkozdon: I don't think that's quite right. I think you want something along the lines of `hypot(J * ξ3x1, J * ξ3x2, J * ξ3x3)' but I wouldn't know for sure unless I worked it out on paper. This is the surface Jacobian calculation I highlight on faces 5 & 6.

If you add a reasonable tests and it passes, then this would show that what you have is correct.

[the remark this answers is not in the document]
slifer50: Oh alright I see what you mean now.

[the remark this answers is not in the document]
jkozdon: `JcV` is for a line integral. For a surface integral you need to consider a parameterized surface integral; see for instance this [wikipedia article](https://en.wikipedia.org/wiki/Surface_integral#Surface_integrals_of_scalar_fields).

Using the notation from the article, you need to compute:


Considering the cross product term gives

where the last equality follows from the standard metric identities; see for instance [Kopriva (2006) equation (5)](https://doi.org/10.1007/s10915-005-9070-8).

Thus we have that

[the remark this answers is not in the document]
slifer50: I went with the following done on Grids.jl. Is there any reason this should not work? before I go ahead try to set up a minimum working case.

MH = kron(1, ntuple(j->ω, dim-1)...)
map!(JH, J, ξ1x3, ξ2x3, ξ3x3
        ) do J, ξ1x3, ξ2x3, ξ3x3
      hypot(J*ξ1x3 ,J*ξ2x3, J*ξ3x3)
    end
    MHJH .= JH .* MH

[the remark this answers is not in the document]
slifer50: @jkozdon the latest push has a test case based on warp fun. The integrals for it should always be =0.5 but it seems the result is always a bit off, I'd appreciate your input.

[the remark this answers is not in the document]
slifer50: Hi @jkozdon I think I may be off because I'm still using "MH = kron(1, ntuple(j->ω, dim)...)" and as such might have the wrong weights, however "MH = kron(1, ntuple(j->ω, dim-1)...)" doesn't work because the array sizes then don't seem to match is this what the syntax should be?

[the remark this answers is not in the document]
jkozdon: I think you need something like: `MH = kron(ones(FT, Nq), ntuple(j->ω, dim-1)...)` since you want to replicate the 2D reference matrix `Nq` times (one for each 2D slab of degrees of freedom).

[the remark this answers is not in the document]
slifer50: Alright I'll try that, thanks

slifer50:
@jkozdon If you get a chance please have a look at the test. It looks normal to me but I'd like a second opinion. If It looks good to you we might want to go ahead and get this reviewed and merged

examples/DGmethods/ex_001_dycoms.jl
```diff
@@ -16,12 +16,13 @@ using LinearAlgebra
 using StaticArrays
 using Logging, Printf, Dates
 using CLIMA.VTK
-
+using DelimitedFiles
```
jkozdon: Do we need all these changes to dycoms in this PR?

slifer50: Oh absolutely not, I'll have that file switched with a clean copy

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,91 @@
...
+  for k in 1:Nqk
+  for j in 1:Nq
+    for i in 1:Nq
+      ijk = i + Nq * ((j-1)+ Nq * (k-1))
+      S[k] += localvgeo[ijk,grid.x1id,e] * localvgeo[ijk,grid.MHid,e]
+      S1[k] += localvgeo[ijk,grid.MHid,e]
+    end
+  end
+  end
+  end
```
jkozdon: This should all go in a GPU kernel, similar to the vertical integral ([1] and [2]) so that it's useable on the GPU and in general (right now this is CPU only code).

[1]: https://github.com/climate-machine/CLIMA/blob/09f2148fee9056bf4acc319badbb3469eae95413/src/DGmethods/DGmodel_kernels.jl#L1012
[2]: https://github.com/climate-machine/CLIMA/blob/09f2148fee9056bf4acc319badbb3469eae95413/src/DGmethods/DGmodel.jl#L163

examples/DGmethods/ex_001_dycoms.jl
slifer50: The new push should be clean

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,91 @@
...
+  for k in 1:Nqk
+  println(0.5-S[k]/S1[k])
+  end
```
jkozdon: You need some sort of concrete test that tells us whether the code works. Only way to really test this is the run same problem at several resolutions and compute a convergence rate. See the vertical integration test for how it was done there [1]

[1]: https://github.com/climate-machine/CLIMA/blob/09f2148fee9056bf4acc319badbb3469eae95413/test/DGmethods/integral_test_sphere.jl#L137-L162

test/DGmethods/Horizontal_integral_test.jl
jkozdon: You will also need an MPI reduce if you want this to work on more than 1 MPI rank. You will have an integral for each MPI rank at each level, then need to reduce the vector across all the ranks to get the actual integral.

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,91 @@
...
+warpfun = (ξ1, ξ2, ξ3) -> begin
+  x1 = ξ1 + (ξ1 - 1/2) * cos(2 * π * ξ2 * ξ3) / 4
+  x2 = ξ2 + exp(sin(2π * (ξ1 * ξ2 + ξ3)))/20
+  x3 = ξ3 + ξ1 / 4 + ξ2^2 / 2 + sin(ξ1 * ξ2 * ξ3)
+  return (x1, x2, x3)
+end
+for N in 3:6
+for Ne in 1:5
+println(N," ",Ne)
+brickrange = (range(FT(0); length=Ne+1, stop=1),
```
jkozdon: Does this need to work in 2D as well, or just 3D. No code has been added for 2D (which would be lines)

test/DGmethods/Horizontal_integral_test.jl
jkozdon: You will also need to add this to the test set (both slurm and/or the DGmehods runtests.jl file).

test/DGmethods/Horizontal_integral_test.jl
slifer50: I see, I'll try to address this as well as your other comments as soon as I can. Thank you for the feedback and input.

jkozdon:
You can probably do this whole thing without writing a kernel. Something along the lines of
```julia
if dim == 2
   mapreduce(identity, +, MH .* Q, dims=(1,2,5))
elseif dim == 3
   mapreduce(identity, +, MH .* Q, dims=(1,5))
end
```
Something like this will work on a single mpirank. with multiple ranks you need an `MPI.reduce` as well.

This assumes that MH is its own array, to avoid the allocation you'd want to use `LazyArrays`

You would need to write a kernel probably if you want to apply an integrand kernel (and not just integrate the state)

slifer50:
Hi Jeremy,

This was my initial intention in terms of what I planned to do for
diagnostics.
However, if there is any use for having a kernel it would certainly be a
good learning experience for me to learn how to write one appropriately.

jkozdon:
I would suggest doing what I suggested above, but writing a kernel similar for the [update aux](https://github.com/climate-machine/CLIMA/blob/09f2148fee9056bf4acc319badbb3469eae95413/src/DGmethods/DGmodel_kernels.jl#L975-L1009) which computes a custom integrand. Here you would launch 1 thread per DOF (similar to update aux), and store the result to a secondary array. You would also need to add a horizontal integrand/kernel functionality similar to the vertical one.

Doing the reduction with a kernel is possible too, but this would probably be a bit complex for a first time kernel.

slifer50:
@jkozdon Could you check the test? it looks to me like this is essentially running as intended.
I think it would be best to have a kernel for integration in a separate PR, we can just merge this so as to have the metrics for slabs available on the master.

jkozdon:
There is no call to `@test` to make sure that the result is correct, it has not been added to 'runtest.jl', or the slurm system either.

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,98 @@
...
+for ArrayType in ArrayTypes
+warpfun = (ξ1, ξ2, ξ3) -> begin
+  x1 = ξ1 + (ξ1 - 1/2) * cos(2 * π * ξ2 * ξ3) / 4
+  x2 = ξ2  + (ξ2 - 1/2) * cos(2 * π * ξ2 * ξ3) / 4
+  x3 = ξ3 + ξ1 / 4 + ξ2^2 / 2 + sin(ξ1 * ξ2 * ξ3) + exp(sin(2π * (ξ1 * ξ2 + ξ3)))/20
+  return (x1, x2, x3)
+end
+for N in 4:4
+for Ne in 1:20
+println(N," ",Ne)
```
jkozdon: Why this print?

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,98 @@
...
+  end
+  Stot = zeros(Nqk)
+  S1tot = zeros(Nqk)
+  Err = 0
+  for k in 1:Nqk
+  Stot[k] = MPI.Reduce(S[k], +, 0, MPI.COMM_WORLD)
+  S1tot[k] = MPI.Reduce(S1[k], +, 0, MPI.COMM_WORLD)
+  Err += (0.5 - Stot[k] / S1tot[k])^2
+  end
+  Err = sqrt(Err / Nqk)
```
jkozdon: Why divide by `Nqk`?

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,98 @@
...
+    for k in 1:Nqk
+      for j in 1:Nq
+        for i in 1:Nq
+          ijk = i + Nq * ((j-1)+ Nq * (k-1))
+          S[k] += localvgeo[ijk,grid.x1id,e] * localvgeo[ijk,grid.MHid,e]
+          S1[k] += localvgeo[ijk,grid.MHid,e]
+        end
+      end
+    end
+  end
```
jkozdon: I do not understand why you do not need to go up the vertical elements?

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,98 @@
...
+for ArrayType in ArrayTypes
+warpfun = (ξ1, ξ2, ξ3) -> begin
+  x1 = ξ1 + (ξ1 - 1/2) * cos(2 * π * ξ2 * ξ3) / 4
+  x2 = ξ2  + (ξ2 - 1/2) * cos(2 * π * ξ2 * ξ3) / 4
+  x3 = ξ3 + ξ1 / 4 + ξ2^2 / 2 + sin(ξ1 * ξ2 * ξ3) + exp(sin(2π * (ξ1 * ξ2 + ξ3)))/20
+  return (x1, x2, x3)
+end
+for N in 4:4
+for Ne in 1:20
```
jkozdon: This seems odd to me, why loop over Ne?

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,98 @@
+using MPI
```
jkozdon: Indenting the file would make it easier to review.

test/DGmethods/Horizontal_integral_test.jl
slifer50: This is to show polynomial order and number of elements in each horizontal direction, can be removed if unnecessary

test/DGmethods/Horizontal_integral_test.jl
slifer50: To divide by the number of contributions.

test/DGmethods/Horizontal_integral_test.jl
slifer50: Because it's been set for only a single vertical element

test/DGmethods/Horizontal_integral_test.jl
slifer50: To see if there is any difference if we increase the number of horizontal elements. Can be removed if unnecessary

test/DGmethods/Horizontal_integral_test.jl
slifer50: Absolutely, will do.

slifer50:
@simonbyrne @kpamnany How should I go about making this an appropriate test ?

test/DGmethods/Horizontal_integral_test.jl
jkozdon: I don't understand why this is needed.

slifer50: To get an average error, since it's taking in one contribution per level k we divide by the number of k levels ( since there is also only one element in the third direction)

jkozdon:
Here is a summary of all the changes I have suggested

- Add calls to `@test` (see examples in the test directory)
- I would add the integration of a function that is
  - different for each horizontal slab (right now all the slabs compute `1/2` so you do not really know if you are correct when there are difference)
  - integrate a function with a geometry that you cannot implement exactly and measure the convergence rate; see for example how the [vertical integration test is done](https://github.com/climate-machine/CLIMA/blob/4294b78a2c24d05e9b51e5e99ed5c4de58f073f2/test/DGmethods/integral_test_sphere.jl)
- Add a test of the 2D metrics

slifer50:
@jkozdon I am having trouble with 

>       * integrate a function with a geometry that you cannot implement exactly and measure the convergence rate;

I am trying to do this by using the metrics to calculate the surface area of a sphere. This is starting from this line https://github.com/slifer50/CLIMA/blob/3270a59e5b9caf19fcc45f70028149a5d7376b59/test/DGmethods/Horizontal_integral_test.jl#L172.

I am currently not getting the expected result but I haven't been able to understand where the large deviation I'm having from my expected result is coming from, I'd appreciate it if you could take a look and see if you can find what I'm doing wrong.

jkozdon:
You need to do several mesh levels, and then measure the rate of convergence. Since it is not exact, you cannot do just one level.

See how we test the [vertical integration functionality](https://github.com/climate-machine/CLIMA/blob/4294b78a2c24d05e9b51e5e99ed5c4de58f073f2/test/DGmethods/integral_test_sphere.jl#L140-L159)

> I am currently not getting the expected result but I haven't been able to understand where the large deviation I'm having from my expected result is coming from

Not sure what you mean by this, hard to estimate analytically what the error should be when the integration is not exact. This is why I suggest you run several levels of refinement, and then look at the convergence rate.

slifer50:
@jkozdon I've tried with progressive refinement in the same way that's done in the case you suggested but there is no convergence. The exact solution remains quite far from the estimation. In this case it's just the surface area of the outer and inner spheres.

smarras79:
> @jkozdon I've tried with progressive refinement in the same way that's done in the case you suggested but there is no convergence. The exact solution remains quite far from the estimation. In this case it's just the surface area of the outer and inner spheres. So what I'm trying to understand is the source of the discrepancy.

@slifer50 @jkozdon  as discussed yesterday at the weekly meeting, I would suggest that you push the working version for the flow in a box and open an issue about the application of your integration scheme to the sphere. As long as you have thoroughly tested it for cartesian grids in a box (LES mode), we will be able to start using it for our current tests.

jkozdon:
> @slifer50 @jkozdon as discussed yesterday at the weekly meeting, I would suggest that you push the working version for the flow in a box and open an issue about the application of your integration scheme to the sphere. As long as you have thoroughly tested it for cartesian grids in a box (LES mode), we will be able to start using it for our current tests.
> Describe the details of what the issue is and mark it as WIP. You can revisit it at a later stage.

This seem very dangerous to me since you're going to trust whether the physics is correct based on possibly incorrect calculation of the statistics.

Based on the tests I've seen, I would say there is some fundamental problem with what's being calculated.

I will not approve the PR and merge it until my [previous comments](https://github.com/climate-machine/CLIMA/pull/504#issuecomment-551168602) are addressed. Someone else is welcome to, but I will not.

src/Mesh/Grids.jl
```diff
@@ -304,6 +312,11 @@ function computegeometry(topology::AbstractTopology{dim}, D, ξ, ω, meshwarp,
       x3ξ3 = J * (ξ1x1 * ξ2x2 - ξ2x1 * ξ1x2)
       hypot(x1ξ3, x2ξ3, x3ξ3)
     end
+    map!(JH, J, ξ1x3, ξ2x3, ξ3x3
+        ) do J, ξ1x3, ξ2x3, ξ3x3
+      hypot(J*ξ1x3 ,J*ξ2x3, J*ξ3x3)
+    end
+    MHJH .= MH .* JH
```
jkozdon:
```julia
map!(MHJH, J, ξ3x1, ξ3x2, ξ3x3) do J, ξ3x1, ξ3x2, ξ3x3
  hypot(J*ξ3x1, J*ξ3x2, J*ξ3x3)
end
MHJH .= MH .* MHJH
```

src/Mesh/Grids.jl
```diff
@@ -296,6 +299,11 @@ function computegeometry(topology::AbstractTopology{dim}, D, ξ, ω, meshwarp,
       x2ξ2 = J * ξ1x1
       hypot(x1ξ1, x2ξ2)
     end
+    map!(JH, J, ξ1x1, ξ1x2) do J, ξ2x1, ξ2x2
+      hypot(J * ξ2x2, J * ξ2x1)
+    end
+    MHJH .= MH .* JH
```
jkozdon:
```julia
map!(MHJH, J, ξ2x1, ξ2x2) do J, ξ2x1, ξ2x2
  hypot(J*ξ2x1, J*ξ2x2)
end
MHJH .= MH .* MHJH
```

src/Mesh/Metrics.jl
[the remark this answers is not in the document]
jkozdon: This [comment](https://github.com/climate-machine/CLIMA/pull/504#discussion_r340769742) had a typo.

The cross product terms should be

(see [Kopriva (2006) equation (5)](https://doi.org/10.1007/s10915-005-9070-8)) which then gives

src/Mesh/Grids.jl
slifer50: @jkozdon this is doesn't seem like it would change anything, is it just a suggestion for cleanup?

slifer50: Oh I see what you mean
Let me try that

slifer50: @jkozdon that was it. Thanks a lot, I really appreciate it. Going through with cleanup and will push for you to verify when it's done.

slifer50:
@jkozdon Have a look at the latest push and let me know if any changes should be made, this should cover all of your points except for the 2d test case which I'll be working on implementing.

slifer50:
@jkozdon Closing this PR and opening a new one which has everything you requested resolved on it.

PR #505, opened 2019-10-28
Remainder model and Float32

jkozdon:
PR to show the breakdown of the remainder model with Float32


simonbyrne:
how did the error manifest itself? what is the easiest way to check in #511?

jkozdon:
I saw problems in the visualization of the Float32 and RemainderModel results.

PR #506, opened 2019-10-29
Add MoistThermo constructor consistency tests

charleskawczynski:
Adds consistency tests to `ThermoDynamicState` constructors:

 - Tests that getter functions return the same value as input arguments to the thermo state.

@tapios 10 of these tests are failing (related to pressure and liquid-ice potential temperature), and I may need your help in resolving some of these tests. This is likely the issue that @smarras79, @slifer50 is running into with initial conditions prescribed by liquid ice potential temperature.

### DIFF
--- a/examples/DGmethods/ex_002_IOstrings.jl
+++ b/examples/DGmethods/ex_002_IOstrings.jl
@@ -98,7 +98,7 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
   H     = Rm_sfc * T_sfc / grav;
   p     = P_sfc * exp(-xvert/H);
   #Density, Temperature
-  TS    = LiquidIcePotTempSHumEquil_no_ρ(θ_liq, q_pt, p)
+  TS    = LiquidIcePotTempSHumEquil(θ_liq, q_pt, p)
   ρ     = air_density(TS)
   T     = air_temperature(TS)
   #Assign State Variables
--- a/examples/DGmethods/ex_002_IOstrings.jl
+++ b/examples/DGmethods/ex_002_IOstrings.jl
@@ -98,7 +98,7 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
   H     = Rm_sfc * T_sfc / grav;
   p     = P_sfc * exp(-xvert/H);
   #Density, Temperature
-  TS    = LiquidIcePotTempSHumNonEquil(θ_liq, q_pt, p)
+  TS    = LiquidIcePotTempSHumNonEquil_given_pressure(θ_liq, q_pt, p)
   ρ     = air_density(TS)
   T     = air_temperature(TS)
   #Assign State Variables
--- a/examples/DGmethods/ex_002_IOstrings.jl
+++ b/examples/DGmethods/ex_002_IOstrings.jl
@@ -29,7 +29,7 @@ using GPUifyLoops
   using CUDAnative
   using CuArrays
   CuArrays.allowscalar(false)
-  const ArrayTypes = (CuArray,) 
+  const ArrayTypes = (CuArray,)
 else
   const ArrayTypes = (Array,)
 end
@@ -46,11 +46,11 @@ const seed = MersenneTwister(0)
 """
   Initial Condition for DYCOMS_RF01 LES
 @article{doi:10.1175/MWR2930.1,
-author = {Stevens, Bjorn and Moeng, Chin-Hoh and Ackerman, 
-          Andrew S. and Bretherton, Christopher S. and Chlond, 
-          Andreas and de Roode, Stephan and Edwards, James and Golaz, 
-          Jean-Christophe and Jiang, Hongli and Khairoutdinov, 
-          Marat and Kirkpatrick, Michael P. and Lewellen, David C. and Lock, Adrian and 
+author = {Stevens, Bjorn and Moeng, Chin-Hoh and Ackerman,
+          Andrew S. and Bretherton, Christopher S. and Chlond,
+          Andreas and de Roode, Stephan and Edwards, James and Golaz,
+          Jean-Christophe and Jiang, Hongli and Khairoutdinov,
+          Marat and Kirkpatrick, Michael P. and Lewellen, David C. and Lock, Adrian and
           Maeller, Frank and Stevens, David E. and Whelan, Eoin and Zhu, Ping},
 title = {Evaluation of Large-Eddy Simulations via Observations of Nocturnal Marine Stratocumulus},
 journal = {Monthly Weather Review},
@@ -74,14 +74,14 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
   T_sfc::FT     = 292.5
   P_sfc::FT     = MSLP
   ρ_sfc::FT     = P_sfc / Rm_sfc / T_sfc
-  # Specify moisture profiles 
+  # Specify moisture profiles
   q_liq::FT      = 0
   q_ice::FT      = 0
   zb::FT         = 600    # initial cloud bottom
   zi::FT         = 840    # initial cloud top
   dz_cloud       = zi - zb
-  q_liq_peak::FT = 0.00045 #cloud mixing ratio at z_i    
-  if xvert > zb && xvert <= zi        
+  q_liq_peak::FT = 0.00045 #cloud mixing ratio at z_i
+  if xvert > zb && xvert <= zi
     q_liq = (xvert - zb)*q_liq_peak/dz_cloud
   end
   if xvert <= zi
@@ -91,14 +91,14 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
     θ_liq = FT(297.5) + (xvert - zi)^(FT(1/3))
     q_tot = FT(1.5e-3)
   end
-    
+
   # Calculate PhasePartition object for vertical domain extent
-  q_pt  = PhasePartition(q_tot, q_liq, q_ice) 
+  q_pt  = PhasePartition(q_tot, q_liq, q_ice)
   #Pressure
   H     = Rm_sfc * T_sfc / grav;
   p     = P_sfc * exp(-xvert/H);
   #Density, Temperature
-  TS    = LiquidIcePotTempSHumEquil(θ_liq, q_pt, p)
+  TS    = LiquidIcePotTempSHumNonEquil(θ_liq, q_pt, p)
   ρ     = air_density(TS)
   T     = air_temperature(TS)
   #Assign State Variables
@@ -107,7 +107,7 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
   e_pot       = grav * xvert
   E           = ρ * total_energy(e_kin, e_pot, T, q_pt)
   state.ρ     = ρ
-  state.ρu    = SVector(ρ*u, ρ*v, ρ*w) 
+  state.ρu    = SVector(ρ*u, ρ*v, ρ*w)
   state.ρe    = E
   state.moisture.ρq_tot = ρ * q_tot
 end
@@ -123,8 +123,8 @@ function run(mpicomm, ArrayType, dim, topl, N, timeend, FT, dt, C_smag, LHF, SHF
   # Problem constants
   # Radiation model
   κ             = FT(85)
-  α_z           = FT(1) 
-  z_i           = FT(840) 
+  α_z           = FT(1)
+  z_i           = FT(840)
   D_subsidence  = FT(3.75e-6)
   ρ_i           = FT(1.13)
   F_0           = FT(70)
@@ -133,17 +133,17 @@ function run(mpicomm, ArrayType, dim, topl, N, timeend, FT, dt, C_smag, LHF, SHF
   f_coriolis    = FT(7.62e-5)
   u_geostrophic = FT(7)
   v_geostrophic = FT(-5.5)
-  
+
   # Model definition
   model = AtmosModel(FlatOrientation(),
                      NoReferenceState(),
                      SmagorinskyLilly{FT}(C_smag),
                      EquilMoist(),
                      StevensRadiation{FT}(κ, α_z, z_i, ρ_i, D_subsidence, F_0, F_1),
-                     (Gravity(), 
-                      RayleighSponge{FT}(zmax, zsponge, 1), 
-                      Subsidence(), 
-                      GeostrophicForcing{FT}(f_coriolis, u_geostrophic, v_geostrophic)), 
+                     (Gravity(),
+                      RayleighSponge{FT}(zmax, zsponge, 1),
+                      Subsidence(),
+                      GeostrophicForcing{FT}(f_coriolis, u_geostrophic, v_geostrophic)),
                      DYCOMS_BC{FT}(C_drag, LHF, SHF),
                      Initialise_DYCOMS!)
   # Balancelaw description
@@ -153,9 +153,9 @@ function run(mpicomm, ArrayType, dim, topl, N, timeend, FT, dt, C_smag, LHF, SHF
                CentralNumericalFluxDiffusive(),
                CentralGradPenalty())
   Q = init_ode_state(dg, FT(0); device=CPU())
-    
+
   lsrk = LSRK54CarpenterKennedy(dg, Q; dt = dt, t0 = 0)
-  # Calculating initial condition norm 
+  # Calculating initial condition norm
 #=  eng0 = norm(Q)
   @info @sprintf """Starting
   norm(Q₀) = %.16e""" eng0
@@ -176,7 +176,7 @@ function run(mpicomm, ArrayType, dim, topl, N, timeend, FT, dt, C_smag, LHF, SHF
                                   Dates.dateformat"HH:MM:SS"))
     end
   end
-  
+
   # Setup VTK output callbacks
   step = [0]
     cbvtk = GenericCallbacks.EveryXSimulationSteps(1) do (init=false)
@@ -184,13 +184,13 @@ function run(mpicomm, ArrayType, dim, topl, N, timeend, FT, dt, C_smag, LHF, SHF
     outprefix = @sprintf("%s/dycoms_%dD_mpirank%04d_step%04d", OUTPATH, dim,
                            MPI.Comm_rank(mpicomm), step[1])
     @debug "doing VTK output" outprefix
-    writevtk(outprefix, Q, dg, flattenednames(vars_state(model,FT)), 
+    writevtk(outprefix, Q, dg, flattenednames(vars_state(model,FT)),
              dg.auxstate, flattenednames(vars_aux(model,FT)))
-        
+
     step[1] += 1
     nothing
   end
-    
+
   solve!(Q, lsrk; timeend=timeend, callbacks=(cbinfo, cbvtk))
 
 end
@@ -212,7 +212,7 @@ let
   for ArrayType in ArrayTypes
     # Problem type
     FT = Float32
-    # DG polynomial order 
+    # DG polynomial order
     N = 4
     # SGS Filter constants
     C_smag = FT(0.15)
@@ -231,9 +231,9 @@ let
       brickrange = (grid1d(xmin, xmax, elemsize=FT(grid_resolution[1])*N),
                     grid1d(ymin, ymax, elemsize=FT(grid_resolution[2])*N),
                     grid1d(zmin, zmax, elemsize=FT(grid_resolution[end])*N))
-      
+
     zsponge = FT(0.75 * zmax)
-    
+
     topl = StackedBrickTopology(mpicomm, brickrange,
                                 periodicity = (true, true, false),
                                 boundary=((0,0),(0,0),(1,2)))
@@ -244,9 +244,9 @@ let
 
     #Create unique output path directory:
     OUTPATH = IOstrings_outpath_name(problem_name, grid_resolution)
-      
+
 #    @info (ArrayType, dt, FT, dim)
-#    result = run(mpicomm, ArrayType, dim, topl, 
+#    result = run(mpicomm, ArrayType, dim, topl,
 #                 N, timeend, FT, dt, C_smag, LHF, SHF, C_drag, grid_resolution, domain_size, zmax, zsponge, problem_name, OUTPATH)
 
   end
--- a/src/Common/MoistThermodynamics/MoistThermodynamics.jl
+++ b/src/Common/MoistThermodynamics/MoistThermodynamics.jl
@@ -20,7 +20,7 @@ export linearized_air_pressure
 export total_energy, internal_energy, internal_energy_sat
 
 # Specific heats of moist air
-export cp_m, cv_m, gas_constant_air, moist_gas_constants
+export cp_m, cv_m, gas_constant_air, gas_constants
 
 # Latent heats
 export latent_heat_vapor, latent_heat_sublim, latent_heat_fusion
@@ -35,7 +35,8 @@ export saturation_excess
 export liquid_fraction_equil, liquid_fraction_nonequil, saturation_adjustment, PhasePartition_equil
 
 # Auxiliary functions, e.g., for diagnostic purposes
-export air_temperature_from_liquid_ice_pottemp, dry_pottemp, dry_pottemp_given_pressure, virtual_pottemp, exner, exner_given_pressure
+export air_temperature_from_liquid_ice_pottemp, air_temperature_from_liquid_ice_pottemp_given_pressure
+export dry_pottemp, dry_pottemp_given_pressure, virtual_pottemp, exner, exner_given_pressure
 export liquid_ice_pottemp, liquid_ice_pottemp_given_pressure, liquid_ice_pottemp_sat, relative_humidity
 
 include("states.jl")
@@ -105,11 +106,8 @@ linearized_air_pressure(ρ::FT, e_tot::FT, e_pot::FT, q::PhasePartition{FT}=q_pt
 linearized_air_pressure(e_kin::FT, e_pot::FT, ts::ThermodynamicState{FT}) where {FT<:Real} =
   linearized_air_pressure(air_density(ts), total_energy(e_kin, e_pot, ts), e_pot, PhasePartition(ts))
 
-function linearized_air_pressure(e_kin::FT, e_pot::FT, ts::PhaseDry{FT}) where {FT<:Real}
-  ρ = air_density(ts)
-  e_tot = total_energy(e_kin, e_pot, ts)
-  return ρ*FT(R_d)*FT(T_0) + FT(R_d)/FT(cv_d)*(ρ*e_tot - ρ*e_pot)
-end
+linearized_air_pressure(e_kin::FT, e_pot::FT, ts::PhaseDry{FT}) where {FT<:Real} =
+  linearized_air_pressure(air_density(ts), total_energy(e_kin, e_pot, ts), e_pot, PhasePartition(ts))
 
 """
     air_density(T, p[, q::PhasePartition])
@@ -198,7 +196,7 @@ cv_m(ts::PhaseDry{FT}) where {FT<:Real} = FT(cv_d)
 
 
 """
-    (R_m, cp_m, cv_m, γ_m) = moist_gas_constants([q::PhasePartition])
+    (R_m, cp_m, cv_m, γ_m) = gas_constants([q::PhasePartition])
 
 Wrapper to compute all gas constants at once, where optionally,
  - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
@@ -212,7 +210,7 @@ The function returns a tuple of
 Without the specific humidity arguments, the results
 are that of dry air.
 """
-function moist_gas_constants(q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
+function gas_constants(q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
     R_gas  = gas_constant_air(q)
     cp = cp_m(q)
     cv = cv_m(q)
@@ -221,7 +219,7 @@ function moist_gas_constants(q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
 end
 
 """
-    (R_m, cp_m, cv_m, γ_m) = moist_gas_constants(ts::ThermodynamicState)
+    (R_m, cp_m, cv_m, γ_m) = gas_constants(ts::ThermodynamicState)
 
 Wrapper to compute all gas constants at once, given a thermodynamic state `ts`.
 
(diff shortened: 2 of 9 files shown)


tapios:
@charleskawczynski Maybe best if we sit down together when you are back, or schedule a call if you want to go over before then.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -109,20 +109,107 @@ using LinearAlgebra
...
+  ρ = collect(range(0.5,2.0,length=n))
+  q_tot_dry = zeros(n)
+  q_tot_moist = collect(range(0.0,0.2,length=n))
+  q_liq = collect(range(0.0,0.1,length=n))
+  q_ice = collect(range(0.0,0.05,length=n))
+
+  e_int = collect(range(10_000.0,200_000.0,length=n))
+  p = collect(range(10_000.0,200_000.0,length=n))
+  # p = ones(n) .* FT(MSLP) # some cases fail using these parameters
+  # ρ = ones(n) .* 1.0      # some cases fail using these parameters
```
charleskawczynski: When uncommented,
```
  ts = LiquidIcePotTempSHumEquil.(θ_liq_ice, q_tot_moist, ρ)
  @test all(liquid_ice_pottemp.(ts) .≈ θ_liq_ice)
```
Fails, so this needs to be made more robust.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -779,32 +779,84 @@ Compute the temperature that is consistent with
 See also [`saturation_adjustment`](@ref).
 """
 function saturation_adjustment_q_tot_θ_liq_ice(θ_liq_ice::FT, q_tot::FT, ρ::FT, p::FT) where {FT<:Real}
```
tapios: Shouldn't you have either pressure `p` or density `rho` as input arguments, but not both, so as not to overspecify the problem?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -779,32 +779,84 @@ Compute the temperature that is consistent with
...
-  T_1 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, p) # Assume all vapor
+  T_1 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, p, PhasePartition(q_tot)) # Assume all vapor
   q_v_sat = q_vap_saturation(T_1, ρ)
   if q_tot <= q_v_sat # If not saturated
     return T_1
   else  # If saturated, iterate
     T_2 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, p, PhasePartition(q_tot, FT(0), q_tot)) # Assume all ice
     T, converged = find_zero(
-      T -> θ_liq_ice - liquid_ice_pottemp_sat(T, p, PhasePartition_equil(T, ρ, q_tot)),
+      T -> θ_liq_ice - liquid_ice_pottemp_sat_given_pressure(T, p, PhasePartition_equil(T, ρ, q_tot)),
```
tapios: I'm confused here that `liquid_ice_pottemp_sat_given_pressure` takes pressure `p` as input argument, but `PhasePartition_equil` uses density `rho`. How is consistency ensured (`p = \rho * R * T`)?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -779,32 +779,84 @@ Compute the temperature that is consistent with
...
+      T_1, T_2, SecantMethod(), FT(1e-3), 10)
+    return T
+  end
+end
+function saturation_adjustment_q_tot_θ_liq_ice(θ_liq_ice::FT, q_tot::FT, ρ::FT) where {FT<:Real}
```
tapios: This version makes sense to me.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -98,56 +97,28 @@ struct PhaseDry{FT} <: ThermodynamicState{FT}
...
+    LiquidIcePotTempSHumEquil(θ_liq_ice, q_tot, ρ)
+    LiquidIcePotTempSHumEquil(θ_liq_ice, q_tot, ρ[, p])
 
 Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
 
  - `θ_liq_ice` - liquid-ice potential temperature
  - `q_tot` - total specific humidity
  - `ρ` - density
+and, optionally
  - `p` - pressure
```
tapios: How is the optional pressure used while ensuring consistency of the overspecified state?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Yes, still in progress..

src/Common/MoistThermodynamics/states.jl
charleskawczynski: I think this has been fixed.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
+  # which is potentially expensive. Other options should
+  # be investigated. Also, the tolerance/iterations are large
+  # when iterated. Perhaps this can be reduced with better guesses.
+
+  # TODO/FIXME: The iteration process seems brittle, and needs to
+  # be reviewed/improved
+  T_1 = air_temperature_initial_guess(θ_liq_ice, ρ, PhasePartition(q_tot)) # Assume all vapor
   q_v_sat = q_vap_saturation(T_1, ρ)
-  if q_tot <= q_v_sat # If not saturated
+  saturated = q_tot <= q_v_sat
```
tapios: Shouldn't this be `unsaturated`?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
+  if saturated
     return T_1
-  else  # If saturated, iterate
-    T_2 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, p, PhasePartition(q_tot, FT(0), q_tot)) # Assume all ice
+  else
+    T_2 = air_temperature_initial_guess(θ_liq_ice, ρ, PhasePartition(q_tot, FT(0), q_tot)) # Assume all vapor
```
tapios: It looks like the `PhasePartition` assumes all ice, rather than all vapor, doesn't it?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
 The liquid-ice potential temperature,
 given a thermodynamic state `ts`.
 """
 liquid_ice_pottemp(ts::ThermodynamicState) =
-  liquid_ice_pottemp(air_temperature(ts), air_pressure(ts), PhasePartition(ts))
+  liquid_ice_pottemp(air_temperature(ts), air_density(ts), PhasePartition(ts))
 liquid_ice_pottemp(ts::PhaseDry) = dry_pottemp(ts)
 
 """
     dry_pottemp(T, p[, q::PhasePartition])
```
tapios:
```suggestion
    dry_pottemp(T, \rho[, q::PhasePartition])
```

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: It seems very odd to get an initial guess by secant iterations, before doing more secant iterations. Why not trying the all vapor initial guess, which in most cases will be quite good:
```suggestion
    T_2 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, ρ, PhasePartition(q_tot, FT(0), FT(0)))
```

tapios:
Looks great overall. Thanks for sorting this out, @charleskawczynski!

examples/DGmethods/ex_002_IOstrings.jl
```diff
@@ -91,14 +91,14 @@ function Initialise_DYCOMS!(state::Vars, aux::Vars, (x,y,z), t)
...
+
   # Calculate PhasePartition object for vertical domain extent
-  q_pt  = PhasePartition(q_tot, q_liq, q_ice) 
+  q_pt  = PhasePartition(q_tot, q_liq, q_ice)
   #Pressure
   H     = Rm_sfc * T_sfc / grav;
   p     = P_sfc * exp(-xvert/H);
   #Density, Temperature
-  TS    = LiquidIcePotTempSHumEquil_no_ρ(θ_liq, q_pt, p)
+  TS    = LiquidIcePotTempSHumNonEquil(θ_liq, q_pt, p)
```
tapios: Might be good to make the default argument here density too, as in other functions, and call the version with pressure as argument `_given_pressure` or something like it.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -98,7 +99,7 @@ The air pressure, linearized around a dry rest state, from the equation of state
 and, optionally,
  - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
 """
-linearized_air_pressure(ρ::FT, e_tot::FT, e_pot::FT, q::PhasePartition{FT}=PhasePartition{FT}(FT(0), FT(0), FT(0))) where {FT<:Real} =
+linearized_air_pressure(ρ::FT, e_tot::FT, e_pot::FT, q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real} =
   ρ*FT(R_d)*FT(T_0) + FT(R_d)/FT(cv_d)*(ρ*e_tot - ρ*e_pot - (ρ*q.tot - ρ*q.liq)*FT(e_int_v0) + ρ*q.ice*(FT(e_int_i0) + FT(e_int_v0)))
 
 linearized_air_pressure(e_kin::FT, e_pot::FT, ts::ThermodynamicState{FT}) where {FT<:Real} =
```
tapios: Instead of retyping the linearized pressure for dry air in l. 110, wouldn't it be better for dry air to call the general form with all specific humidities zero? (That way, we have to change only one line if we need to change this.)

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -343,7 +344,7 @@ and, optionally,
  - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
```
tapios: In line 327 above,

`internal_energy_sat(ts::PhaseDry{FT}) where {FT<:Real} = cv_m(ts) * (air_temperature(ts) - FT(T_0))`

 I suggest to replace `cv_m(ts)` by `cv_d` (since it is for dry air anyway).

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -660,7 +661,7 @@ them.
 Otherwise, the fraction of liquid is a function that is 1 above `T_freeze` and goes to
```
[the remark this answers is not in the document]
tapios: Line 647 looks odd:
`saturation_excess(ts::PhaseDry) = max(0, - q_vap_saturation(ts))`

Why not simply `saturation_excess(ts::PhaseDry) = 0`?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
   end
 end
 
-
```
tapios: Update documentation in line 723: 
> The residual `q.tot - q.liq - q.ice` is the vapor specific humidity.

To make clear where `q.liq` etc. come from.

tapios: Change line 757 (initial guess for secant iteration in saturation adjustment) to:

` T_2 = air_temperature(e_int, PhasePartition(q_tot, FT(0), q_tot-q_v_sat)) # Assume all ice`

This will give a much better initial estimate (though it is not guaranteed to bracket the zero). It is putting only the saturation excess, computed on the basis of `T_1`, into the ice phase.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: Should this be `unsaturated`?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
+  # be investigated. Also, the tolerance/iterations are large
+  # when iterated. Perhaps this can be reduced with better guesses.
+
+  # TODO/FIXME: The iteration process seems brittle, and needs to
+  # be reviewed/improved
+  T_1 = air_temperature_initial_guess(θ_liq_ice, ρ, PhasePartition(q_tot)) # Assume all vapor
   q_v_sat = q_vap_saturation(T_1, ρ)
-  if q_tot <= q_v_sat # If not saturated
+  saturated = q_tot <= q_v_sat
+  if saturated
```
tapios:
```suggestion
  if unsaturated
```

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
+      if !converged
+        error("saturation adjustment did not converge")
+      end
     return T
   end
 end
 
 """
-    liquid_ice_pottemp(T, p, q::PhasePartition)
+    air_temperature_initial_guess(θ_liq_ice, ρ, q::PhasePartition)
```
tapios: Maybe better to call this `air_temperature_from_liquid_ice_pottemp` and change the function that has this name currently to `air_temperature_from_liquid_ice_pottemp_given_pressure`.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
- - `p` pressure
 
 See also [`saturation_adjustment`](@ref).
 """
-function saturation_adjustment_q_tot_θ_liq_ice(θ_liq_ice::FT, q_tot::FT, ρ::FT, p::FT) where {FT<:Real}
-  T_1 = air_temperature_from_liquid_ice_pottemp(θ_liq_ice, p) # Assume all vapor
+function saturation_adjustment_q_tot_θ_liq_ice(θ_liq_ice::FT, q_tot::FT, ρ::FT) where {FT<:Real}
+  # TODO/FIXME: The initial guesses are computed via iteration,
+  # which is potentially expensive. Other options should
+  # be investigated. Also, the tolerance/iterations are large
```
tapios: Yes, we can get them from linearizing the definitions, without iteration.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
-    liquid_ice_pottemp(T, p, q::PhasePartition)
+    air_temperature_initial_guess(θ_liq_ice, ρ, q::PhasePartition)
+
+Solves the non-linear equation
+``
+  T = θ_{liq_ice}*(ρ R_m T/MSLP)^(R_m/cp_m) + (LH_{v0} q_{liq} + LH_{s0} q_{ice}) / cp_m
+``
+for temperature `T`
+"""
+function air_temperature_initial_guess(θ_liq_ice::FT, ρ::FT,
```
tapios: Maybe better to call this `air_temperature_from_liquid_ice_pottemp` and change the function that has this name currently to `air_temperature_from_liquid_ice_pottemp_given_pressure`.

Replace the iterations by a linearized solution for temperature given liquid-ice potential temperature (note that the condensate correction is small, this allows linearization). I'll can write it down later how to do it,

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -849,21 +911,27 @@ The air temperature, where
 and, optionally,
  - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
 """
-air_temperature_from_liquid_ice_pottemp(θ_liq_ice::FT, p::FT, q::PhasePartition{FT}=PhasePartition{FT}(FT(0), FT(0), FT(0))) where {FT<:Real} =
-  θ_liq_ice*exner(p, q) + (FT(LH_v0)*q.liq + FT(LH_s0)*q.ice) / cp_m(q)
+function air_temperature_from_liquid_ice_pottemp(θ_liq_ice::FT,
```
tapios: Change to 
```suggestion
function air_temperature_from_liquid_ice_pottemp_given_pressure(θ_liq_ice::FT,
```

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: That is, first solve for temperature from `liquid_ice_pottemp` assuming no condensate:
```
T_dry = θ_liq_ice * (ρ * R_d θ_liq_ice / FT(MSLP))^(R_d/c_vd)
```
Then add the correction from condensate, assuming the relevant temperature is `T_dry`:
```
T = T_dry + 1/(1-R_d/c_p) * latent_heat_liq_ice(...)/c_pd
```
where `latent_heat_liq_ice` is computed from the saturation excess and phase partitioning given `T_dry`. [Please check my math here. This should be roughly right, but I did not work through the details carefully.]

src/Common/MoistThermodynamics/MoistThermodynamics.jl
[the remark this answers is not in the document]
charleskawczynski: > We no longer have q.liq, q.ice etc. here explicitly in the arguments

I've updated this a bit, but I'm still referencing the properties. Let me know if you prefer additional changes.

> Why not simply saturation_excess(ts::PhaseDry) = FT(0)

Great point, done.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: True. It's a trade-off between fewer flops with the simplified/short-circuited expressions for the dry case vs more flops but re-using the general expressions.

charleskawczynski: For now, it's likely fine to re-use, and when we go to optimize we can make these simplifications.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Should we do the same for the `e_tot` formulation?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
...
   end
 end
 
 """
-    liquid_ice_pottemp(T, p, q::PhasePartition)
+    air_temperature_initial_guess(θ_liq_ice, ρ, q::PhasePartition)
+
+Solves the non-linear equation
+``
+  T = θ_{liq_ice}*(ρ R_m T/MSLP)^(R_m/cp_m) + (LH_{v0} q_{liq} + LH_{s0} q_{ice}) / cp_m
```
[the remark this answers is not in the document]
charleskawczynski: This function is used for computing the initial guess / unsaturated case, which seems to need iterations, too.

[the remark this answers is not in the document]
charleskawczynski: This is just to compute the unsaturated case / initial guess. I can add this docstring to the `saturation_adjustment_q_tot_θ_liq_ice` function, though.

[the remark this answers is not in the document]
tapios: This is moot anyway. You should replace the initial guess function by the linearized inversion I outlined below.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: Note that this is accurate to first order in condensate specific humidities--the same linearization underlying the definition of liquid-ice potential temperature. So no need to iteratively refine this further.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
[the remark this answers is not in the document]
tapios: I added the derivation of how to obtain temperature from `θ_liq_ice` given density to the CLIMA-atmos design document. No iteration necessary for the linear approximation for `θ_liq_ice` we are using. So you just need a function `air_temperature_from_liquid_ice_pottemp` that takes density as an input.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -765,78 +766,139 @@ function saturation_adjustment(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
   end
 end
 
-
 """
```
tapios: L. 748, fix reference: 
```suggestion
See also [`saturation_adjustment_q_tot_θ_liq_ice`](@ref).
```

tapios: Regarding saturation adjustment: I updated the corresponding section in the design document to include a linearized solution for `T_2`. I suggest you implement this, and provide options to either (a) return the linearized `T_2` as the solution without further iteration (should be good enough for what we need), or (b) use this `T_2` as initial guess for secant iterations.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -827,20 +878,65 @@ The dry potential temperature where
...
+  R_m = gas_constant_air(q)
+  # Solve for temperature from liquid_ice_pottemp assuming no condensate:
+  T_dry = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_d)
+
+  # Now, add correction from condensate, assuming the relevant temperature is T_dry:
+  T_linearized = T_dry + latent_heat_liq_ice(q)/cv_d
+
+  T_non_linear = air_temperature_from_liquid_ice_pottemp_non_linear(θ_liq_ice, ρ, q)
+  return T_non_linear
+  # return T_linearized # Fails tests at the moment
```
charleskawczynski: Implementing the linearized temperature, given `θ_liq_ice, ρ, q::PhasePartition` still fails. I'll look through the math tomorrow, but at least we have the non-linear case working to go off of.

tapios: You mean there are accuracy issues if you calculate `θ_liq_ice` given `ρ`, `q`, and `T`, and then use the linearized version to get `T` back given  `θ_liq_ice`,  `ρ`, `q`? Some loss of accuracy is to be expected from the linearization. This is only accurate to order `q`.

charleskawczynski: I'll have to look more closely, I'm not sure if the inaccuracy is due to the saturated or unsaturated case. I'm assuming that it's for the unsaturated case, since this directly translates to temperature, whereas the temperature is only used as an initial guess in the saturated case. Looking into this now.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -827,20 +878,65 @@ The dry potential temperature where
...
+ - `ρ` (moist-)air density
+and, optionally,
+ - `q` [`PhasePartition`](@ref). Without this argument the results are that of dry air.
+"""
+function air_temperature_from_liquid_ice_pottemp(θ_liq_ice::FT, ρ::FT,
+  q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
+
+  R_m = gas_constant_air(q)
+  # Solve for temperature from liquid_ice_pottemp assuming no condensate:
+  T_dry = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_d)
```
tapios: For consistency, you'd want `cv_m` here (with the given `q`);
```suggestion
  T_dry = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_m)
```

tapios: For consistency, you'd want `cv_m` here (with the given `q`);
```suggestion
  T_dry = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_m)
```

tapios: For consistency, you'd want `cv_m` here (with the given `q`);
```suggestion
  T_dry = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_m)
```

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -827,20 +878,65 @@ The dry potential temperature where
...
+
+  # Now, add correction from condensate, assuming the relevant temperature is T_dry:
+  T_linearized = T_dry + latent_heat_liq_ice(q)/cv_d
```
tapios: Same here: Replace `cv_d` by `cv_m`. Perhaps better to write the whole thing in one line 
```suggestion
  T_linearized = θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv_m) + latent_heat_liq_ice(q)/cv_d
```
(or rename `T_dry` to `T_unsaturated` because this is more precisely what it is).

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: I checked the equations in the design doc numerically (which is what I think you have implemented). They are indeed accurate to order `q_c` (condensate specific humidity).

charleskawczynski:
Ok, good news. Changing from `cv_d` to `cv_m` did work for the equilibrium case, but the [non-equilibium test](https://github.com/climate-machine/CLIMA/pull/506/files#diff-2f6bc07146c90e2bb35a5666182fd975R213), which relies on this function [here](https://github.com/climate-machine/CLIMA/pull/506/files#diff-b85af3e13a8241990b17ecb9800d8496R166), still fails. Does this require special treatment since it assumes non-equilibium? @tapios. At worst, we can have the non-equilibrium constructor call the non-linear function.

charleskawczynski:
Ok, I updated the test criteria for `LiquidIcePotTempSHumNonEquil` constructor. The resulting thermo-state is accurate within `0.0004*(latent_heat_liq_ice.(q)./cv_m.(q)).^2`. I left a [test](https://github.com/climate-machine/CLIMA/pull/506/files#diff-2f6bc07146c90e2bb35a5666182fd975R157-R162) that _could_ be used in the `LiquidIcePotTempSHumNonEquil` constructor to compute temperature more accurately, but it will not be used in that constructor by default since it requires iteration.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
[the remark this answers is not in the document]
charleskawczynski: Going to leave this function here for a reference/correctness tests, but we won't use it by default.

charleskawczynski:
Unless there are any more changes needed, this should be ready to merge

smarras79:
> Unless there are any more changes needed, this should be ready to merge. I'll look into saturation adjustment convergence in more detail in a separate PR.

Thanks Charlie, I will test this new implementation with dycoms and let you know.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -191,16 +154,31 @@ struct PhaseNonEquil{FT} <: ThermodynamicState{FT}
...
  - `q_pt` - phase partition
  - `p` - pressure
 """
-function LiquidIcePotTempSHumNonEquil_no_ρ(θ_liq_ice::FT, q_pt::PhasePartition{FT}, p::FT) where {FT<:Real}
-    T = θ_liq_ice * exner(p, q_pt)
+function LiquidIcePotTempSHumNonEquil_given_pressure(θ_liq_ice::FT, q_pt::PhasePartition{FT}, p::FT) where {FT<:Real}
+    T = air_temperature_from_liquid_ice_pottemp_given_pressure(θ_liq_ice, p, q_pt)
     ρ = air_density(T, p, q_pt)
     e_int = internal_energy(T, q_pt)
     PhaseNonEquil(e_int, q_pt, ρ)
```
smarras79: Can you add the following as well:

`"""                                                                                                                                                                                                                                    
    LiquidIcePotTempSHumEquil_given_pressure(θ_liq_ice, q_pt, p)                                                                                                                                                                       
                                                                                                                                                                                                                                       
Constructs a [`PhaseEquil`](@ref) thermodynamic state from:                                                                                                                                                                            
                                                                                                                                                                                                                                       
 - `θ_liq_ice` - liquid-ice potential temperature                                                                                                                                                                                      
 - `q_pt` - phase partition                                                                                                                                                                                                            
 - `p` - pressure                                                                                                                                                                                                                      
"""
function LiquidIcePotTempSHumEquil_given_pressure(θ_liq_ice::FT, q_pt::PhasePartition{FT}, p::FT) where {FT<:Real}
    T = air_temperature_from_liquid_ice_pottemp_given_pressure(θ_liq_ice, p, q_pt)
    ρ = air_density(T, p, q_pt)
    e_int = internal_energy(T, q_pt)
    PhaseEquil(e_int, q_pt, ρ)
end

`

charleskawczynski: I'll try to do this in a later PR (I tried to add this recently, and I had to add a _lot_ more functions).

charleskawczynski:
Ah, this is failing the slurmCI. I'll take a look to update...

charleskawczynski:
@simonbyrne @lcw 
I think I've seen this sort of MPI error a few times:
```
Fatal error in MPI_Init: Other MPI error, error stack:
MPIR_Init_thread(572)..............: 
MPID_Init(224).....................: channel initialization failed
MPIDI_CH3_Init(105)................: 
MPID_nem_init(324).................: 
MPID_nem_tcp_init(178).............: 
MPID_nem_tcp_get_business_card(425): 
MPID_nem_tcp_init(384).............: gethostbyname failed, Mac-124.local (errno 1)
```
Any idea how to fix this?

lcw:
I haven't seen that MPI error message before.  Where/when are you getting it?

charleskawczynski:
[here](https://dev.azure.com/spjbyrne/CLIMA/_build/results?buildId=1534&view=logs&jobId=93056758-5bfb-5750-f113-e720ddefdb4c&taskId=c73b3de0-27be-598b-c922-0c24c64f8585&lineStart=432&lineEnd=433&colStart=1&colEnd=1)

@asraero said that it may just be a broken node. I'll try again.

charleskawczynski:
bots r+

PR #509, opened 2019-10-30
Reformulate ARK timesteppers

mwarusz:
Seems to decrease the number of iterations of GMRES by half at the cost of storing `Nstages - 1` extra vectors. I don't have a good theory why this works, only some hand-wavy arguments. Maybe worth checking if it changes anything for https://github.com/climate-machine/CLIMA/pull/505 @jkozdon

### DIFF
--- a/src/ODESolvers/AdditiveRungeKuttaMethod.jl
+++ b/src/ODESolvers/AdditiveRungeKuttaMethod.jl
@@ -3,6 +3,16 @@ export AdditiveRungeKutta
 export ARK2GiraldoKellyConstantinescu
 export ARK548L2SA2KennedyCarpenter, ARK437L2SA1KennedyCarpenter
 
+# Naive formulation that uses equation 3.8 from Giraldo, Kelly, and Constantinescu (2013) directly.
+# Seems to cut the number of solver iterations by half but requires Nstages - 1 additional storage.
+struct NaiveVariant end
+additional_storage(::NaiveVariant, Q, Nstages) = (Lstages = ntuple(i -> similar(Q), Nstages),)
+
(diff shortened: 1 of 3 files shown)


jkozdon:
> Maybe worth checking if it changes anything for #505 @jkozdon

Unfortunately it seems to make the non-Remainder tests worse and doesn't help the Remainder tests.

jkozdon:
bors try

simonbyrne:
What is this doing exactly? You're storing more of the past solver history between iterations?

mwarusz:
I am formulating a linear problem for the current stage value, basically doing it in the naive way. Previously linearity was used to combine different terms under the linear operator and the linear problem solution variable was defined as a linear combination of the current and previous stages. After solving the linear problem the current stage had to be extracted. Now we solve directly for `Qstages[i]` but need to store evaluations of `L` from previous stages.

jkozdon:
> What is this doing exactly? You're storing more of the past solver history between iterations?

To followup as well, if you look at [Giraldo, Kelly, and Constantinescu (2013)](https://doi.org/10.1137/120876034) the PR implements Equation (3.3) directly by solving for $q^{n+1}$ whereas the code its replacing is Equation (3.5) solving for $q_{tt}$.

It is the same matrix, but with a different right-hand side vector.

simonbyrne:
bors try

mwarusz:
bors try

simonbyrne:
bors try

jkozdon:
Should this be merged?

simonbyrne:
Needs to be updated for the linear solver changes.

mwarusz:
Done

PR #511, opened 2019-11-01
Special-case linear models in RemainderModel

simonbyrne:
Should help reduce cancellation error in #505.

If this works, we can probably work to make this more general.

### DIFF
--- a/src/Atmos/Model/remainder.jl
+++ b/src/Atmos/Model/remainder.jl
@@ -60,8 +60,8 @@ function flux_nondiffusive!(rem::RemainderModel{M,S}, flux::Grad, state::Vars, a
   ρe = state.ρe
   ref = aux.ref_state
   u = ρu / ρ
-  e_pot = gravitational_potential(nlm.atmos.orientation, aux)
-  p = pressure(nlm.atmos.moisture, nlm.atmos.orientation, state, aux)
+  e_pot = gravitational_potential(rem.main.orientation, aux)
+  p = pressure(rem.main.moisture, rem.main.orientation, state, aux)
   # TODO: use MoistThermodynamics.linearized_air_pressure 
   # need to avoid dividing then multiplying by ρ
   pL = ρ * FT(R_d) * FT(T_0) + FT(R_d) / FT(cv_d) * (ρe - ρ * e_pot)
@@ -79,8 +79,8 @@ function flux_nondiffusive!(rem::RemainderModel{M,S}, flux::Grad, state::Vars, a
   ρe = state.ρe
   ref = aux.ref_state
   u = ρu / ρ
-  e_pot = gravitational_potential(nlm.atmos.orientation, aux)
-  p = pressure(nlm.atmos.moisture, nlm.atmos.orientation, state, aux)
+  e_pot = gravitational_potential(rem.main.orientation, aux)
+  p = pressure(rem.main.moisture, rem.main.orientation, state, aux)
   # TODO: use MoistThermodynamics.linearized_air_pressure 
   # need to avoid dividing then multiplying by ρ
   pL = ρ * FT(R_d) * FT(T_0) + FT(R_d) / FT(cv_d) * (ρe - ρ * e_pot)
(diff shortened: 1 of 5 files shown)


simonbyrne:
bors try

simonbyrne:
I realised this doesn't correctly handle the vertical/horizontal splitting. Will need to think this through further.

src/Atmos/Model/remainder.jl
```diff
@@ -47,11 +47,47 @@ function flux_nondiffusive!(rem::RemainderModel, flux::Grad, state::Vars, aux::V
...
+  u = ρu / ρ
+  e_pot = gravitational_potential(rem.main.orientation, aux)
+  p = pressure(rem.main.moisture, rem.main.orientation, state, aux)
+  # TODO: use MoistThermodynamics.linearized_air_pressure 
+  # need to avoid dividing then multiplying by ρ
+  pL = ρ * FT(R_d) * FT(T_0) + FT(R_d) / FT(cv_d) * (ρe - ρ * e_pot)
+
+  flux.ρ = -zero(FT)
+  flux.ρu = ρu .* u' + (p - pL) * I
+  flux.ρe = ((ρe + p) / ρ - (ref.ρe + ref.p) / ref.ρ + e_pot) * ρu
```
tapios: I think this should be 
```suggestion
  flux.ρe = ((ρe - ref.ρe + p) / ρ + ref.p / ref.ρ + e_pot) * ρu
```

src/Atmos/Model/remainder.jl
```diff
@@ -47,11 +47,47 @@ function flux_nondiffusive!(rem::RemainderModel, flux::Grad, state::Vars, aux::V
...
+  u = ρu / ρ
+  e_pot = gravitational_potential(rem.main.orientation, aux)
+  p = pressure(rem.main.moisture, rem.main.orientation, state, aux)
+  # TODO: use MoistThermodynamics.linearized_air_pressure 
+  # need to avoid dividing then multiplying by ρ
+  pL = ρ * FT(R_d) * FT(T_0) + FT(R_d) / FT(cv_d) * (ρe - ρ * e_pot)
+
+  flux.ρ = -zero(FT)
+  flux.ρu = ρu .* u' + (p - pL) * I
+  flux.ρe = ((ρe + p) / ρ - (ref.ρe + ref.p) / ref.ρ) * ρu
```
tapios: Similar to above:
```suggestion
  flux.ρe = ((ρe - ref.ρe + p) / ρ + ref.p / ref.ρ) * ρu
```

src/Atmos/Model/remainder.jl
```diff
@@ -62,10 +98,19 @@ function source!(rem::RemainderModel, source::Vars, state::Vars, aux::Vars, t::R
...
-    m .-= m_s
-  end
+  fill!(m_s, 0)
+  source!(rem.sub, source_s, state, aux, t)
+  m .-= m_s
   nothing
 end
+
+function source!(rem::RemainderModel{M,S}, source::Vars, state::Vars, aux::Vars, t::Real) where {M,S<:AtmosAcousticLinearModel}
+  source!(rem.main, source, state, aux, t)
```
tapios: Shouldn't this have
```suggestion
 source.ρu = -state.ρ * ∇Φ
```
(because the gravitational acceleration is not taken into account in the linear model for acoustic waves)?

src/Atmos/Model/remainder.jl
```diff
@@ -62,10 +98,19 @@ function source!(rem::RemainderModel, source::Vars, state::Vars, aux::Vars, t::R
...
+end
+
+function source!(rem::RemainderModel{M,S}, source::Vars, state::Vars, aux::Vars, t::Real) where {M,S<:AtmosAcousticGravityLinearModel}
+  ∇Φ = ∇gravitational_potential(rem.main.orientation, aux)
+  source!(rem.main, source, state, aux, t)
+  # TODO: avoid subtractive cancellation here
+  source.ρu += state.ρ * ∇Φ
```
tapios: Should l. 115 be omitted (because the gravitational acceleration is taken into account in the linear model)?

tapios:
This looks ok. If we need to go the route of explicitly specifying the remainder model, we should break out just the remainder of the terms to be treated implicitly (as in the design docs). Otherwise, we need to specify the various fluxes and sources separately for each variant of linearization we use (2 currently, but this may increase), leading to potential sources of error. But first let's see if this makes a difference to the cancelations.

bischtob:
@simonbyrne,  is this PR still relevant?

PR #512, opened 2019-11-04
Symmetric treatment of metrics for 2nd derivatives

jkozdon:
The current way we treat metric terms for 2nd order derivatives is funky. This approach is a little more standard, and respects the symmetry of the operator.


jkozdon:
bors try

lcw:
As you noted we just need to do this for the vertical kernels as well.

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
> As you noted we just need to do this for the vertical kernels as well.

Thanks! Done.

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

PR #517, opened 2019-11-06
Add hooks for prefactorization of linear solvers

simonbyrne:
Add a `prefactorize` function that can is applied to linear solvers at setup time. Also changes the argument order to `linearsolve!` to make it consistent.

### DIFF
--- a/src/ODESolvers/AdditiveRungeKuttaMethod.jl
+++ b/src/ODESolvers/AdditiveRungeKuttaMethod.jl
@@ -14,6 +14,32 @@ using ..SpaceMethods
 using ..LinearSolvers
 using ..MPIStateArrays: device, realview
 
+
+
+"""
+    op! = EulerOperator(f!, ϵ)
+
+Construct a linear operator which performs an explicit Euler step ``Q + α f(Q)``,
+where `f!` and `op!` both operate inplace, with extra arguments passed through, i.e.
+```
+op!(LQ, Q, args...)
+```
+is equivalent to
+```
+f!(dQ, Q, args...)
+LQ .= Q .+ ϵ .* dQ
+```
+"""
+mutable struct EulerOperator{F,FT}
+  f!::F
+  ϵ::FT
+end
+
+function (op::EulerOperator)(LQ, Q, args...)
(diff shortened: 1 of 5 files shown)


simonbyrne:
bors try

simonbyrne:
bors try-

simonbyrne:
bors try

src/ODESolvers/AdditiveRungeKuttaMethod.jl
```diff
@@ -14,6 +14,32 @@ using ..SpaceMethods
 using ..LinearSolvers
 using ..MPIStateArrays: device, realview
 
+
+
+"""
+    op! = EulerOperator(f!, ϵ)
+
+Constuct a linear operator which performs an explicit Euler step ``Q + α f(Q)``,
```
lcw:
```suggestion
Construct a linear operator which performs an explicit Euler step ``Q + α f(Q)``,
```

jkozdon:
```suggestion
Construct a linear operator which performs an explicit Euler step ``Q + ϵ f(Q)``,
```

src/ODESolvers/AdditiveRungeKuttaMethod.jl
```diff
@@ -111,7 +144,13 @@ function AdditiveRungeKutta(spacedisc::AbstractSpaceMethod,
                      Q; dt=dt, t0=t0)
 end
 
-ODEs.updatedt!(ark::AdditiveRungeKutta, dt) = (ark.dt = dt)
+function ODEs.updatedt!(ark::AdditiveRungeKutta, dt)
+  ark.dt = dt
+  # this will only work for implicit solves
```
lcw:
```suggestion
  # this will only work for iterative solves
```

jkozdon:
We may want to throw a warning or error if `adjustfinalstep = true` and the matrix is being factored, otherwise the matrix will be built special for the final time step.

Other options:

- adjust final step removed and just number of steps is specified
- adjust final step behavior changed to uniformly change steps for all of the stepping

simonbyrne:
included as part of #518

Issue #523, opened 2019-11-07
slurm CI and args


jkozdon:
It seems that the args are not getting passed through the slurm CI system correctly. (This could very well be my mistake, but since PR #489 is replacing the old slurm CI system, so want to make sure this doesn't continue to happen)

From the slurm system `true` should be passed through on this test [1] but it is not getting propagated [2].  

cc: @kpamnany 

[1]: https://github.com/climate-machine/CLIMA/blob/4294b78a2c24d05e9b51e5e99ed5c4de58f073f2/.slurmci/test_set.jl#L38
[2]: https://gist.github.com/climabot/57e451e3b3dd8669e6e56051dfb6d627#file-out_7701599-L34

kpamnany:
Yes, [this](https://github.com/climate-machine/slurmci/blob/a228fc139eb03c37aba577f905c13eadcf58e8dc/scripts/test-gpu.sh#L18) should be `$*` instead of `$1`. Will fix it here and in the other scripts tomorrow.

PR #526, opened 2019-11-10
Disable failing integration test

simonbyrne:
pseudo1D_advection_diffusion.jl is currently failing. This was previously hidden by #523, but now that is fixed tests are broken.


simonbyrne:
@lcw is this the test you were having trouble with for 1d ilex?

lcw:
Nope it was a different one.

PR #527, opened 2019-11-11
Module for timing

kpamnany:
~~Work in progress.~~ Complete.
- [X] Needs tests
- [X] Needs documentation
- [X] Needs to be used

Currently instrumented:
- DYCOMS driver
- MPI calls
- VTK output

Export `TICTOC_PRINT_RESULTS=1` to display timing output.

### DIFF
--- a/src/InputOutput/VTK/VTK.jl
+++ b/src/InputOutput/VTK/VTK.jl
@@ -6,4 +6,8 @@ include("writemesh.jl")
 include("writevtk.jl")
 include("writepvtu.jl")
 
+function __init__()
+    tictoc()
+end
+
 end
--- a/src/Utilities/TicToc/TicToc.jl
+++ b/src/Utilities/TicToc/TicToc.jl
@@ -1,0 +1,175 @@
+"""
+    TicToc -- timing measurement
+
+Low-overhead time interval measurement via minimally invasive macros.
+
+"""
+module TicToc
+
+using Printf
+
+export @tic, @toc, tictoc
+
+# disable to reduce overhead
+const tictoc_track_memory = true
+
+if tictoc_track_memory
+
+mutable struct TimingInfo
+    ncalls::Int
+    time::UInt64
+    allocd::Int64
+    gctime::UInt64
+    curr::UInt64
+    mem::Base.GC_Num
+end
+TimingInfo() = TimingInfo(0, 0, 0, 0, 0,
+                          Base.GC_Num(0, 0, 0, 0, 0, 0, 0,
+                                      0, 0, 0, 0, 0, 0, 0))
+
+else # !tictoc_track_memory
+
+mutable struct TimingInfo
+    ncalls::Int
+    time::UInt64
+    curr::UInt64
+end
+TimingInfo() = TimingInfo(0, 0, 0)
+
+end # if tictoc_track_memory
+
+const timing_infos = TimingInfo[]
+const timing_info_names = Symbol[]
+const atexit_function_registered = Ref(false)
+
+# `@tic` helper
+function _tic(nm)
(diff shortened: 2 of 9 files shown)


src/Utilities/TicToc/TicToc.jl
```diff
@@ -0,0 +1,163 @@
...
+    end
+end
+
+"""
+    tictoc()
+
+Call at program start to set up the globals used by the macros and
+to register the at-exit callback.
+"""
+function tictoc()
```
vchuravy: As I mentioned on the call you probably want to use:

```julia
    function tictoc()
           for nm in timing_info_names
               exti = Symbol(nm)
               isdefined(@__MODULE__, exti) && continue
               expr = quote
                   const $exti = $TimingInfo()
               end
               eval(Expr(:toplevel, expr))
               push!(timing_infos, getfield(@__MODULE__, exti))
           end
           atexit(print_timing_info)
           return length(timing_info_names)
       end
```

src/Utilities/TicToc/TicToc.jl
```diff
@@ -0,0 +1,163 @@
...
+end # @static if
+
+const timing_infos = TimingInfo[]
+const timing_info_names = Symbol[]
+
+# `@tic` helper
+function _tic(nm)
+    exti = Symbol("tictoc__", nm)
+    global timing_info_names
+    push!(timing_info_names, exti)
```
vchuravy: Check whether the name is already used an throw an error.

src/Utilities/TicToc/TicToc.jl
```diff
@@ -0,0 +1,163 @@
...
+module TicToc
+
+using Printf
+
+export @tic, @toc, tictoc
+
+# disable to reduce overhead
+const tictoc_track_memory = true
+
+@static if tictoc_track_memory
```
vchuravy: `@static` is unnecessary in global scope

src/Utilities/TicToc/TicToc.jl
```diff
@@ -144,16 +144,21 @@ end
...
+macros and to register the at-exit callback.
 """
 function tictoc()
     global timing_info_names
     for nm in timing_info_names
         exti = Symbol(nm)
-        global ex = Expr(:global, Expr(:(=), exti, Expr(:call, Expr(:(.), :TicToc, QuoteNode(:TimingInfo)))))
-        eval(ex)
-        eval(Expr(:call, :push!, :timing_infos, exti))
+        if isdefined(@__MODULE__, exti)
```
vchuravy: This means that repeated calls to `tictoc` will fail. I rather check that only one entity is declaring a global. E.g.only one use of `@tic name`

src/InputOutput/VTK/VTK.jl
```diff
@@ -6,4 +6,8 @@ include("writemesh.jl")
 include("writevtk.jl")
 include("writepvtu.jl")
 
+function __init__()
```
vchuravy: Won't this lead to the multiple init issue I mentioned on the call?
https://github.com/climate-machine/CLIMA/pull/527/files#diff-0c102a9f3b330c12cac54a2a4821a48dR576

Should `tic`/`toc`s be scoped by the module they are defined in?

kpamnany: The `tictoc()` implementation won't re-create timing infos that have already been created -- that's how we the multiple init problem is handled. Or am I forgetting something?

vchuravy: Yea, you are right I am still a bit uncomfortable about namespace clashes, but most likely not something to worry about,

kpamnany:
bors try

PR #529, opened 2019-11-11
Draft PR to isolate MoistThermo issue

smarras79:
#Draft PR to figure out what broke things in the current upstream master:

This branch broke after the latest merge with upstream master.
The issue does not seem tied to MoistThermo bur rather CuArrays but it is not clear by the error printed on screen.


tapios:
@smarras79 Is this still relevant, or can it be closed?

charleskawczynski:
We should probably close it, or rebase it since it was before the git-ocalypse

smarras79:
This can be closed. We are no longer using this branch.

PR #532, opened 2019-11-13
Add new MT constructor

charleskawczynski:
Adds a new moist thermo constructor (`LiquidIcePotTempSHumEquil_given_pressure `), where pressure is given instead of density.


src/Common/MoistThermodynamics/states.jl
```diff
@@ -113,6 +114,24 @@ function LiquidIcePotTempSHumEquil(θ_liq_ice::FT, q_tot::FT, ρ::FT) where {FT<
...
+Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
+
+ - `θ_liq_ice` - liquid-ice potential temperature
+ - `q_tot` - total specific humidity
+ - `p` - pressure
+"""
+function LiquidIcePotTempSHumEquil_given_pressure(θ_liq_ice::FT, q_tot::FT, p::FT) where {FT<:Real}
+    q = PhasePartition(q_tot)
+    Π_dry = exner_given_pressure(p)
+    T = Π_dry*θ_liq_ice + latent_heat_liq_ice(q)/cp_m(q)
```
tapios: Can we make this a function that computes temperature from liquid-ice potential temperature given pressure?

charleskawczynski:
Right now, the constructor is not passing the [consistency tests](https://github.com/climate-machine/CLIMA/pull/532/files#diff-2f6bc07146c90e2bb35a5666182fd975R211-R217), I tried applying [these equations](https://github.com/climate-machine/CLIMA/pull/508#discussion_r342677797), but there's still something wrong.

tapios:
The equations you link to have the wrong specific latent heat (but I’m not sure that is what you are using). Please refer to the design doc for the correct equations (maybe you have them?)

On Nov 12, 2019, at 4:06 PM, Charles Kawczynski <notifications@github.com<mailto:notifications@github.com>> wrote:


Right now, the constructor is not passing the consistency tests<https://github.com/climate-machine/CLIMA/pull/532/files#diff-2f6bc07146c90e2bb35a5666182fd975R211-R217>, I tried applying these equations<https://github.com/climate-machine/CLIMA/pull/508#discussion_r342677797>, but there's still something wrong.

src/Common/MoistThermodynamics/states.jl
charleskawczynski: `air_temperature_from_liquid_ice_pottemp_given_pressure`? Yes, we can use that. I didn't use it initially because it didn't seem to match, exactly, what was [written down](https://github.com/climate-machine/CLIMA/pull/508#discussion_r342677797).

charleskawczynski:
> The equations you link to have the wrong specific latent heat (but I’m not sure that is what you are using). Please refer to the design doc for the correct equations (maybe you have them?) 

Will update

charleskawczynski:
Right now, the temperature equation is implemented as follows:

Screen Shot 2019-11-12 at 4 19 56 PM

But, it seems that it's still not consistent when translating to `e_int`.

tapios:
> Right now, the temperature equation is implemented as follows:
> 
> Screen Shot 2019-11-12 at 4 19 56 PM
> 
> But, it seems that it's still not consistent when translating to `e_int`.

The equation for temperature is correct. I'll write something simple on Slack to get the energy, to check the other constructors.

charleskawczynski:
Reopened in #552

PR #533, opened 2019-11-13
[Part 1] Add Newtons method with analytic derivative

charleskawczynski:
Adds Newton method with analytic derivative, instead of using automatic differentiation


charleskawczynski:
This seems a simple enough extension, I think this is ready to merge @simonbyrne @bischtob

src/Utilities/RootSolvers/RootSolvers.jl
```diff
@@ -41,8 +42,11 @@ that `f(x) ≈ 0`, and a Boolean value `converged` indicating convergence.
 `method` can be one of:
 - `SecantMethod()`: [Secant method](https://en.wikipedia.org/wiki/Secant_method)
 - `RegulaFalsiMethod()`: [Regula Falsi method](https://en.wikipedia.org/wiki/False_position_method#The_regula_falsi_(false_position)_method).
+- `NewtonsMethodAD()`: [Newton's method](https://en.wikipedia.org/wiki/Newton%27s_method) using Automatic Differentiation
+  - The `x1` argument is omitted for Newton's method.
 - `NewtonsMethod()`: [Newton's method](https://en.wikipedia.org/wiki/Newton%27s_method)
   - The `x1` argument is omitted for Newton's method.
+  - The `f′` derivative of root
```
tapios:
```suggestion
  - `f′`: derivative of function `f` whose root is sought
```

src/Utilities/RootSolvers/RootSolvers.jl
```diff
@@ -128,4 +132,18 @@ function find_zero(f::F, x0::T, ::NewtonsMethod,
   return x0, false
 end
 
+function find_zero(f::F, f′::F′, x0::T, ::NewtonsMethod,
+                   xatol=1e-3,
+                   maxiters=10_000) where {F, F′, T<:AbstractFloat}
```
tapios: `maxiters` should be a more reasonable value. For saturation adjustment, we should not need more than 3. Generally, something like a 100 as a ceiling makes sense.

charleskawczynski: These are just default values, which may be better fitted for testing correctness, we'll certainly prescribe 10-100 when called in saturation adjustment.

charleskawczynski:
Reopened in #553

PR #534, opened 2019-11-13
[Part 2] New sat adjust

charleskawczynski:
Adds saturation adjustment using Newton method in #533.

This is not currently passing the [correctness tests](https://github.com/climate-machine/CLIMA/compare/master...charleskawczynski:NewSatAdjust?expand=1#diff-2f6bc07146c90e2bb35a5666182fd975R92-R103), which are passing using Secant method (although it's _almost_ passing).

A bit more of experimentation needs to be done, it could be that the Newton method is more accurate early on. To be more systematic, we should add a test suite for an agreed range of values and test all of the methods to compare performance/convergence.


charleskawczynski:
It might be convenient to get #374 merged to help understand how to improve the robustness of saturation adjustment for whichever method we converge on.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
 PhasePartition(ts::PhaseEquil) = PhasePartition_equil(air_temperature(ts), air_density(ts), ts.q_tot)
 PhasePartition(ts::PhaseNonEquil) = ts.q
 
+function roots_∂e_int_∂T(T::FT, e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
+  cvm = cv_m(PhasePartition(q_tot))
```
tapios: This seems odd. `cv_m` needs to depend on the equilibrium partitioning of `q_tot` into `q_l` and `q_i`, which depends on the input temperature `T`, but `T` is not an input here. It seems to me you need
```suggestion
  cvm = cv_m(PhasePartition_equil(T, ρ, q_tot))
```
here.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
...
+  q_vap_sat = q_vap_saturation(T, ρ)
+  λ = liquid_fraction_equil(T)
+  L = λ*FT(LH_v0) + (1-λ)*FT(LH_s0)
+  ∂q_vap_sat_∂T = q_vap_sat*L/(FT(R_v)*T^2)
+  # T0 = FT(T_min)
+  T0 = FT(T_0)
+  return cvm + ( FT(e_int_v0) + (1-λ)*FT(e_int_i0) + (T - T0) )*∂q_vap_sat_∂T
```
tapios: The last term in parenthesis is missing the `d(cv_m)/d q_vap_sat` derivative [Eqs. (1.26) and (1.27)) in the design doc].

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
...
+ - `e_int` internal energy
+ - `ρ` (moist-)air density
+ - `q_tot` total specific humidity
+
+using Newtons method with analytic gradients.
+
+See also [`saturation_adjustment`](@ref).
+"""
+function saturation_adjustment_NewtonsMethod(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
+  T_1 = max(FT(T_min), air_temperature(e_int, PhasePartition(q_tot))) # Assume all vapor
```
tapios: Why do we need to bound this by `T_min`?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
...
+  q_v_sat = q_vap_saturation(T_1, ρ)
+  unsaturated = q_tot <= q_v_sat
+  if unsaturated
+    return T_1
+  else
+    # FIXME here: need to revisit bounds for saturation adjustment to guarantee bracketing of zero.
+    T_2 = air_temperature(e_int, PhasePartition(q_tot, FT(0), q_tot)) # Assume all ice
```
tapios: Remove these two lines (not needed here):
```
    # FIXME here: need to revisit bounds for saturation adjustment to guarantee bracketing of zero.
    T_2 = air_temperature(e_int, PhasePartition(q_tot, FT(0), q_tot)) # Assume all ice
```

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
...
+    T, converged = find_zero(
+      T -> internal_energy_sat(T, ρ, q_tot) - e_int,
+      T_ -> roots_∂e_int_∂T(T_, e_int, ρ, q_tot),
+      T_1,
+      NewtonsMethod(), FT(1e-3), 10)
+      # NewtonsMethodAD(), FT(1e-3), 3)
```
tapios: Limiting to ~3 iterations should suffice once the gradients are right.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: I was matching the other saturation adjustment function, which I believe had this put in to avoid low/negative temperatures when `e_int` is very low. I plan on adding a PR that tests saturation adjustment for a range of values, and if we can remove it then I'm all for it.

tapios: Please remove the bound in both cases. It is very risky: If it ever is active, and the air is unsaturated, `T_min` is  returned as the temperature, without warning, but it would be wrong. The bound would also degrade the initial condition for Newton iterations if active. If there are convergence issues, especially for the secant method, let's deal with it in different ways.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,52 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
 PhasePartition(ts::PhaseEquil) = PhasePartition_equil(air_temperature(ts), air_density(ts), ts.q_tot)
 PhasePartition(ts::PhaseNonEquil) = ts.q
 
+function roots_∂e_int_∂T(T::FT, e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
+  cvm = cv_m(PhasePartition(q_tot))
+  q_vap_sat = q_vap_saturation(T, ρ)
```
tapios: We should also avoid computing `q_vap_sat` twice, once here and once in `PhasePartition`. Either make it an optional input argument to `PhasePartition`, so it does not need to be computed there, or make it an optional output argument, so it does not need to be computed here.

charleskawczynski: `PhasePartition` is a constructor and doesn't call `q_vap_saturation`. Do you mean in `PhasePartition_equil`? We can do that, I'll add an issue to fix that (it'll involve changing a handful of function calls, but it's a good idea since it may result in saved flops in other function calls).

tapios: Yes, the equilibrium version.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: 3 iterations seem to work really well for low values of `q_tot`, but the error is larger at high values. If we print:
```
  q_tot = FT(0); ρ = FT(1)
  @show abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0)

  q_tot = FT(0.21); ρ = FT(0.1)
  @show abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) - 200.0)

  q_tot = FT(0.78); ρ = FT(1)
  @show abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0)
```

We get

## 3 iterations
```
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0) = 0.0
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) - 200.0) = 1.9237518245063256e-9
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0) = 113.0644848960498 # not converged
```

## 10 iterations

```
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0) = 0.0
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) - 200.0) = 1.9237518245063256e-9
abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0) = 0.004837695608841841 # not converged
```

This seems to be converging to the correct solution given enough iterations quickly at low `q_tot`. I'd suggest leaving `maxiter` and the tolerance as is for now, and we can search for optimal parameters and compare all of the methods systematically in another PR.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: (The equilibrium version is what you need here.)

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Will apply

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: Ok. `q_tot > 0.1` will never occur on Earth (and if such high `q_tot` occur on other planets, saturation adjustment is a bad idea anyway). I think it would be good to add an option of just allowing one iteration, and make that the default. For what we need in LES, that will do, and it avoids the performance problems arising from iterations.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Applying this change breaks several tests:
```
@test saturation_adjustment(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) ≈ 200.0
@test abs(saturation_adjustment_NewtonsMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) - 200.0) < tol_T
@test saturation_adjustment(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) ≈ 300.0
```
They end up getting domain errors (in `saturation_vapor_pressure`)

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Sounds good to me. I'll remove that test and reduce the iterations to 3.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Hmm, turns out that saturation vapor pressure call is breaking when `T_1` is passed to check whether we're at saturation or not (in `q_v_sat = q_vap_saturation(T_1, ρ)`): Inside `saturation_adjustment`, the inputs and resulting temperature are:
```
# saturation_adjustment
(e_int, PhasePartition(q_tot)) = (-143741.164702404, PhasePartition{Float64}(0.21, 0.0, 0.0))
air_temperature(e_int, PhasePartition(q_tot)) = -473.59643691525486
# q_v_sat = q_vap_saturation(T_1, ρ) # breaks due to saturation_vapor_pressure
```
So, temperatures can become _very_ negative with large negative `e_int` (which seem valid since they're computed from `internal_energy_sat` with reasonable looking values). What assumptions are there with the `internal_energy` / `saturation_vapor_pressure` equations?

tapios: The problem here is that for the given internal energy `e_int`, allocating all water to the vapor phase leads to a physically impossible temperature (it has to be ice). So this becomes a problem at saturation at very low temperatures, with unphysically high `q_t`. 

The physical way to deal with this is to have `T_min = eps` (or `sqrt(eps)`), or something else close to absolute zero. (This will never matter in practice, so let's make the bound as small as numerically possible.)

charleskawczynski:
Reopened in #554

PR #535, opened 2019-11-14
Hydrostatic 3D Ocean Model

blallen:
This is a clean branch containing the end results of the `bla/ocean3d`, which is very messy and very inconsistent with master. The only major changes wrt to that branch are in the handle of the secondary balance law and filters through `init_ode_params`. I hacked this in b128462 but we might switch to the interface @jkozdon proposed in 433f39d.

I've verified that the constant temperature case works as long as you don't use an upwind flux. I will be running more tests to make sure all of the features work in the next few days. @jkozdon do you have any ideas for using the `@test` macro to ensure the stability of this Model?


jkozdon:
Can you look at the max and min of a field to test? See https://github.com/climate-machine/CLIMA/commit/88530875cd700e21d34454716925ae4bc6f72904 for what I mean

jkozdon:
My hunch is that the upwind issue is the boundary state not being right for the upwind stuff we hacked in.

Did it work before with upwind? I checked my branch and it didn't work with upwind even before I merged.

blallen:
> Can you look at the max and min of a field to test? See [8853087](https://github.com/climate-machine/CLIMA/commit/88530875cd700e21d34454716925ae4bc6f72904) for what I mean

Yeah, I can do that. I will probably just add the divergence-free test to this branch and then think up a similar thing for the full model. When we run temperature forcing, we only force between 0 and 25, so testing that might be okay.

blallen:
> My hunch is that the upwind issue is the boundary state not being right for the upwind stuff we hacked in.
> 
> Did it work before with upwind? I checked my branch and it didn't work with upwind even before I merged.

We never tested the constant case with upwind before. But with actual gradients, it should be fine? I can also see if just filtering on temperature is enough without also upwinding.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
```diff
@@ -0,0 +1,203 @@
...
+
+  A.τ  = -τₒ * cos(y * 2π / Lʸ)
+  A.f  =  fₒ + β * y
+end
+
+function ocean_init_state!(p::HSBox, state, aux, coords, t)
+  @inbounds z = coords[3]
+  @inbounds H = p.H
+
+  state.u = @SVector [rand(),rand()]
```
blallen: @jkozdon , I get an "unsupported call through a literal pointer" error from this line when running on the GPU. 

I tried to get around it by doing something like `u = rand(); v = rand()` and then passing `u` and `v` through the state vector, but still got the error.

Any suggestions? Besides starting with [0,0] of course.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
```diff
@@ -0,0 +1,203 @@
...
+  norm(Q₀) = %.16e
+  ArrayType = %s""" eng0 ArrayType
+
+  solve!(Q, lsrk, nothing; timeend=timeend, callbacks=(cbinfo,))
+
+  maxQ =  Vars{vars_state(model, FT)}(maximum(Q, dims=(1,3)))
+  minQ =  Vars{vars_state(model, FT)}(minimum(Q, dims=(1,3)))
+
+  @test maxQ.θ ≈ minQ.θ
+  @test maxQ.θ - minQ.θ ≈ 0
```
blallen: @simonbyrne, @charleskawczynski, I get a "LoadError: scalar getindex is disallowed" error when running on the GPU. Everything worked fine when running on a CPU. Do you guys know why?

blallen:
I've updated this with the new `update_aux!` syntax from PR #543, so that the calls to `update_aux!` in the Atmos code are no longer broken. I also added the divergence-free test as a starting point for CI, but it fails on the GPU (see comments).

Once #543 is merged and the test is fixed, I think we can merge this in.

blallen:
bors try

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
```diff
@@ -0,0 +1,203 @@
...
+
+  model = HBModel{typeof(problem),FT}(problem, cʰ, cʰ, cᶻ, αᵀ, λʳ, νʰ, νᶻ, κʰ, κᶻ)
+
+  dg = OceanDGModel(model,
+                    grid,
+                    Rusanov(),
+                    CentralNumericalFluxDiffusive(),
+                    CentralGradPenalty())
+
+  Q = init_ode_state(dg, FT(0))
```
jkozdon:
```suggestion
  Q = init_ode_state(dg, FT(0); device=CPU())
```
In PR #483 @simonbyrne added the ability to initialize on the CPU. This should fix the `rand` issue [above](https://github.com/climate-machine/CLIMA/pull/535/files/246fe87ca6c9489e33f6c04775fa62f359a0a4ef#r348186217) I believe.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
jkozdon: See [comment below](https://github.com/climate-machine/CLIMA/pull/535/files/246fe87ca6c9489e33f6c04775fa62f359a0a4ef#r348190197) on CPU based initialization

test/Ocean/Hydrostatic_Boussinesq/simple_box.jl
```diff
@@ -0,0 +1,233 @@
...
+const Nˣ = 20
+const Nʸ = 20
+const Nᶻ = 20
+const Lˣ = 4e6
+const Lʸ = 4e6
+const H  = 1000
+
+@show xrange = range(FT(0);  length=Nˣ+1, stop=Lˣ)
+@show yrange = range(FT(0);  length=Nʸ+1, stop=Lʸ)
+@show zrange = range(FT(-H); length=Nᶻ+1, stop=0)
```
jkozdon: Remove `@show`?

test/Ocean/Hydrostatic_Boussinesq/simple_box.jl
```diff
@@ -0,0 +1,233 @@
...
+  end
+
+  brickrange = (xrange, yrange, zrange)
+  topl = StackedBrickTopology(mpicomm, brickrange;
+                              periodicity = (false, false, false),
+                              boundary = ((1, 1), (1, 1), (2, 3)))
+
+  dt = 240 # 6 # min( (Lˣ / cʰ) / Nˣ / N², H²/(1000 * κᶻ))
+  @show nout = ceil(Int64, tout / dt)
+  @show dt = tout / nout
```
jkozdon: remove `@show`?

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
```diff
@@ -0,0 +1,203 @@
...
+const Nˣ = 20
+const Nʸ = 20
+const Nᶻ = 20
+const Lˣ = 4e6
+const Lʸ = 4e6
+const H  = 1000
+
+@show xrange = range(FT(0);  length=Nˣ+1, stop=Lˣ)
+@show yrange = range(FT(0);  length=Nʸ+1, stop=Lʸ)
+@show zrange = range(FT(-H); length=Nᶻ+1, stop=0)
```
jkozdon: Remove `@show`?

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
```diff
@@ -0,0 +1,203 @@
...
+  else
+    ArrayType = Array
+  end
+
+  brickrange = (xrange, yrange, zrange)
+  topl = StackedBrickTopology(mpicomm, brickrange;
+                              periodicity = (false, false, false),
+                              boundary = ((1, 1), (1, 1), (2, 3)))
+  @show dt = 240 # (Lˣ / cʰ) / Nˣ / N²
+  @show nout = ceil(Int64, tout / dt)
```
jkozdon: Remove `@show`?

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
jkozdon: You'll need to add `use GPUifyLoops` to the top of the code as well.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
jkozdon: We may need a `mapreduce` implemented in `MPIStateArrays` for this to work on the GPU

blallen: Do the current tests run on CPU or GPU?

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
blallen: Thanks, added these.

test/Ocean/Hydrostatic_Boussinesq/simple_box.jl
blallen: Done.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
blallen: Done.

test/Ocean/Hydrostatic_Boussinesq/simple_box.jl
blallen: Done.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
blallen: Done.

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
jkozdon: Since you just added the test to [`test/Ocean/runtests.jl`] and not [`slurmci-test.toml`], I believe that this will just be run on the CPU.

I think it would be good to have it running on the GPU too if you can.

[`test/Ocean/runtests.jl`]: https://github.com/climate-machine/CLIMA/blob/4bc0b113570e3e089775b9dad19adb2e529fb80f/test/Ocean/runtests.jln
[`slurmci-test.toml`]: https://github.com/climate-machine/CLIMA/blob/4bc0b113570e3e089775b9dad19adb2e529fb80f/slurmci-test.toml

blallen: wait, how are the tests decided? Is it not just `test/runtests.jl`? 

Also, I would like this to run on the GPU as well, but am not sure what is needed to make that happen.

jkozdon: Azure runs the tests listed in [`test/runtests.jl`](https://github.com/climate-machine/CLIMA/blob/4bc0b113570e3e089775b9dad19adb2e529fb80f/test/runtests.jl) and this is CPU only. Since you added `test/Ocean/runtests.jl` there it will be run on Azure.

If you want test to run on the Caltech cluster (e.g., the runnings tests on the GPU) you need to add them to the [`slurmci-test.toml`] file as well in the relevant section with the correct options.

[`test/runtests.jl`]:https://github.com/climate-machine/CLIMA/blob/4bc0b113570e3e089775b9dad19adb2e529fb80f/test/runtests.jl
[`slurmci-test.toml`]: https://github.com/climate-machine/CLIMA/blob/4bc0b113570e3e089775b9dad19adb2e529fb80f/slurmci-test.toml

blallen: Okay, I'll keep that in mind for the future.

But still, I know the test will currently fail on the GPU. What exactly do we need to implement to fix that?

jkozdon: I think that you'll need to do the initialization on the CPU and someone needs to get `maximum` and `minimum` working with `MPIStateArrays` which are built on `CUArray`

simonbyrne: I've added support for this in #547.

simonbyrne:
Assuming tests don't tax the Azure Pipelines too much (i.e. they take only a minute or two), this is fine by me. If you want to run longer tests, please use the Slurm CI as that allows us to run things in parallel.

blallen:
> Assuming tests don't tax the Azure Pipelines too much (i.e. they take only a minute or two), this is fine by me. If you want to run longer tests, please use the Slurm CI as that allows us to run things in parallel.

Ah yeah, these had actual time-stepping going on. I'll move them to the Slurm CI.

blallen:
Now also dependent on PR #560

src/Ocean/Model/HydrostaticBoussinesqModel.jl
```diff
@@ -0,0 +1,515 @@
...
+
+  return nothing
+end
+
+@inline function diffusive!(m::HBModel, D::Vars, G::Grad, Q::Vars,
+                            A::Vars, t)
+  ν = Diagonal(@SVector [m.νʰ, m.νʰ, m.νᶻ])
+  D.ν∇u = ν * G.u
+
+  κ = Diagonal(@SVector [m.κʰ, m.κʰ, A.κᶻ])
```
jkozdon: If this scaling was moved to `flux_diffusive!` could we undo PR #543?

Since the computation of `κᶻ` in [`g!`] just needs `D.κ∇θ[3]` and this would be available in `flux_diffusive!` and known to be consistent with the current state.

Just a thought since #543 still makes me nervous since `diffstate` and `state` are not guaranteed to be consistent at this point (I just see some really nasty bugs arising from that).

[`g!`]: https://github.com/climate-machine/CLIMA/blob/22df165636739cae7ed064dcaa12635c9498b40e/src/Ocean/Model/HydrostaticBoussinesqModel.jl#L285-L295

blallen: Yes, it's easy to do that and then add an `update_aux_diffusive!` that uses the gradients calculated at the same time step.

I'll switch to just storing `∇θ` in the diffusive state and then save the diffusivity scaling for `flux_diffusive!`

blallen: Yes, it's easy to do that and then add an `update_aux_diffusive!` that uses the gradients calculated at the same time step.

I'll switch to just storing `∇θ` in the diffusive state and then save the diffusivity scaling for `flux_diffusive!`

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
blallen: Hi @simonbyrne, I've merged in #547 and #560... but am still getting a map reduce error [1].

Can you take a look at this next week? Thanks for all the help so far

[1] https://www.dropbox.com/s/ypr53qc2pujnngp/stacktrace.txt?dl=0

src/Ocean/Model/HydrostaticBoussinesqModel.jl
blallen: @jkozdon what's your plan for making this change? should we merge this branch as is and then fix it in a separate pull request that undoes #543 ? Or do you want to fix first and then merge this in?

jkozdon: Either way is fine with me. Whatever works best for you.

blallen: Okay, I'll handle this afterwards

test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl
blallen: Latest fixes from @simonbyrne work, we're all ready to go.

simonbyrne:
bors try

blallen:
bors try

PR #536, opened 2019-11-14
apply JuliaFormatter.jl

simonbyrne:
This applies [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl) to the whole repository.

Any thoughts/comments?

- [x] Testing?


jkozdon:
I like this idea a lot.

Suggestions:
- Can you post instructions on how to set up githooks to check this?
- Can we set up the CI to reject non-formatted code? (There is some github [actions](https://github.com/julia-actions/julia-format) for this, but I don't know much about these)
- Add something to README on local usage?
- Is there some style file? I didn't see anything in the commit to tell me how to do this on my commits or what options to set.

vchuravy:
I am a huge fan of this, but I would prefer an incremental approach where every new/change code has to go through the formateé. Otherwise we have a hard break in the got history.

On the other hand just doing it might be a lot less work.

simonbyrne:
I posted this commit just to see if people liked the output. If we're happy with how it looks, I'll add instructions, CI, docs, etc.

>     * Is there some style file? I didn't see anything in the commit to tell me how to do this on my commits or what options to set.

This was just run using the default settings (indent=4, margin=92). Unfortunately it doesn't appear to have a whole lot of other options.

> I am a huge fan of this, but I would prefer an incremental approach where every new/change code has to go through the formateé. Otherwise we have a hard break in the got history.
> 
> On the other hand just doing it might be a lot less work.

Yeah, not sure there is an obvious way to make it incremental. Also, since it changes the indentation from 2 to 4, it would be pretty hard to do incrementally.

blallen:
Do we need to switch the indentation from 2 to 4?

src/DGmethods/DGmodel.jl
```diff
@@ -4,243 +4,398 @@ struct HorizontalDirection <: Direction end
...
+                auxstate.data,
+                vgeo,
+                sgeo,
+                t,
+                vmapM,
+                vmapP,
+                elemtobndy,
+                topology.realelems,
+            )
+        )
```
blallen: My main reservations about the formatter is that it makes some of the function calls that used to fit on 2-3 lines take 20 lines. While this is cleaner, it also means you see less of the algorithm on your screen at once.

simonbyrne:
> Do we need to switch the indentation from 2 to 4?

No this is one of the few options the formatter provides, but it had been suggested by several people as a way to make the nested function calls more obvious. The cases where we have large numbers of nested loops such as
https://github.com/climate-machine/CLIMA/blob/4e72b21ddcc200c998dba964b9af6354ffeaaddc/src/DGmethods/DGmodel_kernels.jl#L99-L116
might be better handled with a macro similar to Oceaninigans.jl:
https://github.com/climate-machine/Oceananigans.jl/blob/866696432a5214ec5f927bf21835f072a8691c78/src/utils.jl#L13-L24

But again, feedback appreciated.

src/DGmethods/DGmodel.jl
simonbyrne: No, unfortunately. Personally I feel this is the main downside as well.

jkozdon:
> This was just run using the default settings (indent=4, margin=92).

Personal I like 2 spaces and 80 character limit, but much prefer uniformity than these little things.

src/DGmethods/DGmodel.jl
jkozdon: For particularly bad places there is always [skip formatting](https://github.com/domluna/JuliaFormatter.jl#skipping-formatting)

simonbyrne:
So the general consensus is that we should enable this. The main concerns:

1. This will break all incoming PRs, which will require a huge amount of work to get back into shape
 - @vchuravy will look to see if we can do this incrementally?

2. How should it be implemented:
 - There is a github action which applies it as a check: https://github.com/julia-actions/julia-format, which would require users to manually use the formatter.
 - Ideally it would be applied to PRs directly, but there doesn't seem to be a way to do this at the moment.


Plan: 
 - We will enable check for the time being, but don't enforce that it passes
 - If PRs make significant changes to a file, then they should apply the formatter to that file.

The remaining question is what formatting options should be applied? If you have opinions chime in below. The options are here: https://github.com/domluna/JuliaFormatter.jl

simonbyrne:
So the general consensus is that we should enable this. The main concerns:

1. This will break all incoming PRs, which will require a huge amount of work to get back into shape
 - @vchuravy will look to see if we can do this incrementally?

2. How should it be implemented:
 - There is a github action which applies it as a check: https://github.com/julia-actions/julia-format, which would require users to manually use the formatter.
 - Ideally it would be applied to PRs directly, but there doesn't seem to be a way to do this at the moment.


Plan: 
 - We will enable check for the time being, but don't enforce that it passes
 - If PRs make significant changes to a file, then they should apply the formatter to that file.

The remaining question is what formatting options should be applied? If you have opinions chime in below. The options are here: https://github.com/domluna/JuliaFormatter.jl

lcw:
It would also be nice to have a git hook that runs the formatter.  Below is what we used in a previous project as a `pre-commit` hook.

```sh
#!/bin/bash
#
# Called by git-commit with no arguments.  This checks to make
# sure that all .c and .h files are indented correctly before
# a commit is made.
#
# To enable this hook, make this file executable and place it
# in $GIT_DIR/hooks.

. git-sh-setup

CHFILES=$(git-diff-index -M --name-only --cached HEAD | \
          grep -E '.*\.(c|cpp|h)$' | grep -v 'node_data\.h')

for CHFILE in $CHFILES;
do
  MKTEMPLATE=`basename $CHFILE`.XXXXXXXX
  TEMPFILE=`mktemp -t "$MKTEMPLATE"` || exit 1
  clang-format $GIT_DIR/../$CHFILE > $TEMPFILE
  if diff $GIT_DIR/../$CHFILE $TEMPFILE
  then
    rm -f $TEMPFILE
  else
    rm -f $TEMPFILE
    NEEDS_FORMAT=1
    echo >&2 "$CHFILE needs to be indented with:"
    echo >&2 "   clang-format -i $GIT_DIR/../$CHFILE"
  fi
done
if [ -z "$NEEDS_FORMAT" ]
then
  exit 0
else
  exit 1
fi
```

lcw:
I also like having a formal style guide, such as https://github.com/jrevels/YASGuide, which JuliaFormatter.jl almost fully supports.

blallen:
> The remaining question is what formatting options should be applied? If you have opinions chime in below. The options are here: https://github.com/domluna/JuliaFormatter.jl

my opinions:
- `indent = 2`
- `margin = 80` (I like to split my screen into four vertical slices)
- `always_for_in = true`
- `whitespace_typedefs = true`
- `whitespace_ops_in_indices = true`
- `remove_extra_newlines = false`

simonbyrne:
I'm not too fussy: I generally use 4 space indent elsewhere, but our kernels have a lot of indentation which makes it cumbersome.

leios:
I agree with almost all the points set in the [Coding Conventions]( https://clima.caltech.edu/) section of the documentation. Switching to 80 columns (instead of 79) is a minor change. Outside of that, it looks like we can follow the YASguide with small exceptions for allowed unicode characters and using 2 spaces instead of 4.

The formatter will fix all the spacing, but we need to make sure we are consistent with function names and everything else as well, and that might need to be left to the reviewer(s).

I think I agree with @blallen's list, except that I would set `remove_extra_newlines = true`. I think this option only removes newlines when there are 2 or more consecutive newines.

Once we have this decided, I will update the Coding Conventions document to reflect the changes here.

kpamnany:
I vote for 4 spaces. The point of indentation is for a clear visual demarcation of blocks, which also means variable scope in many cases. Less than 4 spaces is not clear, IMO.

Apart from that, I'm good with everything else.

simonbyrne:
I've updated this PR to the current changes.

Once this is merged, the process to update an existing PR will look something like:
```
git checkout master .dev/format.jl # to get the most recent formatter script
.dev/format.jl
.dev/format.jl # need to apply twice
git add -u
git commit -m "apply formatter"
git merge master # this should be a relatively clean merge
```

To squash this all into a single commit, you should then be able to
```
git reset --soft master
git commit -m "my new commit message"
```

simonbyrne:
Please chime in if you have any objections!

simonbyrne:
I might try to do this over the weekend.

jkozdon:
> Some of the manual alignment (for example when aligning `=` signs) is now gone and comments are funky. JuliaFormatter will not allow us to do manual alignment anymore, but maybe it's worth fixing some of the comments while we are at it?

You can put stuff in blocks to turn off the formatter:
```julia
#! format: off

Stuff in here the formatter skips

#! format: on
```

leios:
Should we go through and figure out where all the manual formatting is preferred, set syntax off and on before merging? I guess we should at least rebase and apply the formatter again before merging this.

simonbyrne:
If someone wants to figure out how JuliaFormatter works, we could modify it to suit our own needs.

src/Atmos/Parameterizations/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -147,37 +164,37 @@ Computes exchange transfer coefficients:
...
-    crit = Q_b^3 - P_b^2
-    if crit < 0
-      T_b = cbrt(sqrt(-crit) + abs(P_b))                     # Eq. 34 in Ref. Byun1990
-      ζ = zfactor * (1/(3*γ_m)-(T_b + Q_b/T_b))              # Eq. 29 in Ref. Byun1990
+    logz = log(z_b / z_0)
+    zfactor = z_b / (z_b - z_0) * logz
+    s_b = Ri / Pr_0
+    if Ri > 0
+        temp = ((1 - 2 * β_h * Ri) - sqrt(1 + 4 * (β_h - β_m) * s_b))
+        ζ = zfactor / (2 * β_h * (β_m * Ri - 1)) * temp                    # Eq. 19 in Ref. Byun1990
```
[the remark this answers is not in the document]
charleskawczynski: This stuff will be gone when the new surface fluxes is merged (#556)

kpamnany:
Fixes #830.

PR #538, opened 2019-11-14
add state to DG model struct

blallen:
Simon and I had a discussion about making all MPIStateArrays members of the DGModel struct. The main goal with this is to ease the process of accessing various arrays in functions such as `nodal_update_aux!`.  

My PR for the OceanModel #535 is failing because I need to be able to access the `diffstate` values during my call of `nodal_update_aux!` so I modified the function call to pass these through. Unfortunately, this completely broke the functionality for everyone else because the arguments are different now. Instead of dispatching on arguments, we can now use a more generic `nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, t::Real)` and access the necessary MPIStateArrays within the function from the DGModel.

The code currently runs but fails the numeric tests. I have a feeling this is due to the duplication of `Q` and `dg.state` which should be the same thing but might not be exactly that.


jkozdon:
I don't really understand the purpose of this, so not sure what I am supposed to review.

I think that we want to have the ability to have state and model as functionally separate things since ODE solvers sometimes need multiple copies of state.

That's my $0.02 anyway.

src/DGmethods/DGmodel.jl
```diff
@@ -152,7 +154,7 @@ function init_ode_state(dg::DGModel, args...; device=arraytype(dg.grid) <: Array
       initstate!(bl, Val(dim), Val(polyorder), h_state.data, h_auxstate.data, h_vgeo,
           topology.realelems, args...))
     state .= h_state
-  end  
+  end
```
jkozdon:
```suggestion
```

jkozdon:
What I suggested in https://github.com/climate-machine/CLIMA/commit/433f39db7196fb7fe211a177516f50329a904f24 was to just add "params" to the DGmodel (not the full state).

jkozdon:
The way that I view things in the DGmodel is the data should either be static or should be generate it from the current (passed in) state and the data that is static (e.g, topology and physics). Diffstate and aux fall into the generatable data class.

simonbyrne:
I think it was mainly for encapsulation from within the DGModel object: it is odd to have same arrays contained in there, but not others.

> I think that we want to have the ability to have state and model as functionally separate things since ODE solvers sometimes need multiple copies of state.

Ah, that is a good point.

blallen:
Okay, we might not want to do that then. I'm going to take a pass at getting `nodal_update_aux!` to work with just the auxiliary and gradient/diffusive fields being called from the DGModel. If that works, we can close this.

PR #541, opened 2019-11-16
Check for reference state when constructing linear models

mwarusz:


jkozdon:
Is this ready to merge?

PR #543, opened 2019-11-18
Make arguments to update_aux! more general

blallen:
Our current call to `update_aux!` only allows one to use state and auxiliary variables during the function call. Since `auxstate` and `diffstate` are now members of the `DGModel` struct, I'd like to access them this way instead of being passed in as arguments. This would be very useful for the ocean model (PR #535) which needs to access the gradient variables during the `update_aux!` call. 

I'm not 100% happy with the current implementation, as it requires the function passed into `nodal_update_aux!` to always take all three arrays as arguments, but this is still fewer changes than passing the gradient variables as an argument in `update_aux!`. Please give suggestions if you can think of a better way to do this. This is also partially future-proofing for when we add an integral state to the `DGModel` struct. 

Also, why this is passing the tests that explicitly use `update_aux!` but not those that don't is a little obtuse to me. Help here would be much appreciated :)

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -184,18 +184,20 @@ function diffusive!(m::AtmosModel, diffusive::Vars, ∇transform::Grad, state::V
   diffusive!(m.turbulence, diffusive, ∇transform, state, aux, t, ρD_t)
 end
 
-function update_aux!(dg::DGModel, m::AtmosModel, Q::MPIStateArray,
-                     auxstate::MPIStateArray, t::Real)
+function update_aux!(dg::DGModel, m::AtmosModel, Q::MPIStateArray, t::Real)
   FT = eltype(Q)
+  auxstate = dg.auxstate
+
   if num_integrals(m, FT) > 0
     indefinite_stack_integral!(dg, m, Q, auxstate, t)
     reverse_indefinite_stack_integral!(dg, m, auxstate, t)
   end
 
-  nodal_update_aux!(atmos_nodal_update_aux!, dg, m, Q, auxstate, t)
+  nodal_update_aux!(atmos_nodal_update_aux!, dg, m, Q, t)
 end
 
-function atmos_nodal_update_aux!(m::AtmosModel, state::Vars, aux::Vars, t::Real)
+function atmos_nodal_update_aux!(m::AtmosModel, state::Vars, aux::Vars,
+                                 diff::Vars, t::Real)
   atmos_nodal_update_aux!(m.moisture, m, state, aux, t)
   atmos_nodal_update_aux!(m.radiation, m, state, aux, t)
   atmos_nodal_update_aux!(m.turbulence, m, state, aux, t)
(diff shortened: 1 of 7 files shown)


jkozdon:
This interface is fine with me.

jkozdon:
Actually, [diffstate] is set after [update aux] is called. Do we need two update aux function a pre and postdiffstate function calls?

[diffstate]: https://github.com/climate-machine/CLIMA/blob/9b1af81af445bcebc3025e4df354298c94ec40cd/src/DGmethods/DGmodel.jl#L70-L89
[update aux]: https://github.com/climate-machine/CLIMA/blob/9b1af81af445bcebc3025e4df354298c94ec40cd/src/DGmethods/DGmodel.jl#L58-L60

simonbyrne:
> Actually, [diffstate](https://github.com/climate-machine/CLIMA/blob/9b1af81af445bcebc3025e4df354298c94ec40cd/src/DGmethods/DGmodel.jl#L70-L89) is set after [update aux](https://github.com/climate-machine/CLIMA/blob/9b1af81af445bcebc3025e4df354298c94ec40cd/src/DGmethods/DGmodel.jl#L58-L60) is called. Do we need two update aux function a pre and postdiffstate function calls?

This is something @blallen and I discussed earlier: I think this would be a good idea.

blallen:
So I actually need this to happen before the diffusive variables are called (at least with the current set up).

Basically, the algorithm is if `∂θ∂z < 0` then `aux.κᶻ = 1000 * model.κᶻ` at that grid point. Our diffusive variable is `κ∇θ` and needs the updated `κᶻ` to be computed properly. There are some ways around this, but they basically mean ignoring the second gradient transform. Currently, we just take `∂θ∂z` from the previous timestep and it's good enough. 

Maybe @christophernhill or @jm-c can comment on further.

jkozdon:
> So I actually need this to happen before the diffusive variables are called (at least with the current set up).
> 
> Basically, the algorithm is if `∂θ∂z < 0` then `aux.κᶻ = 1000 * model.κᶻ` at that grid point. Our diffusive variable is `κ∇θ` and needs the updated `κᶻ` to be computed properly. There are some ways around this, but they basically mean ignoring the second gradient transform. Currently, we just take `∂θ∂z` from the previous timestep and it's good enough.

This makes sense.

The only thing to be careful of is whether `diffstate` happens to get changed between when you need it and when it was calculated. Not sure if this could/would/will happen with multirate time stepping. It should be safe if every rate builds its own DGmodel with an independent `diffstate`. But something to keep in mind down the road if problems do arise.

simonbyrne:
> The only thing to be careful of is whether `diffstate` happens to get changed between when you need it and when it was calculated.

Another reason to avoid relying of `diffstate` between iterations is that we then don't have to store it in a restart file.

blallen:
> > The only thing to be careful of is whether `diffstate` happens to get changed between when you need it and when it was calculated.
> 
> Another reason to avoid relying of `diffstate` between iterations is that we then don't have to store it in a restart file.

So this can be done, just it means doing the matrix multiplication for `κ∇θ` during `flux_diffusive!` instead of `diffusive!` and just storing `∇θ` in the `diffstate`.  I don't know if there's any efficiency difference between the two versions, but it's an easy change to make.

simonbyrne:
> So this can be done, just it means doing the matrix multiplication for `κ∇θ` during `flux_diffusive!` instead of `diffusive!` and just storing `∇θ` in the `diffstate`. I don't know if there's any efficiency difference between the two versions, but it's an easy change to make.

PR #544, opened 2019-11-19
Remove LazyArrays

simonbyrne:
All the uses of LazyArrays could be expressed as mapreduce operations, which makes the code somewhat cleaner, and avoids a CuArrays issue (https://github.com/JuliaGPU/CuArrays.jl/issues/495).


simonbyrne:
Ah, it's because the weights need to be broadcasted. Nevermind.

vchuravy:
> Ah, it's because the weights need to be broadcasted. Nevermind.

`Ref(weigths)`?

simonbyrne:
Unfortunately no, the weights are an `nnodes` x 1 x `nelems` array, and we need to broadcast them to `nnodes` x `nstates` x `nelems`.

PR #545, opened 2019-11-19
Metrics for integrating on slabs

slifer50:
@jkozdon @smarras79 Metrics for integrating on slabs. Continued from #504. Horizontal_integral_test.jl has all the changes requested including convergence test and 2D integral.


slifer50:
@jkozdon Let me know if there are any more changes that are needed for this to be merged

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -20,6 +20,7 @@ const _ξ1x3, _ξ2x3, _ξ3x3 = Grids._ξ1x3, Grids._ξ2x3, Grids._ξ3x3
 const _M, _MI = Grids._M, Grids._MI
 const _x1, _x2, _x3 = Grids._x1, Grids._x2, Grids._x3
 const _JcV = Grids._JcV
+const _MH = Grids._MH
```
jkozdon:
```suggestion
```

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,313 @@
...
+  base_Nhorz = 4
+  base_Nvert = 2
+  Rinner = 1 // 2  
+  Router = 1
+  expected_result = [-4.5894269717905445e-8 -1.1473566985387151e-8;
+		     -2.0621904184281448e-10  -5.155431637149377e-11;
+	             -8.72191208145523e-13 -2.1715962361668062e-13] 
+  for l in 1:3
+      Nhorz = 2^(l-1) * base_Nhorz
+      Nvert = 2^(l-1) * base_Nvert
```
jkozdon: If you do a fixed number of vertical levels, you could check the error in all the slabs of the mesh (not just the inner and outer)

test/DGmethods/Horizontal_integral_test.jl
```diff
@@ -0,0 +1,313 @@
...
+      Stot = zeros(Nqk)
+      S1tot = zeros(Nqk)
+      Err = 0
+      
+      for k in 1:Nqk
+        Stot[k] = MPI.Reduce(S[k], +, 0, MPI.COMM_WORLD)
+        S1tot[k] = MPI.Reduce(S1[k], +, 0, MPI.COMM_WORLD)
+        Err += (0.5 - Stot[k] / S1tot[k])^2
+      end
+      Err = sqrt(Err / Nqk)
```
jkozdon: I still don't know why this is necessary. Why not just check the error of each slab?

slifer50: @jkozdon sure we could do that instead.

src/DGmethods/DGmodel_kernels.jl
slifer50: @jkozdon The main purpose of the PR is the have the slab metrics available and to test that they are correct. The main application we have for these currently is in the diagnostics, where I've had good results with using these metrics on a personal branch pending this gets merged. 

I'll checkout a fresh Manifest.

smarras79:
@jkozdon @slifer50 the integration used to calculate the diagnostics in Yassine's function is working for what we need it. We need this tool available to all even if it is not general yet. 
The code is getting larger and larger and different users are now using different versions of the diagnostics because this PR is still open.
How can we speed up the merging of the 2D integrals which we know are already working correctly?

jkozdon:
> How can we speed up the merging of the 2D integrals which we know are already working correctly?

I have no opinion on whether this should be merged or not. Just trying to help make sure the math seems right.

I don't seen any obvious problems, but like I've highlighted above and in the previous PR some better test could be added. For instance, tests that would not pass for any odd function integrated with a symmetric quadrature rule.

I'll leave it to @simonbyrne to merge or not.

smarras79:
> > How can we speed up the merging of the 2D integrals which we know are already working correctly?
> 
> I have no opinion on whether this should be merged or not. Just trying to help make sure the math seems right.
> 
> I don't seen any obvious problems, but like I've highlighted above and in the previous PR some better test could be added. For instance, tests that would not pass for any odd function integrated with a symmetric quadrature rule.
> 
> I'll leave it to @simonbyrne to merge or not.

Thanks @jkozdon. Do I understand well that now we are still missing a test that should be added as a control test for the CI but that is not in yet?

jkozdon:
> Thanks @jkozdon. Do I understand well that now we are still missing a test that should be added as a control test for the CI but that is not in yet?

There is a test. I was just suggesting a better test. 3-D I believe is tested due to the use of the sphere, 2-D could use a similar non-exact test (e.g, a field that cannot be integrated exactly with the quadrature rule but shows the right convergence rate under mesh refinement).

slifer50:
bors try

slifer50:
bors try

slifer50:
@simonbyrne, please review this PR when you get chance, I think it's at the point where we can safely merge it.

Issue #410, opened 2019-08-29
Unitful support


simonbyrne:
It would be really interesting to see if we could use [Unitful.jl](https://github.com/PainterQubits/Unitful.jl) quantities in the models.

The usual problem with trying to use them is that at some point you end up with an array of heterogeneous types, which is bad for the compiler. However we should be able to avoid this by using VariableTemplates (since that gives us a heterogeneous view into a homogeneous array).

This should be possible with a few small modifications to VariableTemplates, and making sure that the correct units are used for sources, fluxes/gradients, and integrals (i.e. sources should state/time, fluxes/gradients should be state/length, integrals should be state*length).

rafaqz:
This would be very interesting as a demonstration of Unitful.jl at this scale. 

I'm not sure if this fits your architecture but to get around the problem or parametrizing heterogenous unitful _parameters_ in generic optimisers (for ecophys models) I've been using Flatten.jl to convert between vectors of `Float64/32` and arbitrary, often nested parameter structs or tuples. This solves the performance issue for optimisation.

PR #546, opened 2019-11-20
fix subsidence definition

simonbyrne:
Clearly we need #410!

### DIFF
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -25,7 +25,8 @@ end
 struct Subsidence <: Source
 end
 function atmos_source!(::Subsidence, m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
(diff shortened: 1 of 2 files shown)


src/Atmos/Model/source.jl
```diff
@@ -25,7 +25,8 @@ end
 struct Subsidence <: Source
 end
 function atmos_source!(::Subsidence, m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
-  source.ρu -= state.ρ * m.radiation.D_subsidence
+  n = aux.orientation.∇Φ ./ norm(aux.orientation.∇Φ)
+  source.ρu -= m.radiation.D_subsidence * dot(state.ρ, n) * n
```
akshaysridhar:
```suggestion
  source.ρu -= m.radiation.D_subsidence * dot(state.ρu, n) * n
```

akshaysridhar: good catch @simonbyrne .

PR #549, opened 2019-11-21
Squal Line

charleskawczynski:
Adds Squal line to new framework.


charleskawczynski:
This has gotten very stale, and #601 is farther along.

PR #551, opened 2019-11-21
MT bounds

charleskawczynski:
**This is not a real PR.**

Adds tests for MoistThermodynamics bounds

### DIFF
--- a/docs/src/GettingStarted/RunningClimateMachine.md
+++ b/docs/src/GettingStarted/RunningClimateMachine.md
@@ -143,3 +143,11 @@ Note that unless GPU use is disabled as above, each `ClimateMachine`
 process will use GPU acceleration. If there are insufficient GPUs (four
 in the example above), the `ClimateMachine` processes will share the
 GPU resources available.
+
+## Scripts for end-to-end runs, logging and visualization
+
+The `ClimateMachine` [wiki](https://github.com/CliMA/ClimateMachine.jl/wiki)
+contains detailed examples of [SLURM
+scripts](https://github.com/CliMA/ClimateMachine.jl/wiki/Bash-Run-Scripts)
+that run the `ClimateMachine`, record specified performance metrics and
+produce basic visualization output. 


tapios:
Maybe a better way to summarize error statistics is to compute the temperature error `\Delta T` after saturation adjustment and summarize `rms(\Delta T)`, `max(|\Delta T|)` etc., and plots `\Delta T` vs `T`. If the errors are near freezing, the temperature errors may be small, just not quite small enough to satisfy the convergence criteria.

charleskawczynski:
> Maybe a better way to summarize error statistics is to compute the temperature error `\Delta T` after saturation adjustment and summarize `rms(\Delta T)`, `max(|\Delta T|)` etc., and plots `\Delta T` vs `T`. If the errors are near freezing, the temperature errors may be small, just not quite small enough to satisfy the convergence criteria.

I like this idea, but how can we compute the error if we can't converge to the correct root (`T_correct` in `\Delta T = T_sat_adjust - T_correct`)?

tapios:
> > Maybe a better way to summarize error statistics is to compute the temperature error `\Delta T` after saturation adjustment and summarize `rms(\Delta T)`, `max(|\Delta T|)` etc., and plots `\Delta T` vs `T`. If the errors are near freezing, the temperature errors may be small, just not quite small enough to satisfy the convergence criteria.
> 
> I like this idea, but how can we compute the error if we can't converge to the correct root (`T_correct` in `\Delta T = T_sat_adjust - T_correct`)?

I had in min just using the value `T_sat_adjust` that you get from saturation adjustment after `maxiter` has been reached (even if convergence criteria are not yet satisfied).

test/Common/MoistThermodynamics/thermo_state_iter_sensitivity.jl
```diff
@@ -0,0 +1,95 @@
...
+  dir = joinpath("output","MoistThermoAnalysis")
+
+  for method in (PhaseEquil_NewtonMethod, PhaseEquil)
+    for (p, maxiter) in enumerate(maxiter_range)
+    for (i,RH) in enumerate(RH_range)
+    for (j,ρ) in enumerate(ρ_range)
+    for (k,T) in enumerate(T_range)
+      # q_pt = PhasePartition(q_tot,FT(0),q_tot)s
+      q_sat = q_vap_saturation(T, ρ)
+      q_tot = RH*q_sat
```
tapios:
```suggestion
      q_tot = min(RH*q_sat, 1)
```

charleskawczynski:
This will be moved to [MoistThermodynamics.jl](https://github.com/climate-machine/MoistThermodynamics.jl) once PlanetParameters are replaced with CLIMAParameters.

PR #552, opened 2019-11-21
New MT constructor

charleskawczynski:
Adds a MoistThermo constructor `LiquidIcePotTempSHumEquil_given_pressure`.


charleskawczynski:
Ok, I played the same substitution tricks we did last time to make a `LiquidIcePotTempSHumEquil_given_pressure`, which uses `saturation_adjustment_q_tot_θ_liq_ice_given_pressure` (a mirrored version of `saturation_adjustment_q_tot_θ_liq_ice`). This passes the consistency test for pressure within 1%, and the other variables pass with `isapprox`, so I think this should suffice. cc @tapios 

@smarras79 Please try this for Dycoms and see if it fixes the ICs!

PR #554, opened 2019-11-21
[Part 2] New sat adjust

charleskawczynski:
Adds saturation adjustment using Newton method with analytic gradients.

This is not currently passing the correctness tests, which are passing using Secant method (although it's _almost_ passing).

A bit more of experimentation needs to be done, it could be that the Newton method is more accurate early on. To be more systematic, we should add a test suite for an agreed range of values and test all of the methods to compare performance/convergence.


src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,49 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
 PhasePartition(ts::PhaseEquil) = PhasePartition_equil(air_temperature(ts), air_density(ts), ts.q_tot)
 PhasePartition(ts::PhaseNonEquil) = ts.q
 
+function roots_∂e_int_∂T(T::FT, e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
```
tapios: I'd just call this function ``∂e_int_sat_∂T`` (because it is just the temperature derivative of saturation internal energy).

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -737,6 +738,49 @@ PhasePartition(ts::PhaseDry{FT}) where {FT<:Real} = PhasePartition(FT(0), FT(0),
...
+
+using Newtons method with analytic gradients.
+
+See also [`saturation_adjustment`](@ref).
+"""
+function saturation_adjustment_NewtonsMethod(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real}
+  T_1 = max(FT(T_min), air_temperature(e_int, PhasePartition(q_tot))) # Assume all vapor
+  q_v_sat = q_vap_saturation(T_1, ρ)
+  unsaturated = q_tot <= q_v_sat
+  if unsaturated
```
tapios:
```suggestion
  if unsaturated & T_1 > T_min
```

PR #556, opened 2019-11-21
Improve surface fluxes

charleskawczynski:
This PR:
 - Removed the Byun functions, since we'll want to use a discretization-independent model (as is done in the Nishizawa paper)
 - Combines the functions into a single function call
 - Now offers to pass the potential temperature flux as an optional argument. If not given, the surface fluxes are computed as is done in the Nishizawa paper. Otherwise, the friction velocity and temperature scale are computed in a decoupled system of equations.
 - Adjusted the docs accordingly
 - Removes use of PlanetParameters and uses CLIMAParameters

More tests would be great to add, but I'm not sure how to do that at the moment.


test/Atmos/Parameterizations/SurfaceFluxes/runtests.jl
```diff
@@ -12,69 +10,22 @@ using CLIMA.RootSolvers
...
+  x_s   = [0.0, 300.0]
+
+  Δz = 2.0
+  z = 0.5
+  θ_bar = 300.0
+  a = 4.7
+  pottemp_flux_given = -200.0
+  args = x_initial, x_ave, x_s, z_0, F_exchange, dimensionless_number, θ_bar, Δz, z, a, pottemp_flux_given
+  sfc = surface_conditions(args[1:end-1]...)
+  @show sfc
```
simonbyrne: can you get rid of the `@show`?

test/Atmos/Parameterizations/SurfaceFluxes/runtests.jl
```diff
@@ -12,69 +10,22 @@ using CLIMA.RootSolvers
...
+  sfc = surface_conditions(args...)
+  @show sfc
```
simonbyrne: Same here

simonbyrne:
Needs a rebase

charleskawczynski:
> Needs a rebase or merge + squash

@simonbyrne done.

charleskawczynski:
@simonbyrne Can we merge this after resolving the packages? I'd rather replace CLIMA parameters for surface fluxes after this PR is merged than before. The GPU issue with NLsolve [may be resolved soon](https://github.com/JuliaNLSolvers/NLsolve.jl/issues/234#issuecomment-605536995)

simonbyrne:
JuliaDiffEq includes its own nlsolver
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/nlsolve/type.jl#L11

charleskawczynski:
The surface fluxes module will likely need more iterations but, since @akshaysridhar is working on tests for the current interface, let's merge this and we can work on the next iteration once better tests are hooked in.

PR #557, opened 2019-11-21
Root solver verbose

charleskawczynski:
@simonbyrne This is an alternative to #548. Actually, I think this is preferable to #548, since the history is nice to have, and this requires touching more of the algorithm. This PR does not use dispatch, instead, a `RootSolution` struct is returned, and [VERBOSE_RESULTS](https://github.com/climate-machine/CLIMA/compare/RootSolverVerbose?expand=1#diff-eee2b231487150c96b54a6c41597866fR24) is used to control whether this includes more or less information.


charleskawczynski:
Documentation seems to be breaking. Fixing in #558

charleskawczynski:
Actually, the break seemed related to `MbedTLS`, trying to update toml in #559

charleskawczynski:
Closing this in favor of #548, since #548 can be tested for both cases, whereas this approach cannot.

PR #560, opened 2019-11-21
change mechanism for init on CPU

simonbyrne:
This wasn't tested, and appears to have broken after the recent MPIStateArray changes.


blallen:
This solved the issues I was having yesterday. Thanks @simonbyrne!

PR #561, opened 2019-11-21
Add acoustic wave test on the sphere using 1d IMEX

mwarusz:


mwarusz:
bors try

src/Atmos/Model/ref_state.jl
```diff
@@ -33,9 +34,12 @@ A hydrostatic state specified by a temperature profile and relative humidity.
 struct HydrostaticState{P,F} <: ReferenceState
   temperatureprofile::P
   relativehumidity::F
+  assume_pde_level_balance::Bool
```
simonbyrne: Thanks: can you describe what this option does?

mwarusz: I added a doc string.

simonbyrne: I still don't quite understand, and the comment above "assumes that the *reference state* is in hydrostatic balance" also is confusing (since it is by definition in hydrostatic balance)?

simonbyrne: Are you effectively expressing pressure as a perturbation from reference?

mwarusz: Pressure ends up being a perturbation from reference, but at the same time gravity includes only density perturbation. This means that the hydrostatic assumption was used directly at the pde level, and not simply by creating a reference state that satisfies it to some precision. As for the above comment, I don't think that arbitrary `ReferenceState` has to be in hydrostatic balance, only those created using `HydrostaticState`. Unless we make that assumption ? The main point of this comment was to stress that we assume this balance only for the reference state, i.e. the model as a whole is is not hydrostatic.

jkozdon:
Is this ready to merge?

simonbyrne:
I'm not super keen on adding `assume_pde_level_hydrostatic_balance` to the AtmosModel code: it adds a lot complexity to deal with this one case. Is there any way we can do without it?

src/ODESolvers/AdditiveRungeKuttaMethod.jl
```diff
@@ -280,14 +291,15 @@ Giraldo, Kelly and Constantinescu (2013).
 function ARK2GiraldoKellyConstantinescu(F, L,
                                         linearsolver::AbstractLinearSolver,
                                         Q::AT; dt=nothing, t0=0,
-                                        split_nonlinear_linear=false) where {AT<:AbstractArray}
+                                        split_nonlinear_linear=false,
+                                        version=ARK2PaperVersion()) where {AT<:AbstractArray}
```
simonbyrne: I don't like this version thing: is one of these strictly superior? If not, can we just make it a boolean option (i.e. `paperversion=true`?

mwarusz: I made it a boolean option. I would like to do more tests before declaring any of them superior. Picking the paper version as the default for now, since otherwise I would have to update test reference data.

mwarusz:
I could remove `assume_pde_level_hydrostatic_balance` since this test works without it when using `Float64` (but when using `Float32` it blows up).  However, this issue is not related to this specific test case. For example, I see the same thing in the gravity wave test case that I have been working on (`Float64` works but `Float32` blows up without assume `pde_level_hydrostatic_balance`). The key thing is that this option prevents large cancellation errors that otherwise occur since perturbations around hydrostatic balance are small. This option is used by default in NUMA and it would be useful just for comparison/debugging purposes.

akshaysridhar:
https://gist.github.com/akshaysridhar/335b45be6f2b79e19e419aeddf3b32e6 
I've tried the LES case in Float32 mode (see summary above) - it is interesting to see situational cancellation issues.

src/Atmos/Model/ref_state.jl
tapios: @mwarusz I'm also confused here. The reference state is exactly hydrostatic, by construction, so I do not see why we need `assume_pde_level_balance` in `ref_state.jl`. It seems that what you are doing when `assume_pde_level_balance` is true is to factor out the hydrostatic reference state from the model state in the momentum equation, and then work in the perturbation state. You can do that, but it should be an option at the level of the model, not of the reference state.

test/DGmethods/Euler/acousticwave-1d-imex.jl
```diff
@@ -0,0 +1,231 @@
...
+
+  logger_stream = MPI.Comm_rank(mpicomm) == 0 ? stderr : devnull
+  global_logger(ConsoleLogger(logger_stream, loglevel))
+
+  polynomialorder = 5
+  numelem_horz = 10
+  numelem_vert = 5
+
+  timeend = 60 * 60
+  # timeend = 33 * 60 * 60 # Full simulation
```
simonbyrne: What is the criteria for success in the full simulation?

mwarusz: The first criterion is lack of `NaN`s :) The second one is visual inspection, you should see a wave that propagates similarly to Fig 3. in https://doi.org/10.1016/j.fluiddyn.2004.03.003.

mwarusz:
I removed `assume_pde_level_hydrostatic_balance` from this PR so that it can be merged sooner.

Refactored code for this option (based on top of this branch) is now a separate PR https://github.com/climate-machine/CLIMA/pull/584

Additionally I made the new coefficients of  `ARK2GiraldoKellyConstantinescu` the default but I decided to keep `paperversion` as an option. I have seen the new version enable much larger stable time steps (up to 4x) but it comes with a cost in accuracy. Specifically, the isentropic vortex convergence tests do not show the expected convergence rate with the new version (which means that time truncation errors dominate).

mwarusz:
bors try

simonbyrne:
Thanks @mwarusz, it would be good to figure out the cause of the cancellation issues in Float32.

PR #574, opened 2019-11-26
Fix `liquid_ice_pottemp_given_pressure`

charleskawczynski:
After looking through equations for the Dycoms case with @smarras79, we believe a bug may exist in computing `θ_liq_ice` in `liquid_ice_pottemp_given_pressure`. @smarras79 has a reference that indicates that the dry `c_p` should be used in place of the moist `c_p`. This change nearly fixes the `q_liq` profile in the Dycoms initialization (and may also have an impact on dynamics). I'd like confirmation that this change is accurate @tapios.

References:
 - Cotton, Bryan, van Den Heever   Storm an cloud dynamics (book)
 - Betts (1973) Non-precipitating cumulus convection and its parameterization QJRMS 99:178-196

### DIFF
--- a/src/Common/MoistThermodynamics/MoistThermodynamics.jl
+++ b/src/Common/MoistThermodynamics/MoistThermodynamics.jl
@@ -840,7 +840,7 @@ function liquid_ice_pottemp_given_pressure(T::FT, p::FT,
   q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
     # liquid-ice potential temperature, approximating latent heats
     # of phase transitions as constants
-    return dry_pottemp_given_pressure(T, p, q) * (1 - latent_heat_liq_ice(q)/(cp_m(q)*T))
+    return dry_pottemp_given_pressure(T, p, q) * (1 - latent_heat_liq_ice(q)/(FT(cp_d)*T))
 end
 
 


tapios:
It isn't a bug, just an alternative choice of approximation (which is less accurate than what we are using). Given that the liquid-ice potential temperature is merely a diagnostic for us, feel free to change this to dry `cp` consistently. If you do that, you will also need to change all `R_m` and `cv_m` in the inverse functions `air_temperature_from...` to their dry counterparts.

tapios:
Specifically, you'll also need to change lines 826-828 to
`R_m = gas_constant_air(q)
  return θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_d/cv_d) + latent_heat_liq_ice(q)/cv_d`
This is a little strange, because you'd use the moist air gas constant in the ideal gas law for pressure, but the dry-air coefficients elsewhere. But it is consistent.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -840,7 +840,7 @@ function liquid_ice_pottemp_given_pressure(T::FT, p::FT,
   q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
     # liquid-ice potential temperature, approximating latent heats
     # of phase transitions as constants
-    return dry_pottemp_given_pressure(T, p, q) * (1 - latent_heat_liq_ice(q)/(cp_m(q)*T))
+    return dry_pottemp_given_pressure(T, p, q) * (1 - latent_heat_liq_ice(q)/(FT(cp_d)*T))
```
tapios: To make this consistent, you will need to compute `dry_pottemp_given_pressure(T, p, q)` with zero specific humidity `q`. You will also make analogous changes in all other computations involving the liquid-ice potential temperature.

charleskawczynski:
> It isn't a bug, just an alternative choice of approximation (which is less accurate than what we are using).

Ok, good to know.

> Given that the liquid-ice potential temperature is merely a diagnostic for us, feel free to change this to dry cp consistently.

Ok, I can try changing this consistently, but if using `cp_m` is more accurate then I wonder if this was the best way to fix Dycoms ICs after all.

tapios:
I wouldn’t worry too much about DYCOMS initial conditions. Adjusting q_t to get the desired q_l, given our thermodynamics formulations, is just fine. In any case, the problems we are having with DYCOMS lie elsewhere (not sure yet where).

On Nov 26, 2019, at 6:55 PM, Charles Kawczynski <notifications@github.com<mailto:notifications@github.com>> wrote:


It isn't a bug, just an alternative choice of approximation (which is less accurate than what we are using).

Ok, good to know.

Given that the liquid-ice potential temperature is merely a diagnostic for us, feel free to change this to dry cp consistently.

Ok, I can try changing this consistently, but if using cp_m is more accurate then I wonder if this was the best way to fix Dycoms ICs after all.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
szy21: If we use cp_m here, I wonder if we should use LH_v and LH_s (temperature dependent) instead of LH_v0 and LH_s0 in latent_heat_liq_ice to make it consistent. (This will probably affect the relationship between q_t and q_l the same way as changing cp_m to cp_d).

tapios: Good point—just makes it harder to invert for temperature, which is why we don’t have the temperature dependent latent heats.

tapios: I updated the design docs with constant thermodynamic coefficients for the liquid-ice modification of the potential temperature. If you want to change things, please do as stated there. I left the dry potential temperature with \kappa = R_m/c_{pm}. For the inversion given density, this makes an additional approximation necessary, but it still holds to first order. However, none of this will make much of a difference for DYCOMS--the problems there lie elsewhere. Another option is simply to leave `c_{pm}`, `c_{vm}` etc. as we had it. Either way is fine and will not matter much for what we are interested in.

charleskawczynski: Ok, if this is not a high priority then I'll leave things as is for now, but I'll keep the PR open for reference.

charleskawczynski:
Based on recent discussions, it sounds like we can close this PR, correct? cc: @tapios @smarras79

tapios:
> Based on recent discussions, it sounds like we can close this PR, correct? cc: @tapios @smarras79

Yes, this is obsolete now.

Issue #573, opened 2019-11-26
Add function to provide minimum grid spacing needed to compute Courant numbers 


lcw:
We need a function to provide the minimum grid spacing entering Courant number calculations, separately for the horizontal and vertical directions. So something like min(mesh-size)/N^2 (with some empirical pre-factors), with separate mesh sizes in the horizontal and vertical, will do for now. (We can refine this later if we get better estimates of the effective \Delta x, \Delta z).

PR #575, opened 2019-11-29
Add minimum node distance function

lcw:
This adds a function that approximates the minimum node distance for a given grid in the horizontal and / or vertical directions of the reference element.  In addition, the way direction is specified for the filter is changed to match that of the DGmodel.  This addressed issue #573.

### DIFF
--- a/examples/Atmos/heldsuarez.jl
+++ b/examples/Atmos/heldsuarez.jl
@@ -1,6 +1,6 @@
 using CLIMA
 using CLIMA.Mesh.Topologies: StackedCubedSphereTopology, cubedshellwarp, grid1d
-using CLIMA.Mesh.Grids: DiscontinuousSpectralElementGrid
+using CLIMA.Mesh.Grids
 using CLIMA.Mesh.Filters
 using CLIMA.DGmethods: DGModel, init_ode_state
 using CLIMA.DGmethods.NumericalFluxes: Rusanov, CentralGradPenalty,
@@ -80,8 +80,8 @@ function run(mpicomm, polynomialorder, numelem_horz, numelem_vert,
   # determine the time step
   element_size = (setup.domain_height / numelem_vert)
   acoustic_speed = soundspeed_air(FT(315))
-  lucas_magic_factor = 14
-  dt = lucas_magic_factor * element_size / acoustic_speed / polynomialorder ^ 2
+  lucas_magic_factor = 5
+  dt = lucas_magic_factor * min_node_distance(grid) / acoustic_speed
 
   Q = init_ode_state(dg, FT(0))
   lsrk = LSRK144NiegemannDiehlBusch(dg, Q; dt = dt, t0 = 0)
--- a/src/Mesh/Filters.jl
+++ b/src/Mesh/Filters.jl
@@ -2,6 +2,7 @@ module Filters
 
 using LinearAlgebra, GaussQuadrature, GPUifyLoops
 using ..Grids
+using ..Grids: Direction, EveryDirection
 
 export AbstractSpectralFilter, AbstractFilter
 export ExponentialFilter, CutoffFilter, TMARFilter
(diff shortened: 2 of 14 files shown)


lcw:
bors try

src/Mesh/Filters.jl
```diff
@@ -129,19 +130,19 @@ end
...
+           direction::Direction = EveryDirection())
 
 Applies `filter` to the `states` of `Q`.
 
-The arguments `horizontal` and `vertical` are used to control if the filter
-is applied in the horizontal and vertical reference directions, respectively.
-Note, it is assumed that the trailing dimension is the vertical dimension and
-the rest are horizontal.
+The `direction` arguement controls if the filter is applied in the horizontal
+and vertical reference directions, respectively.  Note, it is assumed that the
```
tapios:
```suggestion
The `direction` argument controls if the filter is applied in the horizontal
and/or vertical directions.  It is assumed that the
```

src/Mesh/Grids.jl
```diff
@@ -173,6 +190,37 @@ function referencepoints(::DiscontinuousSpectralElementGrid{T, dim, N}
   ξ
 end
 
+"""
+    minnodedistance(::DiscontinuousSpectralElementGrid,
```
tapios: Something like `min_node_distance` would be easier to read.

src/Mesh/Grids.jl
```diff
@@ -173,6 +190,37 @@ function referencepoints(::DiscontinuousSpectralElementGrid{T, dim, N}
...
+                         direction::Direction=EveryDirection()
+                        ) where {T, dim, N}
+  topology = grid.topology
+  nrealelem = length(topology.realelems)
+
+  if nrealelem > 0
+    Nq = N + 1
+    Nqk = dim == 2 ? 1 : Nq
+    device = grid.vgeo isa Array ? CPU() : CUDA()
+    minneighbordistance = similar(grid.vgeo, Nq^dim, nrealelem)
```
tapios: Similarly, `min_neighbor_distance` etc.

tapios:
Looks really useful, @lcw! Thanks.

src/Mesh/Filters.jl
lcw: Changed, thanks!

src/Mesh/Grids.jl
lcw: Changed, thanks!

src/Mesh/Grids.jl
lcw: Changed, thanks!

lcw:
bors try

src/Mesh/Grids.jl
```diff
@@ -173,6 +190,37 @@ function referencepoints(::DiscontinuousSpectralElementGrid{T, dim, N}
...
+    @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
+          knl_min_neighbor_distance!(Val(N), Val(dim), direction,
+                                     min_neighbor_distance, grid.vgeo,
+                                     topology.realelems))
+    locmin = minimum(min_neighbor_distance)
+  else
+    locmin = typemax(T)
+  end
+
+  MPI.Allreduce([locmin], MPI.MIN, topology.mpicomm)[1]
```
simonbyrne:
```suggestion
  MPI.Allreduce(locmin, min, topology.mpicomm)
```

examples/Atmos/heldsuarez.jl
```diff
@@ -80,8 +80,8 @@ function run(mpicomm, polynomialorder, numelem_horz, numelem_vert,
   # determine the time step
   element_size = (setup.domain_height / numelem_vert)
   acoustic_speed = soundspeed_air(FT(315))
-  lucas_magic_factor = 14
-  dt = lucas_magic_factor * element_size / acoustic_speed / polynomialorder ^ 2
+  lucas_magic_factor = 5
+  dt = lucas_magic_factor * min_node_distance(grid) / acoustic_speed
```
simonbyrne: @mwarusz @lcw is there rationale on how this factor comes about?

lcw: In practice, I set this factor as large as possible experimentally.  That is, I keep increasing it until the scheme goes unstable and then back off.

If we had an ODE solver that did error control then it would figure out this factor for us.  Otherwise, you might be able to do a linear stability analysis (such as in <doi:10.1016/j.jcp.2008.07.026>) to give you a better estimate.

tapios: Once you have the `min_node_distance`, this factor is the Courant number (call it that). Make sure that when you use 1D IMEX, the `min_node_distance` here is just the *horizontal* minimum distance.

PR #576, opened 2019-12-05
kernel for Restart functionality and MWE in ex_003_dycoms_restart.jl

slifer50:
-The functions : restart_ode_state and restart_state were added to DGmodel.jl and DGmodel_kernels.jl to enable restarting from saved data and work similarly to init_ode_state.
-A minimum working example exists in ex_003_dycoms_restart.jl which is in examples/DGmethods
-The recommended saving and loading framework to use is the jld2 @save and @load. This can be used to restart a failing simulation based on data from a previous callback to save data at a user determined time step. 
-This will also be a Precursor to using and interpolating data from outside sources.


smarras79:
This should PR be merged before it becomes obsolete and it needs to be reworked. 
How do we proceed?

kpamnany:
Superseded by #850.

PR #580, opened 2019-12-10
Pass args in similar order

### LINKED ISSUE #353: Re-arrange args in moist thermo
Moist thermodynamic functions arguments have inconsistent orders, and they should be fixed.

charleskawczynski:
Fixes #353,

Specifically, I think it makes sense to have the same order of arguments in `PhaseDry` as `PhaseEquil`, so that the argument list of `PhaseEquil` extends that of `PhaseDry`, rather than swapping the order of density and specific total humidity.


tapios:
Looking good!

Issue #569, opened 2019-11-25
Update MoistThermo unsaturated condition


charleskawczynski:
Standardize condition for testing saturated/unsaturated results to be
```
...
unsaturated = q_tot <= q_v_sat
if unsaturated && T_1 > FT(T_min)
...
```
This way, if the unsaturated temperature is unrealistic, we'll jump into saturation adjustment.

charleskawczynski:
Fixed by #581

PR #581, opened 2019-12-10
Revamp moist-thermo config & tests

charleskawczynski:
- Makes Newton method with analytic gradients the default saturation adjustment solver
 - Removes `min` call in `q_vap_saturation_from_pressure`

These changes satisfy 100% convergence rate for thermodynamic states in
```
  n = 100
  ρ_range  = range(1e-3, stop=1.6,  length=n); # density range
  RH_range = range(0, stop=1,       length=n); # relative humidity range
  T_range  = range(150.0, stop=400, length=n); # temperature range
```
which translate to thermo state inputs via
```
      q_sat = q_vap_saturation(T, ρ)
      q_tot = min(RH*q_sat, 1)
      q_pt = PhasePartition_equil(T, ρ, q_tot)
      e_int = internal_energy(T, q_pt)
```


tapios:
Awesome! Thanks, @charleskawczynski . A couple comments/questions:

* How about changing function name `saturation_adjustment` -> `saturation_adjustment_SecantMethod` and `saturation_adjustment_NewtonsMethod` -> `saturation_adjustment`?
* I suggest you also change `if unsaturated` to `if unsaturated && T_1 > FT(T_min)` (l. 714) (and close issue #569)
* Please change `p_vap = q_vap * air_density(ts) * gas_constant_air(ts) * air_temperature(ts)` to `p_vap = q_vap * air_density(ts) * FT(R_v) * air_temperature(ts)` (l. 1041, what's there currently is wrong)
* How about putting your new tests in this PR too? They are more physical than we was there before. (Or did you and I missed them somewhere?)
* Close issue #263

charleskawczynski:
Alright, some of the existing tests are breaking with the new ranges, but I think that I have a nice way to revamp things.

charleskawczynski:
Ah! I think I figured out why some cases were breaking, and it was a simple fix: when `T_2` becomes less than `T_1`, the root finder ends up finding NaNs. 

> How about changing function name saturation_adjustment -> saturation_adjustment_SecantMethod and saturation_adjustment_NewtonsMethod -> saturation_adjustment?

Done

> I suggest you also change if unsaturated to if unsaturated && T_1 > FT(T_min) (l. 714) (and close issue #569)

Done

> Please change p_vap = q_vap * air_density(ts) * gas_constant_air(ts) * air_temperature(ts) to p_vap = q_vap * air_density(ts) * FT(R_v) * air_temperature(ts) (l. 1041, what's there currently is wrong)

Done

> How about putting your new tests in this PR too? They are more physical than what was there before. (Or did you and I missed them somewhere?)

This PR was mainly for testing root-finding methods/parameters. I can update the PR to only include the test scripts, but the source changes we should leave out since we won't want to keep the iteration history for production runs.

> Close issue #263

Done

Also, I've included a new function, `guaranteed_convergence_range`, in `states.jl`, which returns vectors of the ranges that we want to guarantee converged thermo states.

charleskawczynski:
@tapios I'm really liking how this is starting to look!

@simonbyrne feel free to merge if this passes!

charleskawczynski:
Closes #569, closes #263

src/Common/MoistThermodynamics/states.jl
```diff
@@ -201,3 +201,35 @@ function LiquidIcePotTempSHumNonEquil_given_pressure(θ_liq_ice::FT, p::FT, q_pt
...
+ - `q_tot`
+ - `q_pt`
+ - `T`
+that are guaranteed converge in saturation adjustment.
+
+Note that the output vectors are of size ``n^3``.
+"""
+function guaranteed_convergence_range(FT, n)
+  ρ_range  = range(FT(1e-3),  stop=FT(1.6), length=n)
+  RH_range = range(FT(0),     stop=FT(1  ), length=n)
```
tapios:
```suggestion
  RH_range = range(FT(0),     stop=FT(1.1  ), length=n)
```
You'd want some RH > 1 here (otherwise you do not need saturation adjustment because air is unsaturated).

src/Common/PlanetParameters/PlanetParameters.jl
```diff
@@ -39,6 +39,7 @@ using ..ParametersType
 @exportparameter cv_i              cp_i          "Isochoric specific heat ice (J/kg/K)"
 @exportparameter T_freeze          273.15        "Freezing point temperature (K)"
 @exportparameter T_min             150.0         "Minimum temperature guess in saturation adjustment (K)"
+@exportparameter ΔT_min            10.0          "Minimum interval for saturation adjustment guess"
```
tapios: I'd leave this parameter in `MoistThermodynamics` since it'll only be used there. It's also quite large (something like 3 K would suffice, I'd think.)

tapios:
Looks great. Thanks so much, and have a good trip!

src/Common/MoistThermodynamics/states.jl
```diff
@@ -201,3 +201,35 @@ function LiquidIcePotTempSHumNonEquil_given_pressure(θ_liq_ice::FT, p::FT, q_pt
...
+"""
+    guaranteed_convergence_range(FT)
+
+A range of input arguments to thermodynamic state constructors
+ - `e_int`
+ - `ρ`
+ - `q_tot`
+ - `q_pt`
+ - `T`
+that are guaranteed converge in saturation adjustment.
```
tapios:
```suggestion
that are guaranteed to converge in saturation adjustment.
```
`guaranteed` is pretty strong--maybe stronger than we can really claim here (e.g., when PlanetParameters change this may not hold). But it's ok for now.

charleskawczynski:
@tapios I added `tol` and `maxiter` as optional arguments to all constructors that users are expected to use with default values that we expect to use in the end, but this way, in the tests, we can use stricter tolerances to ensure that we have convergence within machine precision. Also, methods we expect users not to use (those only needed internally in MoistThermo) will have non-optional `tol` and `maxiter`. In addition, I made the root solver method an optional input argument to the thermo state constructors, so that we can easily test a variety of them.

I tried increasing the range of relative humidity in `tested_convergence_range`. The difficulty with doing this is that the number of failed saturation adjustment convergences increases very quickly as RH increases. The results are outlined here:

## Range 1:

 - `RH_range = range(FT(0),     stop=FT(1.01), length=n)`
 - all tests pass

## Range 2:
 - `RH_range = range(FT(0),     stop=FT(1.015), length=n)`
 - breaks:
   - `ts = PhaseEquil.(e_int, ρ, q_tot, FT(1e-2), 500)`
   - `@test max(abs.(air_pressure.(ts) .- p)...) < 2506`

## Range 3:
 - `RH_range = range(FT(0),     stop=FT(1.02), length=n)`
 - additional breaks:
   - `ts = LiquidIcePotTempSHumEquil.(θ_liq_ice, ρ, q_tot, 1e-5, 80)`
   - `ts_eq              = PhaseEquil.(e_int, ρ, q_tot, FT(1e-1), 100)`

## Comments
I'm not certain how to handle increasing RH, since it results in so many failures (even at a large number of iterations).

src/Common/MoistThermodynamics/states.jl
```diff
@@ -176,28 +194,70 @@ end
...
+ - `T` air temperature
+ - `θ_liq_ice` liquid-ice potential temperature
+that are tested for convergence in saturation adjustment.
+
+Note that the output vectors are of size ``n^3``, and they
+should span the input arguments to all of the constructors.
+"""
+function tested_convergence_range(FT, n)
+  ρ_range  = range(FT(1e-3),  stop=FT(1.6), length=n)
+  RH_range = range(FT(0),     stop=FT(1.02), length=n)
```
tapios: I am concerned that the maximum of `RH_range`is so small that we are not effectively testing saturation adjustment. For `n<50`, we are never testing the nonlinear saturation adjustment here. (Note that saturation adjustment is only tested for `RH > 1`.) I'd prefer a larger maximum, and maybe widening test tolerances so that tests do not fail.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -176,28 +194,70 @@ end
...
+  T_range  = range(FT(T_min), stop=FT(400), length=n)
```
tapios: If test failures on the edges are problems, realistic atmospheric temperatures range from 200 to 320 K, and this would suffice as a test range.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -91,12 +91,12 @@ using LinearAlgebra
   # variables E_int, p, q_tot, compute the temperature and partitioning of the phases
   tol_T = 1e-2
   q_tot = FT(0); ρ = FT(1)
-  @test MT.saturation_adjustment(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) ≈ 300.0
-  @test abs(MT.saturation_adjustment_NewtonsMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot) - 300.0) < tol_T
+  @test MT.saturation_adjustment_SecantMethod(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot, 1e-3, 10) ≈ 300.0
```
tapios: Here too, tolerances of 1e-2 suffice.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -91,12 +91,12 @@ using LinearAlgebra
...
+  @test abs(MT.saturation_adjustment(internal_energy_sat(300.0, ρ, q_tot), ρ, q_tot, 1e-3, 10) - 300.0) < tol_T
```
tapios: Same here: 1e-2.

test/Common/MoistThermodynamics/runtests.jl
tapios: And `tol_T = 1e-1` is good enough for the tests.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -91,12 +91,12 @@ using LinearAlgebra
...
 
   q_tot = FT(0.21); ρ = FT(0.1)
-  @test MT.saturation_adjustment(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) ≈ 200.0
-  @test abs(MT.saturation_adjustment_NewtonsMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot) - 200.0) < tol_T
+  @test MT.saturation_adjustment_SecantMethod(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot, 1e-3, 10) ≈ 200.0
```
tapios: `q_tot = 0.21` is an extremely (and unrealistically) high specific humidity for 200 K and 0.1 kg/m3. More realistic at these values: `q_tot = 2e-4`.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -91,12 +91,12 @@ using LinearAlgebra
...
+  @test abs(MT.saturation_adjustment(internal_energy_sat(200.0, ρ, q_tot), ρ, q_tot, 1e-3, 10) - 200.0) < tol_T
```
tapios: 1e-2 for tolerance should suffice.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -121,218 +121,193 @@ using LinearAlgebra
...
   # PhaseDry
   ts = PhaseDry.(e_int, ρ)
   @test all(internal_energy.(ts) .≈ e_int)
   @test all(air_density.(ts) .≈ ρ)
 
-  # PhaseEquil: Dry case
-  ts = PhaseEquil.(e_int, ρ, q_tot_dry)
+  # PhaseEquil
+  ts = PhaseEquil.(e_int, ρ, q_tot, FT(1e-2), 90, Ref(MT.saturation_adjustment_SecantMethod))
   @test all(internal_energy.(ts) .≈ e_int)
```
tapios: If the approximately equal sign here tests to machine precision, this cannot work when nonlinear saturation adjustment is active. You can expect a similar relative error on energy and other thermodynamic quantities as on temperature. So, this can only be accurate to within `e_int * (tol/T)`, and it should only be tested to this accuracy. Similarly for other thermodynamic quantities.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -121,218 +121,193 @@ using LinearAlgebra
...
+  @test all(getproperty.(PhasePartition.(ts),:tot) .≈ q_tot)
   @test all(air_density.(ts) .≈ ρ)
 
-  # PhaseEquil: Moist case
-  ts = PhaseEquil.(e_int, ρ, q_tot_moist)
-  @test all(internal_energy.(ts) .≈ e_int)
-  @test all(getproperty.(PhasePartition.(ts),:tot) .≈ q_tot_moist)
-  @test all(air_density.(ts) .≈ ρ)
+  # ts = PhaseEquil.(e_int, ρ, q_tot, FT(1e-2), 500)
+  # @test all(internal_energy.(ts) .≈ e_int)
```
tapios: The tests here should test to accuracy within specified relative error (`tot/T`) on temperature.

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -121,218 +121,193 @@ using LinearAlgebra
...
 
-end
+# end
 
 
-@testset "moist thermodynamics - type-stability" begin
+# @testset "moist thermodynamics - type-stability" begin
 
+  # NOTE: `Float32` saturation adjustment tends to have more difficulty
+  # with converging to the same tolerances as `Float64`, so they're relaxed here.
```
tapios: Tolerances can be generally relaxed.

src/Common/MoistThermodynamics/states.jl
charleskawczynski: Fixed!

src/Common/MoistThermodynamics/states.jl
charleskawczynski: Interesting, a max temperature of 400 breaks saturation adjustment with the new ranges (that allow us to RH~1.2), but 350 seems to work fine, so I'll keep it at the higher end.

test/Common/MoistThermodynamics/runtests.jl
charleskawczynski: @tapios If all of the tests are passing with current strict tolerances, should we relax them?

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -685,44 +687,72 @@ function saturation_adjustment_NewtonsMethod(e_int::FT, ρ::FT, q_tot::FT) where
...
+    if !sol.converged
+      error("saturation_adjustment did not converge")
+    end
     return sol.root
   end
 end
 
 """
-    saturation_adjustment(e_int, ρ, q_tot)
+    ΔT_min(::Type{FT})
```
tapios: Do we still need `ΔT_min` and `ΔT_max` with the secant method with the updated (and more physical) tests of saturation adjustment? Otherwise, I'd remove these.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -685,44 +687,72 @@ function saturation_adjustment_NewtonsMethod(e_int::FT, ρ::FT, q_tot::FT) where
...
+
+"""
+    ΔT_max(::Type{FT})
+
+Maximum interval for saturation adjustment using Secant method
+"""
+@inline ΔT_max(::Type{FT}) where FT = FT(10)
+
+"""
+    bound_upper_temperature(T_1::FT, T_2::FT) where {FT<:Real}
```
tapios: Remove unless we still need it with the new tests.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -1036,14 +1096,32 @@ exner(ts::ThermodynamicState) = exner(air_temperature(ts), air_density(ts), Phas
 
 The relative humidity, given a thermodynamic state `ts`.
 """
-function relative_humidity(ts::ThermodynamicState)
+function relative_humidity(ts::ThermodynamicState{FT}) where {FT<:Real}
   q = PhasePartition(ts)
   q_vap = q.tot - q.liq - q.ice
-  p_vap = q_vap * air_density(ts) * gas_constant_air(ts) * air_temperature(ts)
+  p_vap = q_vap * air_density(ts) * FT(R_v) * air_temperature(ts)
```
tapios: Why don't we replace the (redundant) code here with a call to `relative_humidity(T, p, q)`, with temperature and pressure from `air_temperature`, `air_pressure`? That way we reduce possible error sources from having relative humidity defined twice in the code.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -176,28 +195,97 @@ end
...
+end
+
+"""
+    fixed_lapse_rate_ref_state(z::FT,
+                               T_surface::FT,
+                               T_min::FT) where {FT<:AbstractFloat}
+
+Fixed lapse rate hydrostatic reference state
+"""
+function fixed_lapse_rate_ref_state(z::FT,
```
tapios: Any reason not to call the `ref_state.jl` code here?

test/Common/MoistThermodynamics/runtests.jl
tapios: You can try to relax to 1e-2, but it doesn't need to more accurate than that (and 1e-1 is fine too as a test of the code).

src/Common/MoistThermodynamics/states.jl
charleskawczynski: That requires arguments `m::HydrostaticState{P,F}, atmos::AtmosModel, aux::Vars, geom::LocalGeometry`, which is a bit higher level than the primitives we're dealing with here. Maybe the calls should be the other way around though, to avoid code duplication.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: This is still used in `saturation_adjustment_SecantMethod` and `saturation_adjustment_q_tot_θ_liq_ice`

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: I think it makes sense to keep these range bounds, but I can try removing them to see what happens. We may need them again if we increase the number of cases / combinations we test.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Agreed.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Looks like this does improve convergence in the standard cases (for the reference profiles). Specifically:
`ts = PhaseEquil.(e_int, ρ, q_tot, FT(1e-2), 15, Ref(MT.saturation_adjustment_SecantMethod))` passes with `T_2 = bound_upper_temperature(T_1, T_2)` in the saturation adjustment function.

test/Common/MoistThermodynamics/runtests.jl
charleskawczynski: There is no error for `e_int`, since this is an input and should be exactly equal when retrieved :)

charleskawczynski: Temperature is really the only variable that we need to check for a physical tolerance, since this is determined via saturation adjustment. We can test additional variables (e.g., pressure), but it will follow from temperature.

charleskawczynski:
@tapio Added some new tests, specifically to test the accuracy of the default constructors.

src/Common/MoistThermodynamics/MoistThermodynamics.jl
tapios: Ok

src/Common/MoistThermodynamics/states.jl
```diff
@@ -75,8 +75,14 @@ struct PhaseEquil{FT} <: ThermodynamicState{FT}
   "temperature: computed via [`saturation_adjustment`](@ref)"
   T::FT
 end
-PhaseEquil(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real} =
-  PhaseEquil(e_int, ρ, q_tot, saturation_adjustment(e_int, ρ, q_tot))
+function PhaseEquil(e_int::FT,
+                    ρ::FT,
+                    q_tot::FT,
+                    tol::FT=FT(100),
+                    maxiter::Int=5,
```
tapios: Can we have the default `maxiter = 1` here? I'd like to get the performance gains coming from that, if we can, without the users having to specify it. If the tests pass with a tolerance of 0.1 K, this will be good enough.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -75,8 +75,14 @@ struct PhaseEquil{FT} <: ThermodynamicState{FT}
   "temperature: computed via [`saturation_adjustment`](@ref)"
   T::FT
 end
-PhaseEquil(e_int::FT, ρ::FT, q_tot::FT) where {FT<:Real} =
-  PhaseEquil(e_int, ρ, q_tot, saturation_adjustment(e_int, ρ, q_tot))
+function PhaseEquil(e_int::FT,
+                    ρ::FT,
+                    q_tot::FT,
+                    tol::FT=FT(100),
```
tapios: Why 100? I thought these are tolerance in Kelvin, no?

src/Common/MoistThermodynamics/states.jl
tapios: Ok

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -124,215 +124,261 @@ using LinearAlgebra
...
-
-  e_int = collect(range(10_000.0,200_000.0,length=n))
-  p = collect(range(10_000.0,200_000.0,length=n))
-  q_pt_dry = PhasePartition.(q_tot_dry, q_liq .* 0, q_ice .* 0)
-  q_pt_moist = PhasePartition.(q_tot_moist, q_liq, q_ice)
+  e_int, ρ, q_tot, q_pt, T, p, θ_liq_ice = MT.tested_convergence_range(FT, 50)
+  tol = 1e-1
+
+  # PhaseEquil
+  ts_exact = PhaseEquil.(e_int, ρ, q_tot, FT(1e-6), 10000, MT.saturation_adjustment_SecantMethod)
```
tapios: Any reason to test this with secant rather than Newton method here?

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -124,215 +124,261 @@ using LinearAlgebra
...
+  ts = PhaseEquil.(e_int, ρ, q_tot)
+  # Should be machine accurate:
+  @test all(getproperty.(PhasePartition.(ts),:tot) .≈ getproperty.(PhasePartition.(ts_exact),:tot))
+  @test all(internal_energy.(ts) .≈ internal_energy.(ts_exact))
+  @test all(air_density.(ts) .≈ air_density.(ts_exact))
+  # Approximate:
+  @test max(abs.(air_pressure.(ts)    .- air_pressure.(ts_exact))./air_pressure.(ts_exact)...) < tol
```
tapios: Would be good to distinguish `rtol` and `atol` here. What you have here is relative tolerance `rtol`. But it seems to me `tol` is an absolute tolerance, no?

test/Common/MoistThermodynamics/runtests.jl
tapios: Right. But isn't l. 170 testing the internal energy computed from the derived temperature, which is computed by saturation adjustment and hence potentially inaccurate?

charleskawczynski:
Thanks @tapios for going through all of the remaining changes! bors r+

PR #584, opened 2019-12-12
Add AtmosModel option to factor out hydrostatic balance

mwarusz:


tapios:
Looks good, @mwarusz! Do you have any experience what difference it makes to factor out the hydrostatic state or not?

mwarusz:
This option improves stability, especially when working in lower precision (`Float32`). It seems
necessary to run the acoustic wave test stably with `Float32` and I saw it fix the same problem
in the gravity wave test (which is not yet in the repo).

tapios:
Sounds good. It may make sense to make this the default option then.

simonbyrne:
- Is there any downside to doing this?
- Can we add the details to the design docs?

tapios:
> * Is there any downside to doing this?
> * Can we add the details to the design docs?

No downside I can see. I can add the lines to the design docs.

PR #586, opened 2019-12-12
test on Julia 1.3

simonbyrne:


test/DGmethods/integral_test_sphere.jl
```diff
@@ -50,22 +50,18 @@ function init_aux!(m::IntegralTestSphereModel, aux::Vars, g::LocalGeometry)
...
-  aux.θ = atan(y , x)
-  aux.ϕ = asin(z / aux.r)
-
+  θ = atan(y, x)
+  ϕ = asin(z / aux.r)
   # Exact integral
-  a = 1 + sin(aux.θ)^2 + sin(aux.ϕ)^2
-  aux.int.v = exp(-a * aux.r^2) - exp(-a * m.Rinner^2)
-  aux.rev_int.v = exp(-a * m.Router^2) - exp(-a * aux.r^2)
+  aux.a = 1 + cos(ϕ)^2 * sin(θ)^2 + sin(ϕ)^2
```
simonbyrne: I had to reformulate this to make it more stable near the poles (`sin(θ)` can fluctuate considerably).

simonbyrne:
I to also change some tolerances to the integral tests so that they passed on both versions.

Is it possible to derive better bounds for these?

jkozdon:
You could check that the rate of convergence is correct as opposed to checking the numbers themselves.

simonbyrne:
I was hoping there was some sort of absolute bound on Lagrange polynomial interpolation we could use.

What is the rate of convergence? Currently it appears to be a factor of 2^6 for each factor of 2 increase in resolution, but this changes if I change the radius of the sphere.

simonbyrne:
>  but this changes if I change the radius of the sphere.

Ah, that could actually just be due to numerical precision.

blallen:
Any reason this isn't merged yet? Mostly wondering because we are setting up a new machine and would like to use Julia 1.3 if there are no issues with it.

simonbyrne:
I'll merge this, but note that this only changes the Azure Pipelines, we need to change Slurm CI manually (this is one annoying thing from having moved the scripts out of here).

simonbyrne:
I manually changed the version used by Slurm, and it seems to pass, so 1.3 here we come!

PR #587, opened 2019-12-13
Move direction to BalanceLaw

simonbyrne:
This should make it easier to compose models.


simonbyrne:
bors try

simonbyrne:
bors try-

simonbyrne:
bors try

simonbyrne:
@lcw I'm not sure exactly what you mean can you expand on it?

I do know there are still lots of issues we need to figure out (i.e. how we handle multirate stuff, especially 2nd derivatives), but this seemed like a reasonable way forward.

lcw:
I was just thinking about what would happen if you tried to use the remainder model with balance laws with different directions.  Right now I don't think that would work.

simonbyrne:
Yeah, that will require separate kernel calls, which is not ideal.

We can do the necessary projections in the balance laws themselves, but if we're going to do that, it might be worth thinking about a local rotated coordinate system.

charleskawczynski:
Can `direction` be given a more descriptive name?

bischtob:

PR #589, opened 2019-12-13
Fix pressure in LinearTemperatureProfile

mwarusz:


tapios:
Good catch! Thanks, @mwarusz

mwarusz:
bors try

simonbyrne:
Another reason to use Unitful 😄

PR #590, opened 2019-12-14
Enable dims argument for weighted norms

simonbyrne:
Previously this threw an error


simonbyrne:
bors try

src/Arrays/MPIStateArrays.jl
```diff
@@ -488,7 +488,7 @@ function weighted_norm_impl(Q::SubArray{FT, N, A}, W, ::Val{p}, dims::Colon) whe
       if isfinite(p)
         accum += W[i, 1, k] * abs(Q[i, j, k]) ^ p
       else
-        waQ_ijk = W[i, j, k] * abs(Q[i, j, k])
+        waQ_ijk = W[i, 1, k] * abs(Q[i, j, k])
```
charleskawczynski: Why is this being changed? (And why is the line above read with a `1` also: `accum += W[i, 1, k] * abs(Q[i, j, k]) ^ p`?)

simonbyrne: weights are set up to have `size(weights, 2) == 1`

bischtob:

simonbyrne:
yes, we still need to fix it.

test/Arrays/reductions.jl
```diff
@@ -30,6 +30,20 @@ mpirank = MPI.Comm_rank(mpicomm)
...
 
+  QAW = MPIStateArray{Float32}(mpicomm, ArrayType, localsize..., weights=ones(4,1,8))
+  QAW .= A
+
+  @test norm(QAW, 1)   ≈ norm(globalA, 1)
+  @test norm(QAW)      ≈ norm(globalA)
+  @test norm(QAW, Inf) ≈ norm(globalA, Inf)
+
+  if ArrayType == Array
+    # TODO: make this work with CuArrays
```
charleskawczynski, line 41: NLsolve.jl [needs the same thing](https://github.com/JuliaNLSolvers/NLsolve.jl/issues/234#issuecomment-643767576)

PR #592, opened 2019-12-16
Cleanup - towards driver/ configuration structure

akshaysridhar:
1) DYCOMS : Moves radiation and boundary condition components into driver. Extend test time to 100s (towards `slow` tests with long-time-integrated mass conservation metrics). Moves DYCOMS from `examples/DGmethods` to `examples/Atmos`. Case now exists as a standalone run
2) Rayleigh-Benard: Moves boundary condition component into driver. Case now exists as a standalone test 
cc: @simonbyrne  @smarras79


akshaysridhar:
bors try

kpamnany:
Is something similar also going to be done for Held Suarez? That will help make the first cut of the driver/configuration PR (coming soon!) more complete.

PS: @akshaysridhar: Missed [this one](https://github.com/climate-machine/CLIMA/blob/as/cleanup/slurmci-test.toml#L45).

akshaysridhar:
> Is something similar also going to be done for Held Suarez? That will help make the first cut of the driver/configuration PR (coming soon!) more complete.
> 
> PS: @akshaysridhar: Missed [this one](https://github.com/climate-machine/CLIMA/blob/as/cleanup/slurmci-test.toml#L45). 

@kpamnany I'll have a look at `heldsuarez.jl` as well; it uses the generic NoFluxBC() (which I think is reasonable to keep in the `boundaryconditions.jl` file); the problem specific source is already included in the `driver` (Gravity() and Coriolis() are expected to be universally available source terms so I leave them in`source.jl`)

akshaysridhar:
See #672

PR #597, opened 2019-12-31
Driver+configurations

kpamnany:
This is a first cut of the driver+configurations mechanic, to ease the development of experiments. Currently, DYCOMS and Held-Suarez have been ported to use the new mechanic -- see the code in `experiments/`.

This also adds command line argument support -- currently just a few such as enabling/disabling VTK output, etc. See the current supported list [here](https://github.com/climate-machine/CLIMA/blob/kp/driver/src/Driver/Driver.jl#L77).


experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    T_initial::FT = 255
+    domain_height::FT = 30e3
+end
+
+function (setup::HeldSuarezSetup)(state, aux, coords, t) 
+    # callable to set initial conditions
+    FT = eltype(state)
+
+    r = norm(coords, 2)
+    h = r - FT(planet_radius)
```
tapios: We should have an altitude function for this, probably in `orientation.jl`.

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    state.ρu = SVector{3, FT}(0, 0, 0)
+    state.ρe = state.ρ * (internal_energy(setup.T_initial) + aux.orientation.Φ)
+
+    return nothing
+end
+
+function config_heldsuarez(FT, N, resolution, setup)
+    config = CLIMA.GCM_Configuration("HeldSuarez", FT, ArrayType, setup,
+                                     ref_state=HydrostaticState(IsothermalProfile(setup.T_initial), FT(0)),
+                                     turbulence=ConstantViscosityWithDivergence(FT(0)),
```
tapios: For `turbulence`, we'd want to use an anisotropic Smagorinsky filter (until we have horizontal hyperdiffusion). Perhaps @akshaysridhar can weigh in?

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    ΔT_y      = FT(60)
+    Δθ_z      = FT(10)
+    T_equator = FT(315)
+    T_min     = FT(200)
+
+    σ_b          = FT(7 / 10)
+    r            = norm(coord, 2)
+    @inbounds λ  = atan(coord[2], coord[1])
+    @inbounds φ  = asin(coord[3] / r)
+    h            = r - FT(planet_radius)
```
tapios: Same as above, it would be good to provide an altitude function, e.g., in `orientation.jl`.

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    scale_height = FT(7000) # from Smolarkiewicz JAS 2001 paper
```
tapios: Use same `scale_height` as in setup (`R_d T/g`)

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    exner_p       = σ ^ (R_d / cp_d)
+    Δσ            = (σ - σ_b) / (1 - σ_b)
+    height_factor = max(0, Δσ)
+    T_equil       = (T_equator - ΔT_y * sin(φ) ^ 2 - Δθ_z * log(σ) * cos(φ) ^ 2 ) * exner_p
+    T_equil       = max(T_min, T_equil)
+    k_T           = k_a + (k_s - k_a) * height_factor * cos(φ) ^ 4
+    k_v           = k_f * height_factor
+
+    source.ρu += -k_v * ρu
+    source.ρe += -k_T * ρ * cv_d * (T - T_equil) - k_v * ρu' * ρu / ρ
```
tapios:
```suggestion
    source.ρe += -k_T * ρ * cv_d * (T - T_equil)
```
(If the heat generated by friction at the surface is returned to the flow, it has no impact on total energy and should not appear here.)

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+
+    setup = HeldSuarezSetup{FT}()
+
+    # Set up the driver configuration
+    config = config_heldsuarez(FT, N, resolution, setup)
+
+    # Determine the time step
+    element_size = (setup.domain_height / nelem_vert)
+    acoustic_speed = soundspeed_air(FT(315))
+    lucas_magic_factor = 5
```
tapios:
```suggestion
    Courant_number = 0.8
```
(I'm not sure what the right Courant number here is, but it should be less than one.)

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    dt = lucas_magic_factor * min_node_distance(config.grid) / acoustic_speed
```
tapios:
```suggestion
    dt = Courant_number * min_node_distance(config.grid, HorizontalDirection()) / acoustic_speed
```
If we use 1D IMEX, which should be the default for the GCM configuration, we should just use the `min_node_distance` in the horizontal.

min_node_distance(grid,

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+
+    timeend = FT(60) # 400day
```
tapios: Can we give `timeend` in real time (days or seconds)? Or is that what it is already? If it is in SI units, it should probably be something like `60*day`.

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+year = {2005},
+doi = {10.1175/MWR2930.1},
+URL = {https://doi.org/10.1175/MWR2930.1},
+eprint = {https://doi.org/10.1175/MWR2930.1}
+}
+"""
+function init_dycoms!(state::Vars, aux::Vars, (x,y,z), t)
+    FT            = eltype(state)
+
+    xvert::FT     = z
```
tapios: Could we use `altitude` or `z` consistently instead of `xvert`, which is strange name for atmospheric scientists?

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+    Rd::FT        = R_d
+
+    # These constants are those used by Stevens et al. (2005)
+    qref::FT      = FT(9.0e-3)
+    q_tot_sfc::FT = qref
+    q_pt_sfc      = PhasePartition(q_tot_sfc)
+    Rm_sfc::FT    = 461.5 #gas_constant_air(q_pt_sfc)
```
tapios: I think it's better if we use our own thermodynamic constants consistently. Otherwise, we are bound to get inconsistencies in the thermodynamics. (Once we have a way of overriding `PlanetParameters` for the entire code, we can define local thermodynamic constants for specific cases.)

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+    T_sfc::FT     = 292.5
+    P_sfc::FT     = 101780 #MSLP
```
tapios: Use `MSLP`.

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+
+    FT = Float32
+
+    # DG polynomial order
+    N = 4
+
+    # Domain resolution and size
+    Δh = FT(40)
+    Δv = FT(10) #Δh/aspectratio
+    aspectratio = Δh/Δv
```
tapios: Do we need `aspectratio` anywhere? Otherwise, remove.

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+    resolution = (Δh, Δh, Δv)
+
+    xmin, xmax = 0, 5000
+    ymin, ymax = 0, 5000
+    zmin, zmax = 0, 2500
+    domain_size = (xmin, xmax, ymin, ymax, zmin, zmax)
+
+    dt = FT(0.01)
```
tapios: We should have a similar construction here as above, with something like
`dt = Courant_number * min_node_distance(config.grid, HorizontalDirection()) / acoustic_speed`

src/Driver/Configurations.jl
```diff
@@ -0,0 +1,131 @@
...
+
+function GCM_Configuration(name::String, FT::DataType, array_type, init_GCM!;
+                           orientation    = SphericalOrientation(),
+                           ref_state      = NoReferenceState(),
+                           turbulence     = SmagorinskyLilly{FT}(0.15), # TODO: C_smag?
+                           moisture       = EquilMoist(),
+                           radiation      = NoRadiation(),
+                           sources        = (Gravity(),
+                                             Coriolis(),
+                                             GeostrophicForcing{FT}(7.62e-5, 0, 0)), # TODO: f_coriolis?
```
tapios:
```suggestion
                                     
```
There should be no geostrophic forcing here, just the Coriolis force.

src/Driver/Configurations.jl
```diff
@@ -0,0 +1,131 @@
...
+    config.grid = DiscontinuousSpectralElementGrid(config.topology,
+                                                   FloatType=FT,
+                                                   DeviceArray=config.array_type,
+                                                   polynomialorder=N,
+                                                   meshwarp=meshwarp)
+end
+
+function GCM_Configuration(name::String, FT::DataType, array_type, init_GCM!;
+                           orientation    = SphericalOrientation(),
+                           ref_state      = NoReferenceState(),
```
tapios: We want a hydrostatic reference state here by default, with `LinearTemperatureProfile`, `T_min = 200`, `T_surface = 280`, and lapse rate `\Gamma = g/cp`. It would be good if these defaults are easy to override from the drivers.

src/Driver/Configurations.jl
```diff
@@ -0,0 +1,131 @@
...
+
+function LES_Configuration(name::String, FT::DataType, array_type, init_LES!;
+                           orientation    = FlatOrientation(),
+                           ref_state      = NoReferenceState(),
+                           turbulence     = SmagorinskyLilly{FT}(0.15), # TODO: C_smag?
+                           moisture       = EquilMoist(),
+                           radiation      = NoRadiation(),
+                           sources        = (Gravity(),
+                                             Coriolis(),
+                                             GeostrophicForcing{FT}(7.62e-5, 0, 0)), # TODO: f_coriolis?
```
tapios: Yes, better with `f_coriolis` as a parameter specified in the driver.

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    # Set up the filter callback
+    filterorder = 14
+    filter = ExponentialFilter(config.grid, 0, filterorder)
+    cbfilter = EveryXSimulationSteps(1) do
+        Filters.apply!(config.Q, 1:size(Q, 2), config.grid, filter)
+        nothing
+    end
+    =#
+
+    result = CLIMA.invoke(dt, FT(0), timeend, config, LSRK144NiegemannDiehlBusch;
```
tapios: Can we move the time stepper specification to the configuration files, with the possibility of overriding the default there by something that can be specified in the driver? Also, we should make sure that the default for LES and GCM is a 1D IMEX method.

tapios:
Thanks, @kpamnany, for pushing this along. This looks like an awesome start to me. Happy New Year!

experiments/GCM/heldsuarez.jl
```diff
@@ -0,0 +1,140 @@
...
+    ρe    = state.ρe
+    coord = aux.coord
+    Φ     = aux.orientation.Φ
+    e     = ρe / ρ
+    u     = ρu / ρ
+    e_int = e - u' * u / 2 - Φ
+    T     = air_temperature(e_int)
+
+    # Held-Suarez constants
+    k_a       = FT(1 / (40 * day))
```
charleskawczynski: Can we change `day` to `sec_per_day` or whatever it's supposed to be?

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+    qref::FT      = FT(9.0e-3)
+    q_tot_sfc::FT = qref
+    q_pt_sfc      = PhasePartition(q_tot_sfc)
+    Rm_sfc::FT    = 461.5 #gas_constant_air(q_pt_sfc)
+    T_sfc::FT     = 292.5
+    P_sfc::FT     = 101780 #MSLP
+    ρ_sfc::FT     = P_sfc / Rm_sfc / T_sfc
+
+    # Specify moisture profiles
+    q_liq::FT      = 0
```
charleskawczynski:
```suggestion
```
Defined below

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+
+    # Calculate PhasePartition object for vertical domain extent
+    q_pt  = PhasePartition(q_tot, q_liq, q_ice)
+    Rm    = gas_constant_air(q_pt)
+
+    # Pressure
+    H     = Rm_sfc * T_sfc / grav;
+    p     = P_sfc * exp(-xvert/H);
+    # Density, Temperature
+    T = air_temperature_from_liquid_ice_pottemp_given_pressure(θ_liq, p, q_pt)
```
charleskawczynski: I haven't yet removed this function from export but I prefer it not be used outside of MoistThermo since it has a real chance of being misused (usually in the form of an over-defined system, which is tricky to catch). Furthermore, this function is only used when computing non-equilibrium thermodynamic states (which we don't yet support after the first time-step). While all equilibrium thermodynamic states are subsets of non-equilibrium states, the reverse is not true and this can trigger a very large change in thermodynamic quantities after the first timestep.

If you can, try changing this to
```
ts = LiquidIcePotTempSHumEquil_given_pressure(θ_liq_ice, p, q_tot)
T = air_temperature(ts)
ρ = air_density(ts)
```

This will return an equilibrium thermodynamic state, but you may want to confirm that the other variables you wish you prescribed are still close. @smarras79, is this the case that we discussed, and can you confirm that using `LiquidIcePotTempSHumEquil_given_pressure` worked fine?

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,176 @@
...
+    else
+        m = (ziplus - zi)/(u2 - u1)
+        u = (xvert - zi)/m + u1
+
+        m = (ziplus - zi)/(v2 - v1)
+        v = (xvert - zi)/m + v1
+    end
+    e_kin       = FT(1/2) * (u^2 + v^2 + w^2)
+    e_pot       = grav * xvert
+    E           = ρ * total_energy(e_kin, e_pot, T, q_pt)
```
charleskawczynski:
```suggestion
    E           = ρ * total_energy(e_kin, e_pot, ts)
```
Apply this if you were able to switch to using `ts = LiquidIcePotTempSHumEquil_given_pressure(θ_liq_ice, p, q_tot)`

src/Driver/Configurations.jl
```diff
@@ -0,0 +1,131 @@
...
+    # AtmosModel details
+    bl::BalanceLaw
+    numfluxnondiff::NumericalFluxNonDiffusive
+    numfluxdiff::NumericalFluxDiffusive
+    gradnumflux::GradNumericalPenalty
+
+    # initialized after construction
+    #
+    # execution details
+    mpicomm::Union{Missing,MPI.Comm}
```
charleskawczynski: Just curious: is `Missing` preferred over `Nothing` here? And if so, why?

smarras79:
> This is looking nice! My only strong suggestion is about trying to establish a thermo-state. Give what I suggested a try and if it doesn't work, me and (hopefully) @smarras79 can help grind things out. We had worked through this a bit before, so things may be more up-to-date in his latest Dycoms branch.

Thanks @charleskawczynski  I will compare with my latest branch and help @kpamnany on these modifications. Thanks for your insight.

experiments/GCM/heldsuarez.jl
lcw: Why should it be less than one?

tapios:
If we have 1d IMEX in the vertical and then evaluate a Courant number based on the minimum nodal distance in the horizontal, how could it be greater than 1 for stable  horizontally explicit time stepping?

On Jan 2, 2020, at 9:41 AM, Lucas C Wilcox <notifications@github.com> wrote:

﻿

@lcw commented on this pull request.

________________________________

In experiments/GCM/heldsuarez.jl<https://github.com/climate-machine/CLIMA/pull/597#discussion_r362562006>:

> +    N = 5
+
+    # Domain resolution
+    nelem_horz = 6
+    nelem_vert = 8
+    resolution = (nelem_horz, nelem_vert)
+
+    setup = HeldSuarezSetup{FT}()
+
+    # Set up the driver configuration
+    config = config_heldsuarez(FT, N, resolution, setup)
+
+    # Determine the time step
+    element_size = (setup.domain_height / nelem_vert)
+    acoustic_speed = soundspeed_air(FT(315))
+    lucas_magic_factor = 5


Why should it be less than one?

experiments/GCM/heldsuarez.jl
tapios: My preference would be to move all time stepping choices to the configuration files, with an option to override Courant numbers (and time steppers) from the driver.

lcw:
> If we have 1d IMEX in the vertical and then evaluate a Courant number based on the minimum nodal distance in the horizontal, how could it be greater than 1 for stable horizontally explicit time stepping?

It depends on the stability region of the ODE solver but for high-order RK schemes it is definitely possible to get Courant numbers greater than 1.

experiments/GCM/heldsuarez.jl
tapios: From what @smarras79 reports, ARK232 gave best overall performance. @smarras79: what was the horizontal Courant number you used? More generally, we need (a) guidance about stability regions of the various schemes in the design docs, and (b) should associate default Courant numbers to the different schemes, so that the default Courant number decision is not made on the user (driver) level.

smarras79: As of now, the maximum stable Courant number for my runs was 0.2 or less using **LSRK54CarpenterKennedy** but have not yet looked at the max stable values for **ARK2GiraldoKellyConstantinescu** 

I will get back to you on this for values.
This said, I have been in touch with @fxgiraldo about 1D IMEX for dycoms because the solution deviates from the correct one. I am waiting for his corrections on this.

experiments/LES/dycoms.jl
smarras79: Correct, the latest driver that I have been using uses what @charleskawczynski suggests above. 
@charleskawczynski please, use this driver as reference: 

[dycoms driver](https://github.com/climate-machine/CLIMA/blob/sm/yt-DYCOMS-MULTI-RATE-NEW-MOIST-THERMO/test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex-FOR-MACIEJ.jl)

experiments/GCM/heldsuarez.jl
tapios: @smarras79 @lcw @fxgiraldo Could you guys please help Kiran figure out how to set up the various time steppers so that they are associated with default Courant numbers based on their stability regions, and how to incorporate this into the configuration files? We need to make these configuration decisions less "magic" and more transparent. The configurations should also be flexible enough that we can (soon, I hope) accommodate IMEX with multirate, without the user having to specify details at the driver level.

src/Driver/Configurations.jl
kpamnany: Could use either in this case. The explanation of the difference [here](https://news.ycombinator.com/item?id=18600786) doesn't really shed light on this use case. Maybe @simonbyrne has input?

experiments/GCM/heldsuarez.jl
kpamnany: I think https://github.com/climate-machine/CLIMA/pull/593 is introducing this. I'll work with @smarras79 to fold this in when it is worked out.

smarras79: Hi @kpamnany, the Courant calculator is in place in the PR that you indicated, but I am working with @fxgiraldo to get the 1D IMEX best settings for our use. 
@fxgiraldo please, let me know how I can help you to address this.

experiments/LES/dycoms.jl
charleskawczynski: If the solution goes crazy, maybe try reverting this, since it's not precisely the same as what is in the [Dycoms driver](https://github.com/climate-machine/CLIMA/blob/sm/yt-DYCOMS-MULTI-RATE-NEW-MOIST-THERMO/test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex-FOR-MACIEJ.jl#L166)

src/Atmos/Model/orientation.jl
```diff
@@ -15,6 +15,7 @@ end
 gravitational_potential(::Orientation, aux::Vars) = aux.orientation.Φ
 ∇gravitational_potential(::Orientation, aux::Vars) = aux.orientation.∇Φ
 altitude(orientation::Orientation, aux::Vars) = gravitational_potential(orientation, aux) / grav
+altitude(aux::Vars) = altitude(aux.orientation, aux)
```
charleskawczynski: It turns out that this won't work because `orientation` is actually a field in `atmos::AtmosModel`, not the one in `aux::Vars`.

charleskawczynski:
If conflicts become an issue, maybe we can leave/define `get_array_type()` in `misc.jl`, and open a separate PR, since this ends up touching a lot of files.

src/Atmos/Model/orientation.jl
kpamnany: So what's the right way to write the `altitude()` that Held Suarez wants?

kpamnany:
Many of the comments are now addressed, except for a few relating to Held Suarez. DYCOMS is crashing with a kernel error; to be debugged.

The interface is probably pretty settled at this point -- `CLIMA.init()`, `CLIMA.[GCM|LES]_Configuration()`, `CLIMA.setup_solver()`, and `CLIMA.invoke!()`. Further comments are welcome.

src/Atmos/Model/orientation.jl
charleskawczynski: You'll need to pass `atmos::AtmosModel` down to where you want to call it, and then call it with `altitude(atmos, aux)`

charleskawczynski: (or pass `orientation` to where you need it, but we may later find that we need other things, so I figure this will be a nice safety net to avoid changing the interface many times)

experiments/LES/dycoms.jl
```diff
@@ -0,0 +1,191 @@
...
+    C_smag = FT(0.18)
+    C_drag = FT(0.0011)
+    LHF    = FT(115)
+    SHF    = FT(15)
+    bc = DYCOMS_BC{FT}(C_drag, LHF, SHF)
+
+    config = CLIMA.LES_Configuration("DYCOMS", N, resolution, xmax, ymax, zmax,
+                                     init_dycoms!,
+                                     ref_state=ref_state,
+                                     radiation=radiation,
```
charleskawczynski: I believe we may need to add `ConstantSubsidence{FT}(D_subsidence)` for Dycoms. @smarras79 please confirm.

src/Atmos/Model/orientation.jl
kpamnany: This is needed in `init_state!()` and in a source. We could get the model in the latter by defining an `atmos_source!()`, but not sure how we'd get it in the former. Will likely defer this for this PR.

experiments/LES/dycoms.jl
charleskawczynski: Not trying to slow things down, but a sponge may also be needed. I think we're currently working on updating the sponge in #619.

simonbyrne:
I like the overall design. Smaller issues we can iron out as we go.

experiments/GCM/heldsuarez.jl
kpamnany: This is a constant in PlanetParameters, so let's address that question separately. :)

kpamnany:
@simonbyrne: the CI error is consistent across numerous tries: attempt to use an MPI routine after finalizing, right after LinearSolvers small full system. Any idea what's going on?

kpamnany:
@smarras79: using the DYCOMS driver you linked [here](https://github.com/climate-machine/CLIMA/blob/sm/yt-DYCOMS-MULTI-RATE-NEW-MOIST-THERMO/test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex-FOR-MACIEJ.jl) as a template, I get the following at runtime, whether using the 1D IMEX solver or the explicit solver. Did you fix this problem? Can you link a working driver please?
```
ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
ERROR: LoadError: KernelException: exception thrown during kernel execution on device Tesla P100-PCIE-16GB
```

smarras79:
> @smarras79: using the DYCOMS driver you linked [here](https://github.com/climate-machine/CLIMA/blob/sm/yt-DYCOMS-MULTI-RATE-NEW-MOIST-THERMO/test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex-FOR-MACIEJ.jl) as a template, I get the following at runtime, whether using the 1D IMEX solver or the explicit solver. Did you fix this problem? Can you link a working driver please?
> 
> ```
> ERROR: a exception was thrown during kernel execution.
>        Run Julia on debug level 2 for device stack traces.
> ERROR: LoadError: KernelException: exception thrown during kernel execution on device Tesla P100-PCIE-16GB
> ```

Hi @kpamnany  please, stay on hold on testing dycoms because @charleskawczynski and I are currently working at merging my branch into master and that will require some changes.

simonbyrne:
> @simonbyrne: the CI error is consistent across numerous tries: attempt to use an MPI routine after finalizing, right after LinearSolvers small full system. Any idea what's going on?

Not quite sure, but it looks like we rejigged enough so that the GC kicked in and the calls to MPI.Finalize() completed. We should move all MPI-enabled tests to use the `runmpi` wrapper.

kpamnany:
Rebased to include Atmos source change for sponge, but still seeing the kernel error. :-(

experiments/GCM/heldsuarez.jl
kpamnany: @mwarusz: can you confirm this as you wrote the original code?

experiments/GCM/heldsuarez.jl
kpamnany: Will defer this to a later PR.

kpamnany:
Apart from a minimal test case of some sort, to add in `test/Driver/` for code coverage, this is good to go. Both of the drivers here are added into Slurm CI tests.

experiments/GCM/heldsuarez.jl
tapios: It's correct without the drag term in the energy source. There may have been an error in the design document early on, which made it into @mwarusz's code, but the term really shouldn't be there.

simonbyrne:
bors try

kpamnany:
bors try

kpamnany:
bors try-

kpamnany:
bors try

kpamnany:
bors try

kpamnany:
bors try

kpamnany:
bors try

kpamnany:
bors try

kpamnany:
@akshaysridhar: please run `experiments/GCM/heldsuarez.jl` in this branch and validate the results?

@simonbyrne: please have a look at `test/Driver/driver_test.jl`. Can we add an `@test` there that's useful?

dburov190:
How long do you think till this gets merged, guys? a ballpark... I'm sort of relying on this because I'm planning to build a "mini-Earth" example basing on the Held-Suarez, so I've started playing around with it in the master branch but don't want to patch after this gets merged (:

If it might take a week or so I'll fork this branch, no problem

PR #601, opened 2020-01-07
Ck/new squal line

charleskawczynski:
Adds Squal line to new framework. Supersedes #549


trontrytel:
Hi @smarras79  @slifer50. This is the pull request we have been working on with @charleskawczynski. We merged it with master and are in the process of resolving all the conflicts. Wait for another changes to this and an ok from @charleskawczynski. 

This pull request basically adds two new structures: non-equilibrum moisture (for when we want to compute cloud water without saturation adjustment) and precipitation (for handling rain and snow). Maybe precipitation should be a part of moist variables but for now we defined it separately because it has different boundary conditions and velocity. There is still no ice microphysics in it, but lets resolve the issues we have with warm rain and add the ice later.

The major things that still need to be done to make it work are:
- [ ] Boundary conditions for the variables in precipitation and non-equil moisture. We are not sure where to add them and how. The boundary conditions stuff might be changed soon. @smarras79 - do you know more about it than us?
- [ ] Write a driver that actually uses the non-equil moisture and precipitation. Right now we have none so there are prolly some more errors on the way that we haven't seen yet.
- [ ] Apply rain terminal velocity to precipitation. There were some issues with it last time we were working on it so right now it is computed but not applied.

I would suggest this:

- [ ] @slifer50 and @smarras79: Ignore this pull request for now and instead create a pull request with a squall line setup that uses equilibrium moisture and has no precipitation. Make the setup up to date with the current Clima interface and have that merged into Clima. 

- [ ] Once we have that going I can add to it the non-equil moisture. I will still need help with the boundary conditions.

- [ ] Once we have that, whomever is the most available will add the precipitation to the squall line driver.  
 
Once we get through all the coding and debugging there will still be issues with negative values of microphysics variables. Right now we do nothing to ensure positive sign. We can use the squall line setup to tackle that. Having the squall line with warm rain I'll have the setup to add the ice/snow microphysics.

smarras79:
> * [ ]  Boundary conditions for the variables in precipitation and non-equil moisture. We are not sure where to add them and how. The boundary conditions stuff might be changed soon. @smarras79 - do you know more about it than us?

I know that @simonbyrne is working on changing the way b.c. are handled but I don't know how that is being designed.

tapios:
Thanks for this, @charleskawczynski and @trontrytel.

In terms of order of things to do, the following sequence probably makes most sense: 
1a) @kpamnany merges his re-factoring of the driver/configuration infrastructure
1b) @simonbyrne completes and merges the boundary condition interface
2) Adjust the squall line case (and other examples) to the new framework, and merge then. Boundary condition for precipitation etc. should be straightforward once we have the new b.c. interface.

This is provided that 1a) and 1b) can be done within 2-3 weeks.

test/DGmethods/compressible_Navier_Stokes/squall-line-model.jl
```diff
@@ -0,0 +1,284 @@
...
+  end
+
+  solve!(Q, lsrk, param; timeend=timeend, callbacks=(cbinfo, ))
+
+  outprefix = "vtk/squal_line/state_init_$(dim)D_mpirank$(MPI.Comm_rank(mpicomm))_step$(DFloat(0))"
+  writevtk(outprefix, Q, dg, flattenednames(vars_state(model, DFloat)))
+
+  # Print some end of the simulation information
+  engf = norm(Q)
+  Qe = init_ode_state(dg, param, DFloat(timeend), spline_int())
```
simonbyrne: does this need to be splatted?

src/Atmos/Model/moisture.jl
```diff
@@ -73,8 +73,9 @@ vars_aux(::EquilMoist,FT) = @vars(temperature::FT, θ_v::FT, q_liq::FT)
 
 @inline function atmos_nodal_update_aux!(moist::EquilMoist, atmos::AtmosModel,
                                          state::Vars, aux::Vars, t::Real)
+  FT=eltype(state)
   e_int = internal_energy(moist, atmos.orientation, state, aux)
-  TS = PhaseEquil(e_int, state.ρ, state.moisture.ρq_tot/state.ρ, moist.maxiter)
+  TS = PhaseEquil(e_int, state.ρ, state.moisture.ρq_tot/state.ρ, FT(1))
```
charleskawczynski: I would suggest waiting for #711 if you want control over max iterations. Using `FT(1)` will break (MethodError) since this is expected to be an integer. If you meant for this to be the tolerance, `tol` comes after `maxiter`.

src/Atmos/Model/moisture.jl
```diff
@@ -161,6 +162,18 @@ function gradvariables!(moist::NonEquilMoist, atmos::AtmosModel, transform::Vars
...
+  transform.moisture.q_ice = state.moisture.ρq_ice * ρinv
+end
+
+function flux_moisture!(moist::NonEquilMoist, atmos::AtmosModel, flux::Grad, state::Vars, aux::Vars, t::Real)
+  ρ = state.ρ
+  u = state.ρu / ρ
+  z = altitude(atmos.orientation, aux)
+  usub = subsidence_velocity(atmos.subsidence, z)
+  ẑ = vertical_unit_vector(atmos.orientation, aux)
+  u_tot = u .- usub * ẑ
```
charleskawczynski: Subsidence is being reformulated as a source (in #701) so you may want to mirror `flux_moisture!` for `EquilMoist` in that PR.

src/Atmos/Model/moisture.jl
slifer50: Oddly enough it didn't break. though this isn't updated with regards to the current master. The case is still a large work in progress

src/Atmos/Model/moisture.jl
slifer50: Sure I'll have this updated once it's working as intended.

slifer50:
This now has a working non-precipitating squall line running on NonEquilMoist thermodynamics. It runs fully to completion and works with both ConstantDiffusion and Smagorinsky.

slifer50:
A fully 3D non-precipitation squall line case is now working and runs to completion.

trontrytel:
@smarras79 , @slifer50 , @charleskawczynski - Should we close this? I'm fine with closing this PR as long as the `ck/NewSquallLine` branch remains alive.

My plan for microphysics in LES is to work first on the prescribed velocity test based on [Grabowski_1998](https://journals.ametsoc.org/jas/article/55/21/3283/24316/Toward-Cloud-Resolving-Modeling-of-Large-Scale) kinematic test from section 4. I show some initial results/problems from it here: https://github.com/CliMA/ClimateMachine.jl/pull/1330. See especially the last comment from me there. 

Once I have that test ironed out I'll add the microphysics interface to CliMA-LES proper. After that the squall line should be as straightforward as any other LES case.

PR #603, opened 2020-01-07
Add free parameters

charleskawczynski:
New idea for free parameters struct / annotation

### DIFF
--- a/src/CLIMA.jl
+++ b/src/CLIMA.jl
@@ -3,6 +3,7 @@ module CLIMA
 include("misc.jl")
 include("Utilities/TicToc/TicToc.jl")
 include("Utilities/ParametersType/ParametersType.jl")
+include("Utilities/FreeParameters/FreeParameters.jl")
 include("Common/PlanetParameters/PlanetParameters.jl")
 include("Utilities/RootSolvers/RootSolvers.jl")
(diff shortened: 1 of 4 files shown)


tapios:
Thanks, @charleskawczynski! This looks like a good start to me. How about adding another (optional) field to the struct, `std`? We will typically Gaussian priors, so having mean and standard deviation would be good. In fact, for the learning algorithms, we will typically map variables with prior ranges (with uniform priors) to transformed variables with normal priors.

charleskawczynski:
@simonbyrne @trontrytel and @bielim spoke this afternoon, and it seems like the free parameter framework should sit outside of CLIMA, so that `CLIMA`, `Cloudy.jl`, and the UQ packages can all use it.

My understanding is that it will provide a way for us to _annotate_ what parameters are free parameters. Then, at the UQ level, `FreeParameters.jl` will provide methods to extract all of the free parameters in whatever complex model we have in CLIMA (or, e.g., microphysics model) and put them into an array. It may also support transformations.

charleskawczynski:
Closing, based on the discussion. Further development of this module was moved to [FreeParameters.jl](https://github.com/climate-machine/FreeParameters.jl)

PR #604, opened 2020-01-07
Extend init args

charleskawczynski:
Extends arguments for `init_state!`, since we'll need [many additional arguments](https://github.com/climate-machine/CLIMA/pull/601/files#diff-30c477a58b9368ed896dc001fedda8f8R57) in the Squal line setup. Hopefully, this will limit how many files #601 touches.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -250,8 +250,8 @@ boundary_state!(nf, m::AtmosModel, x...) =
 # FIXME: This is probably not right....
 boundary_state!(::CentralGradPenalty, bl::AtmosModel, _...) = nothing
 
-function init_state!(m::AtmosModel, state::Vars, aux::Vars, coords, t)
-  m.init_state(state, aux, coords, t)
+function init_state!(m::AtmosModel, state::Vars, aux::Vars, coords, t, args...)
+  m.init_state(state, aux, coords, t, args...)
 end
 
 end # module


charleskawczynski:

examples/Atmos/heldsuarez.jl
```diff
@@ -159,7 +159,7 @@ Base.@kwdef struct HeldSuarezSetup{FT}
   domain_height::FT = 30e3
 end
 
-function (setup::HeldSuarezSetup)(state, aux, coords, t) 
+function (setup::HeldSuarezSetup)(state, aux, coords, t, args...)
```
simonbyrne: Should only need to add these extra args if they're actually used.

simonbyrne:
I'm okay with this, but would it be better to include the `args` as part of the model object? Alternatively, we could move initialization out of model altogether.

examples/Atmos/heldsuarez.jl
charleskawczynski: Ok, I’ll see if I can remove this without breaking

charleskawczynski:
Not for the case that we need it. Args is a bunch of profiles for the Dycoms initial condition (which I think may be generated from PyCLES, or that matlab script, or both). I’ll take a look tomorrow, if it’s only from the matlab script, then maybe it’s a good time to translate that to Julia.

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -88,14 +88,14 @@ struct InitStateBC <: BoundaryCondition
 end
 function atmos_boundary_state!(::Rusanov, bc::InitStateBC, m::AtmosModel,
                                stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, _...)
-  init_state!(m, stateP, auxP, auxP.coord, t)
+                               auxM::Vars, bctype, t, args...)
+  init_state!(m, stateP, auxP, auxP.coord, t, args...)
```
simonbyrne: I don't think you want this: 
1) some boundary conditions get the interior points as extra args
2) the extra args aren't going to be passed in anyway

test/DGmethods/advection_diffusion/advection_diffusion_model.jl
```diff
@@ -134,29 +134,29 @@ function init_aux!(m::AdvectionDiffusion, aux::Vars, geom::LocalGeometry)
 end
 
 function init_state!(m::AdvectionDiffusion, state::Vars, aux::Vars,
-                     coords, t::Real)
+                     coords, t::Real, args...)
```
simonbyrne: the extra args aren't used

test/DGmethods/advection_diffusion/advection_diffusion_model.jl
```diff
@@ -134,29 +134,29 @@ function init_aux!(m::AdvectionDiffusion, aux::Vars, geom::LocalGeometry)
...
   initial_condition!(m.problem, state, aux, coords, t)
 end
 
 function boundary_state!(nf, m::AdvectionDiffusion, stateP::Vars, auxP::Vars,
-                         nM, stateM::Vars, auxM::Vars, bctype, t, _...)
+                         nM, stateM::Vars, auxM::Vars, bctype, t, args...)
```
simonbyrne: see notes on BC above.

test/DGmethods/compressible_Navier_Stokes/ref_state.jl
```diff
@@ -28,7 +28,7 @@ using CLIMA.Atmos
 using CLIMA.Atmos: internal_energy, thermo_state
 import CLIMA.Atmos: MoistureModel, temperature, pressure, soundspeed
 
-init_state!(state, aux, coords, t) = nothing
+init_state!(state, aux, coords, t, args...) = nothing
```
simonbyrne: again, shouldn't be necessary

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -88,14 +88,14 @@ struct InitStateBC <: BoundaryCondition
...
 end
 function atmos_boundary_state!(::CentralNumericalFluxDiffusive, bc::InitStateBC,
                                m::AtmosModel, stateP::Vars, diffP::Vars,
                                auxP::Vars, nM, stateM::Vars, diffM::Vars,
-                               auxM::Vars, bctype, t, _...)
-  init_state!(m, stateP, auxP, auxP.coord, t)
+                               auxM::Vars, bctype, t, args...)
+  init_state!(m, stateP, auxP, auxP.coord, t, args...)
```
simonbyrne: get rid of this one too.

PR #606, opened 2020-01-08
Initial soil model

simonbyrne:
Need to figure out the boundary conditions here.


examples/Land/soil.jl
```diff
@@ -0,0 +1,58 @@
...
+# Set up DG scheme
+dg = DGModel( # 
+  m, # "PDE part"
+  grid,
+  CentralNumericalFluxNonDiffusive(), # penalty terms for discretizations
+  CentralNumericalFluxDiffusive(),
+  CentralGradPenalty())
+
+
+Δ = min_node_distance(grid)
```
tapios: Do you want to specify that what matters here is the vertical direction?

examples/Land/soilmodel.jl
```diff
@@ -0,0 +1,89 @@
...
+function gradvariables!(m::SoilModel, transform::Vars, state::Vars, aux::Vars, t::Real)
+  transform.T = state.ρcT / (m.ρ * m.c)
+end
+function diffusive!(m::SoilModel, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real)
+  diffusive.∂T∂z = ∇transform.T[2]
+end
+function flux_nondiffusive!(m::SoilModel, flux::Grad, state::Vars, aux::Vars, t::Real)
+end
+function flux_diffusive!(m::SoilModel, flux::Grad, state::Vars, diffusive::Vars, aux::Vars, t::Real)
+  flux.ρcT += SVector(0, m.λ * diffusive.∂T∂z, 0)
```
tapios: Maybe better to make this coordinate independent, using the `vertical_unit_vector` from `orientation.jl`, i.e., 
```suggestion
  flux.ρcT += m.λ * diffusive.∂T∂z .* vertical_unit_vector
```
(To make sure this works on the sphere in the end.)

tapios:
Looks good. You can add the boundary conditions when the more general B.C. Interface is sorted out.

examples/Land/soilmodel.jl
```diff
@@ -0,0 +1,89 @@
...
+                        init_aux!, init_state!,
+                        boundary_state!, wavespeed, LocalGeometry
+using CLIMA.DGmethods.NumericalFluxes: NumericalFluxNonDiffusive,
+                                       NumericalFluxDiffusive,
+                                       GradNumericalPenalty
+
+"""
+    SoilModel
+
+Computes diffusive flux `F` in:
```
tapios:
```suggestion
Computes diffusive flux on the right-hand side of the diffusion equation:
```

tapios:
Can we merge this so others can take over?

charleskawczynski:
@gagnelandmanna and I are working on this

charleskawczynski:
@kmdeck @gagnelandmanna, this has gotten quite stale. Can we close this PR? Or are there things in here we should keep?

kmdeck:
> @kmdeck @gagnelandmanna, this has gotten quite stale. Can we close this PR? Or are there things in here we should keep?
@charleskawczynski 
IMO we can close. It looks like Elias wrote his own finite difference algorithm for soil, which might be useful to hang onto: https://github.com/CliMA/ClimateMachine.jl/blob/852386c495c0261ab40a7b81a408dfa8cb016670/examples/Land/Soil_Physics.jl#L95

Some references mentioned that could be useful to know of:
Dai et al (2019a) and Balland and Arp (2005) 

Other capabilities we have in the new framework.

We can wait for @gagnelandmanna to confirm though.

Does closing  = deleting? Should i grab a copy of that file I mentioned?

charleskawczynski:
> Does closing = deleting? Should i grab a copy of that file I mentioned?

I don't think so, but it would be good to create a local branch so that someone doesn't delete the local branch when house-cleaning.

kmdeck:
> > Does closing = deleting? Should i grab a copy of that file I mentioned?
> 
> I don't think so, but it would be good to create a local branch so that someone doesn't delete the local branch when house-cleaning.

OK, I made  a local copy.

PR #608, opened 2020-01-08
Add quadratic term to air temperature Taylor expansion

### LINKED ISSUE #567: [MoistThermo] Add quadratic term to MT func
We need to add the quadratic term to `air_temperature_from_liquid_ice_pottemp`, and make sure that the error is scaling correctly.

charleskawczynski:
This adds the quadratic term in the Taylor expansion for temperature in `air_temperature_from_liquid_ice_pottemp `. The previous version (with only the linear term, but not quadratic) fails the test that has been added (and is passing with the quadratic term).

Closes #567

### DIFF
--- a/src/Common/MoistThermodynamics/MoistThermodynamics.jl
+++ b/src/Common/MoistThermodynamics/MoistThermodynamics.jl
@@ -927,9 +927,14 @@ and, optionally,
 function air_temperature_from_liquid_ice_pottemp(θ_liq_ice::FT, ρ::FT,
   q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
 
-  cv = cv_m(q)
+  cvm = cv_m(q)
+  cpm = cp_m(q)
   R_m = gas_constant_air(q)
-  return θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv) + latent_heat_liq_ice(q)/cv
+  κ = 1 - cvm/cpm
+  T_u = (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cvm)*θ_liq_ice
+  T_1 = latent_heat_liq_ice(q)/cvm
+  T_2 = -κ/(2*T_u)*(latent_heat_liq_ice(q)/cvm)^2
+  return T_u + T_1 + T_2
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -927,9 +927,14 @@ and, optionally,
 function air_temperature_from_liquid_ice_pottemp(θ_liq_ice::FT, ρ::FT,
   q::PhasePartition{FT}=q_pt_0(FT)) where {FT<:Real}
 
-  cv = cv_m(q)
+  cvm = cv_m(q)
+  cpm = cp_m(q)
   R_m = gas_constant_air(q)
-  return θ_liq_ice * (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cv) + latent_heat_liq_ice(q)/cv
+  κ = 1 - cvm/cpm
+  T_O0 = (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cvm)*θ_liq_ice
```
tapios: I find the O and 0 in the subscript confusing. Why not something like:
```suggestion
  T_u = (ρ*R_m*θ_liq_ice/FT(MSLP))^(R_m/cvm)*θ_liq_ice
```
(for unsaturated) and replace `T_01` by `T_1` etc.?

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -237,6 +237,9 @@ end
 
     # Accurate but expensive `LiquidIcePotTempSHumNonEquil` constructor (Non-linear temperature from θ_liq_ice)
     T_non_linear = air_temperature_from_liquid_ice_pottemp_non_linear.(θ_liq_ice, ρ, FT(1e-3), 10, q_pt)
+    T_expansion = air_temperature_from_liquid_ice_pottemp.(θ_liq_ice, ρ, q_pt)
+    C_1 = 0.0001
+    @test all(abs.(T_non_linear .- T_expansion) .< 15*sqrt(eps(FT)) .+ C_1 .* latent_heat_liq_ice.(q_pt)./cv_m.(q_pt))
```
tapios: Simpler: I'd test for a given relative error (`rtol` a few percent) between `T_non_linear` and `T_expansion`.

tapios:
Looking good! Thanks, @charleskawczynski

src/Common/MoistThermodynamics/MoistThermodynamics.jl
charleskawczynski: Sure, we can do that. I had written it as `O` indicating `Order_1`, `Order_2`, etc. But `T_1`, `T_2` works fine too.

test/Common/MoistThermodynamics/runtests.jl
charleskawczynski: Ah, good point. Adjusted, works, and looks even better.

PR #609, opened 2020-01-08
Add subsidence to `AtmosModel`

charleskawczynski:
Adds `subsidence` to `AtmosModel`, we'll need this in Dycoms and many other cases I believe.


src/Atmos/Model/AtmosModel.jl
```diff
@@ -121,14 +123,18 @@ Where
...
   ρu = state.ρu
   u = ρinv * ρu
+  z  = altitude(m.orientation, aux)
 
   # advective terms
-  flux.ρ   = ρu
-  flux.ρu  = ρu .* u'
-  flux.ρe  = u * state.ρe
+  usub = subsidence_velocity(m.subsidence, z)
+  u_tot = u + usub
```
tapios: The name `subsidence` implies that this is a vertical velocity only, and defined positive downward. So you'd want something like
```suggestion
  u_tot = u - usub * vertical_unit_vector
```
where `vertical_unit_vector` is the upward pointing vertical unit vector.

src/Atmos/Model/subsidence.jl
```diff
@@ -0,0 +1,16 @@
...
+
+abstract type AbstractSubsidence{FT<:AbstractFloat} end
+
+struct NoSubsidence{FT} <: AbstractSubsidence{FT} end
+
+struct ConstantSubsidence{FT} <: AbstractSubsidence{FT}
+  D::FT
+end
+
+subsidence_velocity(::NoSubsidence{FT}, z::FT) where {FT} = SVector{FT}(0,0,0)
```
tapios: I'd define this as a scalar (vertical) velocity here, defined positive downward. Fewer zeros to carry around, and more in line how people use the term.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Should I call this `subsidence_speed` then? Since the direction is already assumed?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -121,14 +123,19 @@ Where
...
+  ρinv = 1/ρ
   ρu = state.ρu
   u = ρinv * ρu
+  z  = altitude(m.orientation, aux)
 
   # advective terms
-  flux.ρ   = ρu
-  flux.ρu  = ρu .* u'
-  flux.ρe  = u * state.ρe
+  usub = subsidence_speed(m.subsidence, z)
```
tapios:
```suggestion
  usub = subsidence_velocity(m.subsidence, z)
```
("subsidence velocity" is the term we usually use)

src/Atmos/Model/AtmosModel.jl
```diff
@@ -121,14 +123,19 @@ Where
...
+  vertical_unit_vector = ∇gravitational_potential(m.orientation, aux)
```
tapios: I think it's better when `vertical_unit_vector` is provided by `orientation.jl`. You may compute it from the gravitational potential there (but need to divide by `g`!).

charleskawczynski: >I think it's better when vertical_unit_vector is provided by orientation.jl. 

`∇gravitational_potential` is a function in `orientation.jl`, but I'll add a new method specifically for the unit vector.

> You may compute it from the gravitational potential there (but need to divide by g!)

Good catch!

src/Atmos/Model/AtmosModel.jl
```diff
@@ -121,14 +123,19 @@ Where
...
   ρu = state.ρu
   u = ρinv * ρu
+  z  = altitude(m.orientation, aux)
 
   # advective terms
-  flux.ρ   = ρu
-  flux.ρu  = ρu .* u'
-  flux.ρe  = u * state.ρe
+  usub = subsidence_velocity(m.subsidence, z)
+  ẑ = altitude_unit_vector(m.orientation, aux)
```
tapios: Why don't we leave it with the common vertical unit vector (I haven't seen altitude unit vector before):
```suggestion
  ẑ = vertical_unit_vector(m.orientation, aux)
```

charleskawczynski:

charleskawczynski:
@simonbyrne, we're passing all of the tests but [timing out](https://gist.github.com/climabot/378a9dd69650d09cf056b17b031a8435#file-out_8517339)

PR #610, opened 2020-01-08
Sm/yt dycoms multi rate new moist thermo master

charleskawczynski:
Just creating this PR to see the code diff

PR #612, opened 2020-01-08
Add `Courant` module

charleskawczynski:
Adds Courant module, broken off from #610

There aren't any unit tests for this module, but it would be nice to add some eventually.

### DIFF
--- a/src/CLIMA.jl
+++ b/src/CLIMA.jl
@@ -31,6 +31,7 @@ include("ODESolvers/MultirateInfinitesimalStepMethod.jl")
 include("ODESolvers/MultirateRungeKuttaMethod.jl")
 include("ODESolvers/GenericCallbacks.jl")
 include("Atmos/Model/AtmosModel.jl")
+include("Diagnostics/Courant.jl")
 include("Diagnostics/Diagnostics.jl")
 include("InputOutput/VTK/VTK.jl")
 
--- a/src/Diagnostics/Courant.jl
+++ b/src/Diagnostics/Courant.jl
@@ -1,0 +1,165 @@
+module Courant
+
+using Dates
+using FileIO
+using JLD2
+using MPI
+using StaticArrays
+using Printf
+
+using ..Atmos
+using ..DGmethods: num_state, vars_state, num_aux, vars_aux, vars_diffusive, num_diffusive, DGModel
+using ..Mesh.Topologies
+using ..Mesh.Grids
+using ..MoistThermodynamics
+using ..MPIStateArrays
+using ..PlanetParameters
+using ..VariableTemplates
+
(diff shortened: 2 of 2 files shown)


charleskawczynski:

tapios:
A couple issues:

1. We only need to distinguish between horizontal and vertical Courant numbers (additional distinctions in the horizontal seem unnecessary to me).
2. This needs clearer variable names (e.g., `Courant_sound_horiz` instead of `Courssx`).
3. It is also unclear to me where 'Dx' etc come from. Is this from `min_node_distance`, or from a local minimum nodal distance on the element where velocities are evaluated? I think the simple evaluation based on a global minimum nodal distance (distinguishing horizontal and vertical, and advective velocities and sound speeds) would suffice.

smarras79:
> A couple issues:
> 
> 1. We only need to distinguish between horizontal and vertical Courant numbers (additional distinctions in the horizontal seem unnecessary to me).
> 2. This needs clearer variable names (e.g., `Courant_sound_horiz` instead of `Courssx`).
> 3. It is also unclear to me where 'Dx' etc come from. Is this from `min_node_distance`, or from a local minimum nodal distance on the element where velocities are evaluated? I think the simple evaluation based on a global minimum nodal distance (distinguishing horizontal and vertical, and advective velocities and sound speeds) would suffice.
> 
> Need @fxgiraldo review.

Perfect, I will change it.

fxgiraldo:
Yes, I agree with only requiring horizontal vs vertical Courant numbers. Regarding whether to use the minimum global spacing or per element could matter for highly unstructured/stretched grids.  Now that @smarras79 has pointed me to a working branch with the Courant numbers, I can add my test cases and compare the time-steps and Courant number computations with NUMA.

tapios:
To simply matters, I suggest you focus on advective and diffusive Courant numbers only. The speed of sound varies only by ~20% in the atmosphere, so the acoustic Courant number is more or less fixed by a reference speed of sound and the minimum nodal distance. This can be calculated offline once. The advective and diffusive Courant numbers are the ones that vary and that we would want to monitor.

kpamnany:
I don't think this should go in `Diagnostics`, rather in `ODESolvers`.

smarras79:
> I don't think this should go in `Diagnostics`, rather in `ODESolvers`.

@kpamnany  don't worry Courant.jl because it is being replaced in full by #618

simonbyrne:
sdiff in #612 seems to be just extracting one of the diagonal components of the tensor to make a scalar (or am I misunderstanding it?) We should probably do something more principled, but I would like to know what that is.

#612 also seems to make the implicit assumption that the dg grid is Cartesian aligned

PR #613, opened 2020-01-08
Add init_aux args, needed for Courant calcs

charleskawczynski:
Adds `Dx,Dy` (`min_node_distance`), broken off from #610

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -224,11 +224,11 @@ function integrate_aux!(m::AtmosModel, integ::Vars, state::Vars, aux::Vars)
 end
 
 # TODO: figure out a nice way to handle this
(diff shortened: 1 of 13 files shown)


simonbyrne:
Where are these actually used? In #610 these variables are only used at the driver level, and not actually used inside `init_aux!`.

simonbyrne:
@smarras79?

charleskawczynski:
They’re used in the Courant module, which is in another small PR

simonbyrne:
#612? `gather_Courant` is called at the drive level, not inside `init_aux`

charleskawczynski:
Yeah, I think you’re right, I’ll double check tomorrow and close if it’s not used.

charleskawczynski:
Yeah, this is unused. Closing.

Issue #565, opened 2019-11-23
aux not being updated 


smarras79:
The only way to update aux.SOMEMODEL.VAR and make it accessible from Diagnostics.jl is by assigning its value from `atmos_init_aux!`

When I am inside a model such as turbulence.jl or moisture.jl,  
the value of any auxiliary quantity `aux.SOMEMODEL.VAR` cannot be update and made available to Diagnostics.jl unless it is done from function `atmos_init_aux!`

Practical example:

In **turbulence.jl** we have the function:

```
function squared_buoyancy_correction(normS, ∇transform::Grad, aux::Vars)
  ∂θ∂Φ = dot(∇transform.turbulence.θ_v, aux.orientation.∇Φ)
  N² = ∂θ∂Φ / aux.moisture.θ_v
  Richardson = N² / (normS^2 + eps(normS))

  buoyancy_correction = sqrt(clamp(1 - Richardson*inv_Pr_turb, 0, 1))

  aux.turbulence.Ri = buoyancy_correction

  return buoyancy_correction
end
```

however, although we have `aux.turbulence.Ri = buoyancy_correction`, aux.turbulence.Ri is zero when shown in **Diagnostics.jl**

We should be able to update the value of `aux.SOMEMODE.VAR` from within any of the functions defined in SOMEMODEL.jl and not just from `atmos_init_aux!`.

How do we modify the code to make this possible?

(Above, SOMEMODEL could be turbulence, moisture, etc.)

@simonbyrne @akshaysridhar @charleskawczynski

akshaysridhar:
@smarras79 : There is an `atmos_nodal_update_aux` function provided for updating auxiliary variables. For example, https://github.com/climate-machine/CLIMA/blob/c046b2f424c8ec59e8b42d76a93de97111940570/src/Atmos/Model/moisture.jl#L11 and https://github.com/climate-machine/CLIMA/blob/c046b2f424c8ec59e8b42d76a93de97111940570/src/Atmos/Model/moisture.jl#L51

smarras79:
> @smarras79 : There is an `atmos_nodal_update_aux` function provided for updating auxiliary variables. For example,
> 
> https://github.com/climate-machine/CLIMA/blob/c046b2f424c8ec59e8b42d76a93de97111940570/src/Atmos/Model/moisture.jl#L11
> 
> and
[...]

PR #617, opened 2020-01-09
Add `update_aux_diffusive!` and remove VSModel from HBModel

blallen:
I implemented a second `update_aux!` function with access to the gradients computed at the current time step as discussed in #565 and #535. 

This also allowed me to remove the VerticalSupplementModel from the HydrostaticBoussinesqModel as we now have access to the partial derivatives used to compute the horizontal divergence of the velocity vector.

### DIFF
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -51,14 +51,17 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
   communicate = !(isstacked(topology) &&
                   typeof(dg.direction) <: VerticalDirection)
 
-  update_aux!(dg, bl, Q, t)
+  aux_comm = update_aux!(dg, bl, Q, t)
+  @assert typeof(aux_comm) == Bool
 
   ########################
   # Gradient Computation #
   ########################
   if communicate
     MPIStateArrays.start_ghost_exchange!(Q)
-    MPIStateArrays.start_ghost_exchange!(auxstate)
+    if aux_comm
+      MPIStateArrays.start_ghost_exchange!(auxstate)
+    end
   end
 
   if nviscstate > 0
@@ -70,7 +73,9 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
 
     if communicate
       MPIStateArrays.finish_ghost_recv!(Q)
-      MPIStateArrays.finish_ghost_recv!(auxstate)
+      if aux_comm
+        MPIStateArrays.finish_ghost_recv!(auxstate)
+      end
     end
 
     @launch(device, threads=Nfp, blocks=nrealelem,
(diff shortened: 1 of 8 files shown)


blallen:
This is a work in progress; the ocean model is currently blowing up. I still wanted comments on it's implemented in case we need something different.

src/DGmethods/DGmodel.jl
```diff
@@ -82,6 +82,8 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
     communicate && MPIStateArrays.start_ghost_exchange!(Qvisc)
   end
 
+  update_aux_diffusive!(dg, bl, Q, t)
```
jkozdon: If you have changed auxstate you'll need to either call update aux again after the viscous state has been communicated or recommunicated the aux state.

Of course you will also probably want some catch for whether aux has changed so you know whether the comm is needed

blallen: hmmm... this function updates the auxstate and relies on knowing the full diffstate. 

Should I put a call to `MPIStateArrays.finish_ghost_recv!(Qvisc)` before L85? Do I need to add another `MPIStateArrays.start_ghost_exchange!(auxstate)` afterwards? I'm not sure what the exact logic is for these.

blallen:
@jkozdon , for some reason the `if` statement in [L150](https://github.com/blallen/CLIMA/blob/bla/update_aux_diffusive/test/DGmethods/Euler/isentropicvortex-multirate.jl#L150) of `isentropicvortex-multirate.jl` is returning `false` for the `ARK2GiraldoKellyConstantinescu` solver and thus initializing with the wrong constructor when I run the tests file. I didn't change anything related to this, so I'm not sure why it's happening. Any ideas?

src/DGmethods/DGmodel.jl
```diff
@@ -79,9 +79,24 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
...
+      if communicate
+        MPIStateArrays.finish_ghost_recv!(Qvisc)
+      end
+
+      update_aux_diffusive!(dg, bl, Q, t)
+
+      if communicate
+        MPIStateArrays.start_ghost_exchange!(auxstate)
+      end
+    end
```
blallen: @jkozdon, is this what you wanted for the communication?

jkozdon: I was thinking something like this

https://github.com/climate-machine/CLIMA/commit/44157d06e78d8b0e4099203450e628afa47846d5

@simonbyrne might have a better suggestion.

jkozdon: Note my code won't run since all the existing `update_aux!` functions don't return `Bool`, but this way we just communicate `auxstate` as needed.

blallen: Okay, I implemented something similar to what you did and it works

blallen:
bors try

blallen:
bors try

blallen:
@jkozdon can you merge this?

jkozdon:
This seems fine to me. @simonbyrne You have an opinion on the interface?

PR #593, opened 2019-12-18
Courant number calculator and addition of SGS viscosity to diffusive state

slifer50:
-A function was made to calculate the courant number: src/Diagnostics/Courant.jl
-SGS viscosity added to diffusive_state in src/Atmos/Model/AtmosModel.jl as it is needed for the diffusive courant number calculation.
-Minimum working example is example/DGmethods/ex_002_dycoms_Courant.jl

PR #618, opened 2020-01-11
Add CFL computation function

lcw:
During the simulation it can be useful to compute the effective CFL
number (`CFL = Δt*wavespeed/Δx`).

This adds a function similar to the minimum node distance function that
returns the maximum CFL number given a function that computes the local
wave speed times the time step size.

### DIFF
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -294,7 +294,7 @@ function cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
   nrealelem = length(topology.realelems)
 
   if nrealelem > 0
-    N = polynomialorder(dg.grid)
+    N = polynomialorder(grid)
     dim = dimensionality(grid)
     Nq = N + 1
     Nqk = dim == 2 ? 1 : Nq
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -277,25 +277,27 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
 end
 
 """
-    cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
-        direction=EveryDirection())
+    courant(local_courant::Function, dg::DGModel, m::BalanceLaw,
+            Q::MPIStateArray, direction=EveryDirection())
 
-Returns the maximum of the evaluation of the function `localcfl` pointwise
-throughout the domain.  The function `localcfl` is given an approximation of
-the local node distance `Δx`.  The `direction` controls which reference
-directions are considered when computing the minimum node distance `Δx`.
+Returns the maximum of the evaluation of the function `local_courant`
+pointwise throughout the domain.  The function `local_courant` is given an
+approximation of the local node distance `Δx`.  The `direction` controls which
+reference directions are considered when computing the minimum node distance
+`Δx`.
 
-An example `localcfl` function is
+An example `local_courant` function is
 
-    function localcfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars,
-                      Δx)
+    function local_courant(m::AtmosModel, state::Vars, aux::Vars,
+                           diffusive::Vars, Δx)
       return Δt * cmax / Δx
     end
 
-where `Δt` is the time step size and `cmax` is the maximum speed of the model.
+where `Δt` is the time step size and `cmax` is the maximum flow speed in the
+model.
 """
-function cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
-             direction=EveryDirection())
+function courant(local_courant::Function, dg::DGModel, m::BalanceLaw,
+                 Q::MPIStateArray, direction=EveryDirection())
   grid = dg.grid
   topology = grid.topology
   nrealelem = length(topology.realelems)
@@ -306,21 +308,21 @@ function cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
     Nq = N + 1
     Nqk = dim == 2 ? 1 : Nq
     device = grid.vgeo isa Array ? CPU() : CUDA()
-    pointwisecfl = similar(grid.vgeo, Nq^dim, nrealelem)
+    pointwise_courant = similar(grid.vgeo, Nq^dim, nrealelem)
     @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
             Grids.knl_min_neighbor_distance!(Val(N), Val(dim), direction,
-                                             pointwisecfl, grid.vgeo,
+                                             pointwise_courant, grid.vgeo,
                                              topology.realelems))
     @launch(device, threads=(Nq*Nq*Nqk,), blocks=nrealelem,
-            knl_local_cfl!(m, Val(dim), Val(N), pointwisecfl, localcfl,
-                           Q.data, dg.auxstate.data, dg.diffstate.data,
-                           topology.realelems))
-    loccflmax = maximum(pointwisecfl)
+            knl_local_courant!(m, Val(dim), Val(N), pointwise_courant,
+                               local_courant, Q.data, dg.auxstate.data,
+                               dg.diffstate.data, topology.realelems))
+    rank_courant_max = maximum(pointwise_courant)
   else
-    loccflmax = typemin(eltype(Q))
+    rank_courant_max = typemin(eltype(Q))
   end
 
-  MPI.Allreduce(loccflmax, max, topology.mpicomm)
+  MPI.Allreduce(rank_courant_max, max, topology.mpicomm)
 end
 
 function copy_stack_field_down!(dg::DGModel, m::BalanceLaw,
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -276,6 +276,46 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
   end
 end
 
+"""
+    cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+        direction=EveryDirection())
+
+Returns an approximation of the CFL number `(c * Δt)/Δx` where `c` is the wave
+speed of interest.
+
+The function `wavespeed_dt` should compute `c * Δt` given the `state`, `aux`
+and `diffusive` variables.  The `direction` controls which reference
+directions are considered when computing the minimum node distance `Δx`.
+"""
+function cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+             direction=EveryDirection())
+  grid = dg.grid
+  topology = grid.topology
+  nrealelem = length(topology.realelems)
+
+  if nrealelem > 0
+    N = polynomialorder(dg.grid)
+    dim = dimensionality(grid)
+    Nq = N + 1
+    Nqk = dim == 2 ? 1 : Nq
+    device = grid.vgeo isa Array ? CPU() : CUDA()
+    local_cfl = similar(grid.vgeo, Nq^dim, nrealelem)
+    @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
+            Grids.knl_min_neighbor_distance!(Val(N), Val(dim), direction,
+                                             local_cfl, grid.vgeo,
+                                             topology.realelems))
+    @launch(device, threads=(Nq*Nq*Nqk,), blocks=nrealelem,
+            knl_local_cfl!(m, Val(dim), Val(N), local_cfl, wavespeed_dt,
+                           Q.data, dg.auxstate.data, dg.diffstate.data,
+                           topology.realelems))
+    locmax = maximum(local_cfl)
+  else
+    locmax = typemin(eltype(Q))
+  end
+
+  MPI.Allreduce(locmax, max, topology.mpicomm)
+end
+
 function copy_stack_field_down!(dg::DGModel, m::BalanceLaw,
                                 auxstate::MPIStateArray, fldin, fldout)
 
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -287,7 +287,7 @@ The function `wavespeed_dt` should compute `c * Δt` given the `state`, `aux`
 and `diffusive` variables.  The `direction` controls which reference
 directions are considered when computing the minimum node distance `Δx`.
 """
-function cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+function cfl(wavespeed_dt::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
              direction=EveryDirection())
   grid = dg.grid
   topology = grid.topology
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -277,17 +277,24 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
 end
 
 """
-    cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+    cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
         direction=EveryDirection())
 
-Returns an approximation of the CFL number `(c * Δt)/Δx` where `c` is the wave
-speed of interest.
-
-The function `wavespeed_dt` should compute `c * Δt` given the `state`, `aux`
-and `diffusive` variables.  The `direction` controls which reference
+Returns the maximum of the evaluation of the function `localcfl` pointwise
+throughout the domain.  The function `localcfl` is given an approximation of
+the local node distance `Δx`.  The `direction` controls which reference
 directions are considered when computing the minimum node distance `Δx`.
+
+An example `localcfl` function is
+
+    function localcfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars,
+                      Δx)
+      return Δt * cmax / Δx
+    end
+
+where `Δt` is the time step size and `cmax` is the maximum speed of the model.
 """
-function cfl(wavespeed_dt::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+function cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
              direction=EveryDirection())
   grid = dg.grid
   topology = grid.topology
@@ -299,21 +306,21 @@ function cfl(wavespeed_dt::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArra
     Nq = N + 1
     Nqk = dim == 2 ? 1 : Nq
     device = grid.vgeo isa Array ? CPU() : CUDA()
-    local_cfl = similar(grid.vgeo, Nq^dim, nrealelem)
+    pointwisecfl = similar(grid.vgeo, Nq^dim, nrealelem)
     @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
             Grids.knl_min_neighbor_distance!(Val(N), Val(dim), direction,
-                                             local_cfl, grid.vgeo,
+                                             pointwisecfl, grid.vgeo,
                                              topology.realelems))
     @launch(device, threads=(Nq*Nq*Nqk,), blocks=nrealelem,
-            knl_local_cfl!(m, Val(dim), Val(N), local_cfl, wavespeed_dt,
+            knl_local_cfl!(m, Val(dim), Val(N), pointwisecfl, localcfl,
                            Q.data, dg.auxstate.data, dg.diffstate.data,
                            topology.realelems))
-    locmax = maximum(local_cfl)
+    loccflmax = maximum(pointwisecfl)
   else
-    locmax = typemin(eltype(Q))
+    loccflmax = typemin(eltype(Q))
   end
 
-  MPI.Allreduce(locmax, max, topology.mpicomm)
+  MPI.Allreduce(loccflmax, max, topology.mpicomm)
 end
 
 function copy_stack_field_down!(dg::DGModel, m::BalanceLaw,
--- a/src/DGmethods/DGmodel_kernels.jl
+++ b/src/DGmethods/DGmodel_kernels.jl
@@ -1080,8 +1080,8 @@ function knl_nodal_update_aux!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, f!, Q,
   end
 end
 
-function knl_local_cfl!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, local_cfl,
-                        wavespeed_dt, Q, auxstate, diffstate,
+function knl_local_cfl!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, pointwisecfl,
+                        localcfl, Q, auxstate, diffstate,
                         elems) where {dim, N}
   FT = eltype(Q)
   nstate = num_state(bl,FT)
@@ -1112,11 +1112,12 @@ function knl_local_cfl!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, local_cfl,
         l_diff[s] = diffstate[n, s, e]
       end
 
-      c_dt = wavespeed_dt(bl, Vars{vars_state(bl,FT)}(l_Q),
-                          Vars{vars_aux(bl,FT)}(l_aux),
-                          Vars{vars_diffusive(bl,FT)}(l_diff))
+      Δx = pointwisecfl[n, e]
+      c = localcfl(bl, Vars{vars_state(bl,FT)}(l_Q),
+                       Vars{vars_aux(bl,FT)}(l_aux),
+                       Vars{vars_diffusive(bl,FT)}(l_diff), Δx)
 
-      local_cfl[n, e] = c_dt / local_cfl[n, e]
+      pointwisecfl[n, e] = c
     end
   end
 end
--- a/src/DGmethods/DGmodel_kernels.jl
+++ b/src/DGmethods/DGmodel_kernels.jl
@@ -1080,6 +1080,47 @@ function knl_nodal_update_aux!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, f!, Q,
   end
 end
 
+function knl_local_cfl!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, local_cfl,
+                        wavespeed_dt, Q, auxstate, diffstate,
+                        elems) where {dim, N}
+  FT = eltype(Q)
+  nstate = num_state(bl,FT)
+  nviscstate = num_diffusive(bl,FT)
+  nauxstate = num_aux(bl,FT)
+
+  Nq = N + 1
+
+  Nqk = dim == 2 ? 1 : Nq
+
+  Np = Nq * Nq * Nqk
+
+  l_Q = MArray{Tuple{nstate}, FT}(undef)
+  l_aux = MArray{Tuple{nauxstate}, FT}(undef)
+  l_diff = MArray{Tuple{nviscstate}, FT}(undef)
+
+  @inbounds @loop for e in (elems; blockIdx().x)
+    @loop for n in (1:Np; threadIdx().x)
+      @unroll for s = 1:nstate
+        l_Q[s] = Q[n, s, e]
+      end
+
+      @unroll for s = 1:nauxstate
+        l_aux[s] = auxstate[n, s, e]
+      end
+
+      @unroll for s = 1:nviscstate
+        l_diff[s] = diffstate[n, s, e]
+      end
+
+      c_dt = wavespeed_dt(bl, Vars{vars_state(bl,FT)}(l_Q),
+                          Vars{vars_aux(bl,FT)}(l_aux),
+                          Vars{vars_diffusive(bl,FT)}(l_diff))
+
+      local_cfl[n, e] = c_dt / local_cfl[n, e]
+    end
+  end
+end
+
 """
     knl_indefinite_stack_integral!(::Val{dim}, ::Val{N}, ::Val{nstate},
                                             ::Val{nauxstate}, ::Val{nvertelem},
--- a/src/DGmethods/DGmodel_kernels.jl
+++ b/src/DGmethods/DGmodel_kernels.jl
@@ -1080,9 +1080,9 @@ function knl_nodal_update_aux!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, f!, Q,
   end
 end
 
-function knl_local_cfl!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, pointwisecfl,
-                        localcfl, Q, auxstate, diffstate,
-                        elems) where {dim, N}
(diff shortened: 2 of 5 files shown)


test/DGmethods/cfl.jl
```diff
@@ -0,0 +1,165 @@
...
+
+        grid = DiscontinuousSpectralElementGrid(topl,
+                                                FloatType = FT,
+                                                DeviceArray = ArrayType,
+                                                polynomialorder = N,
+                                                meshwarp = warpfun)
+
+        ξ = referencepoints(grid)
+        hmnd = (ξ[2] - ξ[1]) / (2Neh)
+        vmnd = (ξ[2] - ξ[1]) / (2Nev)
```
jkozdon: Could be removed.

jkozdon:
bors try

lcw:
bors try+

lcw:
bors try-

lcw:
bors try+

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,47 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+            Grids.knl_min_neighbor_distance!(Val(N), Val(dim), direction,
+                                             local_cfl, grid.vgeo,
+                                             topology.realelems))
+    @launch(device, threads=(Nq*Nq*Nqk,), blocks=nrealelem,
+            knl_local_cfl!(m, Val(dim), Val(N), local_cfl, wavespeed_dt,
+                           Q.data, dg.auxstate.data, dg.diffstate.data,
+                           topology.realelems))
+    locmax = maximum(local_cfl)
+  else
+    locmax = typemax(eltype(Q))
```
simonbyrne: when does this branch get used?

lcw: When an MPI rank doesn't have any elements.

test/DGmethods/cfl.jl
```diff
@@ -0,0 +1,149 @@
...
+                           Gravity(),
+                           PeriodicBC(),
+                           initialcondition!)
+
+        dg = DGModel(model, grid, Rusanov(), CentralNumericalFluxDiffusive(),
+                     CentralGradPenalty())
+
+        Δt = FT(1//2)
+
+        function wavespeed_dt(m::AtmosModel, state::Vars, aux::Vars,
```
simonbyrne: should we make this a function in `balancelaw.jl` and move the method into `AtmosModel.jl`?

lcw: I don't think so since.  As I understand it, @fxgiraldo wants to compute the effective CFL for various wave speed functions, not just one authoritative one.

src/DGmethods/DGmodel.jl
simonbyrne: then should it be `typemin` (since otherwise this will become the result from `Allreduce`)?

test/DGmethods/cfl.jl
simonbyrne: Ah, ok. We could make this the default though.

lcw: Sure, I am open to that.

smarras79:
Hi @lcw I may be missing something, but isn't this what  [#612 ](https://github.com/climate-machine/CLIMA/pull/612) already does?

jkozdon:
It is similar, but has some advantages: more general (does not assume atmos physics nor mesh stretching being uniform in vertical and horizontal), works on the gpu with minimal data transfer, and works when there are no elements on an MPI rank.

@fxgiraldo needed something like this for his testing, so @lcw realized it might be useful for others so he made general. If others do not need it we can throw it away when @fxgiraldo is done if #612 is the preferred approach.

smarras79:
Then why not including the small differences into [#612 ](https://github.com/climate-machine/CLIMA/pull/612) instead of duplicating a similar calculation in two different modules?

test/DGmethods/cfl.jl
tapios: What will primarily need in the end is the CFL number based on the advective velocity. The speed of sound only varies by ~20% in the atmosphere, so the acoustic Courant number can be computed at initialization time and will not vary much. 

It would help to add a diffusive CFL number too (can easily be accomplished with this set of functions by making the input more general than wavespeed * dt). This will be limiting in many cases.

fxgiraldo:
I was having issues with  #612 so in discussion with how to get something up and running for my testing, @lcw offered to write something for me.  We can discuss how best to proceed.  I need to understand how they both work but my understanding is that #618 is a bit more general.

smarras79:
> I was having issues with #612 so in discussion with how to get something up and running for my testing, @lcw offered to write something for me. We can discuss how best to proceed. I need to understand how they both work but my understanding is that #618 is a bit more general.

Then I would keep #618 but fully remove #612 to avoid confusion in the future. 
I can close #612  it if everyone agrees.

tapios:
> > I was having issues with #612 so in discussion with how to get something up and running for my testing, @lcw offered to write something for me. We can discuss how best to proceed. I need to understand how they both work but my understanding is that #618 is a bit more general.
> 
> Then I would keep #618 but fully remove #612 to avoid confusion in the future.
> I can close #612 if everyone agrees.

It looks like #618 may perform better and hence may be preferable. But we need the diffusive Courant number too. So my suggestion: add functionality for diffusive Courant number to #618, and complete the structure around it so that advective and diffusive Courant numbers in the horizontal and vertical are returned (borrowing from #612).

tapios:
That is: Move some of what is currently in the test into the model core (Atmos model of DG methods), so that we have an option to have complete (horizontal vs vertical, advective vs diffusive) Courant number monitoring in the model. From that we can build adaptive timestepping eventually.

jkozdon:
> That is: Move some of what is currently in the test into the model core (Atmos model of DG methods), so that we have an option to have complete (horizontal vs vertical, advective vs diffusive) Courant number monitoring in the model. From that we can build adaptive timestepping eventually.

I believe @lcw intent was to just add an example / test of the functionality. Guessing that some folks using the atmos and ocean models should decide on the actual outward facing interface. that said what @lcw added will work for any speed of interest.

My recommendation would be to merge this then have the physics teams play with the interface to find what they’d like.

tapios:
> That is: Move some of what is currently in the test into the model core (Atmos model of DG methods), so that we have an option to have complete (horizontal vs vertical, advective vs diffusive) Courant number monitoring in the model. From that we can build adaptive timestepping eventually.
> I believe @lcw intent was to just add an example / test of the functionality. Guessing that some folks using the atmos and ocean models should decide on the actual outward facing interface. that said what @lcw added will work for any speed of interest. My recommendation would be to merge this then have the physics teams play with the interface to find what they’d like.

I think a better way to do this is to build out the functionality completely and then merge it. This is not a physics issue, but a numerics issues, i.e., it’s better if you guys own it and integrate it into the code base so others can use it easily for diagnostics, without having to add their own code.

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,47 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+directions are considered when computing the minimum node distance `Δx`.
+"""
+function cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
+             direction=EveryDirection())
+  grid = dg.grid
+  topology = grid.topology
+  nrealelem = length(topology.realelems)
+
+  if nrealelem > 0
+    N = polynomialorder(dg.grid)
```
charleskawczynski:
```suggestion
    N = polynomialorder(grid)
```

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,47 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+        direction=EveryDirection())
+
+Returns an approximation of the CFL number `(c * Δt)/Δx` where `c` is the wave
+speed of interest.
+
+The function `wavespeed_dt` should compute `c * Δt` given the `state`, `aux`
+and `diffusive` variables.  The `direction` controls which reference
+directions are considered when computing the minimum node distance `Δx`.
+"""
+function cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
```
charleskawczynski:
```suggestion
function cfl(wavespeed_dt::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
```

test/DGmethods/cfl.jl
```diff
@@ -0,0 +1,149 @@
...
+  T∞::FT = 300
+  translation_speed::FT = 150
+  translation_angle::FT = pi / 4
+  α = translation_angle
+  u∞ = SVector(translation_speed * cos(α), translation_speed * sin(α), 0)
+
+  u = u∞
+  T = T∞
+  # adiabatic/isentropic relation
+  p = p∞ * (T / T∞) ^ (FT(1) / kappa_d)
```
charleskawczynski: @tapios I know that this is just a test, but maybe we can add `struct IsentropicProcess end` and relevant functions to MoistThermo (`pressure(p∞,T,T∞,::IsentropicProcess) p∞ * (T / T∞) ^ (FT(1) / kappa_d)`).

charleskawczynski:
This looks great to me. It makes more sense for this computation to be located in the `DGmodel_kernels` layer.

test/DGmethods/cfl.jl
tapios: Sure. That would be good. We'll need them also for some aspects of the boundary conditions (see design doc), then with the proper moist generalizations where `1/kappa_d` becomes `c_pm/R_m`.

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,47 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
           knl_nodal_update_aux!(m, Val(dim), Val(polyorder), f!,
                           Q.data, dg.auxstate.data, dg.diffstate.data, t,
                           topology.realelems))
+
+end
+
+"""
+    cfl(wavespeed_dt, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
```
tapios: A detail: I wouldn't call this wavespeed. Most of the time, it will not be a wavespeed but an advective velocity (e.g., jet speed).

test/DGmethods/cfl.jl
tapios: You'd want to be a bit careful with `isentropic`, though. These relations really only hold for *unsaturated* isentropic processes.

src/DGmethods/DGmodel.jl
lcw: I generalized this callback so that it computes the local cfl constant directly given the local minimum node spacing Δx.  This should allow for the computation of hyperbolic and diffusive cfl constants.

lcw:
bors try

charleskawczynski:
Should we add a complementary function to this PR that computes `Δt` given a max CFL?

simonbyrne:
Is there any functionality missing from #612? What do we need for the diffusive Courant number?

charleskawczynski:
> Is there any functionality missing from #612? What do we need for the diffusive Courant number?

#612 computes `dt_inout`, which updates `dt` based on the max CFL. These kernels could be very easily modified/duplicated to do the same thing--which I'd recommend over #612 anyway since these kernels are more elegant. I'd be fine with opening a separate PR, or just updating #612 to look more like these kernels if that's easier.

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,54 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
           knl_nodal_update_aux!(m, Val(dim), Val(polyorder), f!,
                           Q.data, dg.auxstate.data, dg.diffstate.data, t,
                           topology.realelems))
+
+end
+
+"""
+    cfl(localcfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
```
tapios:
```suggestion
    cfl(local_cfl::Function, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
```
to improve readability

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,54 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+        direction=EveryDirection())
+
+Returns the maximum of the evaluation of the function `localcfl` pointwise
+throughout the domain.  The function `localcfl` is given an approximation of
+the local node distance `Δx`.  The `direction` controls which reference
+directions are considered when computing the minimum node distance `Δx`.
+
+An example `localcfl` function is
+
+    function localcfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars,
```
tapios:
```suggestion
    function local_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars,
```

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,54 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+                      Δx)
+      return Δt * cmax / Δx
+    end
+
+where `Δt` is the time step size and `cmax` is the maximum speed of the model.
```
tapios:
```suggestion
where `Δt` is the time step size and `cmax` is the maximum flow speed in the model.
```

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,54 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+  topology = grid.topology
+  nrealelem = length(topology.realelems)
+
+  if nrealelem > 0
+    N = polynomialorder(grid)
+    dim = dimensionality(grid)
+    Nq = N + 1
+    Nqk = dim == 2 ? 1 : Nq
+    device = grid.vgeo isa Array ? CPU() : CUDA()
+    pointwisecfl = similar(grid.vgeo, Nq^dim, nrealelem)
```
tapios:
```suggestion
    pointwise_cfl = similar(grid.vgeo, Nq^dim, nrealelem)
```

src/DGmethods/DGmodel.jl
```diff
@@ -242,6 +242,54 @@ function nodal_update_aux!(f!, dg::DGModel, m::BalanceLaw, Q::MPIStateArray,
...
+    @launch(device, threads=(Nq, Nq, Nqk), blocks=nrealelem,
+            Grids.knl_min_neighbor_distance!(Val(N), Val(dim), direction,
+                                             pointwisecfl, grid.vgeo,
+                                             topology.realelems))
+    @launch(device, threads=(Nq*Nq*Nqk,), blocks=nrealelem,
+            knl_local_cfl!(m, Val(dim), Val(N), pointwisecfl, localcfl,
+                           Q.data, dg.auxstate.data, dg.diffstate.data,
+                           topology.realelems))
+    loccflmax = maximum(pointwisecfl)
```
tapios:
```suggestion
    loc_cfl_max = maximum(pointwisecfl)
```

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -1037,6 +1037,48 @@ function knl_nodal_update_aux!(bl::BalanceLaw, ::Val{dim}, ::Val{N}, f!, Q,
...
+
+      @unroll for s = 1:nauxstate
+        l_aux[s] = auxstate[n, s, e]
+      end
+
+      @unroll for s = 1:nviscstate
+        l_diff[s] = diffstate[n, s, e]
+      end
+
+      Δx = pointwisecfl[n, e]
```
tapios:
```suggestion
      Δx = pointwise_cfl[n, e]
```

test/DGmethods/cfl.jl
```diff
@@ -0,0 +1,149 @@
...
+                           Gravity(),
+                           PeriodicBC(),
+                           initialcondition!)
+
+        dg = DGModel(model, grid, Rusanov(), CentralNumericalFluxDiffusive(),
+                     CentralGradPenalty())
+
+        Δt = FT(1//2)
+
+        function localcfl(m::AtmosModel, state::Vars, aux::Vars,
```
tapios: I suggest moving this out of the test and into `DGmethods`, and define `local_cfl` functions for advective terms (i.e., like here, but without the `soundspeed` term) and diffusive terms, with direction arguments for vertical and horizontal.

tapios:
> Is there any functionality missing from #612? What do we need for the diffusive Courant number?

I suggest building on this: move the definition of `local_cfl` numbers out of tests, into `DGMethods`, and build this out to compute advective (with and without soundspeed) and diffusive CFL numbers, for either vertical or horizontal directions. It would be good for users to easily be able to access any of these Courant numbers.

simonbyrne:
> #612 computes `dt_inout`, which updates `dt` based on the max CFL.

Isn't this just `cfl_max / (result / dt)`

charleskawczynski:
> Isn't this just `cfl_max / (result / dt)`

Yes, I'm pretty sure we can just re-use this, it may be nice to at least write a wrapper function, though.

simonbyrne:
It might make sense to pass the `direction` object through, so that we can dispatch on it to handle the directionality of the velocity, i.e.
```julia
function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::EveryDirection)
  u = state.ρu/state.ρ
  return Δt * (norm(u) + soundspeed(m.moisture, m.orientation, state, aux)) / Δx
end

function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::VerticalDirection)
  u = state.ρu/state.ρ
  n = vertical_unit_vector(m.orientation, aux)
  return Δt * (abs(dot(n, u)) + soundspeed(m.moisture, m.orientation, state, aux)) / Δx
end

function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::HorizontalDirection)
  u = state.ρu/state.ρ
  n = vertical_unit_vector(m.orientation, aux)
  return Δt * (norm(u .- dot(n, u) .* n) + soundspeed(m.moisture, m.orientation, state, aux)) / Δx
end
```
and it sounds like we would want a variant without `soundspeed` as well (I don't quite understand when it should or should not be incorporated, so someone who knows more on this should probably advise).

Also I'm not sure about the diffusive case: is that based on the momentum flux tensor? In that case, would we need to pass in `Δx` as a vector?

charleskawczynski:
How about
```julia
function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::EveryDirection, u_sound)
  u = state.ρu/state.ρ
  return Δt * (norm(u) + u_sound) / Δx
end

function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::VerticalDirection, u_sound)
  u = state.ρu/state.ρ
  n = vertical_unit_vector(m.orientation, aux)
  return Δt * (abs(dot(n, u)) + u_sound) / Δx
end

function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, ::HorizontalDirection, u_sound)
  u = state.ρu/state.ρ
  n = vertical_unit_vector(m.orientation, aux)
  return Δt * (norm(u .- dot(n, u) .* n) + u_sound) / Δx
end

function advective_cfl(m, state, aux, diffusive, Δx, direction,
              u_sound=soundspeed(m.moisture, m.orientation, state, aux))
  return advective_cfl(m, state, aux, diffusive, Δx, u_sound, direction)
end
```
?

tapios:
Any of the above options for the advective Courant number seem fine to me. Here's a catalog of what we need:

- [ ] Advective Courant number in the horizontal and vertical (just based on advective velocity, without sound speed). We need this for stability monitoring for explicit timesteppers in directions where the sound waves are treated implicitly or are treated with a higher-rate explicit time step.
- [ ] Acoustic Courant number in the horizontal and vertical. This is based on the sum of the advective velocity and sound speed, as above. We need this for stability monitoring for explicit timesteppers where sound waves are included in the explicit timestep. 
- [ ] Diffusive Courant number in the horizontal and vertical (though we primarily need vertical, where diffusion can be limiting): `D .* Δt ./ Δx.^2)` where `D` is the diffusivity or viscosity for the quantity of interest (water, energy, momentum--it could be any of them; this is not restricted to the momentum flux tensor). We need this to monitor stability of explicit solves for diffusion (especially vertical diffusion can limit timesteps in our models).

As for terminology, the term Courant number for this nondimensional number is commonly used. Courant-Friedrichs-Lewy (CFL) is used for the stability condition; less commonly it is also for the number itself (e.g., https://en.wikipedia.org/wiki/Courant–Friedrichs–Lewy_condition). I've seen Courant number more than CFL for the number itself, so Courant alone might be preferable.

tapios:
> > Also I'm not sure about the diffusive case: is that based on the momentum flux tensor? In that case, would we need to pass in `Δx` as a vector?
> 
> Actually it occurs to me that we might be able to handle the directionality using this as well: e.g. if `Δx` was a vector then we could define:
> 
> ```julia
> function advective_cfl(m::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, Δx, u_sound)
>   u = state.ρu/state.ρ
>   return Δt * (maximum(u ./ Δx)  + u_sound)
> end
> ```

Ok in principle. But you'd need to include the soundspeed in the velocity before dividing by `Δx`, i.e., `return Δt * (maximum(abs(u) + u_sound)./ Δx)`

tapios:
@fxgiraldo It would be good to have your input here. We need to build this out so that (a) we can monitor stability, and (b) build adaptive timestepping on top of it.

fxgiraldo:
I agree on the points already made: what we are computing is called the Courant number (CFL is the condition on stability). And, yes, we need Courant numbers based on the advective speed, advective + acoustic speed (for horizontal and vertical) and based diffusion. The Courant number is a non-dimensional parameter so since speed has m/s units then we need to multiply by dt (secs) and divide by dx (m), dy, dz, etc.  For the diffusive case since kinematic viscosity has units of m^2/s then we need to multiply by dt (s) and divide by dx^2 (m^2).

Once we develop guidance on the maximum allowable Courant number for a given method, then we can use this function to compute DT given the mesh spacing and the speeds and viscosities.

simonbyrne:
@fxgiraldo I have two questions:

1. How is the advective Courant number defined in multiple dimensions, and on a non-isotropic grid? 
2. How is the diffusive Courant number defined for our equation set?

smarras79:
> @fxgiraldo I have two questions:
> 
> 1. How is the advective Courant number defined in multiple dimensions, and on a non-isotropic grid?
> 2. How is the diffusive Courant number defined for our equation set?

@simonbyrne I can answer this (@fxgiraldo feel free to jump in). The formulas above, even if maybe they will be coded slightly differently in the CFL PR, show you how these quantities are calculated in each direction:

```

    Courant_sound_x = dt*max_sound_speed/Dx
    Courant_sound_y = dt*max_sound_speed/Dy
    Courant_sound_z = dt*max_sound_speed/Dz

    Cour_advective_x = dt*MPI.Allreduce(sau,MPI.MAX,mpicomm)/Dx
    Cour_advective_y = dt*MPI.Allreduce(sav,MPI.MAX,mpicomm)/Dy
    Cour_advective_z = dt*MPI.Allreduce(saw,MPI.MAX,mpicomm)/Dz

    max_sdiff = MPI.Allreduce(sdiff,MPI.MAX,mpicomm)
    Cour_DIFF_x = dt*max_sdiff/Dx^2
    Cour_DIFF_y = dt*max_sdiff/Dy^2
    Cour_DIFF_z = dt*max_sdiff/Dz^2
```

slifer50:
> > @fxgiraldo I have two questions:
> > 
> > 1. How is the advective Courant number defined in multiple dimensions, and on a non-isotropic grid?
> > 2. How is the diffusive Courant number defined for our equation set?
> 
> @simonbyrne I can answer this (@fxgiraldo feel free to jump in). The formulas above, even if maybe they will be coded slightly differently in the CFL PR, show you how these quantities are calculated in each direction:
> 
> ```
> 
>     Courant_sound_x = dt*max_sound_speed/Dx
>     Courant_sound_y = dt*max_sound_speed/Dy
>     Courant_sound_z = dt*max_sound_speed/Dz
> 
>     Cour_advective_x = dt*MPI.Allreduce(sau,MPI.MAX,mpicomm)/Dx
>     Cour_advective_y = dt*MPI.Allreduce(sav,MPI.MAX,mpicomm)/Dy
>     Cour_advective_z = dt*MPI.Allreduce(saw,MPI.MAX,mpicomm)/Dz
> 
>     max_sdiff = MPI.Allreduce(sdiff,MPI.MAX,mpicomm)
>     Cour_DIFF_x = dt*max_sdiff/Dx^2
>     Cour_DIFF_y = dt*max_sdiff/Dy^2
>     Cour_DIFF_z = dt*max_sdiff/Dz^2
> ```

Just in case it isn't clear from the above lines. `max_sdiff` is a kinematic viscosity, in #612 this was initially taken to be the maximum turbulent viscosity however as @tapios said, this isn't the only viscosity to be taken into account.

simonbyrne:
thanks. My question is more how do we make this sufficiently general to work on non-Cartesian aligned grids, eg for the cube sphere. Do we need to take the tensor structure of the diffusion tensor into account?

slifer50:
> thanks. My question is more how do we make this sufficiently general to work on non-Cartesian aligned grids, eg for the cube sphere. Do we need to take the tensor structure of the diffusion tensor into account?

I don't think we need to do that in terms of `max_sdiff` since I believe it should be a scalar quantity with orientation already being in taken into account when it is computed. @smarras79 @fxgiraldo please correct me if I'm mistaken here.

simonbyrne:
sdiff in #612 seems to be just extracting one of the diagonal components of the tensor to make a scalar (or am I misunderstanding it?) We should probably do something more principled, but I would like to know what that is.

#612 also seems to make the implicit assumption that the dg grid is Cartesian aligned

slifer50:
what is extracted in #612 should be the turbulent viscosity which as far as I know should be the diffusivity we need for the diffusive courant number. It is calculated here: https://github.com/climate-machine/CLIMA/blob/5ba66c4a0ef4fa4723efb67371ba6f50864fbf70/src/Atmos/Model/turbulence.jl#L146 and its calculation takes into account the type of grid being used. It's then used to get the momentum flux tensor here: https://github.com/climate-machine/CLIMA/blob/5ba66c4a0ef4fa4723efb67371ba6f50864fbf70/src/Atmos/Model/turbulence.jl#L156.
In so far as I can tell it isn't taking a diagonal component of the tensor to make a scalar, rather it's calculating the turbulent viscosity which is multiplied by `S = symmetrize(∇u)` to get the scaled momentum flux tensor.

simonbyrne:
It's taking the 3rd component of the `diffusive` state:
https://github.com/climate-machine/CLIMA/blob/15b1d9ef7ad767bbe4b014fb2e1267df00d76aa1/src/Diagnostics/Courant.jl#L112
which at the moment is `ρτ[2,2]`:
https://github.com/climate-machine/CLIMA/blob/5ba66c4a0ef4fa4723efb67371ba6f50864fbf70/src/Atmos/Model/AtmosModel.jl#L70
(though this will change with #626).

slifer50:
> It's taking the 3rd component of the `diffusive` state:
> 
> https://github.com/climate-machine/CLIMA/blob/15b1d9ef7ad767bbe4b014fb2e1267df00d76aa1/src/Diagnostics/Courant.jl#L112
> 
> which at the moment is `ρτ[2,2]`:
> 
> https://github.com/climate-machine/CLIMA/blob/5ba66c4a0ef4fa4723efb67371ba6f50864fbf70/src/Atmos/Model/AtmosModel.jl#L70
> 
> (though this will change with #626).

No it's taking the what's set in #612 as the third component of the diffusive state array ( the diffusive state has a bunch of different components, in #612 the turbulent viscosity was placed in the diffusive state because it was simpler than making it accessible to the aux state) this is not the tensor rather it is exactly what is being calculated here:

https://github.com/climate-machine/CLIMA/blob/5ba66c4a0ef4fa4723efb67371ba6f50864fbf70/src/Atmos/Model/AtmosModel.jl#L192

simonbyrne:
Ah, it was part of #593, but didn't make it into #612. 

I would still appreciate a write up of what exactly we want, that correctly handles a non-Cartesian grid, since currently we seem to have several subtly different interpretations.

slifer50:
I think in terms of handling non cartesian grids, so long as we are sure the min-node distance functions take the geometry into consideration we won't need any particular generalizations since all the other terms we need already take geometry into account. @smarras79 can you confirm this?

smarras79:
> I think in terms of handling non cartesian grids, so long as we are sure the min-node distance functions take the geometry into consideration we won't need any particular generalizations since all the other terms we need already take geometry into account. @smarras79 can you confirm this?

Based on what @lcw did when he created #618  yes, the geometry should be taken care of correctly also when the geometry is non-cartesian.

test/DGmethods/cfl.jl
blallen: Shouldn't the definition of `local_cfl` be done in each model individually? Same as things like `flux_nondiffusive` and `diffusive!` and `source!`? That way, the `local_cfl` is specified uniquely for each model and doesn't require all models to have the same state vector.

src/DGmethods/DGmodel.jl
lcw: thanks

src/DGmethods/DGmodel.jl
lcw: thanks

src/DGmethods/DGmodel_kernels.jl
lcw: thanks

lcw:
bors try

lcw:
bors try

slifer50:
@lcw Is the user supposed to specify courant number functions in the same way it is done for the test case? If this is the case I think it would be helpful to have some presets (courant for advection, diffusion etc, while taking into account directions

lcw:
> @lcw Is the user supposed to specify courant number functions in the same way it is done for the test case? If this is the case I think it would be helpful to have some presets (courant for advection, diffusion etc)

Yes, the interface is designed so that each set of physics can define their own Courant functions.

tapios:
> > @lcw Is the user supposed to specify courant number functions in the same way it is done for the test case? If this is the case I think it would be helpful to have some presets (courant for advection, diffusion etc)
> 
> Yes, the interface is designed so that each set of physics can define their own Courant functions.

As outlined above, we need advective and diffusive Courant numbers, separately for the horizontal and vertical. Who will provide the functions?

slifer50:
@lcw @tapios I could add the preset functions if you'd like. Should these be done here on this PR or separately after this is merged.

tapios:
Thanks for the offer, @slifer50. Really appreciate it. We need the functions and the integration of the Courant number monitoring into the timestepping, so we can build adaptive timesteppers on top of this.  @simonbyrne, ideally with @fxgiraldo, should decide how best to do this.

test/DGmethods/courant.jl
```diff
@@ -0,0 +1,149 @@
...
+let
+  # boiler plate MPI stuff
+  CLIMA.init()
+  mpicomm = MPI.COMM_WORLD
+  ll = uppercase(get(ENV, "JULIA_LOG_LEVEL", "INFO"))
+  loglevel = ll == "DEBUG" ? Logging.Debug :
+  ll == "WARN"  ? Logging.Warn  :
+  ll == "ERROR" ? Logging.Error : Logging.Info
+  logger_stream = MPI.Comm_rank(mpicomm) == 0 ? stderr : devnull
+  global_logger(ConsoleLogger(logger_stream, loglevel))
```
kpamnany: [`CLIMA.init()`](https://github.com/climate-machine/CLIMA/blob/master/src/Driver/Driver.jl#L133) now does all this.

simonbyrne:
Merged as part of #712.

PR #620, opened 2020-01-13
Update turbulence models

charleskawczynski:
Updates turbulence model, broken off from #610.

Needs review.

### DIFF
--- a/src/Atmos/Model/turbulence.jl
+++ b/src/Atmos/Model/turbulence.jl
@@ -2,6 +2,8 @@
 using DocStringExtensions
 using CLIMA.PlanetParameters
 using CLIMA.SubgridScaleParameters
+using CLIMA.Mesh.Grids
+using CLIMA.Mesh.Grids: VerticalDirection, HorizontalDirection, EveryDirection
 export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss
 
 abstract type TurbulenceClosure end
@@ -21,10 +23,10 @@ function gradvariables!(::TurbulenceClosure, transform::Vars, state::Vars, aux::
 end
 
 """
-  PrincipalInvariants{FT} 
+  PrincipalInvariants{FT}
 
-Calculates principal invariants of a tensor. Returns struct with fields first,second,third 
-referring to each of the invariants. 
+Calculates principal invariants of a tensor. Returns struct with fields first,second,third
+referring to each of the invariants.
 """
 struct PrincipalInvariants{FT}
   first::FT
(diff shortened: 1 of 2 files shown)


src/Atmos/Model/turbulence.jl
```diff
@@ -63,7 +65,7 @@ end
 """
     SmagorinskyLilly <: TurbulenceClosure
 
-  § 1.3.2 in CliMA documentation 
+  § 1.3.2 in CliMA documentation
```
tapios: Either update or remove the references to the design documents. These are no longer correct.

charleskawczynski: I'll do this everywhere, we can't expect that they'll stay in sync.

src/Atmos/Model/turbulence.jl
```diff
@@ -253,20 +275,29 @@ url = {https://link.aps.org/doi/10.1103/PhysRevFluids.1.041701}
...
+vars_aux(::AnisoMinDiss,FT) = @vars(Δ::FT,ρν::FT,BR::FT,Freq::FT)
+vars_gradient(::AnisoMinDiss,FT) = @vars(θ_v::FT)
 function atmos_init_aux!(::AnisoMinDiss, ::AtmosModel, aux::Vars, geom::LocalGeometry)
   aux.turbulence.Δ = lengthscale(geom)
+  aux.turbulence.ρν = 0
+  aux.turbulence.BR = 0
+end
+function atmos_nodal_update_aux!(::AnisoMinDiss, ::AtmosModel, state::Vars, aux::Vars, diffusive::Vars, t::Real)
+  aux.turbulence.ρν = diffusive.ρν
+  aux.turbulence.BR = diffusive.turbulence.BR
```
simonbyrne: where are these used?

charleskawczynski: Just searched. It's used in diagnostics, but then the variable that is assigned to it is never used, so we can remove it.

charleskawczynski: Actually..... It _is_ used in [diagnostics](https://github.com/climate-machine/CLIMA/pull/610/files#diff-f26cd0b9cbdc8868d486fccf30ef3d2aR260-R261), which is subsequently called [here](https://github.com/climate-machine/CLIMA/pull/610/files#diff-8210716d7cccb7f5fcbb29c7b43aae59R69).

charleskawczynski: To avoid wasting significant resources, we should compute this on the fly. But computing on the fly leads to the possibility of inconsistent formulas.

We need to incorporate `AtmosModel` into the diagnostics to have consistent flow control of computed terms. Otherwise, we will continue to deal with the possibility of inconsistent formulas.

smarras79: Right, they are quantities that are written out to file for post-processing. 
They may be not necessary operationally, but we certainly need them while debugging our turbulence models.

charleskawczynski: @simonbyrne up to you whether you want to leave this there for diagnostics, or remove and update the diagnostics later.

simonbyrne: These should be implemented in diagnostics.

charleskawczynski:
Closing this since it didn't seem to be needed for DYCOMS.

PR #621, opened 2020-01-13
Attempt to fix docs

charleskawczynski:
The docs are currently disabled because they were hanging during the build. Trying to fix them here, it may need to be in steps, since some of the docs have really fallen out of sync, and the generated examples from `Literate` seem to be getting tripped up in parsing the examples. Also, `Documenter.jl` has changed a bit and some functionality seems to be broken (e.g., LaTeX equations).

I'm going to rebase once I figure out the fewest changes necessary to fix things.


ali-ramadhan:
> Also, `Documenter.jl` has changed a bit and some functionality seems to be broken (e.g., LaTeX equations).

I think Documenter.jl v0.24.0 changed the default LaTeX backend from MathJax to KaTeX but it can be changed back: https://juliadocs.github.io/Documenter.jl/stable/lib/internals/writers/#Documenter.Writers.HTMLWriter.HTML

simonbyrne:
Why is KaTeX broken? In my experience it produces nicer output.

ali-ramadhan:
I don't think it's necessarily broken, it's just missing some features that MathJax has like equation numbering: https://github.com/KaTeX/KaTeX/issues/350

I do like that it's faster than MathJax.

charleskawczynski:
Ok, @simonbyrne docs are building again.

I think it would be nice to still have equation numbering or at least some way to reference equations (the EDMF equations utilize referencing equations a lot). But the EDMF equations are just one doc page of many. Also, I'm not sure if it was the generated examples or the `checkdocs` parameter, but the docs do seem to be building a lot faster now.

simonbyrne:
@charleskawczynski can you uncomment the doc builder in azure pipelines?

charleskawczynski:
> .. can you uncomment the doc builder in azure pipelines?

@simonbyrne done.

azure-pipelines.yml
```diff
@@ -124,45 +124,45 @@ jobs:
...
-#      ./julia-$(JULIA_VERSION)/bin/julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
-#                                                                        Pkg.develop(PackageSpec(path=pwd()));
-#                                                                        Pkg.build()'
-#      ./julia-$(JULIA_VERSION)/bin/julia --project=docs/ docs/make.jl
-#    env:
-#      DOCUMENTER_KEY: $(documenter_key)
-#    displayName: 'Build and deploy docs'
+- job: Documentation
+
+ timeoutInMinutes: 60
```
simonbyrne: I think you deleted too many spaces.

PR #626, opened 2020-01-14
Make diffusive! terms linear

### LINKED ISSUE #605: diffusive! should be linear in ∇transform
After some discussion with @fxgiraldo, it seems that for the `diffusive_penalty!` to be applied correctly, we require that `diffusive!` be a linear function of `∇transform`. Any other operations should be moved into `flux_diffusive!`.

simonbyrne:
Fixes #605, overlaps with #620.

### DIFF
--- a/src/Atmos/Model/turbulence.jl
+++ b/src/Atmos/Model/turbulence.jl
@@ -7,41 +7,81 @@ export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss
 abstract type TurbulenceClosure end
 
 vars_state(::TurbulenceClosure, FT) = @vars()
-vars_gradient(::TurbulenceClosure, FT) = @vars()
-vars_diffusive(::TurbulenceClosure, FT) = @vars()
 vars_aux(::TurbulenceClosure, FT) = @vars()
 
 function atmos_init_aux!(::TurbulenceClosure, ::AtmosModel, aux::Vars, geom::LocalGeometry)
 end
 function atmos_nodal_update_aux!(::TurbulenceClosure, ::AtmosModel, state::Vars, aux::Vars, t::Real)
 end
-function diffusive!(::TurbulenceClosure, diffusive, ∇transform, state, aux, t, ν)
-end
 function gradvariables!(::TurbulenceClosure, transform::Vars, state::Vars, aux::Vars, t::Real)
 end
-
-"""
-  PrincipalInvariants{FT} 
-
-Calculates principal invariants of a tensor. Returns struct with fields first,second,third 
-referring to each of the invariants. 
-"""
-struct PrincipalInvariants{FT}
-  first::FT
-  second::FT
-  third::FT
+function diffusive!(::TurbulenceClosure, diffusive, ∇transform, state, aux, t)
 end
-function compute_principal_invariants(X::StaticArray{Tuple{3,3}})
+
+"""
+    ν, τ = turbulence_tensors(::TurbulenceClosure, state::Vars, diffusive::Vars, aux::Vars, t::Real)
+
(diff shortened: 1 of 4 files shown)


simonbyrne:
bors try

jkozdon:
bors try

src/Atmos/Model/turbulence.jl
```diff
@@ -191,34 +228,45 @@ struct Vreman{FT} <: TurbulenceClosure
...
+function diffusive!(::Vreman, diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real)
+  diffusive.turbulence.∇u = ∇transform.u
+  diffusive.turbulence.N² = dot(∇transform.turbulence.θ_v, aux.orientation.∇Φ) / aux.moisture.θ_v
 end
-function scaled_momentum_flux_tensor(m::Vreman, ρν, S)
-  (-2*ρν) * S
+
+function turbulence_tensors(m::Vreman, state::Vars, diffusive::Vars, aux::Vars, t::Real)
+  FT = eltype(state)
+  α = diffusive.∇u
```
jkozdon:
```suggestion
  α = diffusive.turbulence.∇u
```

simonbyrne:
Yeah that's where I'm stuck: will try to get it sorted tomorrow.

jkozdon:
Now that I understand the issue better, I fully expect the CI tests do you need to be updated for any test that uses dynamic viscosity.  You probably already realize that, but figured I should mention it just in case.

jkozdon:
Been thinking about this more and understanding things better about what is currently in the code, but I have a hunch that this is a (major?) contributing factor to the instability issues being seen with IMEX and Dycoms. Not sure that we can do much to address that, until we have this PR sorted out.

The reason I think this is that dynamic viscosity is getting set based on the wrong quantities, so it could either be too small or large since it is really being set by an erroneous quantity.

smarras79:
> Been thinking about this more and understanding things better about what is currently in the code, but I have a hunch that this is a (major?) contributing factor to the instability issues being seen with IMEX and Dycoms. Not sure that we can do much to address that, until we have this PR sorted out.
> 
> The reason I think this is that dynamic viscosity is getting set based on the wrong quantities, so it could either be too small or large since it is really being set by an erroneous quantity.

@jkozdon thanks for giving some thoughts on the topic. 
The viscosity that we currently use is the Smagorinsky eddy viscosity. It is based on the strain rate tensor (among other things). You can think of it as a viscosity that is only present when there is strain and is proportional to the square of the grid size (i.e. the top-hat filter width of the LES approximation). In other words, it is a physically-driver viscosity. 
Can you explain a bit more what you think may be happening?

fxgiraldo:
@smarras79, I am looking into this with @jkozdon but the issue is that for the LDG method to correctly represent the derivatives (gradients), we need that what goes into them be a linear function of the state vector (or of the primitive variables). If they are nonlinear functions, then the map G(q1 + q2) is not equal to G(q1) + G(q2) where q1 and q2 are state vector quantities and G is some map (that must be linear).

jkozdon:
@smarras7o it’s not that we cannot do this, we just need to move it to later, e.g., to the numerical flux. @simonbyrne has been working on this, I was just pointing out that it could be that this is causing some of the other numerical instability problems that have been seen recently and might be part of why IMEX isn’t working right for DYCOMS (or it might not be, I don’t know)

smarras79:
> @smarras7o it’s not that we cannot do this, we just need to move it to later, e.g., to the numerical flux. @simonbyrne has been working on this, I was just pointing out that it could be that this is causing some of the other numerical instability problems that have been seen recently and might be part of why IMEX isn’t working right for DYCOMS (or it might not be, I don’t know)

Thanks for the details, @jkozdon and @fxgiraldo

jkozdon:
closing in favor of #648

PR #627, opened 2020-01-15
Working DYCOMS (latest driver from Simone)

kpamnany:
#### This is not a real PR. ####
This is a "working" version of the DYCOMS driver being used by Simone. "Working" in the sense that it doesn't crash; @smarras79 will confirm here if the results produced are good.

The only difference between this and `master` is moist thermo. @charleskawczynski will fix this once Simone confirms that the results are good. That fix should automatically fix the driver/configurations PR as well.

One other thing: @smarras79, this driver is not using an IMEX solver?


smarras79:
> #### This is not a real PR.
> This is a "working" version of the DYCOMS driver being used by Simone. "Working" in the sense that it doesn't crash; @smarras79 will confirm here if the results produced are good.
> 
> The only difference between this and `master` is moist thermo. @charleskawczynski will fix this once Simone confirms that the results are good. That fix should automatically fix the driver/configurations PR as well.
> 
> One other thing: @smarras79, this driver is not using an IMEX solver?

Yes, this driver does use IMEX. It uses the LSRK144NiegemannDiehlBusch solver.
@kpamnany @charleskawczynski thanks for narrowing the issue down.

simonbyrne:
Isn’t LSRK144NiegemannDiehlBusch an explicit silver?

smarras79:
The code runs, but the b.c. for dycoms are incorrect. 

1) This Dirichlet b.c. should be removed:
https://github.com/climate-machine/CLIMA/blob/17f8bd63191b298ad7a04dab84b24833b7e6ddb8/src/Atmos/Model/boundaryconditions.jl#L139 

2) same here:
https://github.com/climate-machine/CLIMA/blob/17f8bd63191b298ad7a04dab84b24833b7e6ddb8/src/Atmos/Model/boundaryconditions.jl#L213

smarras79:
> Isn’t LSRK144NiegemannDiehlBusch an explicit solver?

It should be 1D Imex. (@mwarusz)

mwarusz:
@simonbyrne is correct, `LSRK144NiegemannDiehlBusch` is an explicit solver

charleskawczynski:
@smarras79 Can you validate results in `9c023d2b03c19c4d5303ac596d18891e91dc08c2`?

charleskawczynski:
@smarras79 Can you validate results in `f0e04e5cd2f52f554ba536d4070020a0e130b547`? I was able to get those results to run for 15 min.

smarras79:
> > @smarras79 Can you validate results in `f0e04e5cd2f52f554ba536d4070020a0e130b547`? I was able to get those results to run for 15 min.
> 
> Validated. It also works.

Sorry, I take it back. It blew up (instability) at 10,000 s. I am checking a few parameters now. Will let you know if I find anything.

smarras79:
> > > @smarras79 Can you validate results in `f0e04e5cd2f52f554ba536d4070020a0e130b547`? I was able to get those results to run for 15 min.
> > 
> > 
> > Validated. It also works.
> 
> Sorry, I take it back. It blew up (instability) at 10,000 s. I am checking a few parameters now. Will let you know if I find anything.

OK, it works now and the results look the best we achieved so far for this resolution. 
I had to make small changes to the parameters in the driver. 
The new driver should be the reference.  
@charleskawczynski I don't want to push to the branch: should I send it to you as a stand alone file and you incorporate it with a new PR?

charleskawczynski:
@smarras79 can you validate results in `b17ce2f`? This commit ran for 20 minutes.

charleskawczynski:
@smarras79 can you validate results in `1625e39`? This ran fine for 28 minutes. We're not going to get much closer than this!

kpamnany:
I have squashed and rebased this branch on master and pushed it [here](https://github.com/climate-machine/CLIMA/tree/kp/working-dycoms-driver). Apart from the driver itself, there are only two changes wrt `master` -- the Dirichlet b.c. has been removed for `DYCOMS_BC` per Simone's [comment](https://github.com/climate-machine/CLIMA/pull/627#issuecomment-574873551), and `maxiters` is adjusted to 5 in thermo. The driver itself has been moved (from `test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex.jl` to replace `examples/DGmethods/ex_001_dycoms.jl`).

@smarras79: please check out this latest branch and verify the results? Multi-GPU should also work there. If you are satisfied with what you see there, we can close this PR and delete this branch.

kpamnany:
Another point that's confusing me. `LSRK144NiegemannDiehlBusch` is an explicit solver method, but in the driver, it is being utilized under the IMEX solver condition (because `exp_step` is 0 [here](https://github.com/climate-machine/CLIMA/blob/kp/working-dycoms-driver/examples/DGmethods/ex_001_dycoms.jl#L388)).

Due to this, the `dt` computation is coming from the IMEX solver part and is computed as:
```dt             = Courant_number * min_node_distance(dg.grid, HorizontalDirection())/soundspeed_air(FT(290))```

Under the explicit solver part, `dt` is computed as:
```dt             = Courant_number * min_node_distance(dg.grid, VerticalDirection())/soundspeed_air(FT(340))```

So, what are the right ways to compute `dt` for IMEX solvers and explicit solvers?

**Comment by fxgiraldo** [comment]
@kpamnany, to help explain this, I have added a section to the bottom of the Numerics design documents regarding the explicit stability of some of the time-integrators used in CLIMA.  This will give you a sense of the relative difference in the stability region between, say, LSRK(14,4), LSRK(5,4), and ARK(2,3,2).  

Comments welcomed.

**Comment by kpamnany** [comment]
That is helpful, @fxgiraldo, thanks. It would be good to have these implemented with the solvers. E.g.:
```get_initial_dt(::AdditiveRungeKutta, C, ∆s_H, ∆s_V, ...)```
And:
```get_initial_dt(::LowStorageRungeKutta2N, C, ...)```

Then, it can be abstracted away from the driver, and we can simplify [this](https://github.com/climate-machine/CLIMA/blob/kp/driver/src/Driver/Driver.jl#L217-L238) bit of code.

File: test/DGmethods/compressible_Navier_Stokes/ex_001_dycoms-imex.jl [hunk 1]
```diff
@@ -0,0 +1,472 @@
...
+  T_s = FT(290.4)
+  Γ_lapse = FT(grav/cp_d)
+  Temp = LinearTemperatureProfile(T_min,T_s,Γ_lapse)
+  RelHum = FT(0)
+
+  # Model definition
+  model = AtmosModel(FlatOrientation(),
+                     HydrostaticState(Temp,RelHum),
+                     SmagorinskyLilly{}(C_smag),
+                     EquilMoist(),
```
charleskawczynski:
```suggestion
                     EquilMoist(5),
```

kpamnany:
The most recent commit that works is `1625e39` which uses a `maxiter` of 5. I've updated [this branch](https://github.com/climate-machine/CLIMA/tree/kp/working-dycoms-driver) to this working commit.

More recent commits here (that are closer to `master`) are failing.

kpamnany:
Latest driver on master, `experiments/LES/dycoms.jl` has been validated.

PR #630, opened 2020-01-16
diffusive_penalty! -> numerical_flux_gradient!

jkozdon:
In `faceviscterms!` change from requiring a diffusive penalty function (strong flux)
```
  F^{*} - F^{-}
```
to requiring the (weak) numerical flux
```
   F^{*} 
```
The subtraction of `F^{-}` then happens in the calling kernel.

This unifies `faceviscterms!` with` facevisc!` and should simplify flux
based boundary conditions.

### DIFF
--- a/src/Ocean/Model/HydrostaticBoussinesqModel.jl
+++ b/src/Ocean/Model/HydrostaticBoussinesqModel.jl
@@ -11,7 +11,7 @@ using ..PlanetParameters: grav
 using ..Mesh.Filters: CutoffFilter, apply!, ExponentialFilter
 using ..Mesh.Grids: polynomialorder, VerticalDirection
 
-using ..DGmethods.NumericalFluxes: Rusanov, CentralGradPenalty,
+using ..DGmethods.NumericalFluxes: Rusanov, CentralNumericalFluxGradient,
                                    CentralNumericalFluxDiffusive,
                                    CentralNumericalFluxNonDiffusive
 
@@ -296,7 +296,8 @@ end
 end
 
 @inline function ocean_boundary_state!(::HBModel, ::CoastlineFreeSlip,
-                                       ::Union{Rusanov, CentralGradPenalty},
+                                       ::Union{Rusanov,
+                                               CentralNumericalFluxGradient},
                                        Q⁺, A⁺, n⁻, Q⁻, A⁻, t)
   return nothing
 end
@@ -313,7 +314,8 @@ end
 end
 
 @inline function ocean_boundary_state!(::HBModel, ::CoastlineNoSlip,
-                                       ::Union{Rusanov, CentralGradPenalty},
+                                       ::Union{Rusanov,
+                                               CentralNumericalFluxGradient},
                                        Q⁺, A⁺, n⁻, Q⁻, A⁻, t)
   Q⁺.u = -Q⁻.u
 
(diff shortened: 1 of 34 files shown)


jkozdon:
bors try

jkozdon:
@fxgiraldo This should be what we were talking about earlier today.

fxgiraldo:
Thanks @jkozdon !  With this we can impose flux bcs directly and only have to define the numerical flux for either the diffusive or non-diffusive terms.

simonbyrne:
> Thanks @jkozdon ! With this we can impose flux bcs directly and only have to define the numerical flux for either the diffusive or non-diffusive terms.

Could you explain how this might work in the case of the atmospheric BCs in the design docs?

src/Ocean/Model/HydrostaticBoussinesqModel.jl
```diff
@@ -313,7 +314,8 @@ end
 end
 
 @inline function ocean_boundary_state!(::HBModel, ::CoastlineNoSlip,
-                                       ::Union{Rusanov, CentralGradPenalty},
+                                       ::Union{Rusanov,
+                                               CentralNumericalFluxGradient},
                                        Q⁺, A⁺, n⁻, Q⁻, A⁻, t)
   Q⁺.u = -Q⁻.u
```
blallen: Do these boundary state functions need to be updated to reflect the fact that we are now calculating `F^{*}` instead of `F^{*} - F^{-}`?

jkozdon: No. The state currently specified should be fine. We're just moving `F^{-}` out of the flux function into the kernel. So this should not effect any current functionality or results. The hope is that this will simplify more general bcs later, since the user doesn't need to figure out what the plus side should be if that's tricky (e.g., you do not need to invert a complicated boundary condition).

jkozdon:
> Could you explain how this might work in the case of the atmospheric BCs in the design docs?

There are really two things here:

1. The models previously needed to know that a strong form was used for the DG gradient equation
2. The model had to come up with a plus-side state to feed into the numerical flux routine that gave back the desired value

An example of 2 is the the no-slip boundary condition
```
u = 0
```
Currently we support setting the plus-side state and then calling the numerical flux. So in the gradient numerical flux you need to set
```
u^+ = -u^-
```
so that after averaging you get `u^+ + u^- = 0` instead of just saying `u = 0`. (For the non-diffusive fluxes you still want to set `u^+ = 0` then call the numerical flux routine, though for the diffusive flux you can just set the flux to `0` directly since we are not specifying bcs on the boundary gradient).

There were some kludgy ways to get around 2 without 1, but it just seemed easer to just do the cleaner fix.

To actually make it all cleaner, the default [`numerical_boundary_flux_gradient!`](https://github.com/climate-machine/CLIMA/blob/449c4279ad5acb463b95e1d4b606d940885e7b80/src/DGmethods/NumericalFluxes.jl#L56-L73) will need to be specialized for different balance laws. On approach would be to pack the `bctype` into a `Val` so that the default (setting the `boundary_state` is called sometimes and setting the flux directly others).

A more radical idea is to just get rid of `boundary_state` all together for all numerical flux, and just say that the user models have to do what they want for all cases.

simonbyrne:
> A more radical idea is to just get rid of `boundary_state` all together for all numerical flux, and just say that the user models have to do what they want for all cases.

We can do this by making it easier to overload `numerical_boundary_flux_nondiffusive!`: I'm working on a PR to make this easier.

jkozdon:
We can do this by making it easier to overload numerical_boundary_flux_nondiffusive!: I'm working on a PR to make this easier.

Great! The big things this helps is that the overloaded function doesn’t need to bother with the subtraction of the minus side flux, which seems cleaner to me.

jkozdon:
This PR needs #626 to go through first

jkozdon:
To elaborate: this should not change the CI results, but it does because `diffusive` is a nonlinear function so `D(state^* - state^-) \ne D(state^*) - D(state^-)` which is required here as discussed in #605.

jkozdon:
bors try

jkozdon:
now depends on #648

PR #631, opened 2020-01-16
Move Vars wrappers out from NumericalFluxes to DGmodel_kernels.jl

simonbyrne:


simonbyrne:
This should make it easier for BalanceLaws to define boundary numerical fluxes directly.

blallen:
Hi @simonbyrne, thanks for doing this; it makes the `NumericalFlux.jl` kernels a lot clearer!

Do we need to separately call the `Vars` wrapper in each flux function call though? Is it possible to just use the wrapper before the loops, or failing that, at the start of the innermost part of each loop?

jkozdon:
When this is merged I can rebase #630

simonbyrne:
> Do we need to separately call the `Vars` wrapper in each flux function call though? Is it possible to just use the wrapper before the loops, or failing that, at the start of the innermost part of each loop?

There shouldn't be a performance difference, I only kept it this way to make clear what works with `Vars` and what works with arrays.

simonbyrne:
bors try

test/LinearSolvers/poisson.jl
```diff
@@ -51,21 +51,17 @@ end
...
-  tau = FT(1.0)
-  for s = 1:nstate
-    @inbounds F[s] -= tau * (QM[1] - QP[1])
-  end
+    bl, fluxᵀn, n, state⁻, diff⁻, aux⁻, state⁺, diff⁺, aux⁺, t)
+
+  Fᵀn = parent(fluxᵀn)
+  FT = eltype(Fᵀn)
+  tau = FT(1)
+  Fᵀn .-= tau*(parent(state⁻) - parent(state⁺))
```
simonbyrne: @jkozdon is this sign correct? It's the opposite of the sign in the Rusanov flux . Or is this to do with being a diffusive numerical flux?

jkozdon:
I'd just cherry-pick this commit to fix the advection-diffusion. I think we are hitting machine epsilon with Float32 on that test

https://github.com/climate-machine/CLIMA/commit/0a4b7a3252239db0e302b1108970ce0e5141f3cd

test/LinearSolvers/poisson.jl
jkozdon: I believe that @mwarusz implemented `-\Del u = 0`, so the sign is correct.

blallen:
> > Do we need to separately call the `Vars` wrapper in each flux function call though? Is it possible to just use the wrapper before the loops, or failing that, at the start of the innermost part of each loop?
> 
> There shouldn't be a performance difference, I only kept it this way to make clear what works with `Vars` and what works with arrays.

Just for my education, this means that if you did `l_F = Vars{vars_state(bl,FT)}(l_F)` before 
https://github.com/climate-machine/CLIMA/blob/d5f7d2059dd5e1a4345292a9c65be83a4e8a3564/src/DGmethods/DGmodel_kernels.jl#L593 
then this line would fail? 

If that's true, it might still be worth it to do something like `vars_F = Vars{vars_state(bl,FT)}(l_F)` before L552 and just pass `vars_F` in L533, L558, L576, L582 to improve the code readability.

jkozdon:
Third option would be to not make the kernels ever know about vars (only arrays) and define wrapper functions to do this along the lines of:
```julia
numerical_flux_diffusive!(nf, bl, F::AbstractArray, ...) = numerical_flux_diffusive(nf, bl, Vars{vars_state(bl,FT)}(l_F), ...)
```
That way we do not have hidden aliasing of the arrays and vars.

jkozdon:
Of course downside of that is even more indirection and possibly making things even more opaque(?).

If we did do this, probably should be done across all the kernels, so similarly for `flux_X`, `source`, `diffusive`, etc.

simonbyrne:
> Of course downside of that is even more indirection and possibly making things even more opaque(?).
> 
> If we did do this, probably should be done across all the kernels, so similarly for `flux_X`, `source`, `diffusive`, etc.

That is my only concern: if we're going to do this, we need to go through and rename all the variables in the kernels to be consistent.

PR #635, opened 2020-01-17
throw more informative error if timestep cannot be adjusted

simonbyrne:


simonbyrne:
Okay, this should now throw the error at the start, rather than at the end (which is presumably more useful)

Issue #393, opened 2019-08-23
Make `error` calls GPU-friendly


charleskawczynski:
There's at least one in `MoistThermo`

vchuravy:
Errors should be okay on the GPU. As long as they don't use capture objects.

charleskawczynski:
Ok, I'll change the title

vchuravy:
Since Charlie just asked me to clarify what I meant:

With a captured object I mean a mutable struct where a reference to said struct is passed to a function. This can lead to that object escaping the local function and cause the object to be heap allocated. 

As an example take:

```julia
using StaticArrays
function f(i)
  A = MArray{Tuple{3}}((0,0,0))
  return A[i]
end
```

We create a mutable struct `A` that we index into. It looks like the reference to `A` does not escape the function,
but if we call it with  an out of bounds index. 

```julia
julia> f(4)
ERROR: BoundsError: attempt to access 3-element MArray{Tuple{3},Int64,1,3} with indices SOneTo(3) at index [4]
Stacktrace:
 [1] throw_boundserror(::MArray{Tuple{3},Int64,1,3}, ::Tuple{Int64}) at ./abstractarray.jl:538
 [2] checkbounds at ./abstractarray.jl:503 [inlined]
 [3] getindex at /home/vchuravy/.julia/packages/StaticArrays/3KEjZ/src/MArray.jl:91 [inlined]
 [4] f(::Int64) at ./REPL[8]:3
 [5] top-level scope at REPL[9]:1
```

We can see that there is a call to `throw_boundserror` that is not inlined and captures the reference to the `MArray`.

Inspecting the `@code_llvm` for this example, we can see the creation of a GCFrame and the heap allocation of the object.

```llvm
;  @ REPL[8]:2 within `f'
define i64 @julia_f_16107(i64) {
top:
  %gcframe = alloca %jl_value_t addrspace(10)*, i32 3
...
  %11 = call noalias nonnull %jl_value_t addrspace(10)* @jl_gc_pool_alloc(i8* %10, i32 1696, i32 32) #1
```

Julia's escape analysis could not remove the allocation, which is  something we heavily rely on for GPU compilation.

If we change the code to include an `@inbounds`
```julia
function f(i)
         A = MArray{Tuple{3}}((0,0,0))
         @inbounds return A[i]
end 
```

The GCFrame and the object allocation is gone.

```llvm
define i64 @julia_f_16145(i64) {
top:
[...]

PR #636, opened 2020-01-17
Add more iterations to default saturation adjustment

charleskawczynski:
#627 runs for 30 minutes with 5 saturation adjustment iterations but breaks for 3. So, let's bump it up.

### DIFF
--- a/src/Common/MoistThermodynamics/states.jl
+++ b/src/Common/MoistThermodynamics/states.jl
@@ -79,7 +79,7 @@ function PhaseEquil(e_int::FT,
                     ρ::FT,
                     q_tot::FT,
                     tol::FT=FT(1e-1),
-                    maxiter::Int=3,
+                    maxiter::Int=5,
                     sat_adjust::F=saturation_adjustment) where {FT<:Real,F}
     return PhaseEquil{FT}(e_int, ρ, q_tot, sat_adjust(e_int, ρ, q_tot, tol, maxiter))
 end


charleskawczynski:
Assuming there's nothing wrong with the inputs to `PhaseEquil` from #627, this indicates that we should somehow increase the [test set of `saturation_adjustment`](https://github.com/climate-machine/CLIMA/blob/master/test/Common/MoistThermodynamics/runtests.jl#L139)

tapios:
I don't see why inaccuracies in saturation adjustment would break DYCOMS (rather than just leading to inaccurate solutions). There probably is a different underlying problem (e.g., large oscillations from numerical instability).

charleskawczynski:
The problem should not be related to the accuracy since we've left `tol` the same. Saturation adjustment will return the same values for all inputs with the same tolerance (the exit condition is only based on `tol`). The issue here is that DYCOMS is yielding inputs to `saturation_adjustment` that require more than 3 iterations in order to converge, and we're not testing these inputs in the test set.

charleskawczynski:
We could remove the [error calls](https://github.com/climate-machine/CLIMA/blob/master/src/Common/MoistThermodynamics/MoistThermodynamics.jl#L691-L693) when `saturation_adjustment` does not converge, and I suspect the code may still run fine, but then we won't know the difference between nearly not converging and drastically not converging.

tapios:
Yes, I'd prefer to remove the error calls and replace them by warnings that write out the input values that let to the lack of convergence. I do suspect this lack of convergence indicate deeper issues (e.g., oscillations generating physically strange or unrealizable inputs).

charleskawczynski:
This would be fine on the CPU, but I believe that trying to print these values on the GPU, I believe, will result in kernel crashes #393. I'd be fine raising the ceiling on `maxiter` to help get #597 through with DYCOMS working. We can always return to the moist thermo test-set to lower `maxiter` back down later.

tapios:
Can you increase `maxiter` within the DYCOMS driver only? Where I'm coming from is this: I'd like saturation adjustment to run with just 1 iteration, as it does in many (if not most) current models. We can sacrifice some accuracy for that. When it needs ~5 iterations, something is wrong, and we should figure out what it is. 

Can you write out the input values within the `if` statement before you throw the error?

charleskawczynski:
>Can you increase maxiter within the DYCOMS driver only?

Yes, we can do this by passing these parameters into the `AtmosModel`.

> Can you write out the input values within the if statement before you throw the error?

Yes we can, but the `print` call itself is what crashes the GPU kernel, so we'll only get useful information for CPU runs. Want me to add in print statements of the inputs before the `error` call?

tapios:
> > Can you write out the input values within the if statement before you throw the error?
> 
> Yes we can, but the `print` call itself is what crashes the GPU kernel, so we'll only get useful information for CPU runs. Want me to add in print statements of the inputs before the `error` call?

Not if it leads to crashes. But I would like to know under which circumstances the algorithm does not converge. If you can reproduce the error quickly, just writing out the relevant values once and sending them to me would be good enough.

charleskawczynski:
> Not if it leads to crashes. But I would like to know under which circumstances the algorithm does not converge. If you can reproduce the error quickly, just writing out the relevant values once and sending them to me would be good enough.

I can start a run of #627 on the CPU and print the input values for which saturation adjustment does not converge. But I'm not sure how long it will take to see failed inputs. I can start it now and let it run over the weekend.

simonbyrne:
Can we use the previous value as the initial guess? Would that help things?

tapios:
> Can we use the previous value as the initial guess? Would that help things?

Yes. We talked about this in the past. I am not sure how easy it is to implement (you'd need to pass previous temperatures from the dycore into `MoistThermodynamics`), but it may help.

charleskawczynski:
It turns out that adding `maxiter=5` for DYCOMS only is currently difficult without hard-coding `maxiter` in multiple places that must be kept in sync.

tapios:
In that case, can you just throw a non-informative warning when it does not converge and otherwise ignore it?

charleskawczynski:
> In that case, can you just throw a non-informative warning when it does not converge and otherwise ignore it?

We can't without breaking the GPU implementation. But I think the point is now mute. It turned out that we were able to allow DYCOMS to use `maxiter=5`, which was applied in #639 and #637. Since the only remaining issue is related to testing moist thermo with inputs from DYCOMS, I've noted this in #263. Given this information, closing this.

PR #637, opened 2020-01-19
Add `maxiter` to `EquilMoist`, reorder (`tol,maxiter`)

charleskawczynski:
- Adds `maxiter` to `EquilMoist`
 - Reorders `maxiter` and `tol` to protect against bug fix caught in #632

This was added so that `EquilMoist(5)` can be used for the DYCOMS driver in #597 without changing the default `maxiter` in saturation adjustment, based on discussions in #636.

The `maxiter` in `EquilMoist` should probably be reverted at some point since there will be two default `maxiter`'s (one in `states.jl`, and another in `moisture.jl`), which is not an ideal design.

### DIFF
--- a/src/Atmos/Model/moisture.jl
+++ b/src/Atmos/Model/moisture.jl
@@ -63,7 +63,8 @@ thermo_state(moist::DryModel, orientation::Orientation, state::Vars, aux::Vars) 
 
 Assumes the moisture components are computed via thermodynamic equilibrium.
 """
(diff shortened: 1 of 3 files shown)


charleskawczynski:
@kpamnany pointed out that there is a call to `PhaseEquil` in diagnostics, which doesn't have access to `AtmosModel` yet, and so doing this is actually not as simple as we originally thought without hard coding constants in multiple places that must be kept in sync.

PR #639, opened 2020-01-19
Update `compute_thermo` in diagnostics to use balance law

charleskawczynski:
This will make sure that diagnostics will use the same `ThermodynamicState` as the solver, determined by its `MoistureModel`.


charleskawczynski:
If this works, we can re-open #637, which will maintain consistent `maxiter`.

PR #642, opened 2020-01-22
Add DYCOMS data for MT tests

charleskawczynski:
- Adds `testdata.jl` to store/pull input data that breaks `saturation_adjustment` (using `DataDeps.jl`)
 - Adds `update_constructor_data.jl` to streamline updating new data that breaks `saturation_adjustment`
 - Updates `runtests.jl` to pull in this data for testing (but the test currently fails)


charleskawczynski:
bors try

PR #648, opened 2020-01-23
Make diffusive! terms linear 

jkozdon:
Proposed scheme to close out #626

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -20,8 +20,11 @@ import CLIMA.DGmethods: BalanceLaw, vars_aux, vars_state, vars_gradient,
                         update_aux!, integrate_aux!, LocalGeometry, lengthscale,
                         resolutionmetric, DGModel, num_integrals,
                         nodal_update_aux!, indefinite_stack_integral!,
-                        reverse_indefinite_stack_integral!
-using ..DGmethods.NumericalFluxes
+                        reverse_indefinite_stack_integral!, num_state
+import ..DGmethods.NumericalFluxes: boundary_state!, Rusanov,
+                                    CentralGradPenalty,
+                                    CentralNumericalFluxDiffusive,
+                                    numerical_boundary_flux_diffusive!
 
 """
     AtmosModel <: BalanceLaw
@@ -169,12 +172,17 @@ end
 
 @inline function flux_diffusive!(atmos::AtmosModel, flux::Grad, state::Vars,
                                  diffusive::Vars, aux::Vars, t::Real)
-
   ν, τ = turbulence_tensors(atmos.turbulence, state, diffusive, aux, t)
-
   D_t = (ν isa Real ? ν : diag(ν)) * inv_Pr_turb
   d_h_tot = -D_t .* diffusive.∇h_tot
+  flux_diffusive!(atmos, flux, state, diffusive, aux, t, ν, τ, d_h_tot, D_t)
+end
 
+
+
+@inline function flux_diffusive!(atmos::AtmosModel, flux::Grad, state::Vars,
+                                 diffusive::Vars, aux::Vars, t::Real, ν, τ,
+                                 d_h_tot, D_t)
   flux.ρu += τ * state.ρ
   flux.ρe += τ * state.ρu
   flux.ρe += d_h_tot * state.ρ
@@ -243,6 +251,44 @@ boundary_state!(::CentralGradPenalty, bl::AtmosModel, _...) = nothing
 
 function init_state!(m::AtmosModel, state::Vars, aux::Vars, coords, t, args...)
   m.init_state(state, aux, coords, t, args...)
+end
+
+
+function numerical_boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive,
+    atmos::AtmosModel, fluxᵀn::Vars{S}, n::SVector,
+    state⁻::Vars{S}, diff⁻::Vars{D}, aux⁻::Vars{A},
+    state⁺::Vars{S}, diff⁺::Vars{D}, aux⁺::Vars{A},
+    bctype, t,
+    state1⁻::Vars{S}, diff1⁻::Vars{D}, aux1⁻::Vars{A}) where {S,D,A}
+
+  FT = eltype(fluxᵀn)
+  nstate = num_state(atmos,FT)
+  Fᵀn = parent(fluxᵀn)
+
+  F⁻ = similar(Fᵀn, Size(3, nstate))
+  fill!(F⁻, -zero(FT))
+  flux_diffusive!(atmos, Grad{S}(F⁻), state⁻, diff⁻, aux⁻, t)
+
+  F⁺ = similar(Fᵀn, Size(3, nstate))
+  fill!(F⁺, -zero(FT))
+  atmos_boundary_flux_diffusive!(nf, atmos.boundarycondition, atmos,
+                                 Grad{S}(F⁺), state⁺, diff⁺, aux⁺, n,
+                                 Grad{S}(F⁻), state⁻, diff⁻, aux⁻,
+                                 bctype, t,
+                                 state1⁻, diff1⁻, aux1⁻)
+
+  Fᵀn .+= (F⁻ + F⁺)' * (n/2)
+end
+
+function atmos_boundary_flux_diffusive!(nf, bc, atmos,
+                                        F⁺, state⁺, diff⁺, aux⁺, n,
+                                        F⁻, state⁻, diff⁻, aux⁻,
+                                        bctype, t,
+                                        state1⁻, diff1⁻, aux1⁻)
+
+  atmos_boundary_state!(nf, bc, atmos, state⁺, diff⁺, aux⁺, n, state⁻, diff⁻, aux⁻,
+                        bctype, t, state1⁻, diff1⁻, aux1⁻)
+  flux_diffusive!(atmos, F⁺, state⁺, diff⁺, aux⁺, t)
 end
 
 end # module
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -24,7 +24,7 @@ import CLIMA.DGmethods: BalanceLaw, vars_aux, vars_state, vars_gradient,
 import ..DGmethods.NumericalFluxes: boundary_state!, Rusanov,
                                     CentralGradPenalty,
                                     CentralNumericalFluxDiffusive,
-                                    numerical_boundary_flux_diffusive!
+                                    boundary_flux_diffusive!
 
 """
     AtmosModel <: BalanceLaw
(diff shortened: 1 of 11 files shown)


jkozdon:
bors try

jkozdon:
@simonbyrne: @mwarusz, @thomasgibson, and I worked on this today. Didn't want to push to your branch so hence the new PR.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -263,4 +253,42 @@ function init_state!(m::AtmosModel, state::Vars, aux::Vars, coords, t, args...)
   m.init_state(state, aux, coords, t, args...)
 end
 
+
+function numerical_boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive,
```
simonbyrne: Maybe we can avoid overloading this by exposing a variant of flux_diffusive! which takes a normal vector

jkozdon:
bors try

src/Atmos/Model/AtmosModel.jl
jkozdon: Just pushed a slightly different interface. Complication arises from the fact we want to specify BCs in two different ways. I imagine all this could change soon anyway after we have a bit more time to sort out BCs.

Mainly want to get this PR through so that other BCs issues and instability issues can be addressed.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -263,4 +253,15 @@ function init_state!(m::AtmosModel, state::Vars, aux::Vars, coords, t, args...)
   m.init_state(state, aux, coords, t, args...)
 end
 
+boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive, atmos::AtmosModel,
+                         F⁺, state⁺, diff⁺, aux⁺, n⁻,
```
simonbyrne: Why `n⁻` instead of just `n`?

src/Atmos/Model/AtmosModel.jl
simonbyrne: Definitely, thanks for wrapping this up.

jkozdon:
bors try

src/Atmos/Model/AtmosModel.jl
jkozdon: It's the normal with respect to the minus side of the face as opposed to the normal with respect to the plus side (which is equal and opposite).

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
@smarras79 How can I test that I didn't introduce an error into dycoms?

fxgiraldo:
@smarras79 I sent you an email about running dycoms.  It is essentially the same question that @jkozdon is asking here. Just letting you know that this is part of the same question.

smarras79:
> @smarras79 I sent you an email about running dycoms. It is essentially the same question that @jkozdon is asking here. Just letting you know that this is part of the same question.

Ho @jkozdon and @fxgiraldo I've just  seen your email but I will send the same reply here in case other need the driver:

A new dycoms driver was built to run with master. It is here: 

https://github.com/climate-machine/CLIMA/blob/master/examples/DGmethods/ex_001_dycoms-imex.jl

Try it and let me know if you have problems. 
To visualize the data, this graph_diagnostics.jl  should replace the one currently in master (I am opening a PR for it now): https://github.com/climate-machine/CLIMA/blob/sm/graph_diagnostics.jl/src/Diagnostics/graph_diagnostic.jl

jkozdon:
How long does the simulation take to run and on what hardware?

smarras79:
> How long does the simulation take to run and on what hardware?

This specific setup completes in approximately 40 minutes. It runs stably with up to Courant 0.95.
1 GPU (I only tested it on a Titan V)

smarras79:
dt is calculated based on a pre-fixed Courant here: 

https://github.com/climate-machine/CLIMA/blob/db5e67225d7e1c5bfb4185fe7f3de38abb752cf1/examples/DGmethods/ex_001_dycoms-imex.jl#L327

increase it as needed.

jkozdon:
Thanks.

fxgiraldo:
thanks @smarras79.  40 minutes to run a simulation length of how long?  Also, is this still the 10x smaller problem than the one suggested in the dycoms 2005 paper?

smarras79:
> thanks @smarras79. 40 minutes to run a simulation length of how long? Also, is this still the 10x smaller problem than the one suggested in the dycoms 2005 paper?

40 minutes for a simulated time of 14400 s (4 hours) at Courant 0.95 for a coarse resolution of dx=dy=40m X dz=20m.

Yes, this set up is for a domain that is 9 times (1/3 smaller in each horizontal direction) smaller than what it should be.

fxgiraldo:
OK, that is good news. If this scaling holds then we should be able to do the big simulation in 40 x 3(x-direction) x 3(y-direction) x 3(time) = 1080 minutes or am I forgetting something?  This gives us a simulation in 18 hours of wallclock time on 1 GPU.

smarras79:
> OK, that is good news. If this scaling holds then we should be able to do the big simulation in 40 x 3(x-direction) x 3(y-direction) x 3(time) = 1080 minutes or am I forgetting something? This gives us a simulation in 18 hours of wallclock time on 1 GPU.

At coarse resolution yes; however, we would ideally like to have a fine resolution (dx=dy=35 X dz=5) to run in less than 24 hours. 
As of now, I could only keep stability for the fine resolution up to Courant = 0.4. 
Do you think that the new B.C. and handling of viscosity will allow to run at a larger Courant value?

jkozdon:
> Do you think that the new B.C. and handling of viscosity will allow to run at a larger Courant value?

We haven't changed the BC treatment yet, but the viscosity was being mishandled so there is a chance it will be more stable now (or it might not be)

smarras79:
> > Do you think that the new B.C. and handling of viscosity will allow to run at a larger Courant value?
> 
> We haven't changed the BC treatment yet, but the viscosity was being mishandled so there is a chance it will be more stable now (or it might not be)

Sounds good. Keep me informed and let me know if I can help.

fxgiraldo:
> > OK, that is good news. If this scaling holds then we should be able to do the big simulation in 40 x 3(x-direction) x 3(y-direction) x 3(time) = 1080 minutes or am I forgetting something? This gives us a simulation in 18 hours of wallclock time on 1 GPU.
> 
> At coarse resolution yes; however, we would ideally like to have a fine resolution (dx=dy=35 X dz=5) to run in less than 24 hours.
> As of now, I could only keep stability for the fine resolution up to Courant = 0.4.
> Do you think that the new B.C. and handling of viscosity will allow to run at a larger Courant value?

I'm trying to get a sense of what the current wallclock time is for the target simulation.   Am I understanding this correctly that the fine problem you want to run is then a factor of 16 bigger than I approximated?  4x in the vertical and say 4x in time-step.

smarras79:
> > > OK, that is good news. If this scaling holds then we should be able to do the big simulation in 40 x 3(x-direction) x 3(y-direction) x 3(time) = 1080 minutes or am I forgetting something? This gives us a simulation in 18 hours of wallclock time on 1 GPU.
> > 
> > 
> > At coarse resolution yes; however, we would ideally like to have a fine resolution (dx=dy=35 X dz=5) to run in less than 24 hours.
> > As of now, I could only keep stability for the fine resolution up to Courant = 0.4.
> > Do you think that the new B.C. and handling of viscosity will allow to run at a larger Courant value?
> 
> I'm trying to get a sense of what the current wallclock time is for the target simulation. Am I understanding this correctly that the fine problem you want to run is then a factor of 16 bigger than I approximated? 4x in the vertical and say 4x in time-step.

That's correct.

jkozdon:
@fxgiraldo I am running this right now on our TitanV. It's been running 45 minutes, looks like it will take 2.3 hours on our box. @smarras79 Are you sure that the version of the driver in master really runs in 40 minutes on a TitanV for you? Anyway you can check this?

smarras79:
> @fxgiraldo I am running this right now on our TitanV. It's been running 45 minutes, looks like it will take 2.3 hours on our box. @smarras79 Are you sure that the version of the driver in master really runs in 40 minutes on a TitanV for you? Anyway you can check this?

Yep, 40 to 45 min on my titan V. Same driver. What Courant number did you set it up for?

src/Atmos/Model/AtmosModel.jl
simonbyrne: I though it was the minus face of the side wrt to the normal?

jkozdon:
> Yep, 40 to 45 min on my titan V. Same driver. What Courant number did you set it up for?

I am running the [ex_001_dycoms-imex.jl] currently on  master

[ex_001_dycoms-imex.jl]: https://github.com/climate-machine/CLIMA/blob/db5e67225d7e1c5bfb4185fe7f3de38abb752cf1/examples/DGmethods/ex_001_dycoms-imex.jl

jkozdon:
The current driver on master just crashed on me after time `5.6047793127537825e03`

@smarras79 Can you check out a fresh version of master and confirm that you can run this exact file in 40 minutes without it crashing?

smarras79:
> The current driver on master just crashed on me after time `5.6047793127537825e03`
> 
> @smarras79 Can you check out a fresh version of master and confirm that you can run this exact file in 40 minutes without it crashing?

I confirm that this crashed at 5,000 second for me as well on `master`. I pointed this out to @charleskawczynski and @kpamnany who are looking into the differences with their branch that was the working precursor to the current master.

simonbyrne:
Let’s not keep this held up on DYCOMS, that will require some manual testing

src/DGmethods/NumericalFluxes.jl
```diff
@@ -238,8 +236,33 @@ function numerical_flux_diffusive!(::CentralNumericalFluxDiffusive,
...
+  nstate = num_state(bl,FT)
+  Fᵀn = parent(fluxᵀn)
+
+  F⁻ = similar(Fᵀn, Size(3, nstate))
+  fill!(F⁻, -zero(FT))
+  flux_diffusive!(bl, Grad{S}(F⁻), state⁻, diff⁻, aux⁻, t)
+
+  F⁺ = similar(Fᵀn, Size(3, nstate))
+  fill!(F⁺, -zero(FT))
+  boundary_flux_diffusive!(nf, bl,
```
simonbyrne: So if we want to prescribe a specific diffusive boundary flux, is this function we should extend?

simonbyrne: cc: @fxgiraldo

src/Atmos/Model/AtmosModel.jl
jkozdon: Both interpretations work. I tend to think of a minus and plus normal, though others only consider that they have a normal

src/DGmethods/NumericalFluxes.jl
jkozdon: That's my thought right now. Something like what I did in the [dycoms bc]

[dycoms bc]:  https://github.com/climate-machine/CLIMA/blob/26ce8f722f07cfaf8f97a4f4624ccf288a0bc0ca/src/Atmos/Model/boundaryconditions.jl#L126-L241

simonbyrne: This seems to conflict with the Numerics design docs:
Screen Shot 2020-01-24 at 10 34 49 AM

jkozdon: I was just trying to preserve close to what we were currently doing, not to bring things in to conformity with Bassi and Rebay (which is our ultimate goal). So you are correct that we should directly overload the function.

simonbyrne: That's fine, I'm trying to figure out how things should be done.

jkozdon: Do you want to BCs fixed before this is merged?

jkozdon: I think if we merge this, then #630, I have a pretty good idea what we need to do for the BCs.

jkozdon: We could also just merge the BCs update into #630 if you wanted (since that's the main motivation for #630)

simonbyrne: Lets get this in.

simonbyrne: Whatever you think is best: would be great to get them both wrapped up today.

PR #651, opened 2020-01-24
Termination conditionGMRES matches IterativeSolvers.jl

rohanmclure:
Change `residual_norm < norm(b)` to `residual_norm < norm(inital_residual) * rel_tolerance`.
Adding larger tests soon.

### DIFF
--- a/src/LinearSolvers/GeneralizedMinimalResidualSolver.jl
+++ b/src/LinearSolvers/GeneralizedMinimalResidualSolver.jl
@@ -142,6 +142,7 @@ function LS.doiteration!(linearoperator!, Q, Qrhs,
 
   # if not converged restart
   converged || LS.initialize!(linearoperator!, Q, Qrhs, solver, args...)
+  # @show residual_norm, j, threshold 
 
   (converged, j, residual_norm)
 end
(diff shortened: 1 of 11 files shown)


simonbyrne:
To provide some context: we noticed that GMRES didn't appear to be converging in a reasonable number of iterations. After some investigation, @rohanmclure was able to narrow it down to the tolerance used in the termination criteria. Unfortunately, this isn't really discussed much in the literature: this criteria matches that of IterativeSolvers.jl and PETSc.

src/LinearSolvers/GeneralizedMinimalResidualSolver.jl
```diff
@@ -67,11 +67,12 @@ function LS.initialize!(linearoperator!, Q, Qrhs,
...
     krylov_basis[1] .+= Qrhs
 
-    threshold = solver.tolerance[1] * norm(Qrhs, weighted)
+    threshold = solver.tolerance[1] * norm(krylov_basis[1], weighted)
     residual_norm = norm(krylov_basis[1], weighted)
 
     converged = false
-    if residual_norm < threshold
+    # FIXME: Should only be true for threshold zero
+    if threshold <=10eps(eltype(Q))
```
simonbyrne: lets allow the user to specify an `atol`/`rtol`.

src/LinearSolvers/GeneralizedMinimalResidualSolver.jl
```diff
@@ -67,11 +67,12 @@ function LS.initialize!(linearoperator!, Q, Qrhs,
     krylov_basis[1] .*= -1
     krylov_basis[1] .+= Qrhs
```
simonbyrne: can fuse these into 
```julia
    krylov_basis[1] .= Qrhs .- krylov_basis[1]
```

PR #657, opened 2020-01-24
Fix boundary conditions to account for #617

blallen:
When committing #617, I removed the viscosity and diffusivity from the diffusive state for the ocean model without updating the boundary conditions to accommodate that. This PR is to properly update them.


src/Ocean/Model/HydrostaticBoussinesqModel.jl
```diff
@@ -89,8 +89,8 @@ function vars_aux(m::HBModel, T)
     θʳ::T           # SST given    # TODO: Should be 2D
     f::T            # coriolis
     τ::T            # wind stress  # TODO: Should be 2D
-    # κ::SMatrix{3, 3, T, 9} # diffusivity tensor (for convective adjustment)
-    κᶻ::T
+    ν::SVector{3, T}
+    κ::SVector{3, T}
```
blallen: @charleskawczynski is it possible to update the `Vars` macro so that we can write out an`SMatrix` to the vtk files? 

Ideally, these would be 3x3 matrices so that I don't have to keep calling `Diagonal` whenever I want to use them, but they aren't because then I can't output any of the auxiliary variables to the vtk files.

blallen:
> Not sure what you want me to review.

Mostly to check that I implemented the BC `κ∇θ = f(x)` properly since we are now enforcing continuity on just `∇θ` instead of `κ∇θ`.

blallen:
bors try

src/Ocean/Model/HydrostaticBoussinesqModel.jl
charleskawczynski: Sorry for the late response. @simonbyrne was actually the master-mind behind the `Vars` type and VariableTemplates, so he's best positioned to modify this. That said, I'm pretty sure that `Vars` already uses `SMatrix`. I think for vtk export, we just need to modify [flattenednames](https://github.com/climate-machine/CLIMA/blob/master/src/Utilities/VariableTemplates/var_names.jl#L6-L9) but I'm not sure how the names/formatting plays here.

blallen:
@jkozdon @simonbyrne can one of you check why the pseudo1D advection GPU test is failing? I thought it was a temporary error, but it happened on two separate attempts. I can run the test on my laptop and it succeeds, so I'm not sure why it isn't working on the cluster.

jkozdon:
I believe this is due to the upgrade from Julia 1.2 to 1.3 in the slurm CI. See https://github.com/climate-machine/CLIMA/pull/660#issuecomment-580985631

This afternoon I can put together a quick PR on top of master that fixes this.

This reveals that our CI has a hidden dependency on https://github.com/climate-machine/slurmci not sure if we want to make this more explicit some how (@simonbyrne, @kpamnany)

simonbyrne:
I think we should consider moving the slurm scripts back into this repo, as it will make them easier to change.

cc @kpamnany

kpamnany:
My thinking was/is that SlurmCI is very specific to the Central cluster and unnecessary for anyone who wants to clone and run CLIMA, or even to contribute. Also, there's a good bit of work still to do in building out the performance analysis and it'll be easier to do there, without needing to open PRs to the main repo.

IMO, we should be moving towards splitting the main repo into more parts rather than the opposite!

jkozdon:
> This afternoon I can put together a quick PR on top of master that fixes this.

see #674

jkozdon:
Moving Slurm CI discussion to issue #675

blallen:
should we wait for this to be merged before we add the commit to other outstanding PRs?

PR #659, opened 2020-01-25
Allow turbulence models with NoOrientation

mwarusz:
This would help with diagnosing issues at the intersection of numerics and turbulence models by
allowing creation of simpler test cases, e.g. isotropic decaying turbulence.


mwarusz:
bors try

mwarusz:
bors try

Issue #652, opened 2020-01-24
Better CI with integration tests


jkozdon:
Right now we just run a few time steps of the atmos model for integration tests (dycoms, RTB, etc.), but it is a little concerning that the CI didn't break with the changes from #648

jkozdon:
Currently the [Dycoms] CI passes, but that just means that the code [didn't crash](https://github.com/climate-machine/CLIMA/blob/461657afa53ad608015a27f7c91dd8c15e605810/experiments/LES/dycoms.jl#L154-L185) since there is no `@test` in the test file.

_Originally posted by @jkozdon in https://github.com/climate-machine/CLIMA/pull/660#issuecomment-581980491_

Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.

_Originally posted by @kpamnany in https://github.com/climate-machine/CLIMA/pull/660#issuecomment-581981548_

> Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.

We need something to compare against when we perform the time-integrator study to make sure we know the stability condition parameters for the inviscid and viscous processes. Having a working DYCOMS up to the point that the stability measures (e.g., horizontal and vertical Courant numbers and Peclet numbers) equilibrate will be helpful.

_Originally posted by @fxgiraldo in https://github.com/climate-machine/CLIMA/pull/660#issuecomment-581996076_

> > Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.
> 
> We need something to compare against when we perform the time-integrator study to make sure we know the stability condition parameters for the inviscid and viscous processes. Having a working DYCOMS up to the point that the stability measures (e.g., horizontal and vertical Courant numbers and Peclet numbers) equilibrate will be helpful.

Since we now have a way to I/O restart files, one thing that could be good to test only 100s of DYCOMS is to start the test from a restart file in the middle of the simulation so that all the initial transient is gone. Run it for 100s, and look at the norms difference for that slab of the run.
@akshaysridhar what do you think of this option?

_Originally posted by @smarras79 in https://github.com/climate-machine/CLIMA/pull/660#issuecomment-582001464_

blallen:
Can someone link me to the way to restart files?

This will help us implement tests for the ocean model beyond checking that we preserve incompressibility.

kpamnany:
[...]

PR #660, opened 2020-01-25
Update boundary conditions to Bassi and Rebay (1997)

jkozdon:
- [x] Advection-Diffusion model
- [x] Heat eqn
- [x] Atmos model
- [x] Ocean model
- [x] poisson test

### DIFF
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -19,9 +19,10 @@ end
 #TODO: figure out a better interface for this.
 # at the moment we can just pass a function, but we should do something better
 # need to figure out how subcomponents will interact.
-function atmos_boundary_state!(::Rusanov, f::Function, m::AtmosModel,
-                               stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, _...)
+function atmos_boundary_state!(::Union{Rusanov, CentralNumericalFluxGradient},
+                               f::Function, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, _...)
   f(stateP, auxP, nM, stateM, auxM, bctype, t)
 end
 
@@ -33,9 +34,10 @@ function atmos_boundary_state!(::CentralNumericalFluxDiffusive, f::Function,
 end
 
 # lookup boundary condition by face
-function atmos_boundary_state!(nf::Rusanov, bctup::Tuple, m::AtmosModel,
-                               stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, _...)
+function atmos_boundary_state!(nf::Union{Rusanov, CentralNumericalFluxGradient},
+                               bctup::Tuple, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, _...)
   atmos_boundary_state!(nf, bctup[bctype], m, stateP, auxP, nM, stateM, auxM,
                         bctype, t)
 end
@@ -74,9 +76,10 @@ Set the momentum at the boundary to be zero.
 struct NoFluxBC <: BoundaryCondition
 end
 
-function atmos_boundary_state!(::Rusanov, bc::NoFluxBC, m::AtmosModel,
-                               stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, _...)
+function atmos_boundary_state!(::Union{Rusanov, CentralNumericalFluxGradient},
+                               bc::NoFluxBC, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, _...)
   FT = eltype(stateM)
   stateP.ρ = stateM.ρ
   stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
@@ -104,9 +107,10 @@ mainly useful for cases where the problem has an explicit solution.
 """
 struct InitStateBC <: BoundaryCondition
 end
-function atmos_boundary_state!(::Rusanov, bc::InitStateBC, m::AtmosModel,
-                               stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, _...)
+function atmos_boundary_state!(::Union{Rusanov, CentralNumericalFluxGradient},
+                               bc::InitStateBC, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, _...)
   init_state!(m, stateP, auxP, auxP.coord, t)
 end
 function atmos_boundary_state!(::CentralNumericalFluxDiffusive, bc::InitStateBC,
@@ -126,9 +130,10 @@ struct DYCOMS_BC{FT} <: BoundaryCondition
   LHF::FT
   SHF::FT
 end
-function atmos_boundary_state!(::Rusanov, bc::DYCOMS_BC, m::AtmosModel,
-                               stateP::Vars, auxP::Vars, nM, stateM::Vars,
-                               auxM::Vars, bctype, t, state1::Vars, aux1::Vars)
+function atmos_boundary_state!(::Union{Rusanov, CentralNumericalFluxGradient},
+                               bc::DYCOMS_BC, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, state1::Vars, aux1::Vars)
   # stateM is the 𝐘⁻ state while stateP is the 𝐘⁺ state at an interface.
   # at the boundaries the ⁻, minus side states are the interior values
   # state1 is 𝐘 at the first interior nodes relative to the bottom wall
@@ -257,7 +262,8 @@ struct RayleighBenardBC{FT} <: BoundaryCondition
   T_top::FT
 end
 # Rayleigh-Benard problem with two fixed walls (prescribed temperatures)
-function atmos_boundary_state!(::Rusanov, bc::RayleighBenardBC, m::AtmosModel,
+function atmos_boundary_state!(::Union{Rusanov, CentralNumericalFluxGradient},
+                               bc::RayleighBenardBC, m::AtmosModel,
                                stateP::Vars, auxP::Vars, nM, stateM::Vars,
                                auxM::Vars, bctype, t,_...)
   # Dry Rayleigh Benard Convection
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -1,7 +1,7 @@
 using CLIMA.PlanetParameters
 export PeriodicBC, NoFluxBC, InitStateBC, DYCOMS_BC, RayleighBenardBC
 
-function atmos_boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive,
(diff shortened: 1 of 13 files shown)


jkozdon:
Had to create new PR, since PR #656 got closed automatically when branch jek/flux was merged (had the PR set to that branch to get a better diff). Could not reopen #656, hence this new PR.

jkozdon:
bors try

simonbyrne:
I mentioned to @fxgiraldo on the call this morning, our biggest challenge is that we mostly want to apply boundary conditions (particularly diffusive ones) directly by specifying the boundary flux.

[Something like this example](https://mathematica.stackexchange.com/a/127425/37818) might be a good test case?

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try

simonbyrne:
This is looking really good, thanks!

src/DGmethods/NumericalFluxes.jl
```diff
@@ -252,19 +238,33 @@ function numerical_boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive,
...
-                           Grad{S}(F⁻), state⁻, diff⁻, aux⁻,
+                           Grad{S}(F),
+                           state⁺, diff⁺, aux⁺,
+                           n⁻,
+                           state⁻, diff⁻, aux⁻,
                            bctype, t,
                            state1⁻, diff1⁻, aux1⁻)
 
-  Fᵀn .+= (F⁻ + F⁺)' * (n⁻/2)
+  Fᵀn .+= F' * n⁻
```
simonbyrne: Can we update the design docs to reflect this change?

simonbyrne: cc @fxgiraldo

src/DGmethods/NumericalFluxes.jl
```diff
@@ -252,19 +238,33 @@ function numerical_boundary_flux_diffusive!(nf::CentralNumericalFluxDiffusive,
...
 
-  F⁻ = similar(Fᵀn, Size(3, nstate))
-  fill!(F⁻, -zero(FT))
-  flux_diffusive!(bl, Grad{S}(F⁻), state⁻, diff⁻, aux⁻, t)
-
-  F⁺ = similar(Fᵀn, Size(3, nstate))
-  fill!(F⁺, -zero(FT))
+  F = similar(Fᵀn, Size(3, nstate))
+  fill!(F, -zero(FT))
   boundary_flux_diffusive!(nf, bl,
```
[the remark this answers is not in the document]
jkozdon: Wouldn't this just be overloading `numerical_boundary_flux_diffusive!` completely?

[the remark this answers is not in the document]
jkozdon: i.e., this is what `numerical_boundary_flux_diffusive!` sets.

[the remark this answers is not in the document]
simonbyrne: It might be worth inserting an extra function for dispatch purposes.

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
@blallen Can you grab the branch for this PR and make sure that all your ocean simulations look OK? I had to update the ocean model with the new boundary condition treatment.

simonbyrne:
@jkozdon what's the status of this?

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -71,12 +76,17 @@ Set the momentum at the boundary to be zero.
...
-                               auxM::Vars, bctype, t, _...)
+function atmos_boundary_state!(nf::Union{Rusanov, CentralNumericalFluxGradient},
+                               bc::NoFluxBC, m::AtmosModel, stateP::Vars,
+                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
+                               t, _...)
   FT = eltype(stateM)
   stateP.ρ = stateM.ρ
-  stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
+  if typeof(nf) == Rusanov
+    stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
```
simonbyrne: why is this necessary?

jkozdon: This is because we run this non-diffussive flux back through the numerical flux function, wheras for the Gradient terms it is used directly in the flux function.

@fxgiraldo can you confirm this is right?

jkozdon:
> @jkozdon what's the status of this?

We have a weird Float32 issue difference between the slurm CI and our local system that I am trying to sort out. I have an idea, but haven't been able to test it. (We can comment out these weird results from the tests and open an issue if that's better)

Also would like @blallen (or someone else) to confirm the changes I made to the ocean morel are correct.

Otherwise this should be good to go.

src/Atmos/Model/boundaryconditions.jl
simonbyrne: Do we need this special handling for any other variables, or just momentum?

jkozdon: Just momentum for this boundary condition type

Can you confirm @fxgiraldo

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
Now `Float32` is working for all tests. From looking at the [latest] slurm results and [previous] slurm results, suggests the the issue was Julia version 1.2 versus 1.3; Slurm CI upgraded to with climate-machine/slurmci#11

[latest]: https://gist.github.com/climabot/351b30191d17dc9dfe20d649caac0049
[previous]: https://gist.github.com/climabot/a9718f0db9d2d5475519b6d4631fd2f1
[PR]: https://github.com/climate-machine/slurmci/pull/11

jkozdon:
bors try

jkozdon:
Would be nice is someone (@kpamnany, @simonbyrne, @charleskawczynski, @smarras79?) could confirm that this PR does not break Dycoms; based on #664 it seems to be working now(?).

cc @blallen

kpamnany:
DYCOMS appears to be running fine for this PR (from the SlurmCI try above: https://gist.github.com/climabot/d48549f5727f9f35ceb433ac98e981d8#file-out_8654789).

jkozdon:
bors try

jkozdon:
This is ready to go from my perspective.

fxgiraldo:
> DYCOMS appears to be running fine for this PR (from the SlurmCI try above: https://gist.github.com/climabot/d48549f5727f9f35ceb433ac98e981d8#file-out_8654789).

Has someone confirmed that DYCOMS is running fine?

jkozdon:
> > DYCOMS appears to be running fine for this PR (from the SlurmCI try above: https://gist.github.com/climabot/d48549f5727f9f35ceb433ac98e981d8#file-out_8654789).
> 
> Has someone confirmed that DYCOMS is running fine?

Currently the CI passed, but that just means that the code [didn't crash](https://github.com/climate-machine/CLIMA/blob/461657afa53ad608015a27f7c91dd8c15e605810/experiments/LES/dycoms.jl#L154-L185) since there is no `@test` in the test file.

kpamnany:
Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.

fxgiraldo:
> Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.

We need something to compare against when we perform the time-integrator study to make sure we know the stability condition parameters for the inviscid and viscous processes. Having a working DYCOMS up to the point that the stability measures (e.g., horizontal and vertical Courant numbers and Peclet numbers) equilibrate will be helpful.

smarras79:
> > Right. I'm hoping @akshaysridhar or @smarras79 can suggest a useful `@test` for the current 100sec simulation at some point.
> 
> We need something to compare against when we perform the time-integrator study to make sure we know the stability condition parameters for the inviscid and viscous processes. Having a working DYCOMS up to the point that the stability measures (e.g., horizontal and vertical Courant numbers and Peclet numbers) equilibrate will be helpful.

Since we now have a way to I/O restart files, one thing that could be good to test only 100s of DYCOMS is to start the test from a restart file in the middle of the simulation so that all the initial transient is gone. Run it for 100s, and look at the norms difference for that slab of the run.
@akshaysridhar what do you think of this option?

jkozdon:
This is likely a discussion to move to Issue #652

blallen:
From a preliminary look, the ocean boundary conditions look good to me. I need to take a deeper look later today, but you can go ahead and merge this and I'll just make a small fix PR if things aren't right.

jkozdon:
Thanks @blallen.

@simonbyrne should this be merged then?

simonbyrne:
While I don't have any problem with the changes, I still don't think this addresses the core problem in that it is still somewhat unclear how they are supposed to be used, e.g. https://github.com/climate-machine/CLIMA/pull/660/files#r373245511.

That said, given the complexity of this PR, I don't want to hold it up, so if you feel that those points can be addressed in the design docs, i'm happy to merge it.

blallen:
piggybacking off @simonbyrne :

my interpretation of this change is that we've switched from specific numerical fluxes (Rusanov, CentralNonDiff, CentralDiff, NumericalGradPenalty) to a generic NumericalFlux (either Diff or NonDiff) where the user defines both the numerical flux at the element interfaces and along the domain boundary (through two separate functions, but one calls the other). At the same time, additional methods were added so that this is backward compatible and you can still define boundary conditions in terms of ghost points instead of the expected flux.

is this generally correct?

jkozdon:
The main goal was the bring the BC treatment in line with a [published method](https://doi.org/10.1006/jcph.1996.5572). The main changes are:

- for the gradient flux on boundaries, instead of calling the `numerical_flux_gradient!` with the minus side state and plus side (ghost) state, we just call `gradvariables!` and `diffusive!` on the plus side (ghost state); see [here](https://github.com/climate-machine/CLIMA/blob/d808788447a2354d1e5927b986d212780f078278/src/DGmethods/NumericalFluxes.jl#L49-L63)
- for the diffusive flux on the boundary we no longer call the `numerical_flux_diffussive!` on the minus and plus side states, but instead just call `flux_diffusive!` on the plus side (ghost) state; see [here](https://github.com/climate-machine/CLIMA/blob/d808788447a2354d1e5927b986d212780f078278/src/DGmethods/NumericalFluxes.jl#L241-L279)
- for the diffusive flux on the boundary it should be easier to specify the flux directly as opposed to having to define a plus side (ghost) state; see [here](https://github.com/climate-machine/CLIMA/blob/d808788447a2354d1e5927b986d212780f078278/src/DGmethods/NumericalFluxes.jl#L230-L263)

The treatment of the non-diffusive flux should be largely unchanged (i.e., we still define a plus side state and call the numerical flux).

If folks do not think this is useful, I'm fine abandoning the PR until we understand better what is needed.

blallen:
Okay, that makes sense to me. For second order/parabolic terms, we only use the plus side when constructing the numerical flux on the boundary, while for first order/hyperbolic terms, we use both the plus and minus sides. 

I'll take a look through the paper to review everything and then see how the ocean boundary conditions fit into it. But it should be pretty straight forward to adapt to.

simonbyrne:
Thanks @jkozdon for the summary, that does make it much clearer, and I do think this is a much improved interface for working with diffusive fluxes, which is certainly a current pain point.

My remaining concern is the handling of non-diffusive and gradient fluxes: e.g. it is still not clear to me how these conditions are derived, their relation to the Bassi & Rebay paper (e.g. the paper seems to one should use the same plus side for both?), or how we should prescribe alternatives (which I believe we will need for the LES case). But this sounds like it requires more input from @fxgiraldo, so I'm happy to merge this for now and try to address these other points later.

Issue #655, opened 2020-01-24
CLIMA.init breaks Jupyter notebook


simonbyrne:
Calling `CLIMA.init()` in a Jupyter notebook causes a kernel error. This seems to be due to changes in #597.

kpamnany:
The kernel error arose from `ArgParse.jl` exiting if arguments aren't recognized and the fact that Jupyter also uses a command line argument. Until `ArgParse.jl` supports partial parsing, I've fixed this by ignoring all command line arguments if there are any unrecognized ones iff `isinteractive()`.

kpamnany:
I've tested this locally in a Jupyter notebook, but can you verify also that `kp/fix-init` does fix this?

simonbyrne:
Yep, that works!

PR #661, opened 2020-01-27
Clean up CLIMA initialization

kpamnany:
Removes the need to call `CLIMA.array_type()` at global scope which was previously necessary due to the dynamic definition of `_init_array()` for `CuArrays` causing a world age problem.

Once this is merged, `CLIMA.init()` should be the first function called. This also enables a `--disable-gpu` command line argument besides the `CLIMA_GPU=false` environment variable as well as a `--mpi-knows-cuda` command line argument that will be used to control https://github.com/climate-machine/CLIMA/pull/439.


src/Driver/Driver.jl
```diff
@@ -18,25 +18,39 @@ using ..MoistThermodynamics
 using ..MPIStateArrays
 using ..TicToc
 
-const cuarray_pkgid = Base.PkgId(Base.UUID("3a865a2d-5b23-5a0f-bc46-62713ec82fae"), "CuArrays")
+mutable struct CLIMA_Settings
```
simonbyrne: Can use `Base.@kwdef` to provide a keyword constructor and default values.

kpamnany:
Will fix #655 before `r+`ing.

PR #662, opened 2020-01-28
Columnwise GMRES Sovler

rohanmclure:
GMRES Solver permitting a parameter with configurable number of element columns, assuming a StackedBox grid topology.

No preconditioning provided, so far, but updates all iterative solver constructors to permit configuring both absolute and relative tolerances.


src/LinearSolvers/ColumnwiseGMRESSolver.jl
```diff
@@ -0,0 +1,174 @@
...
+    H  = ntuple(x -> (@MArray zeros(MP1, M)), N)
+    g0 = ntuple(x -> (@MArray zeros(MP1)), N)
+    stop_iter = @MArray fill(-1, N)
+
+    new{M, N, M + 1, M * (M + 1), eltype(Q), Int64, AT}(
+        krylov_basis, H, g0, stop_iter, (rtol, atol))
+  end
+end
+
+@inline function StackGMRES(Q::AT; rtol=√eps(eltype(AT)), atol=eps(eltype(AT)),
```
rohanmclure: Need a way for sizes of solver local arrays to be known at compile time.

src/Arrays/MPIStateArrays.jl
```diff
@@ -57,6 +57,21 @@ struct MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
...
+                                                    nabrtorank, nabrtovmaprecv, nabrtovmapsend,
+                                                    device_send_buffer, device_recv_buffer,
+                                                    weights, commtag
+                                                    ) where {FT, DATN, DAI1, DAV, DAT2}
+    new{FT, DATN, DAI1, DAV, DAT2}(mpicomm, data, realdata, realelems, ghostelems,
+                                   vmaprecv, vmapsend, sendreq, recvreq, host_send_buffer,
+                                   host_recv_buffer, nabrtorank, nabrtovmaprecv, nabrtovmapsend,
+                                   device_send_buffer, device_recv_buffer, weights, commtag)
+  end
+
```
simonbyrne: remove this if not necessary

src/LinearSolvers/ColumnwiseGMRESSolver.jl
```diff
@@ -0,0 +1,177 @@
...
+    StackGMRES(Q::AT; M, rtol, atol, nhorzelem) where AT
+
+This solver performs the matrix-free generalised minimal residual method for
+nodal columns residing in the same stack of elements. The stack-global
+convergence criterion requires convergence in all nodal columns.
+"""
+mutable struct StackGMRES{M, N, MP1, T, I, AT} <: AbstractColumnGMRESSolver{M}
+  krylov_basis::NTuple{MP1, AT}
+  "Hessenberg matrix"
+  H::Vector{Matrix{T}}
```
simonbyrne: Try a `Vector{S}` where `S<:SMatrix`, and same below, then in the main loop you can do `MMatrix(solver.H[1])` to get an `MMatrix` that you can mutate.

blallen:
I ported this over to the branch bla/ocean_gmres and while it runs, I think it is still too slow to be useful for the ocean model without a preconditioner. I'm currently just running a test where the SingleColumnLU solver could be used instead and it is taking a very very long time to do the individual steps. 

I'll keep playing around with it tho. Maybe I can find something that works. It might be worth chatting later this week.

bischtob:

PR #663, opened 2020-01-29
Driver-config: patch argument in diagnostics callback

akshaysridhar:
Quick fix for diagnostics callback argument 
@kpamnany 
@simonbyrne


simonbyrne:
This is fixed in #661, so shouldn't be necessary.

PR #666, opened 2020-01-31
Fix keyword arguments in SolverTypes

dburov190:
Otherwise it's impossible to specify time-stepper in IMEX without setting `linear_model` and `linear_solver` first.


dburov190:
Thank you, both!

PR #667, opened 2020-01-31
Preset courant functions for atmos Model

slifer50:
The PR defines a module for some preset courant functions that make use of #618. These currently enable non-diffusive and diffusive courant number calculations in the vertical or horizontal directions.


src/Courant/Courant_presets.jl
```diff
@@ -0,0 +1,54 @@
...
+  k̂ = aux.orientation.∇Φ / norm(aux.orientation.∇Φ)
+  if direction isa VerticalDirection
+    u = dot(state.ρu/state.ρ, k̂) * k̂
+          return Δt * (norm(u) + soundspeed(m.moisture, m.orientation, state,aux)) / Δx
+  elseif direction isa HorizontalDirection
+    u = cross(k̂, cross(state.ρu/state.ρ,k̂))      
+    return Δt * (norm(u) + soundspeed(m.moisture, m.orientation, state,
+                                            aux)) / Δx
+  end
+end
```
smarras79: This is very clean and looks good. 

Can we have a third function just to calculate the advective (i.e. without soundspeed) courant?

src/Courant/Courant_presets.jl
```diff
@@ -0,0 +1,54 @@
...
+
+function Diffusive_CFL(m::AtmosModel, state::Vars, aux::Vars,
+                               diffusive::Vars, Δx, Δt, direction=VerticalDirection())
+  ν, τ, sdiff = turbulence_tensors(m.turbulence, state, diffusive, aux, 0)
```
smarras79: Can't we pass the value of the eddy viscosity which is already computed elsewhere instead of re-computing it?

src/Courant/Courant_presets.jl
slifer50: Yes I'll add that in shortly

src/Courant/Courant_presets.jl
slifer50: I'm looking into seeing if I can do this with update-diffusive.

jkozdon: Note that there is no guarantee that the current `diffusive` and `aux` arrays is consistent with the `state` outside of the dg model [1], so for safety reason you may want to recompute these updates before calculating the CFL.

> Can't we pass the value of the eddy viscosity which is already computed elsewhere instead of re-computing it?

It is never stored anymore, so you must recompute it [2].

[1]: https://github.com/climate-machine/CLIMA/blob/7977f64a67cb745f3f23a6c90c0c4ad12328c915/src/DGmethods/DGmodel.jl#L20-L175
[2]: https://github.com/climate-machine/CLIMA/blob/c117b4c7eb8b33b3133fcd5b4886269d1356bc08/src/Atmos/Model/AtmosModel.jl#L175-L182

src/Atmos/Model/turbulence.jl
```diff
@@ -104,11 +104,14 @@ end
 
 function turbulence_tensors(m::ConstantViscosityWithDivergence,
     state::Vars, diffusive::Vars, aux::Vars, t::Real)
-
+  FT = eltype(state)
+  k̂ = aux.orientation.∇Φ / norm(aux.orientation.∇Φ)
```
charleskawczynski: Should this be
```
k̂ = vertical_unit_vector(orientation::Orientation, aux::Vars)
```
from `orientation.jl`? In which case, you'll need to pass in the `AtmosModel`.

If so, please repeat this elsewhere, where needed.

src/Courant/Courant_presets.jl
```diff
@@ -0,0 +1,54 @@
...
+using ..Mesh.Grids
+using ..Mesh.Grids: VerticalDirection, HorizontalDirection, EveryDirection
+using Logging
+using Printf
+using LinearAlgebra
+using CLIMA.Atmos
+
+
+
+function Advective_CFL(m::AtmosModel, state::Vars, aux::Vars,
```
jkozdon: Note that both of these definitions of Advection CLF and Diffusive CFL are atmos specific functions. Might be better to have them live in the atmos model and create similar functions for the ocean?

Style-wise `Advective_CFL` and `Diffusive_CFL` should be lower case since they are not types.

slifer50: @jkozdon I see what you mean, I'll make the changes as requested

slifer50:
> @slifer50 Thanks for taking this on.
> 
> There seems to be a bunch of changes unrelated to the PR. probably good to clean that up before merging, and might be good to extract your functionality and add it to the exist #618 PR.

@jkozdon Could you point out the unrelated changes you mentioned? some small changes were made in Atmos Model and turbulence to be able to split the viscosity into horizontal and vertical components as well as test that it was being done correctly.

src/Atmos/Model/turbulence.jl
slifer50: Does this work as intended with Spherical orientation?

experiments/LES/dry_rayleigh_benard.jl
```diff
@@ -0,0 +1,133 @@
+using Distributions
```
jkozdon: Are these changes needed?

experiments/LES/dycoms.jl
```diff
@@ -63,16 +63,16 @@ function init_dycoms!(state, aux, (x,y,z), t)
...
+    #r1 = FT(rand(Uniform(-0.002, 0.002)))
+    #r2 = FT(rand(Uniform(-0.00001, 0.00001)))
+    #r3 = FT(rand(Uniform(-0.001, 0.001)))
+    #r4 = FT(rand(Uniform(-0.001, 0.001)))
+    #if z <= 400.0
+     #   θ_liq += r1 * θ_liq
+      #  q_tot += r2 * q_tot
+       # u     += r3 * u
+       # v     += r4 * v
+    #end
```
jkozdon: Are these changes needed?

experiments/LES/dycoms.jl
```diff
@@ -140,7 +140,7 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
                                      solver_type=CLIMA.ExplicitSolverType(LSRK144NiegemannDiehlBusch),
                                      ref_state=ref_state,
                                      C_smag=C_smag,
-                                     moisture=EquilMoist(5),
+                                     moisture=EquilMoist(),
```
jkozdon: Are these changes needed?

src/Atmos/Model/turbulence.jl
jkozdon: Can we move this whole computation to the CFL computation? The reason I suggest this is that we changing the interface and computation for just the CFL (which isn't called at every time step but only occasionally).

src/Atmos/Model/turbulence.jl
```diff
@@ -89,7 +89,7 @@ $(DocStringExtensions.FIELDS)
 """
 struct ConstantViscosityWithDivergence{FT} <: TurbulenceClosure
   "Dynamic Viscosity [kg/m/s]"
-  ρν::FT
+  ρν::SVector{3,FT}
```
jkozdon: Why is this needed?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -174,7 +174,7 @@ end
 
 @inline function flux_diffusive!(atmos::AtmosModel, flux::Grad, state::Vars,
                                  diffusive::Vars, aux::Vars, t::Real)
-  ν, τ = turbulence_tensors(atmos.turbulence, state, diffusive, aux, t)
+  ν, τ, sdiff = turbulence_tensors(atmos.turbulence, state, diffusive, aux, t)
```
jkozdon: See comment below

experiments/LES/dycoms.jl
charleskawczynski: This will break dycoms, please revert this.

src/Atmos/Model/turbulence.jl
jkozdon: The reason this is a concern really is that we might kill our register usage on the GPU doing stuff that isn't required.

experiments/LES/dry_rayleigh_benard.jl
slifer50: No I will remove this, thanks.

experiments/LES/dycoms.jl
slifer50: No, I'll remove this.

experiments/LES/dycoms.jl
slifer50: No I'll clean that up.

src/Atmos/Model/turbulence.jl
slifer50: This was done for testing purposes, so we could set different constant viscosities in different directions.

src/Atmos/Model/AtmosModel.jl
slifer50: I'm currently looking for an effective way of not having this. Since we need a split viscosity for the courant numbers and non split viscosity for the general calculations. "sdiff" is needed  exclusively for the courant calculations.

slifer50: Since it doesn't look like I can avoid having to do a recomputation I think I will write separate function in turbulence to get the split viscosity.

src/Atmos/Model/turbulence.jl
```diff
@@ -241,14 +250,15 @@ end
 function diffusive!(::Vreman, orientation::Orientation,
                     diffusive::Vars, ∇transform::Grad, state::Vars, aux::Vars, t::Real)
   diffusive.turbulence.∇u = ∇transform.u
+  diffusive.turbulence.S = symmetrize(diffusive.turbulence.∇u)
```
jkozdon: This is storing the same thing twice, which will increase our GPU memory usage. Possibly we should just be storing the symmeterized version?

slifer50: The Vreman subgridscale model seems to need both, however it computes the symmeterized version locally. However I currently need S to be accessible to do the Courant calculation. This would be simple if all the subgridscale models either typically stored only of the two tensors but this seems to vary depending on the choice of turbulence closure. It might be best in this case to possibly only store the non symmeterized version for all closures and have them compute the symmetrized version locally so that it is never saved before this is we must be certain it is not needed elsewhere.

src/Atmos/Model/turbulence.jl
slifer50: @jkozdon noted, I will attempt to move this into the CFL computation itself as you suggest.

src/Atmos/Model/turbulence.jl
jkozdon: Recomputing simple algebraic quantities is almost alway preferential in general (loading data is the most expensive thing we do, so loading the same data multiple times in the same functions doesn't make sense).

When we start needing to do iterations to computing values or using special functions the picture changes a bit.

src/Atmos/Model/turbulence.jl
slifer50: Since I've already done tests with this, I can remove this and revert the changes for the constant viscosity closure, would that be alright @jkozdon ?

jkozdon:
> might be good to extract your functionality and add it to the exist #618 PR

Talked to @lcw and he is fine with you pushing commits to his branch.

The other advantage of this is that you would preserve his commit history, which means if code breaks that he wrote we will know who to talk to (right now since you took his code but not his commits it seems that you wrote all the code in the git history).

jkozdon:
> > might be good to extract your functionality and add it to the exist #618 PR
> 
> Talked to @lcw and he is fine with you pushing commits to his branch.

You can also get everything working here then we can create a diff commit that we push to his branch before merging the PR

src/Atmos/Model/turbulence.jl
jkozdon: I would suggest reverting it then if its not needed in the final PR

src/Atmos/Model/turbulence.jl
slifer50: It looks S isn't used outside of turbulence.jl I'm going to modify it so that only the non-symmeterized version is saved and S can be computed for each closure as needed.

slifer50:
> > > might be good to extract your functionality and add it to the exist #618 PR
> > 
> > 
> > Talked to @lcw and he is fine with you pushing commits to his branch.
> 
> You can also get everything working here then we can create a diff commit that we push to his branch before merging the PR

Sounds good, we can do that once I've addressed everything here.

src/Courant/Courant_presets.jl
tapios: Better yet: make the functions universal here, i.e., calculate the relevant soundspeed and viscosity/diffusivity in the atmosphere/ocean models, then call the functions here with those model-specific quantities. 

Please also change `_CFL` to `Courant_number` for consistency with the rest. 

BTW: we want to share the turbulence module between atmosphere and ocean in the end. The functions we need there are the same for atmosphere and ocean; just parameters are different.

slifer50: @tapios @jkozdon I could probably change the functions to use any BalanceLaw instead of AtmosModel allowing them to be used by any model, would this be alright?

tapios: Yes, that sounds good to me.

slifer50:
@jkozdon @tapios @lcw @charleskawczynski @smarras79 I believe I've adressed everything that's been mentioned so far in the latest 2 commits, please let me know if I've missed something.

src/Atmos/Model/turbulence.jl
```diff
@@ -321,13 +322,17 @@ end
 
 function turbulence_tensors(m::AnisoMinDiss, state::Vars, diffusive::Vars, aux::Vars, t::Real)
   FT = eltype(state)
+end
+function scaled_momentum_flux_tensor(m::AnisoMinDiss, ρν, S)
```
skandalaCLIMA: @slifer50 : Hi Yassine, I am working on merging the CFL pull requests #618 & #667. Can I request you to provide a clarification on this revision. 

With this revision, the function turbulence_tensors, just returns FT = eltype(state). This seems to be called in the "courant_presets" as ν, τ = turbulence_tensors(m.turbulence, state, diffusive, aux, 0).  

- please advise if the diffusive_courant function has been approved from the physics/numerics point of view. 
- it may be a good idea to make any revisions to the Turbulence model in a different pull request.

Thanks.
cc: @simonbyrne, @kpamnany @akshaysridhar @bischtob

simonbyrne:
Merged as part of #712.

PR #671, opened 2020-01-31
Bound `q_tot` in thermo state

charleskawczynski:
Since we've been having issues with `q_tot` not satisfying `0 <= q_tot <= 1`, let's force this constraint upon creating a thermo state for the default constructor.

Note that this does not apply to the initial conditions if defined with a different constructor (e.g., liquid ice potential temperature)


charleskawczynski:
cc: @tapios 
bors r+

tapios:
Looks good to me. Thanks so much for doing this so promptly, @charleskawczynski!

PR #674, opened 2020-02-03
Update advection diffusion test to Julia 1.3

jkozdon:
Update advection diffusion to use Julia 1.3 results; Slurm CI upgraded to with climate-machine/slurmci#11

### DIFF
--- a/test/DGmethods/advection_diffusion/pseudo1D_advection_diffusion.jl
+++ b/test/DGmethods/advection_diffusion/pseudo1D_advection_diffusion.jl
@@ -164,58 +164,51 @@ let
   base_num_elem = 4
 
   expected_result = Dict()
-  expected_result[2, 1, Float64, EveryDirection] = 1.2228434091602128e-02
-  expected_result[2, 2, Float64, EveryDirection] = 8.8037798002260420e-04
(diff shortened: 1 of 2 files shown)


charleskawczynski:
Thank you!

blallen:
should we wait for this to be merged before we add the commit to other outstanding PRs?

blallen:
just pushed a commit where I cut the endtime for `test_divergence_free.jl` in half. Hopefully that will prevent it from timing out again

charleskawczynski:
> just pushed a commit where I cut the endtime for test_divergence_free.jl in half. Hopefully that will prevent it from timing out again

Won't that break the results being checked?

blallen:
> > just pushed a commit where I cut the endtime for test_divergence_free.jl in half. Hopefully that will prevent it from timing out again
> 
> Won't that break the results being checked?

it's just checking that a constant scalar tracer stays constant. the test doesn't depend on the exact run time because it's just checking that min(θ) = max(θ) when θ=20 everywhere at the start

jkozdon:
> just pushed a commit where I cut the endtime for `test_divergence_free.jl` in half. Hopefully that will prevent it from timing out again

Weird that this has started to time out. Did something else change on the Slurm CI? Or is Julia 1.3 slower for CLIMA some how?

PR #677, opened 2020-02-04
Add isentropic temperature profile

mwarusz:
This adds a constant potential temperature profile for visualization of perturbations in the rising thermal bubble and density current test cases. This is also probably a better state for linearization in the IMEX rising thermal bubble test.

@thomasgibson

### DIFF
--- a/src/Atmos/Model/ref_state.jl
+++ b/src/Atmos/Model/ref_state.jl
@@ -1,6 +1,7 @@
 ### Reference state
 using DocStringExtensions
-export NoReferenceState, HydrostaticState, IsothermalProfile, LinearTemperatureProfile
+export NoReferenceState, HydrostaticState, IsothermalProfile, LinearTemperatureProfile,
+       DryAdiabaticProfile
 
 """
     ReferenceState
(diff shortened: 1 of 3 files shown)


mwarusz:
bors try

tapios:
A simpler way to get this is to use `LinearProfile` with lapse rate `grav/cp`. You can write a separate isentropic constructor for this (in which case I would call is `DryIsentropic` because it does not take moisture into account). Or simply use `LinearProfile` with g/cp for the lapse rate.

tapios:
And just to add context why I strongly suggest using the `LinearProfile` setup that is already there: We will only use strictly isentropic (adiabatic) profiles in idealized cases with relatively small vertical domain extents. More generally, we want to at least truncate at a minimum temperature because the adiabatic temperature profiles become unrealistic higher up (eventually reaching negative temperatures). And in cases with moisture, we want more stable lapse rates in the reference state. See the atmospheric design doc.

mwarusz:
@tapios @charleskawczynski 

Changed to reuse `LinearProfile`. Renamed to `DryAdiabaticProfile`.

mwarusz:
bors try

src/Atmos/Model/ref_state.jl
```diff
@@ -90,6 +91,27 @@ function (profile::IsothermalProfile)(orientation::Orientation, aux::Vars)
   return (profile.T, p)
 end
 
+"""
+    DryAdiabaticProfile{F} <: TemperatureProfile
+
+A uniform potential temperature profile.
```
tapios:
```suggestion
A temperature profile that has uniform potential temperature up to the minimum temperature `T_min`
```

mwarusz: Done. Thanks.

tapios:
@mwarusz @charleskawczynski Can we complete and merge this? It looks good to me.

mwarusz:
bors try

PR #680, opened 2020-02-04
Add isentropic relations

charleskawczynski:
- Adds isentropic relations to moist thermo.

I'm not sure what others we need, but it would be nice to change all instances of these hard-coded equations into function calls.

### DIFF
--- a/src/Common/MoistThermodynamics/isentropic.jl
+++ b/src/Common/MoistThermodynamics/isentropic.jl
@@ -1,0 +1,48 @@
+#####
+##### Formulas assuming dry (unsaturated) adiabatic (i.e. isentropic) processes
+#####
+
+export DryAdiabaticProcess
+
+export air_pressure_given_θ, air_pressure, air_temperature
+
+"""
+    DryAdiabaticProcess
+
+For dispatching to isentropic formulas
+"""
+struct DryAdiabaticProcess end
+
+"""
+    air_pressure_given_θ(θ::FT, Φ::FT, ::DryAdiabaticProcess)
+
+The air pressure for an isentropic process, where
+
+ - `θ` potential temperature
+ - `Φ` gravitational potential
+"""
+air_pressure_given_θ(θ::FT, Φ::FT, ::DryAdiabaticProcess) where {FT} =
+  FT(MSLP) * (1 - Φ / (θ * FT(cp_d))) ^ (FT(cp_d) / FT(R_d))
+
+"""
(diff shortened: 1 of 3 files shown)


src/Common/MoistThermodynamics/isentropic.jl
```diff
@@ -0,0 +1,44 @@
...
+"""
+struct IsentropicProcess end
+
+"""
+    air_pressure_given_θ(θ::FT, Φ::FT, ::IsentropicProcess)
+
+The air pressure for an isentropic process, where
+
+ - `θ` potential temperature
+ - `Φ` gravitation potential
```
tapios:
```suggestion
 - `Φ` gravitational potential
```

src/Common/MoistThermodynamics/isentropic.jl
```diff
@@ -0,0 +1,44 @@
+export IsentropicProcess
```
tapios:
```suggestion
export DryAdiabaticProcess
```
(and similar throughout). I'd make clear that these are all dry isentropic (or adiabatic) processes.

src/Common/MoistThermodynamics/isentropic.jl
```diff
@@ -0,0 +1,44 @@
...
+"""
+    air_pressure_given_θ(θ::FT, Φ::FT, ::IsentropicProcess)
+
+The air pressure for an isentropic process, where
+
+ - `θ` potential temperature
+ - `Φ` gravitation potential
+"""
+air_pressure_given_θ(θ::FT, Φ::FT, ::IsentropicProcess) where {FT} =
+  FT(MSLP) * (1 - Φ / (θ * FT(cp_d))) ^ (FT(cp_d) / FT(R_d))
```
tapios: Consider using moist specific heats, gas constants etc. here and elsewhere?

tapios:
Thanks, @charleskawczynski. This looks useful. We'll need such functions in boundary condition/surface flux implementations. I'd use moist thermodynamic quantities (`R_m` etc.) in the calculations, for somewhat greater accuracy. Otherwise, LGTM.

src/Common/MoistThermodynamics/isentropic.jl
charleskawczynski: Wouldn't using moist specific heats contradict `DryAdiabaticProcess`?

tapios: By ‘dry’ we usually mean unsaturated and non-condensing (as in an ocean is dry :). If you want to be more specific, you can say ‘unsaturated’.

PR #684, opened 2020-02-05
Clean up dycoms BC

jkozdon:
@mwarusz found a bug in the DYCOMS bc that no boundary conditions was being enforced on the top boundary on the diffusive variables.

This PR fixes that as well as removes unused variables in the boundary conditions (such as some unused moist physics calls) and also refactors the `DYCOMS_BC` to reuse values computed for the `NoFluxBC` whenever possible.

Co-authored-by: Maciej Waruszewski <mwarusz@igf.fuw.edu.pl>

### DIFF
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -134,33 +134,27 @@ struct DYCOMS_BC{FT} <: BoundaryCondition
   LHF::FT
   SHF::FT
 end
-function atmos_boundary_state!(nf::Union{NumericalFluxNonDiffusive, NumericalFluxGradient},
-                               bc::DYCOMS_BC, m::AtmosModel, stateP::Vars,
-                               auxP::Vars, nM, stateM::Vars, auxM::Vars, bctype,
-                               t, state1::Vars, aux1::Vars)
-  # stateM is the 𝐘⁻ state while stateP is the 𝐘⁺ state at an interface.
-  # at the boundaries the ⁻, minus side states are the interior values
-  # state1 is 𝐘 at the first interior nodes relative to the bottom wall
-  FT = eltype(stateP)
-  # Get values from minus-side state
-  ρM = stateM.ρ
-  UM, VM, WM = stateM.ρu
-  EM = stateM.ρe
-  QTM = stateM.moisture.ρq_tot
-  uM, vM, wM  = UM/ρM, VM/ρM, WM/ρM
-  q_totM = QTM/ρM
 
-  # Assign reflection wall boundaries (top wall)
-  if typeof(nf) <: NumericalFluxNonDiffusive
-    stateP.ρu -= 2dot(stateM.ρu, nM) * SVector(nM)
-  else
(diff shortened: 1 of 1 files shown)


jkozdon:
bors try

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -228,27 +190,33 @@ function atmos_boundary_flux_diffusive!(nf::NumericalFluxDiffusive,
     # Boundary momentum fluxes
     # ----------------------------------------------------------
     # Case specific for flat bottom topography, normal vector is n⃗ = k⃗ = [0, 0, 1]ᵀ
-    # A more general implementation requires (n⃗ ⋅ ∇A) to be defined where A is replaced by the appropriate flux terms
+    # A more general implementation requires (n⃗ ⋅ ∇A) to be defined where A is
+    # replaced by the appropriate flux terms
     C_drag = bc.C_drag
-    τ13⁺  = - C_drag * windspeed_FN * u_FN
-    τ23⁺  = - C_drag * windspeed_FN * v_FN
+    τ13⁺ = - C_drag * windspeed_FN * u_FN[1]
```
mwarusz: Add `@inbounds` here.

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -228,27 +190,33 @@ function atmos_boundary_flux_diffusive!(nf::NumericalFluxDiffusive,
...
+    τ23⁺ = - C_drag * windspeed_FN * u_FN[2]
+    τ21⁺ = τ⁻[2,1]
```
mwarusz: Add `@inbounds` here.

jkozdon: Good catch. Thanks!

smarras79:
Thanks for that. 
However, the top should only be controlled by the relaxation due to the sponge that should allow a free slip with a no-penetration b.c. on the vertical component on velocity. It may be that a viscous b.c. on the top will actually cause problems for this specific tests. I will test it as soon as we know that the simulation preserves stability all the way to completion.

jkozdon:
Even with a sponge layer, I’m pretty sure you need to use a well-posed boundary condition which it wasn’t before.

lcw:
In my experience with linear wave propagation we always set a boundary condition where the domain is artificially truncated.  This is not a physical boundary condition but is needed to prevent numerical instability.

If the sponge is doing its job, in my experience, it doesn't matter what boundary condition you set for the artificially truncated domain.

jkozdon:
> If the sponge is doing its job, in my experience, it doesn't matter what boundary condition you set for the artificially truncated domain.

But in your experience, we need to set something reasonable (e.g., enough boundary conditions)?

lcw:
> But in your experience, we need to set something reasonable (e.g., enough boundary conditions)?

That is right, something well posed.

smarras79:
> > But in your experience, we need to set something reasonable (e.g., enough boundary conditions)?
> 
> That is right, something well posed.

Perfect, thanks for clarifying. I'll test dycoms with those b.c.

PR #685, opened 2020-02-06
Make numerical boundary flux gradient handling consistent with nondiffusive

simonbyrne:
Slight tweak to changes from #660

cc @jkozdon

### DIFF
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -82,11 +82,7 @@ function atmos_boundary_state!(nf::Union{NumericalFluxNonDiffusive, NumericalFlu
                                t, _...)
   FT = eltype(stateM)
   stateP.ρ = stateM.ρ
-  if typeof(nf) <: NumericalFluxNonDiffusive
-    stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
-  else
-    stateP.ρu -=  dot(stateM.ρu, nM) * SVector(nM)
-  end
+  stateP.ρu -= 2 * dot(stateM.ρu, nM) * SVector(nM)
 end
 
 function atmos_boundary_state!(::NumericalFluxDiffusive, bc::NoFluxBC,
@@ -96,7 +92,7 @@ function atmos_boundary_state!(::NumericalFluxDiffusive, bc::NoFluxBC,
   FT = eltype(stateM)
   stateP.ρ = stateM.ρ
   stateP.ρu -= dot(stateM.ρu, nM) * SVector(nM)
-  
+
   fill!(getfield(diffP, :array), FT(0))
 end
 
@@ -151,11 +147,7 @@ function atmos_boundary_state!(nf::Union{NumericalFluxNonDiffusive, NumericalFlu
   q_totM = QTM/ρM
 
   # Assign reflection wall boundaries (top wall)
-  if typeof(nf) <: NumericalFluxNonDiffusive
-    stateP.ρu -= 2dot(stateM.ρu, nM) * SVector(nM)
-  else
-    stateP.ρu -= dot(stateM.ρu, nM) * SVector(nM)
-  end
+  stateP.ρu -= 2dot(stateM.ρu, nM) * SVector(nM)
 
   # Assign scalar values at the boundaries
   stateP.ρ = ρM
@@ -275,11 +267,7 @@ function atmos_boundary_state!(nf::Union{NumericalFluxNonDiffusive, NumericalFlu
   @inbounds begin
     FT = eltype(stateP)
     stateP.ρ = ρP = stateM.ρ
-    if typeof(nf) <: NumericalFluxNonDiffusive
-      stateP.ρu = -stateM.ρu
-    else
-      stateP.ρu = SVector{3,FT}(0,0,0)
-    end
+    stateP.ρu = -stateM.ρu
     if bctype == 1
       E_intP = ρP * cv_d * (bc.T_bot - T_0)
     else
--- a/src/DGmethods/NumericalFluxes.jl
+++ b/src/DGmethods/NumericalFluxes.jl
@@ -58,8 +58,9 @@ function numerical_boundary_flux_gradient!(nf::CentralNumericalFluxGradient,
   boundary_state!(nf, bl, state⁺, aux⁺, n, state⁻, aux⁻, bctype, t, state1⁻,
                   aux1⁻)
   gradvariables!(bl, transform⁺, state⁺, aux⁺, t)
-  G = n .* parent(transform⁺)'
-  diffusive!(bl, fluxᵀn, Grad{T}(G), state⁻, aux⁻, t)
+  numerical_flux_gradient!(nf, bl, fluxᵀn, n,
+    transform⁻, state⁻, aux⁻,
+    transform⁺, state⁺, aux⁺, t)
 end
 
 


jkozdon:
So I worked out the analysis for advection-diffusion, and this would be wrong in that case and you really want what is currently in the code, as this way would be linearly unstable.

@fxgiraldo and I were looking again at Bassi and Rebay, and they would suggest in the case of a solid wall setting the flux directly (as opposed to calling the numerical flux function), but in the case of advection-diffusion with Dirichlet BC you do want to call the numerical flux.

@fxgiraldo looked at the case of dropping the if statement and setting `stateP.ρu -= dot(stateM.ρu, nM) * SVector(nM)` and things were stable, but we lose mass conservation.

simonbyrne:
> So I worked out the analysis for advection-diffusion, and this would be wrong in that case and you really want what is currently in the code, as this way would be linearly unstable.

Can you expand on this, either here or in the design docs?

> @fxgiraldo and I were looking again at Bassi and Rebay, and they would suggest in the case of a solid wall setting the flux directly (as opposed to calling the numerical flux function), but in the case of advection-diffusion with Dirichlet BC you do want to call the numerical flux.

I'm not sure I understand: which of the numerical fluxes (nondiff, gradient, diff) should we set?

fxgiraldo:
@simonbyrne let me know if this is still an issue.  I added to the discussion on slack but am not sure if this conversation here is OBE or still active.

simonbyrne:
@fxgiraldo could you address my questions above about the advection diffusion case?

jkozdon:
@simonbyrne I added something to the design docs. Wasn't sure where to put it, so I dumped it before the BC section. Feel free to move it around or ask question (best to ping me directly rather than through the doc).

The heart of the matter is that boundary conditions are not trivial to implement in DG, and unfortunately naive choices can lead to problems. In the case of advection-diffusion we need to do different things on the two equations to impose the boundary condition.

For non-linear equation and systems the choices are even trickier and what you can do for a stable scheme can depend on which boundary you are on. For advection-diffusion, we are using a unified treatment for both boundaries, but you could do something different since one boundary is an inflow boundary and the other an outflow; the inflow has to be Dirichlet but the outflow could be Dirichlet or Neumann thus this can allow for different numerical implementation choices even in the Dirichlet-Dirchlet case.

simonbyrne:
Thanks Jeremy, that is very much appreciated.

> The heart of the matter is that boundary conditions are not trivial to implement in DG, and unfortunately naive choices can lead to problems.

I get this, but unfortunately boundary conditions are actually going to be at the core of our work, as they will form the interfaces by which the component models interact (via transfer of water, energy, etc), so it is essential we get this right. And unfortunately the complexity of the models and the boundary conditions is only going to increase.

Maybe we can turn this problem around: what are the principals we can use to implement boundary conditions? how can we verify that what we're doing makes sense? Are there numerical diagnostics we can use? 

I do like your example of considering a single element: perhaps if we can build up a library of such simple cases, it might highlight what we need

lcw:
> Maybe we can turn this problem around: what are the principals we can use to implement boundary conditions? how can we verify that what we're doing makes sense? Are there numerical diagnostics we can use?
> 
> I do like your example of considering a single element: perhaps if we can build up a library of such simple cases, it might highlight what we need

There is a nice paper by Jan Nordström <https://doi.org/10.1007/s10915-016-0303-9> that has a similar energy based analysis to what Jeremy did but for Navier–Stokes.  It has a section (3.1 The Roadmap) that give a roadmap for developing stable boundary conditions.  Is this what you are looking for @simonbyrne?

blallen:
Question:

Can we switch back to using a central flux for the `numerical_boundary_flux_gradient!` to solve this? Basically instead of setting the numerical flux for the gradient to be `H_s^* = u_bc n`, we set it to be `H_s^* = n/2 (u^- + u_bc)`.

Thus if we set `u^+ = u^-` for the boundary state to impose no-slip boundary conditions, we will recover `H_s^* = 0` without having to specify a different boundary state for the `numerical_boundary_flux_gradient!`.

I know this takes us out of agreement with Bassi-Rabay, but it achieves the result we need for stability, semi-automatically (at least for this case).

blallen:
Also:

@sandreza has worked through some examples in https://github.com/sandreza/DG_Playground and has seen that he can specify the unstable choices of ghost points and not have 1D advection-diffusion blow up because of using Rusanov fluxes.

Is there a way of doing the stability analysis with Rusanov fluxes?

simonbyrne:
My idea is to add a Trait that allows us to specify how a boundary condition will be applied in each case
```julia
abstract type NumericalBoundaryMethod end

# Specify plus state, pushes through numerical flux
# e.g. how nondiffusive flux is currently applied,
struct NumericalBoundaryStateOpposite <: BoundaryTrait end

# Specify boundary state directly
# e.g. how gradient flux and diffusive flux currently operate
struct NumericalBoundaryStateDirect <: BoundaryTrait end

# Specify flux directly along the normal component
# e.g. what #733 does for diffusive fluxes
struct NumericalBoundaryFluxDirectNormal <: BoundaryTrait end
```

jkozdon:
> Can we switch back to using a central flux for the `numerical_boundary_flux_gradient!` to solve this? Basically instead of setting the numerical flux for the gradient to be `H_s^* = u_bc n`, we set it to be `H_s^* = n/2 (u^- + u_bc)`.
> 
> Thus if we set `u^+ = u^-` for the boundary state to impose no-slip boundary conditions, we will recover `H_s^* = 0` without having to specify a different boundary state for the `numerical_boundary_flux_gradient!`.

Seems like you would get `H_s^* = u^-` which is essentially no boundary condition. Seems like it would be preferable to just set `H_s^* = 0`.

But calling the numerical flux on the gradient boundary state will lead to possible instability; see what I put in the design doc, unless you do something like `u^+ = 2ubc - u^-`, which seems more ackward to me than `u^+ = ubc`

Keep in mind, once this BC is known how to set, it doesn't seem like its that difficult to mimmic later. Ultimately, whatever you do in the kernel needs to be consistent with what's done in the model. For each BC type, you may do things a bit differently like is discussed in the paper @lcw linked do [above](https://github.com/climate-machine/CLIMA/pull/685#issuecomment-583555509) by Jan Nordström <https://doi.org/10.1007/s10915-016-0303-9>.

> @sandreza has worked through some examples in https://github.com/sandreza/DG_Playground and has seen that he can specify the unstable choices of ghost points and not have 1D advection-diffusion blow up because of using Rusanov fluxes.

When you say stable, do you mean that code didn't blow up or that you looked at the eigenvalues of the ODE system and they did not have a positive real part. The energy analysis I outlined in the design doc essentially shows that the eigenvalue spectrum has a non-positive real part.

Even if the code doesn't blow up for a given test, it might still be unstable in the sense that there are are eigenvalues with positive real part. One possibility is that the ODE solver could be damping these modes (particularly if dynamically adaptive time stepping is used). Alternatively, you may not have sufficient energy in these modes for the simulation to have blow up. If the positive real part is small, it can take a long time to blow (but would eventually.

> Is there a way of doing the stability analysis with Rusanov fluxes?

You should be able to modify what I outlined in the design doc. I believe that Rusanov for advection-diffusion should reduce to upwinding (at least in 1-D).

blallen:
> > Can we switch back to using a central flux for the `numerical_boundary_flux_gradient!` to solve this? Basically instead of setting the numerical flux for the gradient to be `H_s^* = u_bc n`, we set it to be `H_s^* = n/2 (u^- + u_bc)`.
> > Thus if we set `u^+ = u^-` for the boundary state to impose no-slip boundary conditions, we will recover `H_s^* = 0` without having to specify a different boundary state for the `numerical_boundary_flux_gradient!`.
> 
> Seems like you would get `H_s^* = u^-` which is essentially no boundary condition. Seems like it would be preferable to just set `H_s^* = 0`.

Ah, I have a very unfortunate typo. I meant to say set `u^+ = -u^-`. The rest of your points still hold though.

simonbyrne:
Also, we should use separate functions for specifying the "boundary state" and "ghost state"

simonbyrne:
Closing for now.

PR #672, opened 2020-01-31
Remove old drivers

kpamnany:
~~Based on #592 but does not include the Rayleigh Benard part~~. Updated version of #592.

This doesn't work. @akshaysridhar: need your help on this one!

Closes #616.

PR #686, opened 2020-02-06
Make AtmosModel use keywords

charleskawczynski:
- Modifies the drivers so that `model::AtmosModel` is an input
 - Make AtmosModel use kwdef

This is a useful first step before adding free parameters


charleskawczynski:
@simonbyrne ready for review!

tapios:
Can we just get rid of `examples/DGmethods/ex_001_dycoms-imex.jl `? It's confusing to have this setup specified in two different places (and continue to update it in both).

charleskawczynski:
> Can we just get rid of examples/DGmethods/ex_001_dycoms-imex.jl ? It's confusing to have this setup specified in two different places (and continue to update it in both).

@akshaysridhar is removing that example in #672

simonbyrne:
LGTM:  @kpamnany does this work for you?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -52,6 +52,40 @@ struct AtmosModel{O,RS,T,M,P,R,SU,S,BC,IS} <: BalanceLaw
...
+                         precipitation::P=NoPrecipitation(),
+                         radiation::R=NoRadiation(),
+                         subsidence::SU=NoSubsidence{FT}(),
+                         source::S=( Gravity(),
+                                     Coriolis(),
+                                     GeostrophicForcing{FT}(7.62e-5, 0, 0)),
+                         # TODO: Probably want to have different bc for state and diffusion...
+                         boundarycondition::BC=NoFluxBC(),
+                         init_state::IS=nothing) where {FT<:AbstractFloat,O,RS,T,M,P,R,SU,S,BC,IS}
+
```
kpamnany: The defaults chosen in `LES_Configuration()` and `GCM_Configuration()` were discussed extensively with @akshaysridhar and were 'reasonable' for the two types of configurations. I suggest two constructors, with maybe something like:
```
abstract type AtmosConfiguration end
struct AtmosLESConfiguration <: AtmosConfiguration end
struct AtmosGCMConfiguration <: AtmosConfiguration end
function AtmosModel{FT}(::Type{AtmosLESConfiguration}; ...) ... end
function AtmosModel{FT}(::Type{AtmosGCMConfiguration}; ...) ... end
```

src/Driver/Configurations.jl
```diff
@@ -131,31 +110,15 @@ function GCM_Configuration(name::String,
...
-                           turbulence         = SmagorinskyLilly{FT}(C_smag),
-                           moisture           = EquilMoist(),
-                           precipitation      = NoPrecipitation(),
-                           radiation          = NoRadiation(),
-                           subsidence         = NoSubsidence{FT}(),
-                           sources            = (Gravity(), Coriolis()),
-                           bc                 = NoFluxBC(),
+                           model              = AtmosModel{FT}(;orientation=SphericalOrientation(),
+                                                                sources=(Gravity(), Coriolis()),
+                                                                init_state=init_GCM!),
```
kpamnany: Then this would just use `model = AtmosModel{FT}(AtmosGCMConfiguration())`.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Ah, that's exactly what we, thanks!

src/Driver/Configurations.jl
charleskawczynski: Yes, but we'll still need the `init_state`, for now

akshaysridhar:
> This is no longer using `Base.@kwdef` despite the title. Nonetheless, I'll reproduce my comment from Slack here: I think constructors are a good idea anyway, because they let you add keyword arguments that allow changing the parameters to the contained types. E.g. https://github.com/climate-machine/CLIMA/blob/master/src/Driver/Configurations.jl#L80-L87. Or https://github.com/climate-machine/CLIMA/blob/master/src/Driver/Configurations.jl#L88-L89. And we can add more specialized kinds of constructors.
> 
> Having said that, although it is in current `master` in the configuration, I'm not sure it's a good thing to have e.g. `C_smag = 0.21` as a keyword parameter because although it is convenient, it is particular to `SmagorinskyLilly()` and meaningless (?) if you use a different turbulence model.
> 
> This PR removes those specific parameters, which is cleaner but perhaps slightly less convenient. Either way, I like the way this is done.

Right - other turbulent models will have their own coefficient specifications; C_smag is specific to `Constant/Dynamic Smagorinsky` / `Vreman` type models.`; its default is perhaps best specified within `turbulence.jl`

charleskawczynski:
bors try

charleskawczynski:
Since this touches a lot of drivers, it would be nice to get a second set of eyes to make sure that we haven't ruined results for drivers that don't have tests

simonbyrne:
bors try

charleskawczynski:
bors try

test/Mesh/interpolation.jl
```diff
@@ -192,16 +187,13 @@ function run_cubed_sphere_interpolation_test()
...
-                       NoReferenceState(),
-                       ConstantViscosityWithDivergence(FT(0)),
-                       DryModel(),
-                       NoPrecipitation(),
-                       NoRadiation(),
-                       NoSubsidence{FT}(),
-                       nothing,
-                       NoFluxBC(),
-                       setup)
+    model = AtmosModel{FT}(AtmosLESConfiguration;
```
simonbyrne: I realise this is just used as part of the setup, but does it make sense to allow an `AtmosLESConfiguration` to use `SphericalOrientation`?

charleskawczynski: It seems like a `SphericalOrientation` is used in `test/DGmethods/Euler/acousticwave-1d-imex.jl`, I just used LES here to not change anything.

simonbyrne: Why not use GCM?

charleskawczynski: I think I decided based on the orientation (this originally used a `FlatOrientation()`), but I guess this does make more sense in the context of a spherical orientation/gcm. Ultimately, I was trying to not change the flow control. Want me to try changing this to `GCM` with `SphericalOrientation`?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -52,6 +54,78 @@ struct AtmosModel{O,RS,T,M,P,R,SU,S,BC,IS} <: BalanceLaw
   init_state::IS
 end
 
+abstract type AtmosConfiguration end
+struct AtmosLESConfiguration <: AtmosConfiguration end
+struct AtmosGCMConfiguration <: AtmosConfiguration end
```
kpamnany: Unfortunately, we should probably reconsider these names because #689 will rename `LES_Configuration()` to `Atmos_LES_Configuration()`, etc. Maybe `AtmosLESConfigType`?

kpamnany:
Just to note: #694 should merge today and will shrink this PR a good bit.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I'm fine with `AtmosLESConfigType`.

charleskawczynski:
> Just to note: #694 should merge today and will shrink this PR a good bit.

Ok, should we wait for #694 then?

charleskawczynski:
bors try

charleskawczynski:
bors try-

PR #687, opened 2020-02-07
Simplify using `CLIMA.Atmos` in drivers

charleskawczynski:
Changing `using CLIMA.Atmos: ...` in all of the drivers is a bit of a pain. This PR exports a few extra things from AtmosModel, and makes drivers just have `using CLIMA.Atmos`.

### DIFF
--- a/examples/Atmos/heldsuarez.jl
+++ b/examples/Atmos/heldsuarez.jl
@@ -11,13 +11,9 @@ using CLIMA.VTK: writevtk, writepvtu
 using CLIMA.GenericCallbacks: EveryXWallTimeSeconds, EveryXSimulationSteps
 using CLIMA.MPIStateArrays: euclidean_distance
 using CLIMA.PlanetParameters: R_d, grav, MSLP, planet_radius, cp_d, cv_d, day
-using CLIMA.MoistThermodynamics: air_density, total_energy, soundspeed_air, internal_energy, air_temperature
-using CLIMA.Atmos: AtmosModel, SphericalOrientation, NoReferenceState,
-                   DryModel, NoPrecipitation, NoRadiation, NoSubsidence, NoFluxBC,
-                   ConstantViscosityWithDivergence,
-                   vars_state, vars_aux,
-                   Gravity, Coriolis,
-                   HydrostaticState, IsothermalProfile
+using CLIMA.MoistThermodynamics
+using CLIMA.Atmos
+using CLIMA.Atmos: vars_state, vars_aux
 using CLIMA.VariableTemplates: flattenednames
 
 using MPI, Logging, StaticArrays, LinearAlgebra, Printf, Dates, Test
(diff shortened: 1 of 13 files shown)


kpamnany:
Need to ensure we aren't polluting the namespace for the other models.

charleskawczynski:
> Need to ensure we aren't polluting the namespace for the other models.

I completely agree; this does violate the "take only what you need" principle. But, all drivers seem to need to be updated every time new pieces are adjusted in the AtmosModel. Perhaps this won't be an issue after #686. I'm fine waiting to see if it's still a hassle after #686.

src/Atmos/Model/moisture.jl
```diff
@@ -1,4 +1,5 @@
-export DryModel, EquilMoist
+import ..MoistThermodynamics: internal_energy
+export DryModel, EquilMoist, thermo_state
```
blallen: `thermo_state` might potentially be used by Ocean later on?

src/Atmos/Model/remainder.jl
```diff
@@ -1,3 +1,5 @@
+export RemainderModel
```
blallen: this entire struct / file should probably be moved into DGmethods as nothing here is atmosphere specific. if you remove L35 (which doesn't seem to be used for anything), this becomes entirely generic

src/Atmos/Model/turbulence.jl
```diff
@@ -2,7 +2,7 @@
 using DocStringExtensions
 using CLIMA.PlanetParameters
 using CLIMA.SubgridScaleParameters
-export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss
+export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss, turbulence_tensors
```
blallen: Ocean half will probably want to use `turbulence_tensors` in the future

examples/Atmos/heldsuarez.jl
```diff
@@ -11,13 +11,9 @@ using CLIMA.VTK: writevtk, writepvtu
 using CLIMA.GenericCallbacks: EveryXWallTimeSeconds, EveryXSimulationSteps
 using CLIMA.MPIStateArrays: euclidean_distance
 using CLIMA.PlanetParameters: R_d, grav, MSLP, planet_radius, cp_d, cv_d, day
-using CLIMA.MoistThermodynamics: air_density, total_energy, soundspeed_air, internal_energy, air_temperature
-using CLIMA.Atmos: AtmosModel, SphericalOrientation, NoReferenceState,
-                   DryModel, NoPrecipitation, NoRadiation, NoSubsidence, NoFluxBC,
-                   ConstantViscosityWithDivergence,
-                   vars_state, vars_aux,
-                   Gravity, Coriolis,
-                   HydrostaticState, IsothermalProfile
```
blallen: Should I also be checking if we might need to use any of the names that were removed from the specific `using` calls?

kpamnany: Such were potentially not being used, so it would be better to check what is being exported across the board.

src/Atmos/Model/moisture.jl
charleskawczynski: I believe this is the plan

charleskawczynski:
This has gotten quite stale, and it seems a bit less of an issue now that #686 is in.

PR #688, opened 2020-02-07
Take ODESolvers outside of module in ODESolvers

charleskawczynski:


test/DGmethods/Euler/isentropicvortex-multirate.jl
```diff
@@ -143,15 +139,16 @@ function run(mpicomm, ArrayType, polynomialorder, numelems, setup,
...
+  # arbitrary and not needed for stability, just for testing
   fast_dt = slow_dt / 3
 
   Q = init_ode_state(dg, FT(0))
 
   slow_ode_solver = LSRK144NiegemannDiehlBusch(slow_dg, Q; dt = slow_dt)
 
   # check if FastMethod is ARK, is there a better way ?
-  if isdefined(AdditiveRungeKuttaMethod, Symbol(FastMethod))
+  if FastMethod == ARK2GiraldoKellyConstantinescu
```
charleskawczynski: @mwarusz I wasn't sure how to translate this since this PR merges the modules. Is this okay for now? I added an `AbstractAdditiveRungeKutta` type, but I'm still not sure how to maintain the intended flow control.

mwarusz: That's fine for now. Actually, this is more correct than what was there before. I call `FastMethod` with  the `paperversion` keyword argument which is only an argument of `ARK2GiraldoKellyConstantinescu`.

PR #689, opened 2020-02-08
First attempt at running an Ocean model with Driver + Configuration setup.

blallen:
I converted test_divergence_free.jl to run with the new set up. Haven't tested the setup with IMEX yet, only fully explicit. Includes a fleshing out of the Problem concept, as this enabled an easy way to dispatch on initialization and boundary conditions functions. 

The two inelegant things in the current iteration are specifying the timestep and passing through filters. I've handled both of these with user-defined functions with nothing defaults, but there should be a better way to do this.

Ideally, I'd like to be able to merge four scripts from `test/Ocean/Hydrostatic_Boussinesq` into a single script, currently called `experiments/HBM/test_divergence_free.jl` (likely to be renamed `simple_box.jl` once that is accomplished). This requires an easy way to switch solver_type and problem, ideally independent of each other, as well as potentially a way to turn on and off tests depending on those settings.


christophernhill:
@blallen a few comments/thoughts

 * dt not defined - maybe e.g.
          ```
             ERROR: LoadError: UndefVarError: dt not defined
             at /home/vsts/work/1/s/src/Driver/Driver.jl:257
         ```
   in Azure, Linux. 
  That fails merge tests, but I am not sure from code why it is happening i.e.
       ``dt`` looks at first glance like it is there for all cases? 
   

 * could we split Driver/Configurations.jl so that each of
    LES_Configuration
    GCM_Configuration
    HBM_Configuration
     are in separate files. I feel like the pattern of having them all in one file won't
     be sustainable.

* In high-level driver here https://github.com/climate-machine/Oceananigans.jl/blob/master/examples/ocean_wind_mixing_and_convection.jl
    there are many more comments in driver code. What do we think about starting to use Documenter.jl? This can give docs like
     https://climate-machine.github.io/Oceananigans.jl/latest/generated/ocean_wind_mixing_and_convection/
     when comments are included etc... The drivers in this repo don't have a lot of narrative comments, so knowledge and thinking may get lost over time - or not get shared easily. 

* in

    CLIMA/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl

     is there a way to organize all the different signature ocean_boundary_state!()
     functions.
     There aren't any comments saying much about what each one is. Some comments
     might be useful. Some (for example https://github.com/blallen/CLIMA/blob/1bff61b7399361763e53ac30a91768e901f80420/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl#L303 or https://github.com/blallen/CLIMA/blob/1bff61b7399361763e53ac30a91768e901f80420/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl#L321) 
    have signatures that aren't really consistent with what the function does. Its a little
    unclear if that is deliberate, or just something left over from development. 
   
    Should things like this be tidied as you merge into master? That slows things down, but without
    some tidying things can get hard to reason about and can be confusing when
    revisited later etc....


 *  in CLIMA/src/Driver/Driver.jl is there any interest in using init(), run(), finalize() for  the main "life-cycle" stages. We trained a lot of the climate modeling community to use those phrases in previous project and it is fairly widely used terminology.

tapios:
Thanks, @christophernhill, these are all good comments, especially about documentation, splitting the configurations file, and using init(), run() etc. Could you guys please iterate with @kpamnany to work these into the driver/configuration setup broadly?

tapios:
To make names clearer, consistent with what people use, and unambiguous, how about using

GCM -> AGCM
LES -> ALES
HBM -> OGCM

(Or something similar. “HBM” is difficult to parse.)

kpamnany:
@blallen and I have just discussed how to integrate this, taking into consideration other open PRs that touch on related things, and we've worked out a plan. The `dt` issue is a minor bug and the configurations file will be split as suggested. We do need more and better documentation in the CLIMA code base across the board.

Re: function naming: we do have an `init()`, but Julia already defines a `run()` in `Base` which is why `invoke!()` was chosen. We can change it to `run()` if that's preferred anyway? There is no real `finalize()` step at the moment.

kpamnany:
> GCM -> AGCM
> LES -> ALES

I'll change these.

> HBM -> OGCM

I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?

tapios:
> > GCM -> AGCM
> > LES -> ALES
> 
> I'll change these.
> 
> > HBM -> OGCM
> 
> I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?

Perhaps @rafferrari can weigh in here. I think all ocean GCM configurations that are planned will be HBM. Ocean LES will be nonhydrostatic. I was primarily looking to make names consistent with what people in the community are used to.

blallen:
> > > GCM -> AGCM
> > > LES -> ALES
> > 
> > 
> > I'll change these.
> > > HBM -> OGCM
> > 
> > 
> > I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?
> 
> Perhaps @rafferrari can weigh in here. I think all ocean GCM configurations that are planned will be HBM. Ocean LES will be nonhydrostatic. I was primarily looking to make names consistent with what people in the community are used to.

The reason I didn't call it Ocean GCM was to distinguish the model that runs on the sphere and the model that runs in a box. While the underlying physics/PDEs are the same for the two, the types of problems you want to solve are different. This distinction might not matter long term, but the way `Configurations.jl` is currently written requires splitting the two. 

If you look at L190-L197, you can see that `HBM_Configuration` will always set up a regular, rectangular `StackedBrickTopology` which is not generally what people mean when they say GCM, as far as I know.

tapios:
> > > > GCM -> AGCM
> > > > LES -> ALES
> > > 
> > > 
> > > I'll change these.
> > > > HBM -> OGCM
> > > 
> > > 
> > > I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?
> > 
> > 
> > Perhaps @rafferrari can weigh in here. I think all ocean GCM configurations that are planned will be HBM. Ocean LES will be nonhydrostatic. I was primarily looking to make names consistent with what people in the community are used to.
> 
> The reason I didn't call it Ocean GCM was to distinguish the model that runs on the sphere and the model that runs in a box. While the underlying physics/PDEs are the same for the two, the types of problems you want to solve are different. This distinction might not matter long term, but the way `Configurations.jl` is currently written requires splitting the two.
> 
> If you look at L190-L197, you can see that `HBM_Configuration` will always set up a regular, rectangular `StackedBrickTopology` which is not generally what people mean when they say GCM, as far as I know.

Then why not making this explicit, e.g., 'GCM_Box' vs 'GCM_Sphere' or something like it, with both using what is generic to the GCM configuration in a configuration file? (After all, both sphere and box will be hydrostatic Boussinesq, so HBM would become ambiguous.)

blallen:
> > > > > GCM -> AGCM
> > > > > LES -> ALES
> > > > 
> > > > 
> > > > I'll change these.
> > > > > HBM -> OGCM
> > > > 
> > > > 
> > > > I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?
> > > 
> > > 
> > > Perhaps @rafferrari can weigh in here. I think all ocean GCM configurations that are planned will be HBM. Ocean LES will be nonhydrostatic. I was primarily looking to make names consistent with what people in the community are used to.
> > 
> > 
> > The reason I didn't call it Ocean GCM was to distinguish the model that runs on the sphere and the model that runs in a box. While the underlying physics/PDEs are the same for the two, the types of problems you want to solve are different. This distinction might not matter long term, but the way `Configurations.jl` is currently written requires splitting the two.
> > If you look at L190-L197, you can see that `HBM_Configuration` will always set up a regular, rectangular `StackedBrickTopology` which is not generally what people mean when they say GCM, as far as I know.
> 
> Then why not making this explicit, e.g., 'GCM_Box' vs 'GCM_Sphere' or something like it, with both using what is generic to the GCM configuration in a configuration file? (After all, both sphere and box will be hydrostatic Boussinesq, so HBM would become ambiguous.)

Okay, I implemented something along those lines in 4e4eb7bc031b0b2dabb523b1eae3be0041fdec7d

rafferrari:
> > > > > > GCM -> AGCM
> > > > > > LES -> ALES
> > > > > 
> > > > > 
> > > > > I'll change these.
> > > > > > HBM -> OGCM
> > > > > 
> > > > > 
> > > > > I think that other Ocean models are planned, which is why the specific model name is being used. I'll let @blallen comment on this proposed change. Perhaps `HydrostaticBoussinesqModel` should be a subtype of an abstract `OceanModel` and the configuration would be `OGCM_Configuration()` and common to all ocean models?
> > > > 
> > > > 
> > > > Perhaps @rafferrari can weigh in here. I think all ocean GCM configurations that are planned will be HBM. Ocean LES will be nonhydrostatic. I was primarily looking to make names consistent with what people in the community are used to.
> > > 
> > > 
> > > The reason I didn't call it Ocean GCM was to distinguish the model that runs on the sphere and the model that runs in a box. While the underlying physics/PDEs are the same for the two, the types of problems you want to solve are different. This distinction might not matter long term, but the way `Configurations.jl` is currently written requires splitting the two.
> > > If you look at L190-L197, you can see that `HBM_Configuration` will always set up a regular, rectangular `StackedBrickTopology` which is not generally what people mean when they say GCM, as far as I know.
> > 
> > 
> > Then why not making this explicit, e.g., 'GCM_Box' vs 'GCM_Sphere' or something like it, with both using what is generic to the GCM configuration in a configuration file? (After all, both sphere and box will be hydrostatic Boussinesq, so HBM would become ambiguous.)
> 
> Okay, I implemented something along those lines in [4e4eb7b](https://github.com/climate-machine/CLIMA/commit/4e4eb7bc031b0b2dabb523b1eae3be0041fdec7d)

I think that Ocean-BoxGCM is a good name for the configuration we are using. I have no strong opinions on whether it should be Ocean-BoxGCM or Ocean-GCM-Box, but the former is consistent with the atmospheric naming convention in CliMa.

experiments/Atmos_SphereGCM/heldsuarez.jl
```diff
@@ -33,14 +33,16 @@ function init_heldsuarez!(state, aux, coords, t)
 end
 
 function config_heldsuarez(FT, N, resolution)
-    config = CLIMA.GCM_Configuration("HeldSuarez", N, resolution, domain_height, init_heldsuarez!,
-                                     ref_state=HydrostaticState(IsothermalProfile(T_initial),
+    config = CLIMA.Atmos_SphereGCM_Configuration("HeldSuarez", N,
```
tapios: Maybe we can leave it at 
```suggestion
    config = CLIMA.Atmos_GCM_Configuration("HeldSuarez", N,
```
(don't think we would want a non-spherical GCM configuration anytime soon).

rafferrari: Agreed.

kpamnany:
There are 2 open PRs that will affect this one, so we should probably defer moving the configuration functions into their own files for the moment.

kpamnany:
Now that #686 amd #694 are merged, we should rebase. Before that, @blallen: can you mention any other objectives you have for this PR? It looks pretty good to me already, apart from some needed cleanup.

src/Driver/Driver.jl
```diff
@@ -210,16 +210,21 @@ function setup_solver(t0::FT, timeend::FT,
...
     # create DG model, initialize ODE state
-    dg = DGModel(driver_config.bl, driver_config.grid, driver_config.numfluxnondiff,
-                 driver_config.numfluxdiff, driver_config.gradnumflux)
+    dg = DGModel(driver_config.bl, driver_config.grid,
+                 driver_config.numfluxnondiff, driver_config.numfluxdiff,
+                 driver_config.gradnumflux, modeldata=modeldata)
     @info @sprintf("Initializing %s", driver_config.name)
     Q = init_ode_state(dg, FT(0), init_args...; forcecpu=forcecpu)
+    update_aux!(dg, driver_config.bl, Q, FT(0))
+    update_aux_diffusive!(dg, driver_config.bl, Q, FT(0))
```
kpamnany: These seem out of place. Does `init_aux()` not suffice?

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
```diff
@@ -1,7 +1,7 @@
 module HydrostaticBoussinesq
 
 export HydrostaticBoussinesqModel, HydrostaticBoussinesqProblem, OceanDGModel,
-       LinearHBModel
+       LinearHBModel, calculate_dt
```
kpamnany: `calculate_dt` is a pretty generic name for CLIMA -- this should probably not be exported.

src/Driver/Driver.jl
blallen: `update_aux!` applies the filters to the state variables and `update_aux_diffusive!` calculates the vertical component of velocity using the integral kernels. 

Both of these features probably deserve their own interface instead of being done during the `update_aux` calls, but I think adding those interfaces is a few steps down the road

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
blallen: yeah, I'm still figuring out how modules and exporting work in Julia. I can probably remove the export and just add a `CLIMA.` in `simple_box.jl`

kpamnany:
@blallen: please add/adjust `slurmci-test.toml` entries as appropriate.

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
blallen: So I made this a method defined in `DGmethods/balancelaw.jl` that we can use multiple dispatch on with the type of model. I put some examples in `Atmos.jl` and `linear.jl` but these probably need to be updated

blallen:
bors try

slurmci-test.toml
```diff
@@ -31,7 +31,6 @@ cpu_gpu = [
   { file = "test/LinearSolvers/poisson.jl", slurmargs = ["--ntasks=2"], args = [] },
   { file = "test/LinearSolvers/columnwiselu.jl", slurmargs = ["--ntasks=1"], args = []},
   { file = "test/LinearSolvers/bandedsystem.jl", slurmargs = ["--ntasks=3", "--time=02:00:00"], args = [] },
-  { file = "test/Ocean/shallow_water/GyreDriver.jl", slurmargs = ["--ntasks=1"], args = [] },
```
kpamnany: I removed this since it isn't there anymore, but if it has just moved/been renamed, please re-add.

slurmci-test.toml
```diff
@@ -47,7 +46,8 @@ gpu = [
   { file = "test/Ocean/Hydrostatic_Boussinesq/test_divergence_free.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/Ocean/Hydrostatic_Boussinesq/test_divergence_free_IMEX.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "examples/Atmos/dry_rayleigh_benard.jl", slurmargs = ["--ntasks=3"], args = [] },
-  { file = "experiments/GCM/heldsuarez.jl", slurmargs = ["--ntasks=3"], args = [] },
-  { file = "experiments/LES/risingbubble.jl", slurmargs = ["--ntasks=3"], args = [] },
-  { file = "experiments/LES/dycoms.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "experiments/Atmos_GCM/heldsuarez.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "experiments/Atmos_LES/risingbubble.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "experiments/Atmos_LES/dycoms.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "experiments/Ocean_BoxGCM/homogeneous_box.jl", slurmargs = ["--ntasks=3"], args = [] },
```
kpamnany: How about `ocean_gyre.jl`?

src/DGmethods/balancelaw.jl
```diff
@@ -138,3 +138,7 @@ function create_diffstate(bl, grid, commtag=111)
 
   return diffstate
 end
+
+function calculate_dt(grid, model::BalanceLaw)
+    return 0.1
+end
```
kpamnany: `BalanceLaw` is an abstract type -- I think this function should be empty?

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
kpamnany: I added `ode_dt` back, as the ability to override the timestep computation can be useful. Shouldn't affect you.

kpamnany:
bors try

kpamnany:
Azure CI is apparently [not running](https://github.com/blallen/CLIMA/blob/bla/ocean_driver/test/Ocean/runtests.jl#L6-L7) any Ocean tests.

src/DGmethods/balancelaw.jl
blallen: yeah we can do that instead

slurmci-test.toml
blallen: Unfortunately, we don't have a good test for `ocean_gyre.jl`. It really needs to start from a checkpoint file to perform a useful test

slurmci-test.toml
blallen: ah I completely broke all the shallow water stuff, I'll fix it

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
blallen: sounds good to me

kpamnany: Aaannd I added `Courant_number` back too, after some discussion with the Atmos folks. Again, this shouldn't affect you, but let's chat if needed.

blallen:
> Azure CI is apparently [not running](https://github.com/blallen/CLIMA/blob/bla/ocean_driver/test/Ocean/runtests.jl#L6-L7) any Ocean tests.

The only test we have right now is the divergence-free test, which was too long to run on Azure before. since I have seen that it fails very quickly if something is wrong, I might be able to move it there instead of running it as an experiment on slurm CI. 

But in general, we do need a set of tests for Azure CI that don't require timestepping. @christophernhill @jm-c do you guys have any ideas? (this will be for a separate PR but might as well start the discussion)

slurmci-test.toml
blallen: actually didn't break it, just moved things around.

I need to convert this to use the Driver + Config framework too, but I think I'll do that in a separate PR

PR #691, opened 2020-02-08
RTB in driver-config; cleanup old drivers; fix b.c. superscripts

akshaysridhar:
1) Removes old drivers from DGmethods/compressible_Navier_Stokes/
     - Rising bubble : reintroduced in this PR 
     - Rayleigh-Benard : reintroduced in #672 
     - Deletes old dycoms drivers from examples/DGmethods/...
2) Cleans up superscripts for b.c. function args
@thomasgibson @kpamnany @simonbyrne


src/Atmos/Model/boundaryconditions.jl
```diff
@@ -134,53 +134,91 @@ struct DYCOMS_BC{FT} <: BoundaryCondition
...
+  # at the boundaries the ⁻, minus side states are the interior values
+  # state1 is 𝐘 at the first interior nodes relative to the bottom wall
+  FT = eltype(state⁺)
+  # Get values from minus-side state
+  ρM = state⁻.ρ
+  UM, VM, WM = state⁻.ρu
+  EM = state⁻.ρe
+  QTM = state⁻.moisture.ρq_tot
+  uM, vM, wM  = UM/ρM, VM/ρM, WM/ρM
+  q_totM = QTM/ρM
```
simonbyrne: looks like this can go too

akshaysridhar: Done.

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -77,27 +77,27 @@ struct NoFluxBC <: BoundaryCondition
...
-  stateP.ρu -= dot(stateM.ρu, nM) * SVector(nM)
+                               m::AtmosModel, state⁺::Vars, diff⁺::Vars,
+                               aux⁺::Vars, n⁻, state⁻::Vars, diff⁻::Vars,
+                               aux⁻::Vars, bctype, t, _...)
+  FT = eltype(state⁻)
+  state⁺.ρ = state⁻.ρ
+  state⁺.ρu -= dot(state⁻.ρu, n⁻) * SVector(n⁻)
   
-  fill!(getfield(diffP, :array), FT(0))
+  fill!(getfield(diff⁺, :array), FT(0))
```
simonbyrne: @jkozdon @fxgiraldo out of curiousity: should we be setting `diff⁺ = 0` or `diff⁺ = -diff⁻`

jkozdon: I believe that we want a zero Neumann BC here, so zero would be correct. If you were setting a Dirichlet you would set `diff⁺ = -diff⁻`

kpamnany:
I'm confused.
>        * Rising bubble : reintroduced in this PR
It looks like these have been removed, but I don't see the reintroduction -- is it still on the way?
>        * Rayleigh-Benard : reintroduced in #672
>        * Deletes old dycoms drivers from examples/DGmethods/...

Both these are already done in #672.

src/Atmos/Model/boundaryconditions.jl
simonbyrne: > If you were setting a Dirichlet you would set `diff⁺ = -diff⁻`

Why not `diff⁺ = diff⁻`?

akshaysridhar:
Merged into #672

PR #692, opened 2020-02-08
Update CUDAdrv and small cleanup

vchuravy:
Adds explicit compat bounds for MPI and CUDAdrv (we can't choose older versions than those stated)
This allows for the code in CMBuffers to be cleaned up a bit.

### DIFF
--- a/src/Arrays/CMBuffers.jl
+++ b/src/Arrays/CMBuffers.jl
@@ -43,10 +43,10 @@ used for staging data and for MPI transfers. When running on:
 - `pin`:  register the `transfer` buffer with CUDA
 """
 struct CMBuffer{T, Arr, Buff}
-    stage :: Arr     # Same type as Q.data, used for staging
-    transfer :: Buff # Union{Nothing,Buff}
+    stage::Arr     # Same type as Q.data, used for staging
+    transfer::Buff # Union{Nothing,Buff}
 
-    function CMBuffer{T}(::Type{Arr}, kind, dims...; pin=true) where {T, Arr}
+    function CMBuffer{T}(::Type{Arr}, kind, dims...; pin = true) where {T, Arr}
         if kind == SingleCMBuffer
             transfer = nothing
         elseif kind == DoubleCMBuffer
@@ -90,7 +90,7 @@ function prepare_transfer!(buf::CMBuffer)
     if buf.transfer === nothing
         # nothing to do here
     else
-        copybuffer!(buf.transfer, buf.stage, async=false)
+        copybuffer!(buf.transfer, buf.stage, async = false)
     end
 end
 
@@ -98,7 +98,9 @@ function prepare_stage!(buf::CMBuffer)
     if buf.transfer === nothing
         # nothing to do here
     else
-        copybuffer!(buf.stage, buf.transfer, async=false)
+        # This is `async=true`, since we will launch
+        # a kernel on the same stream afterwards.
+        copybuffer!(buf.stage, buf.transfer, async = true)
     end
 end
 
@@ -117,7 +119,7 @@ Copy a buffer from device to host or vice-versa. Internally this uses
 """
 function copybuffer! end
 
-function copybuffer!(A::AbstractArray, B::AbstractArray; async=true)
+function copybuffer!(A::AbstractArray, B::AbstractArray; async = true)
     copy!(A, B)
 end
 
@@ -134,7 +136,12 @@ end
         end
         GC.@preserve arr begin
             # XXX: is HOSTREGISTER_DEVICEMAP useful?
-            Mem.register(Mem.HostBuffer, pointer(arr), sizeof(arr), Mem.HOSTREGISTER_DEVICEMAP)
+            Mem.register(
+                Mem.HostBuffer,
+                pointer(arr),
+                sizeof(arr),
+                Mem.HOSTREGISTER_DEVICEMAP,
+            )
         end
     end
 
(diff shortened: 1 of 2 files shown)


src/Arrays/CMBuffers.jl
```diff
@@ -98,6 +98,8 @@ function prepare_stage!(buf::CMBuffer)
     if buf.transfer === nothing
         # nothing to do here
     else
+        # This could be `async=true`, since we will launch
+        # a kernel on the same stream afterwards.
         copybuffer!(buf.stage, buf.transfer, async=false)
```
vchuravy: @kpamnany did you run into issues with this set to true?

vchuravy:
bors try

vchuravy:
bors try

kpamnany:
This is the mysterious problem with the updated CU* packages for just the Kessler example.

src/Arrays/CMBuffers.jl
kpamnany: No. It should be set to `true`; it works fine and offers a small but measurable speedup.

vchuravy:
bors try

kpamnany:
This is pretty strange. Why does using an async copy in `prepare_stage!()` cause numerical changes? @vchuravy: any ideas? @jkozdon?

jkozdon:
Smells like a race condition, that the copy is being done out of sync with the kernel that uses the data.

jkozdon:
All the problems are in the old DGmodel code. Would be good to understand why, even though we are throwing this code away (in case its a bug that permeated to the new model)

- https://gist.github.com/climabot/b3eb24057945677a58d27ee71ade2b95#file-out_9259778-L36
- https://gist.github.com/climabot/b3eb24057945677a58d27ee71ade2b95#file-out_9259780-L36
- https://gist.github.com/climabot/b3eb24057945677a58d27ee71ade2b95#file-out_9259777-L36
- https://gist.github.com/climabot/b3eb24057945677a58d27ee71ade2b95#file-out_9259775-L36

jkozdon:
I don't know enough about the CUDA aware MPI, but could the problem actually be in the function `prepare_transfer! `

https://github.com/climate-machine/CLIMA/blob/1884070110d9c1c9643ef1f4923bb5371390ffa1/src/Arrays/CMBuffers.jl#L89-L95

which is used in `start_ghost_exchange `

https://github.com/climate-machine/CLIMA/blob/1884070110d9c1c9643ef1f4923bb5371390ffa1/src/Arrays/MPIStateArrays.jl#L275-L296

Namely can the `Isend` happen before the `fillsendbuf!` has executed? (I don't know how the synchronization works with cuda aware MPI. When we are double buffering, synchronization is happening in the `prepare_transfer!`, but is it happening in the single buffer case?)

jkozdon:
On second thought, if this is the issue I don't know why we don't see it in more tests...

vchuravy:
This is going to be superseded by #791

but yes it loos and smells like a data-race. Which is odd since everything should still happen on the `CuDefaultStream`, and we should be synchronizing through that.

jkozdon:
> This is going to be superseded by #791

True. But still concerning in case it's a larger logic issue.

Is it related to the [changes](https://github.com/climate-machine/CLIMA/pull/791/commits/09f09ae1068faea1c5b9645838883ce145c04c07#diff-d71062f5becfbb4ce9301e78f1c67f2e) that are in #791 related to the `DGmethods_old`?

kpamnany:
While it may appear like one, I don't think this is a race. The errors are in the same tests, 9 days apart. Unlikely to be caused by a particular interleaving.

Unfortunately I have no ideas on what could be behind this. Are the expected errors computed analytically? If so, can we stretch this out -- run for longer and see if the divergence increases?

jkozdon:
They are measured, but our code should be floating point stable if the same mpi distribution is used, the same version of julia, and the same package version.

If you want to check divergence you'd have to run master longer then compare with this. But since the numbers in this test have been stable for 9 months (https://github.com/climate-machine/CLIMA/commit/fc35418a235e3b212a17658deb733f6c2a0e3edd) something wacky is going on.

jkozdon:
> While it may appear like one, I don't think this is a race. The errors are in the same tests, 9 days apart. Unlikely to be caused by a particular interleaving.

The same tests are breaking, but they are generating different numbers from one another which suggests are race condition to me: [older one](https://gist.githubusercontent.com/climabot/7b7f53c31d13051eacdedf1f0f4bd010/raw/e3ab12833698a5bdca6d791f694acb2b8d7fbc5b/out_9104666) versus [newer one](https://gist.githubusercontent.com/climabot/b3eb24057945677a58d27ee71ade2b95/raw/9a7eb512ccb8a699f4e8a1dce04b0fc13fb380f6/out_9259775)

mwarusz:
These lines in `DGmethods_old` seem to be doing the same `recv` twice which probably does not play well with async copies. https://github.com/climate-machine/CLIMA/blob/5989d1f13dffd46dd21ed5bbe78c309638481db4/src/DGmethods_old/DGBalanceLawDiscretizations.jl#L615-L620
Commenting out one `recv` seems to fix things.

vchuravy:
> These lines in DGmethods_old seem to be doing the same recv twice which probably does not play well with async copies. 

Could we get into the situation where one rank is far enough ahead so that they already posted the next send of ISends? So the recvs would indeed receive to early.

vchuravy:
There are a lot of unrelated formatting changes in here, which will cause merge conflicts with #791

I would rather hold this back until that is merged (or split out the formatting changes into their own commit/PR).

jkozdon:
>Could we get into the situation where one rank is far enough ahead so that they already posted the next send of ISends? So the recvs would indeed receive to early?

Thanks for tracking this down @mwarusz, and for the interpretation @vchuravy.

kpamnany:
Yes, good find @mwarusz !

vchuravy:
Chatting with Kiran, I will incorporate these changes with #791 or a follow on PR thereof.

PR #695, opened 2020-02-11
Change default linear solver in IMEXSolverType to ManyColumnLU

mwarusz:
This PR changes the default linear solver in `IMEXSolverType` to `ManyColumnLU`. This is a better
default since `SingleColumnLU` won't work on the sphere. This also fixes a bug where the linear solver specified in `solver_type` was not used.


mwarusz:
bors try

bischtob:
Good job. This took me until now to find out as well :(.

PR #701, opened 2020-02-12
Fix subsidence

### LINKED ISSUE #700: Update subsidence specification
The way subsidence is specified needs to be modified. We need subsidence to give  a source in the specific humidity and energy equations, i.e.,
`source.e = \rho w_sub k \cdot \nabla h_tot`
where `h_tot` is total enthalpy and 
`source.q_t = \rho w_sub k \cdot \nabla q_t`
where `k` is the vertical unit vector and `w_sub` is a prescribed subsidence velocity (defined positive downward).

This needs to replace the current flux modification.

charleskawczynski:
- Fixes expression for subsidence, as specified in the design docs

Closes #700


src/Atmos/Model/AtmosModel.jl
```diff
@@ -252,8 +248,14 @@ end
   ν, τ = turbulence_tensors(atmos.turbulence, state, diffusive, aux, t)
   D_t = (ν isa Real ? ν : diag(ν)) * inv_Pr_turb
   d_h_tot = -D_t .* diffusive.∇h_tot
+
+  z  = altitude(atmos.orientation, aux)
+  w_sub = subsidence_velocity(atmos.subsidence, z)
+  k̂ = vertical_unit_vector(atmos.orientation, aux)
+  flux.ρe += state.ρ*w_sub .* k̂ .* diffusive.∇h_tot
```
tapios: This needs to go not into the flux, but into the source (currently in `source.jl`): 
```suggestion
  source.ρe += state.ρ*w_sub .* k̂ .* diffusive.∇h_tot
```

src/Atmos/Model/moisture.jl
```diff
@@ -107,8 +107,14 @@ function flux_moisture!(moist::EquilMoist, atmos::AtmosModel, flux::Grad, state:
...
 end
 
-function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, diffusive::Vars, aux::Vars, t::Real, D_t)
+function flux_diffusive!(moist::EquilMoist, atmos::AtmosModel, flux::Grad, state::Vars, diffusive::Vars, aux::Vars, t::Real, D_t)
   d_q_tot = (-D_t) .* diffusive.moisture.∇q_tot
+
+  z  = altitude(atmos.orientation, aux)
+  w_sub = subsidence_velocity(atmos.subsidence, z)
+  k̂ = vertical_unit_vector(atmos.orientation, aux)
+  flux.moisture.ρq_tot += state.ρ*w_sub .* k̂ .* diffusive.∇q_tot
```
tapios: Same as for energy, it needs to go into source:
```suggestion
  source.moisture.ρq_tot += state.ρ*w_sub .* k̂ .* diffusive.∇q_tot
```

tapios:
Thanks so much, @charleskawczynski. So fast!

src/Atmos/Model/source.jl
```diff
@@ -33,6 +33,24 @@ function atmos_source!(::Coriolis, atmos::AtmosModel, source::Vars, state::Vars,
...
+  ρ = state.ρ
+  z = altitude(atmos.orientation, aux)
+  w_sub = subsidence_velocity(subsidence, z)
+  k̂ = vertical_unit_vector(atmos.orientation, aux)
+
+  source.ρe += ρ*w_sub .* k̂ .* diffusive.∇h_tot
+  source.moisture.ρq_tot += ρ*w_sub .* k̂ .* diffusive.moisture.∇q_tot
+end
+
+subsidence_velocity(subsidence::Subsidence{FT}, z::FT) where {FT} = subsidence.D*z
```
smarras79: I think this should have a minus sign @tapios @charleskawczynski  ? can you confirm

src/Atmos/Model/moisture.jl
```diff
@@ -100,11 +100,7 @@ end
 function flux_moisture!(moist::EquilMoist, atmos::AtmosModel, flux::Grad, state::Vars, aux::Vars, t::Real)
   ρ = state.ρ
   u = state.ρu / ρ
-  z = altitude(atmos.orientation, aux)
-  usub = subsidence_velocity(atmos.subsidence, z)
-  ẑ = vertical_unit_vector(atmos.orientation, aux)
-  u_tot = u .- usub * ẑ
-  flux.moisture.ρq_tot += u_tot * state.moisture.ρq_tot
+  flux.moisture.ρq_tot += u * state.moisture.ρq_tot
```
charleskawczynski: This moisture flux should still exist, correct? @tapios

tapios: Yes, we still need the flux.

src/Atmos/Model/source.jl
tapios: I think we have `subsidence_velocity` defined as positive downward, in which case the sign is correct. This definition makes sense because subsidence implies it is down. However, it may be confusing. I'd be fine with changing the sign of the subsidence velocity so it is positive upward, and then changing the sign of the source term correspondingly.

charleskawczynski: Based on the design docs, you're correct. But I also added this term (which should be negative), so the source turns out to be correct. I'll fix this.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
@akshaysridhar it might be nice to see how Dycoms changes as a result of this PR. Aside from that, I think this is ready to go

src/Atmos/Model/AtmosModel.jl
```diff
@@ -312,16 +302,16 @@ function init_aux!(m::AtmosModel, aux::Vars, geom::LocalGeometry)
 end
 
 """
-    source!(m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
+    source!(m::AtmosModel, source::Vars, state::Vars, aux::Vars, diffusive::Vars, t::Real)
```
simonbyrne: Can we keep the argument order consistent with `flux_diffusive!` (i.e. `diffusive` before `aux`)?

charleskawczynski: Yes, I'll swap the order

PR #710, opened 2020-02-13
Add balance law to init funcs

charleskawczynski:
Adds balance law to initial condition functions, this will allow us to use `orientation` methods, and not re-implement them in all of the drivers.

### DIFF
--- a/experiments/Atmos_GCM/heldsuarez.jl
+++ b/experiments/Atmos_GCM/heldsuarez.jl
@@ -16,13 +16,15 @@ const p_ground = MSLP
 const T_initial = 255
 const domain_height = 30e3
 
-function init_heldsuarez!(state, aux, coords, t)
+function init_heldsuarez!(bl, state, aux, coords, t)
     global p_ground, T_initial
 
     FT = eltype(state)
 
+    # TODO: change to altitude function
     r = norm(coords, 2)
     h = r - FT(planet_radius)
+    # h = altitude(bl.orientation, aux)
 
     scale_height = R_d * FT(T_initial) / grav
     p            = FT(p_ground) * exp(-h / scale_height)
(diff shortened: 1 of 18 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

test/runtests.jl
```diff
@@ -3,7 +3,8 @@ using Test, Pkg
 ENV["DATADEPS_ALWAYS_ACCEPT"] = true
 ENV["JULIA_LOG_LEVEL"] = "WARN"
 
-for submodule in ["Utilities/TicToc",
+for submodule in [
+                  "Utilities/TicToc",
```
kpamnany: This change looks a bit weird to me, but it's a stylistic choice...

charleskawczynski: Ah, yes. It does look a bit funny, but I find this convenient when I want to locally test specific modules (and comment the rest). I can revert if you'd like

kpamnany: Don't mind much either way, and your explanation makes sense.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -16,13 +16,15 @@ const p_ground = MSLP
...
-function init_heldsuarez!(state, aux, coords, t)
+function init_heldsuarez!(bl, state, aux, coords, t)
     global p_ground, T_initial
 
     FT = eltype(state)
 
+    # TODO: change to altitude function
     r = norm(coords, 2)
     h = r - FT(planet_radius)
+    # h = altitude(bl.orientation, aux)
```
kpamnany: Need to add `using CLIMA.Atmos: altitude` for this to work. Or export `altitude` in `orientation.jl`.

charleskawczynski: I was thinking that I'd do that in a separate PR

PR #711, opened 2020-02-13
Allow driver specific saturation adjustment tolerance in EquilMoist

smarras79:
Based on the latest finding about Dycoms stability, this PR modifies moisture.jl to allow for a user defined tolerance value from the driver:

`Base.@kwdef struct EquilMoist <: MoistureModel
  maxiter::Int=3
end`
to 
`Base.@kwdef struct EquilMoist <: MoistureModel
  maxiter::Int=3
  tolerance::Float64=1e-1
end`


experiments/LES/dycoms.jl
```diff
@@ -337,7 +337,7 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
     model = AtmosModel{FT}(AtmosLESConfiguration;
                            ref_state=ref_state,
                           turbulence=SmagorinskyLilly{FT}(C_smag),
-                            moisture=EquilMoist(5),
+                            moisture=EquilMoist(5, 1e-1),
```
charleskawczynski:
```suggestion
                            moisture=EquilMoist(5, FT(1e-1)),
```

src/Atmos/Model/moisture.jl
```diff
@@ -64,17 +64,20 @@ thermo_state(moist::DryModel, orientation::Orientation, state::Vars, aux::Vars)
 Assumes the moisture components are computed via thermodynamic equilibrium.
 """
 Base.@kwdef struct EquilMoist <: MoistureModel
-  maxiter::Int = 3
+  maxiter::Int=3
+  tolerance::Float64=1e-1
```
charleskawczynski:
```suggestion
  tolerance::FT=FT(1e-1)
```

bischtob:
Does this mean we have a stable DYCOMS :)?

charleskawczynski:
The default value for the saturation adjustment tolerance is [already 1e-1](https://github.com/climate-machine/CLIMA/blob/master/src/Common/MoistThermodynamics/states.jl#L82). Is the intent to increase or decrease this tolerance for dycoms?

smarras79:
> Does this mean we have a stable DYCOMS :)?

Yes if we use a tolerance that is larger than 0.1

Also, I am currently adding @thomasgibson implementation of the multi-rate from his density current to dycoms. It requires some changes to Driver.jl though.

experiments/LES/dycoms.jl
simonbyrne:
```suggestion
                            moisture=EquilMoist(maxiter=5),
```

src/Atmos/Model/moisture.jl
```diff
@@ -64,17 +64,20 @@ thermo_state(moist::DryModel, orientation::Orientation, state::Vars, aux::Vars)
 Assumes the moisture components are computed via thermodynamic equilibrium.
 """
 Base.@kwdef struct EquilMoist <: MoistureModel
```
simonbyrne:
```suggestion
Base.@kwdef struct EquilMoist{FT} <: MoistureModel
```

src/Atmos/Model/moisture.jl
simonbyrne:
```suggestion
  tolerance::FT=1e-1
```

smarras79:
I need this change to be merged to be able to run Dycoms to completion. The one in Master now depends on this change or it blows up at 5000s.
How should we proceed?
thanks!
@simonbyrne

smarras79:
Thanks @charleskawczynski !

PR #717, opened 2020-02-15
Remove weighted infinity norm

mwarusz:
The CPU version was buggy. The GPU version didn't work. It didn't make much sense to begin with, since l_inf is not an integral.


mwarusz:
bors try

PR #722, opened 2020-02-18
Test horizontal/vertical splitting and remove skew symmetric

mwarusz:
This adds a test for horizontal/vertical splitting in the box and on the sphere using advection-diffusion
equation. It also removes split treatment of metric terms in `volumerhs`, which caused some small inconsistencies in the splitting that could become important as we move towards actually solving equations in the horizontal/vertical (as opposed to only using it for 1D IMEX where this didn't seem to matter).

### DIFF
--- a/slurmci-test.toml
+++ b/slurmci-test.toml
@@ -16,6 +16,7 @@ cpu_gpu = [
   { file = "test/DGmethods/compressible_Navier_Stokes/mms_bc_atmos.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/DGmethods/compressible_Navier_Stokes/mms_bc_dgmodel.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/DGmethods/compressible_Navier_Stokes/density_current-model.jl", slurmargs = ["--ntasks=3"], args = [] },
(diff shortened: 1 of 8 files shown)


mwarusz:
bors try

simonbyrne:
Which matrix was skew-symmetric?

jkozdon:
> Which matrix was skew-symmetric?

We were using a skew-symmetric formulation of DG for the treatment of the metric terms (essentially, the appear half of the time inside the derivative and half of the time outside).

In the other situations (linear problems) there are stability reasons for doing this, but for non-linear problems the story is less clear. Also, our current formulation doesn't get the benefits even for the linear part.

@mwarusz discovered that when we have the metrics outside, the horizontal is not truly horizontal so it made sense to remove until we had a stronger reason for doing this. This wasn't a problem for 1D-IMEX per se, where you just have to split enough of the horizontal and vertical, but was a problem if you wanted to solve a PDE in the horizontal in the 3D model.

Even now (I believe) our horizontal, is only horizontal on the sphere without topography.

simonbyrne:
failure is in integration test for pseudo1D_heat_eqn.jl 

Would be good to get rid of these hard-coded norms in the tests.

jkozdon:
> failure is in integration test for pseudo1D_heat_eqn.jl
> 
> Would be good to get rid of these hard-coded norms in the tests.

One option would be to check that things are the right order of magnitude and the the rate of convergence is correct.

PR #726, opened 2020-02-19
Add PlanetConstants

charleskawczynski:
- Adds `PlanetConstants`

This allows for planet constants to be initialized as:

```julia
earth = PlanetConstants{FT}(EarthConstants)
big_earth = PlanetConstants{FT}(EarthConstants; grav=100.0, planet_radius=FT(6.371e8))
```

This also separates what I've now called `UniversalConstants`, which do not change from planet to planet.


src/Common/PlanetParameters/EarthConstants.jl
```diff
@@ -0,0 +1,83 @@
+function PlanetConstants{FT}(::Type{EarthParameters};
+                             k_Karman =          FT(0.4),
```
tapios: Can we please retain the long-form definition and information on units we had in PlanetParameters, at least as a comment? This is important for readability.

tapios: I mean: have the long-form definitions stated next to the numerical values, rather than in a separate file?

src/Common/PlanetParameters/PlanetaryConstants.jl
```diff
@@ -0,0 +1,105 @@
...
+  "Gas constant dry air (J/kg/K)"
+  R_d::FT
+  "Adiabatic exponent dry air"
+  kappa_d::FT
+  "Isobaric specific heat dry air"
+  cp_d::FT
+  "Isochoric specific heat dry air"
+  cv_d::FT
+  "Density of liquid water (kg/m^3)"
+  ρ_cloud_liq::FT
```
tapios:
```suggestion
  ρ_liq::FT
```
(and then introduce ` ρ_cloud_liq` and set `ρ_cloud_liq =  ρ_liq` (point is that we use the same density of liquid water in soil, oceans, clouds etc., although we may wish to retain the option of having a different `ρ_cloud_liq`)

src/Common/PlanetParameters/PlanetaryConstants.jl
```diff
@@ -0,0 +1,105 @@
...
+  "Density of ice water (kg/m^3)"
```
tapios:
```suggestion
  "Density of water ice (kg/m^3)"
```

src/Common/PlanetParameters/PlanetaryConstants.jl
```diff
@@ -0,0 +1,105 @@
...
+  ρ_cloud_ice::FT
```
tapios:
```suggestion
  ρ_ice::FT
```
(and as for liquid, then introduce `ρ_cloud_ice = ρ_ice`)

src/Common/PlanetParameters/PlanetaryConstants.jl
```diff
@@ -0,0 +1,105 @@
...
+  "Molecular weight (kg/mol)"
```
tapios:
```suggestion
  "Molecular weight water (kg/mol)"
```

src/Common/PlanetParameters/EarthConstants.jl
tapios: But it's also ok to leave it if it is difficult to incorporate the constant names with the numerical values.

charleskawczynski: The documentation for the parameters are in the `PlanearytConstants.jl`. I think it'd be easier to keep the docs there and focus on values in the planet files, but I can add them to both if you'd like.

tapios: Right, I saw that. What I mean is that it would be nice to keep documentation near where the numerical values are defined (e.g., to see units next to values).

src/Common/PlanetParameters/PlanetaryConstants.jl
```diff
@@ -0,0 +1,83 @@
...
+
+using ..UniversalConstants
+
+export PlanetConstants,
+       AbstractPlanetConstants,
+       EarthConstants
+       # MarsConstants
+
+Base.@kwdef struct PlanetConstants{FT<:AbstractFloat}
+  k_Karman::FT        # Von Karman constant (1)
```
simonbyrne: Can add docstings
```suggestion
  "Von Karman constant (1)"
  k_Karman::FT
```

simonbyrne: Why not just add the default values here, i.e.
```
```suggestion
  k_Karman::FT= 0.4
```

charleskawczynski: I'm not sure this is a good idea because it will be difficult to catch if you accidentally miss a keyword for one of the other constructors (that's why I don't prescribe any for the default).

charleskawczynski: This way, if a constructor for a particular planet is incomplete, we'll know right away.

src/Common/PlanetParameters/EarthConstants.jl
charleskawczynski: I'll add them back in, I think it'll be fine either way.

src/Common/PlanetParameters/EarthConstants.jl
```diff
@@ -0,0 +1,49 @@
+function PlanetConstants{FT}(::Type{EarthConstants}; kwargs...) where {FT}
+  settings = (; dummy=1) # cannot apply merge to (; nothing)
+
+  @append_nt settings k_Karman =          FT(0.4)                             # Von Karman constant (1)
```
simonbyrne: You can do this without a macro, 
```julia
function PlanetConstants{FT}(::Type{EarthConstants}; kwargs...) where {FT}
   defaults = (k_Karman = FT(0.4), ...)
   PlanetConstants(pairs(merge(defaults, kwargs))...)
end
```

simonbyrne:
This looks good. Can we integrate this with `AtmosModel` to replace PlanetParameters?

tapios:
Any reason not to merge this soon?

charleskawczynski:
> Any reason not to merge this soon?

It's not currently being used yet, but I can do that in a separate PR.

PR #727, opened 2020-02-20
New approach to doing integrals

jkozdon:
Allows the user to say where integrals are stored and what integrals are reversed

### DIFF
--- a/experiments/Atmos_LES/dycoms.jl
+++ b/experiments/Atmos_LES/dycoms.jl
@@ -15,8 +15,13 @@ using CLIMA.MoistThermodynamics
 using CLIMA.PlanetParameters
 using CLIMA.VariableTemplates
 
-import CLIMA.DGmethods: vars_state, vars_aux, vars_integrals,
-                        integrate_aux!
+import CLIMA.DGmethods: vars_state, vars_aux,
+                        vars_integrals, vars_reverse_integrals,
+                        indefinite_stack_integral!,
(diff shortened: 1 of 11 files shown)


jkozdon:
@blallen This might address some of your concerns. If you like this approach, we'd need to fix all the other codes to use this format (right now I just did the integral test).

jkozdon:
Also need to test on the GPU

blallen:
@jkozdon I like this approach. I will test it out in the HydrostaticBoussinesqModel tomorrow.

jkozdon:
@blallen Just forced pushed some commits (GPU was broken since you cannot create views into shared memory). Also updated file `integral_test_sphere`

blallen:
@jkozdon is the atmos model integral call the only one left that we need to update? I think there are just the two integral tests, the ocean model, and atmos model that use integrals right now and everything else has the `_old` suffix

jkozdon:
@jkozdon<https://github.com/jkozdon> is the atmos model integral call the only one left that we need to update? I think there are just the two integral tests, the ocean model, and atmos model that use integrals right now and everything else has the _old suffix

I believe that this is correct.

blallen:
bors try

blallen:
@jkozdon , time to merge this? is there anyone else we should ask to look at it?

jkozdon:
bor r+

jkozdon:
@simonbyrne Sorry, rebased to cleanup the commits

PR #737, opened 2020-02-24
[Atmos] Utilize existing functions in atmos init

charleskawczynski:
Now that the `init_state` methods are aware of the balance law, we should utilize the following methods in the `init_state` methods:

 - `altitude`
 - `gravitational_potential`
 - `internal_energy(::ThermodynamicState)`
 - `total_energy(..., ::ThermodynamicState)`

Since several `DryPhase` models use `p` and `T` for initialization, I've also added a constructor to moist thermo that accepts these args:
 - Adds a `PhaseDry_given_pT` constructor in moist thermo.
I’ve also added 
 - A `latitude` function for `SphericalOrientation`
 - A `longitude` function for `SphericalOrientation`

### DIFF
--- a/experiments/Atmos_GCM/heldsuarez.jl
+++ b/experiments/Atmos_GCM/heldsuarez.jl
@@ -24,17 +24,19 @@ function init_heldsuarez!(bl, state, aux, coords, t)
 
     FT = eltype(state)
 
-    # TODO: change to altitude function
-    r = norm(coords, 2)
-    h = r - FT(planet_radius)
-    # h = altitude(bl.orientation, aux)
+    z = altitude(bl.orientation, aux)
+    e_pot = gravitational_potential(bl.orientation, aux)
 
-    scale_height = R_d * dc.T_initial / grav
-    p            = dc.p_ground * exp(-h / scale_height)
+    scale_height = FT(R_d) * dc.T_initial / FT(grav)
+    p            = dc.p_ground * exp(-z / scale_height)
 
-    state.ρ  = air_density(dc.T_initial, p)
+    ts = PhaseDry_given_pT(p, dc.T_initial)
+    e_int = internal_energy(ts)
+    ρ = air_density(ts)
+
+    state.ρ  = ρ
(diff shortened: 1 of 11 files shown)


experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -21,17 +21,19 @@ function init_heldsuarez!(bl, state, aux, coords, t)
...
 
-    state.ρ  = air_density(FT(T_initial), p)
+    ts = PhaseDry_given_pT(p, FT(T_initial))
+    e_int = internal_energy(ts)
+    ρ = air_density(ts)
+
+    state.ρ  = ρ
     state.ρu = SVector{3, FT}(0, 0, 0)
-    state.ρe = state.ρ * (internal_energy(FT(T_initial)) + aux.orientation.Φ)
+    state.ρe = state.ρ * (e_int + e_pot)
```
tapios: Need to add noise (e.g., in energy or density) to break symmetry of initial condition. See @bischtob’s HS branch.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -21,17 +21,19 @@ function init_heldsuarez!(bl, state, aux, coords, t)
 
     FT = eltype(state)
 
-    # TODO: change to altitude function
-    r = norm(coords, 2)
-    h = r - FT(planet_radius)
-    # h = altitude(bl.orientation, aux)
+    h = altitude(bl.orientation, aux)
```
tapios: I’d prefer to use `z` for altitude (it’s what we usually do; we use `h` for enthalpy of sometimes a fixed height)

tapios:
Looks good to me. There are important updates to the HS setup in #746. @bischtob will need to update #746 once this is merged.

experiments/Atmos_GCM/heldsuarez.jl
charleskawczynski: I'll leave this for #746, since this PR currently should not change behavior.

Issue #491, opened 2019-10-15
DG tests that need to be moved to the new framework


mwarusz:
I discussed with @lcw and @jkozdon which DG methods tests/examples need to be moved to the new framework so that we can finally kill the old one. I marked those that have been moved but not yet merged. 

- [ ] test/DGmethods_old/conservation/sphere.jl
- [x] test/LinearSolvers/poisson.jl
- [x] test/DGmethods_old/util/filter_test.jl
- [x] test/DGmethods_old/util/nonnegative_test.jl
- [x] test/DGmethods_old/sphere/advection_sphere_*
- [ ] examples/DGmethods_old/ex_004_nonnegative.jl

The last two can be combined into one test and I will try to do that this week.

Additionally, the following could be moved but there is no functionality to take local gradient of auxiliary state in the new framework.
- ~~test/DGmethods_old/util/grad_test.jl~~ (not grad aux in new framework)
- ~~test/DGmethods_old/util/grad_test_sphere.jl~~  (not grad aux in new framework)

@simonbyrne

trontrytel:
I'm working on 

- [ ] examples/Microphysics/ex_1_saturation_adjustment.jl
- [ ] examples/Microphysics/ex_2_Kessler.jl

jkozdon:
Should this be part of [milestone 1](https://github.com/climate-machine/CLIMA/milestone/1)?

charleskawczynski:
We're so close that, it seems, it would be a shame if we couldn't include it. Does @mwarusz have any bandwidth to tackle the last example? (that is the last one, right?)

charleskawczynski:
Also, if we do close this issue out, we'd be able to remove `CLIMA/src/DGmethods_old`, correct?

jkozdon:
> We're so close that, it seems, it would be a shame if we couldn't include it. Does @mwarusz have any bandwidth to tackle the last example? (that is the last one, right?)

I can do try to do it in the next week or two if @mwarusz doesn't get to it.

> Also, if we do close this issue out, we'd be able to remove `CLIMA/src/DGmethods_old`, correct?

Yes we can remove `DGmethods_old`.

mwarusz:
"Also, if we do close this issue out, we'd be able to remove CLIMA/src/DGmethods_old, correct?"

Yes.

[...]

PR #738, opened 2020-02-24
removing old kernels

leios:
Can we just remove these?

@vchuravy Forced me to do this and takes all responsibility


leios:
bors try

lcw:
I think we were waiting for https://github.com/climate-machine/CLIMA/issues/491 before removing these.

leios:
@lcw Thanks for that.

I didn't know this was already somewhat documented. It seems like there are also a few microphysics examples that rely on these old kernels.

Would anyone mind if I attacked that issue in this PR? It's low-priority, so maybe next week.

lcw:
@leios it would be great if you want to attack #491.  Let me know if you run into issues, I would be happy to help.

bischtob:
The comments/documentation in DGMethods_old is outstanding. Can we make sure the wisdom carries over to the updated DGMethods?

jkozdon:
> The comments/documentation in DGMethods_old is outstanding. Can we make sure the wisdom carries over to the updated DGMethods?

Thanks. We did spend a bit of time on this initially, but some of that got lost in the transition to the current framework.

We initially modeled our docs on

- https://github.com/JuliaFEM/JuliaFEM.jl
- https://dealii.org/current/doxygen/deal.II/Tutorial.html

But this never really got fully flushed out and never transitioned to the newest framework.

charleskawczynski:
@leios We need to also remove these files from being included in `src/CLIMA.jl`

Issue #740, opened 2020-02-25
Compat settings


simonbyrne:
- Keep Manifest.toml (for now)
- Move version constraints to Project.toml
- Enable CompatHelper to check for updates

blallen:
Do we need to do anything else for this now that we have CompatHelper?

PR #743, opened 2020-02-25
Add versions to compat

charleskawczynski:
Needed for #742

A step towards #740


charleskawczynski:
@vchuravy do you have suggestions on what to do for `docs/Project.toml`? Should we add the same `[compat]` versions?

Project.toml
```diff
@@ -40,7 +40,28 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [compat]
+Combinatorics = "1.0.0"
```
simonbyrne: We probably should use tilde operators:
https://julialang.github.io/Pkg.jl/v1/compatibility/index.html#Tilde-specifiers-1

simonbyrne: @vchuravy is that correct?

charleskawczynski: According to [some General docs](https://github.com/JuliaRegistries/General/blob/master/README.md#automatic-merging-of-pull-requests), it seems that we want to include upper bounds, but using the tilde seems to not provide upper bounds. It's not clear to me which to follow.

vchuravy: Yes.If the patch version is significant otherwise I would use just `1.0`

charleskawczynski:
Is this even needed now that Compat has all of the PRs?

simonbyrne:
Fixed by #789.

PR #745, opened 2020-02-25
Replace 'M' and 'P' with '⁻' and '+'

blallen:
Marking things that come from the interior/minus side and exterior/plus side with minus/plus signs instead of M/P. 

This should make the variable names easier to read and make their intent clearer. 

This PR only affects the DG Kernels.


blallen:
bors try

blallen:
bors try

PR #746, opened 2020-02-25
Re-engineer Held-Suarez setup

bischtob:
We need a few fixes to the Held-Suarez setup as it currently stands. This PR will address the following problems:

- [x] Use random density perturbations in initial condition
- [x] Use bottom drag only tangential to the surface 
- [x] Use sponge layer to absorb waves at the top of the domain
- [x] Replace `IsothermalProfile` with `LinearTemperatureProfile`

### DIFF
--- a/experiments/Atmos_GCM/heldsuarez.jl
+++ b/experiments/Atmos_GCM/heldsuarez.jl
@@ -1,11 +1,14 @@
+using Distributions: Uniform
 using LinearAlgebra
 using StaticArrays
+using Random: rand
 using Test
 
 using CLIMA
 using CLIMA.Atmos
 using CLIMA.GenericCallbacks
 using CLIMA.ODESolvers
+using CLIMA.ColumnwiseLUSolver: ManyColumnLU
 using CLIMA.Mesh.Filters
 using CLIMA.Mesh.Grids
 using CLIMA.MoistThermodynamics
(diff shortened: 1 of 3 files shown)


experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,61 +1,84 @@
...
-    # h = altitude(bl.orientation, aux)
 
-    scale_height = R_d * FT(T_initial) / grav
-    p            = FT(p_ground) * exp(-h / scale_height)
-
-    state.ρ  = air_density(FT(T_initial), p)
+function init_heldsuarez!(bl, state, aux, coords, t)
+    FT            = eltype(state)
+    pressure_sfc  = FT(MSLP)
+    temp_init     = FT(300.0)
```
akshaysridhar: `const T_init` vs `temp_init` within this function - is there a way to keep just one variable with this information ? (We can now pass additional arguments to the solver setup that are made available to the init_heldsuarez! function for example https://github.com/climate-machine/CLIMA/blob/7956017fb59bd1e566716c5d7f8d2816df57d77b/src/Driver/Driver.jl#L212)

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -97,43 +120,65 @@ function held_suarez_forcing!(bl, source, state, diffusive, aux, t::Real)
...
+    timeend       = FT(days*24*60*60) # end time (seconds)
+    
+    # Set up driver configuration
+    driver_config = config_heldsuarez(FT, poly_order, (n_horz, n_vert))
+    
+    # Set up ODE solver configuration
+    #ode_solver_type = CLIMA.ExplicitSolverType(
+    #  solver_method=LSRK144NiegemannDiehlBusch
+    #)
+    ode_solver_type = CLIMA.IMEXSolverType(
```
akshaysridhar: Is this equivalent to `CLIMA.DefaultSolverType()` ?

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -97,43 +120,65 @@ function held_suarez_forcing!(bl, source, state, diffusive, aux, t::Real)
...
+    )
+
+    # Set up experiment
+    solver_config = CLIMA.setup_solver(
+      timestart, 
+      timeend, 
+      driver_config,
+      ode_solver_type=ode_solver_type,
+      Courant_number=0.4,
+      forcecpu=true
```
akshaysridhar: @bischtob You pointed out that `forcecpu` is perhaps a bit ambiguous (i.e. the action here is to force initialisation, not the run itself, on the CPU)- we can address the keyword change in a different PR - does that seem reasonable ?

simonbyrne: cc @kpamnany

experiments/Atmos_GCM/heldsuarez.jl
bischtob: Perhaps :)

experiments/Atmos_GCM/heldsuarez.jl
bischtob: @akshaysridhar : works for me

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,61 +1,89 @@
...
-const p_ground = MSLP
-const T_initial = 255
-const domain_height = 30e3
 
-function init_heldsuarez!(bl, state, aux, coords, t)
-    global p_ground, T_initial
-
-    FT = eltype(state)
+const pressure_ground = MSLP
+const T_init = 300.0 # unit: Kelvin
```
tapios: It doesn't matter much what you choose as the initial temperature, but `T_init = 255` will probably equilibrate a bit faster (because the stratosphere, where relaxation times are longer, will end up with a temperature around 200 K).

experiments/Atmos_GCM/heldsuarez.jl
tapios: Agreed. It's odd to have the same temperature be defined twice.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,61 +1,89 @@
...
+function config_heldsuarez(FT, poly_order, resolution)
+    name          = "HeldSuarez"
+    domain_height = FT(30e3)
+    T_ref         = FT(300.0)
+    Rh_ref        = FT(0.0)
+    turb_visc     = FT(0.0)
+    
+    # Rayleigh sponge 
+    zsponge = FT(15e3)
+    τ_relax = FT(0.5 / 60 / 60 / 24)
```
tapios:
```suggestion
    τ_relax = FT(0.5 / day)
```
(We used to have `day` (seconds per day) defined in `PlanetParameters`. I'm not sure we still have it, but we want something like this here.)

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,61 +1,89 @@
...
+    u_relax = SVector(FT(0), FT(0), FT(0))
+    rayleigh_sponge = RayleighSponge{FT}(domain_height, zsponge, τ_relax, u_relax, 2)
+
+    # Configure the model setup
+    model = AtmosModel{FT}(
+      AtmosGCMConfiguration;
+      
+      ref_state  = HydrostaticState(
+                     IsothermalProfile(T_ref),
```
tapios: This will not be a good reference state with `T_ref = 300` (temperature deviations in the upper atmosphere will be 50%, leading to >20% error for sound speed in linear implicit timestepping). Better: Use `LinearTemperatureProfile` with `T_surface = 300`. lapse rate `0.7 g/c_p`, and `T_min = 200`. This will be close to equilibrated state.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,61 +1,89 @@
...
+
+    # Configure the model setup
+    model = AtmosModel{FT}(
+      AtmosGCMConfiguration;
+      
+      ref_state  = HydrostaticState(
+                     IsothermalProfile(T_ref), 
+                     Rh_ref
+                   ),
+      turbulence = ConstantViscosityWithDivergence(turb_visc),
```
tapios: Constant (isotropic) viscosity can lead to stability problems: With a viscosity large enough to damp enstrophy in the horizontal, you may run into CFL bounds in the vertical if you use IMEX. 

We need to add the Smagorinsky-Lilly SGS model that has anisotropic mixing (see design doc), i.e., set the vertical viscosity/diffusivity in Smagorinsky-Lilly to zero (or something small). It should be an easy thing to add and may be good to add here. We'll need it if hyperdiffusion is not available.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -83,7 +111,7 @@ function held_suarez_forcing!(bl, source, state, diffusive, aux, t::Real)
     @inbounds λ  = atan(coord[2], coord[1])
```
tapios: Update to longitude/latitude once #737  is merged.

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -97,43 +125,65 @@ function held_suarez_forcing!(bl, source, state, diffusive, aux, t::Real)
...
-    solver_config = CLIMA.setup_solver(t0, timeend, driver_config,
-                                       ode_solver_type=ode_solver_type)
-
-    # Set up the filter callback
+    # Driver configuration parameters
+    FT            = Float32           # floating type precision
+    poly_order    = 5                 # discontinuous Galerkin polynomial order
+    n_horz        = 5                 # horizontal element number  
+    n_vert        = 5                 # vertical element number
+    days          = 30                 # experiment day number
```
tapios:
```suggestion
    run_days          = 30                 # experiment day number
```

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -97,43 +125,65 @@ function held_suarez_forcing!(bl, source, state, diffusive, aux, t::Real)
...
+    timestart     = FT(0)             # start time (seconds)
+    timeend       = FT(days*24*60*60) # end time (seconds)
```
tapios:
```suggestion
    timeend       = FT(run_days*day) # end time (seconds)
```
(or something like it: point is to use definition of length of day from PlanetParameters; we do not want to re-type it and have novices have to figure out what this is)

src/Atmos/Model/orientation.jl
```diff
@@ -18,6 +18,13 @@ gravitational_potential(::Orientation, aux::Vars) = aux.orientation.Φ
 ∇gravitational_potential(::Orientation, aux::Vars) = aux.orientation.∇Φ
 altitude(orientation::Orientation, aux::Vars) = gravitational_potential(orientation, aux) / grav
 vertical_unit_vector(orientation::Orientation, aux::Vars) = ∇gravitational_potential(orientation, aux) / grav
+function projection_normal(orientation::Orientation, aux::Vars, u⃗::AbstractVector) 
+  n̂ = vertical_unit_vector(orientation, aux)
+  return n̂ * (n̂' * u⃗)
+end
+function projection_tangential(orientation::Orientation, aux::Vars, u⃗::AbstractVector)
```
tapios: These functions are good to have. But I wonder whether we should make them more general. We'll need these (or functions like these) at boundaries in a number of contexts. Then the projections should be normal and tangential to the boundary, i.e., with `n̂` being the inward pointing normal vector on the boundary. (This notation for `n̂` is also consistent with the design docs. There and elsewhere, we used `k` for the vertical unit vector.)

experiments/Atmos_GCM/heldsuarez.jl
bischtob: Thanks. Was planning on swapping this out once I can confirm that we don't have the funky oscillations anymore.

src/Atmos/Model/orientation.jl
LenkaNovak: Sriharsha is working on the projection functions today.

bischtob:
Simulation breaks after several hundred days. Oscillations in the sponge layer seems to break it. No idea why yet. Still investigating :(.

bischtob:
@tapios & @akshaysridhar, is there a chance that having such a low resolution set up leads to problems with resolving the sphere properly, leading to stability issues?

tapios:
Unlikely. You’ll just have problems with angular momentum conservation in our case, but that shouldn’t affect stability.

bischtob:
bors try

bischtob:
bors try

src/Driver/Driver.jl
```diff
@@ -100,14 +101,14 @@ function parse_commandline()
         "--diagnostics-interval"
             help = "interval in simulation steps for gathering diagnostics"
             arg_type = Int
-            default = 10000
+            default = 1
```
kpamnany: I don't think it's a good idea to make this the default. :)

src/Driver/Driver.jl
```diff
@@ -100,14 +101,14 @@ function parse_commandline()
...
         "--enable-vtk"
             help = "output VTK to <output-dir> every <vtk-interval> simulation steps"
             action = :store_true
         "--vtk-interval"
             help = "interval in simulation steps for VTK output"
             arg_type = Int
-            default = 10000
+            default = 4
```
kpamnany: Or this.

Since you have to explicitly `--enable-vtk` to turn this on, it isn't blowing things up in the tests now, but such a short interval is going to dump far too much data for the ordinary case (especially for testing).

kpamnany:
There are no diagnostics for the Ocean model implemented. I'll fix this with something empty in the upcoming diagnostics PR, but until then we may want to add `--disable-diagnostics` as a parameter for all the Ocean tests in `slurmci-test.toml`.

Alternatively, we can make `--disable-diagnostics` the default and you have to explicitly add `--enable-diagnostics` to generate them?

src/Driver/Driver.jl
bischtob: 🥇

src/Driver/Driver.jl
bischtob: My bad

bischtob:
> There are no diagnostics for the Ocean model implemented. I'll fix this with something empty in the upcoming diagnostics PR, but until then we may want to add `--disable-diagnostics` as a parameter for all the Ocean tests in `slurmci-test.toml`.
> 
> Alternatively, we can make `--disable-diagnostics` the default and you have to explicitly add `--enable-diagnostics` to generate them?

Agreed and changed.

bischtob:
bors try

bischtob:
@simonbyrne, can you check this build failure?

kpamnany:
bors try

experiments/Atmos_GCM/heldsuarez.jl
```diff
@@ -1,143 +1,214 @@
...
+using CLIMA.ColumnwiseLUSolver: ManyColumnLU
 using CLIMA.Mesh.Filters
 using CLIMA.Mesh.Grids
 using CLIMA.MoistThermodynamics
 using CLIMA.PlanetParameters
 using CLIMA.VariableTemplates
 
 
+# Import directional keywords (CLIMA.Mesh.Grids)
+import CLIMA.Mesh.Grids: VerticalDirection, HorizontalDirection, EveryDirection
```
kpamnany: This is now no longer necessary due to #784.

src/Driver/Driver.jl
```diff
@@ -25,7 +25,7 @@ Base.@kwdef mutable struct CLIMA_Settings
     mpi_knows_cuda::Bool = false
     show_updates::Bool = true
     update_interval::Int = 60
-    enable_diagnostics::Bool = true
+    enable_diagnostics::Bool = false
```
kpamnany: Unfortunately this doesn't actually change the default. I thought I'd added a comment there to explain this, but clearly not. I'll fix this in #810 if you're decided that diagnostics should not be enabled by default?

bischtob: I am fine either way :).

kpamnany:
Please squash before merging? :)

bischtob:
> Please squash before merging? :)

Forgot about this! Thanks for reminding me :).

PR #774, opened 2020-02-26
Fixing runtests for systems with OpenMPI and less than 5 physical cores

leios:
# Fixing CLIMA tests for systems with low processor number

CLIMA tests fail on systems with fewer processors without the `--oversubscribe` option. We could add another exception for these cases or make the `--oversubscribe` option default.

## Description

The following error will appear on systems with fewer processors:

```
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 4
slots that were requested by the application:

  /usr/bin/julia

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
--------------------------------------------------------------------------
```

## For review by CLIMA Developers

- [x] There are no open PR's for this already
- [ ] The code conforms to the [style guidelines](We should have this) and has consistent naming conventions
- [ ] This code does what it is technically intended to do (All numerics make sense either physically or computationally)

### DIFF
--- a/test/testhelpers.jl
+++ b/test/testhelpers.jl
@@ -18,8 +18,14 @@ function runmpi(tests, file)
   coverage_opt = coverage_opts[Base.JLOptions().code_coverage]
   testdir = dirname(file)
 
+  if !Sys.iswindows() && occursin( "OpenRTE", read(`mpiexec --version`, String))
+    oversubscribe = `--oversubscribe`
+  else
+    oversubscribe = ``
+  end
+
   for (n, f) in tests
-    cmd = `mpiexec --oversubscribe -n $n $(Base.julia_cmd()) --startup-file=no --project=$(Base.active_project()) --code-coverage=$coverage_opt $(joinpath(testdir, f))`
+    cmd = `mpiexec $oversubscribe -n $n $(Base.julia_cmd()) --startup-file=no --project=$(Base.active_project()) --code-coverage=$coverage_opt $(joinpath(testdir, f))`
 
     @info "Running MPI test..." n f cmd
     # Running this way prevents:
--- a/test/testhelpers.jl
+++ b/test/testhelpers.jl
@@ -18,14 +18,8 @@ function runmpi(tests, file)
   coverage_opt = coverage_opts[Base.JLOptions().code_coverage]
   testdir = dirname(file)
 
-  if haskey(ENV, "SLURM_JOB_ID")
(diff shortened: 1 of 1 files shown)


leios:
bors try

bischtob:
Simulation breaks after several hundred days. Oscillations in the sponge layer seems to break it. No idea why yet. Still investigating :(.

leios:
Apparently the `--oversubscribe` option does not exist in MPICH, which is what we use for the tests:
https://github.com/precice/precice/pull/443

I'll try to fix this soon

kpamnany:
Essentially, we should not be checking `SLURM_JOB_ID` to decide whether or not to use `--oversubscribe`. Unfortunately, there's nothing else we can easily/safely use to check which MPI implementation we're on, or which special flags to pass in to `mpiexec`.

Maybe we should pull this from the environment from something like `CLIMA_MPI_ARGS`?

On a slightly different note, I'm a bit confused by this file altogether. Why is it `using MPI` and checking if it's [initialized](https://github.com/climate-machine/CLIMA/blob/0e68ffb55a907ecbd7cb2c77e43ef7dcc359af15/test/testhelpers.jl#L4-L6) and all that? It's just spawning `mpiexec` (which may not belong to the same MPI that MPI.jl is built with).

Cc: @simonbyrne

simonbyrne:
My plan is to add an enum to MPI.jl to identify the MPI implementation.

MPI.jl runtests currently does the following:
https://github.com/JuliaParallel/MPI.jl/blob/7536834fd4e36daaa9efb4b0b54eeb39a6b8deee/test/runtests.jl#L31-L33

leios:
bors try

PR #782, opened 2020-02-28
Interpolated velocity is projected along the rad, lat, and long direc…

skandalaCLIMA:
…tions for the cubed sphere mesh.

Changes made to the interface to provide additional flexibility to the user.

### DIFF
--- a/src/Mesh/Interpolation.jl
+++ b/src/Mesh/Interpolation.jl
@@ -14,7 +14,10 @@ using CUDAnative
 using GPUifyLoops
 
 export InterpolationBrick, write_interpolated_data,
-       InterpolationCubedSphere, interpolate_local!
+       InterpolationCubedSphere, interpolate_local!,
+       InterpolationTopology
+
+abstract type InterpolationTopology  end
 #--------------------------------------------------------
 """
     InterpolationBrick(grid::DiscontinuousSpectralElementGrid{FT}, xbnd::Array{FT,2}, xres) where FT <: AbstractFloat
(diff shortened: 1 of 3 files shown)


src/Mesh/Interpolation.jl
```diff
@@ -831,26 +827,31 @@ This interpolation function works for cubed spherical shell geometry.
...
-    if device==CPU()
+    if device == CPU()
         #------------------------------------------------------------------------------------------
         Nel   = length(offset) - 1
 
         vout    = zeros(FT,nvars) #FT(0)
         vout_ii = zeros(FT,nvars) #FT(0)
         vout_ij = zeros(FT,nvars) #FT(0)
 
+        _ρu, _ρv, _ρw = 2, 3, 4
```
simonbyrne: This is really specific to AtmosModel.

I think it would make more sense to put this into a separate function: ideally it would be integrated with the stuff in src/Diagnostics.

src/Mesh/Interpolation.jl
```diff
@@ -996,13 +1025,59 @@ function interpolate_cubed_sphere_CUDA!(offset::AbstractArray{T,1}, m1_r::Abstra
...
+    bs     = blockDim().x  # block dim
+    idx    = ti + (bi-1)*bs
+    np_tot = size(v,1)
+
+    _ρu, _ρv, _ρw = 2, 3, 4
+
+    # projecting velocity onto unit vectors in rad, lat and long directions
+    # assumed u, v and w are located in columns 2, 3 and 4
+    if idx ≤ np_tot
+#        vrad =  v[idx,_ρu] * CUDAnative.cosd(lat_grd[lati[idx]]) * CUDAnative.cosd(long_grd[longi[idx]]) +
```
simonbyrne: @vchuravy I'm assuming that this breaks because CUDA doesn't have a `cosd` function?

src/Mesh/Interpolation.jl
simonbyrne: Since this fixes a somewhat urgent problem, I'm happy to merge it in for the time being. We need to come up with a more robust solution though.

src/Mesh/Interpolation.jl
skandalaCLIMA: In this case, the code does not break down. With the commented lines enabled, it runs to completion but produces the wrong result.  We ran the CUDA cosd function on a sample argument earlier today. While we noticed a difference, it was too small to make a big impact. Still trying to identify what is going wrong.

simonbyrne: Okay, we can keep what you have for now.

simonbyrne:
We can do this until we figure out a better solution.

Add a GPU test to the Slurm CI and then merge

PR #790, opened 2020-03-02
Remove some random perturbations for DYCOMS

kpamnany:
# Description

This removes some of the random perturbations that were being applied to the initial state of DYCOMS, leaving behind only one for `θ_liq`.

Submitted on behalf of @smarras79.

### DIFF
--- a/experiments/Atmos_LES/dycoms.jl
+++ b/experiments/Atmos_LES/dycoms.jl
@@ -267,14 +267,8 @@ function init_dycoms!(bl, state, aux, (x,y,z), t)
 
     # Perturb initial state to break symmetry and trigger turbulent convection
     r1 = FT(rand(Uniform(-0.002, 0.002)))
-    r2 = FT(rand(Uniform(-0.00001, 0.00001)))
-    r3 = FT(rand(Uniform(-0.001, 0.001)))
-    r4 = FT(rand(Uniform(-0.001, 0.001)))
-    if z <= 400.0
+    if z <= 200.0
         θ_liq += r1 * θ_liq
-        q_tot += r2 * q_tot
-        u     += r3 * u
-        v     += r4 * v
     end
 
     # Pressure


vchuravy:
In general we should inject randomness in a very controlled way, so that we do have reproducible runs as much as possible which aids debugging.

kpamnany:
Is there a better way to "trigger turbulent convection"? @charleskawczynski?

With regard to the current implementation, we can simply seed the RNG with a constant to get reproducibility. But this is not exactly a test -- it is an experiment. IOW, it should run for a few hours, _with_ a random perturbation, to verify that the physics is working correctly. So it would be a weekly test.

So I guess what we want to do is keep the short run with a constant seed to the RNG for a quick test, and add a variation that uses a random variation for a weekly test (when we have those). Does that sound right?

smarras79:
> Is there a better way to "trigger turbulent convection"? @charleskawczynski?
> 
> With regard to the current implementation, we can simply seed the RNG with a constant to get reproducibility. But this is not exactly a test -- it is an experiment. IOW, it should run for a few hours, _with_ a random perturbation, to verify that the physics is working correctly. So it would be a weekly test.
> 
> So I guess what we want to do is keep the short run with a constant seed to the RNG for a quick test, and add a variation that uses a random variation for a weekly test (when we have those). Does that sound right?

Low intensity random perturbations (given a specified spectrum or not) are pretty much the easiest and probably only way to do it.

simonbyrne:
Could we just sample some low-order frequencies and hard-code them? e.g.
```
θ_liq = 0.00232 * sin(x) + 0.0018 * sin(2x) + 0.00052 * cos(x) + θ_liq + 0.0002 * cos(2x)
```

bischtob:
> Could we just sample some low-order frequencies and hard-code them? e.g.
> 
> ```
> θ_liq = 0.00232 * sin(x) + 0.0018 * sin(2x) + 0.00052 * cos(x) + θ_liq + 0.0002 * cos(2x)
> ```

That's how I would do it, too.

charleskawczynski:
> Is there a better way to "trigger turbulent convection"? @charleskawczynski?

I agree with Simon, sin/cos waves are typically effective at achieving the same thing-- you may need to tune the frequency/amplitude of these waves. This might be a nice utility function available to initialization functions.

smarras79:
If we want to use some analytic perturbations for reproducibility, then I would recommend to follow Bryan's definitions. We want to trigger a turbulent field that has a pre-defined energy spectrum. 

Options (including code) are nicely described by George Bryan here:

[](https://www2.mmm.ucar.edu/people/bryan/Code/getperts.html)

kpamnany:
This PR reduces the amount of randomness, so I'll merge it and open an issue to track removing the remaining element of randomness.

PR #791, opened 2020-03-02
Move to KernelAbstractions

mwarusz:


mwarusz:
bors try

ali-ramadhan:
This is awesome! Will be a good source of examples.

Would be curious to see before/after performance benchmarks (although I assume they'll be pretty similar or even faster now[?] considering all the work being done in KernelAbstractions.jl).

mwarusz:
As Ali pointed out, this needs benchmarks and some cleanup. I started this PR just to see the CI.

mwarusz:
bors try

src/DGmethods/DGmodel.jl
```diff
@@ -110,43 +115,47 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment=false)
...
     communicate && MPIStateArrays.finish_ghost_recv!(Qhypervisc_div)
 
-    @launch(device, threads=Nfp, blocks=nrealelem,
-            facehyperviscterms!(bl, Val(dim), Val(N), dg.diffusion_direction,
-                                CentralHyperDiffusiveFlux(),
-                                Qhypervisc_grad.data, Qhypervisc_div.data,
-                                Q.data, auxstate.data,
-                                grid.vgeo, grid.sgeo, grid.vmap⁻, grid.vmap⁺,
-                                grid.elemtobndy, topology.realelems, t))
+    event = facehyperviscterms!(device, workgroups_face...)
```
vchuravy:
```suggestion
    event = facehyperviscterms!(device, workgroups_surface...)
```

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
bors try

mwarusz:
Hmm, it seems that `wait` on the GPU only synchronizes with the default stream and doesn't wait before launching subsequent kernels. We are failing because of unwanted overlap. Was the idea to specify everything using `dependencies` @vchuravy ?

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -31,923 +31,852 @@ const _n1, _n2, _n3 = Grids._n1, Grids._n2, Grids._n3
...
+                    Vars{vars_hyperdiffusive(bl,FT)}(l_Qhypervisc),
+                    Vars{vars_aux(bl,FT)}(l_aux), t)
+
+    @unroll for s = 1:nstate
+      s_F[1,i,j,k,s] += l_F[1,s]
+      s_F[2,i,j,k,s] += l_F[2,s]
+      s_F[3,i,j,k,s] += l_F[3,s]
+    end
+    
+    # Build "inside metrics" flux
```
simonbyrne: While you're updating this, would you be able to expand on what each block is doing? e.g. the equation it is evaluating?

mwarusz:
bors try

mwarusz:
bors try

lcw:
bors try

src/Arrays/MPIStateArrays_kernels.jl
```diff
@@ -2,29 +2,28 @@ using Requires
...
-  @inbounds @loop for i in (1:nvmapsend;
-                            threadIdx().x + blockDim().x * (blockIdx().x-1))
+  i = @index(Global, Linear)
+  @inbounds begin
     e, n = fldmod1(vmapsend[i], Np)
     @unroll for s = 1:nvar
       sendbuf[s, i] = buf[n, s, e]
     end
   end
-  nothing
```
blallen: From the YAS Guide (https://github.com/jrevels/YASGuide)

> - The return keyword should always be omitted from return statements within short-form method definitions (f(...) = ...). The return keyword should never be omitted from return statements within any other context (function ... end, macro ... end, etc.).
> - If a function does not have a clearly appropriate return value, then explicitly return nothing.

If we are really planning on using the YAS guide for this project, can we please start following the advice on return statements?

mwarusz: It is not easy to follow YAS here since the `@kernel` macro rewrites this function into three separate functions. After the macro has done its magic, `knl_fillsendbuf!` return a kernel object, and the kernel object returns an event. I would prefer if function annotated with `@kernel` do not have explicit return statements.

blallen: Ah okay, that makes sense. I think leaving out the `return` statement in `@kernel` functions is fine as long as we document that somewhere and stick to it throughout the project.

mwarusz:
bors try

mwarusz:
bors try

src/Arrays/MPIStateArrays_kernels.jl
vchuravy: Also this made me think about what it means to have a return statement and I think KernelAbstraction will generally forbid that. https://github.com/JuliaGPU/KernelAbstractions.jl/issues/50

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try-

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try-

mwarusz:
bors try

src/LinearSolvers/ColumnwiseLUSolver.jl
```diff
@@ -53,9 +54,12 @@ end
 
 function LS.linearsolve!(clu::ColumnwiseLU{F}, ::AbstractColumnLUSolver,
                          Q, Qrhs, args...) where {F <: DGModel}
+  device = typeof(Q.data) <: Array ? CPU() : CUDA()
+
   dg = clu.f
   A = clu.A
   Q .= Qrhs
+  sync_device(device)
```
lcw: This should not be needed anymore.

src/Utilities/Kernels/Kernels.jl
```diff
@@ -0,0 +1,13 @@
...
+using KernelAbstractions
+
+sync_device(::CPU) = nothing
+using Requires
+@init @require CUDAdrv = "c5f51814-7f29-56b8-a69c-e4d8f6be1fde" begin
+  import .CUDAdrv
+  sync_device(::CUDA) = CUDAdrv.synchronize()
+end
+
+end
```
lcw: Do we still need this file?

src/Arrays/MPIStateArrays.jl
```diff
@@ -4,7 +4,8 @@ using LinearAlgebra
 using DoubleFloats
 using LazyArrays
 using StaticArrays
-using GPUifyLoops
+using KernelAbstractions
+using ..Kernels
```
lcw: Not needed if we remove the Kernels file

src/CLIMA.jl
```diff
@@ -1,5 +1,6 @@
 module CLIMA
 
+include(joinpath("Utilities", "Kernels", "Kernels.jl"))
```
lcw: Not needed if we remove the kernels file

src/DGmethods/DGmethods.jl
```diff
@@ -8,7 +8,9 @@ using StaticArrays
 using ..SpaceMethods
 using ..VariableTemplates
 using DocStringExtensions
-using GPUifyLoops
+using KernelAbstractions
+using ..Kernels
```
lcw: Not needed if we remove the Kernels file

mwarusz:
bors try

mwarusz:
bors try-

mwarusz:
bors try

src/DGmethods/DGmodel_kernels.jl
mwarusz: Considering the size of this PR I think this should be addressed in a separate one. I don't mind doing this but I am not sure which equations you have in mind (physical equations in strong form, weak DG formulation, or fully discrete ? )

mwarusz:
@simonbyrne 

I think this ready to go. The build failed because of timeouts. The time limit for `test-cpu-tests`
and `test-gpu-tests` will need to be adjusted because of increased compilation times and additional overhead when launching small kernels a ton of times (e.g. `ODESolvers` tests).

src/DGmethods/DGmodel_kernels.jl
blallen: This is probably good to do in a separate PR after I finish updating the numerics document. We can then make sure the design docs, code, and comments all agree in notation and nomenclature.

vchuravy:
Let see how it deals with the extended timelimits on bors

bors try

simonbyrne:
bors try-

simonbyrne:
bors try-

simonbyrne:
bors ping

simonbyrne:
bors try

vchuravy:
Ah yes turning on threading of course causes MPI to think that the world is
burning down.

@simonbyrne is this OpenMPI with UCX?

simonbyrne:
>  @simonbyrne is this OpenMPI with UCX?

Yup. At least we figured this one out already.

https://github.com/JuliaParallel/MPI.jl/issues/337

simonbyrne:
bors try

vchuravy:
```
A call to mkdir was unable to create the desired directory:

  Directory: /tmp/ompi.hpc-90-03.5184/pid.432953
  Error:     No such file or directory

Please check to ensure you have adequate permissions to perform
the desired operation.
```

@simonbyrne maybe a node on Central is running out of space?

PR #795, opened 2020-03-03
Move NetCDF writing into a `Writers` module

kpamnany:
# Description

Extract NetCDF writing functionality from `Mesh/Interpolation.jl` into its own module (so that it can be reused by Diagnostics, others...)


src/InputOutput/Writers/netcdf.jl
```diff
@@ -0,0 +1,32 @@
...
+        for n in 1:nd
+            ds.dim[dimnames[n]] = dims[n]
+        end
+        for n in 1:nd
+            defVar(ds, dimnames[n], axes[n], (dimnames[n],))
+        end
+        for v in 1:nv
+            @eval (defVar($ds, $(varnames[v]), $(varvals[clns...,v]),
+                          $(dimnames)))
+    end
```
simonbyrne:
```suggestion
        end
```

src/InputOutput/Writers/netcdf.jl
```diff
@@ -0,0 +1,32 @@
...
+    Dataset(filename, "c") do ds
+        for n in 1:nd
+            ds.dim[dimnames[n]] = dims[n]
+        end
+        for n in 1:nd
+            defVar(ds, dimnames[n], axes[n], (dimnames[n],))
+        end
+        for v in 1:nv
+            @eval (defVar($ds, $(varnames[v]), $(varvals[clns...,v]),
+                          $(dimnames)))
```
simonbyrne: does this work without the `@eval`?

kpamnany: 🤔 ... I don't see why it wouldn't. I just copied @skandalaCLIMA's code over without thinking about it. Why did you use `@eval` @skandalaCLIMA?

skandalaCLIMA: I was using it to save the handle for the variable. Comments and other attirbutes can be added based on the handle. However, it is not currently used and I agree that it can be removed.

PR #800, opened 2020-03-04
Unroll loop over sources

simonbyrne:
# Description

Using `map` seemed to cause problems for > 4 sources.

This manually unrolls the loop using the `@nexprs` macro.

cc: @akshaysridhar 

### DIFF
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -9,8 +9,12 @@ end
 function atmos_source!(::Nothing, atmos::AtmosModel, source::Vars, state::Vars, diffusive::Vars, aux::Vars, t::Real)
 end
 # sources are applied additively
-function atmos_source!(stuple::Tuple, atmos::AtmosModel, source::Vars, state::Vars, diffusive::Vars, aux::Vars, t::Real)
-  map(s -> atmos_source!(s, atmos, source, state, diffusive, aux, t), stuple)
+@generated function atmos_source!(stuple::Tuple, atmos::AtmosModel, source::Vars, state::Vars, diffusive::Vars, aux::Vars, t::Real)
+  N = fieldcount(stuple)
+  return quote
+    Base.Cartesian.@nexprs $N i -> atmos_source!(stuple[i], atmos, source, state, diffusive, aux, t)
+    return nothing
+  end
 end
 
 abstract type Source


charleskawczynski:
How many sources can we unroll with this method? What's the scaling like? How far can we push this?

Let's go atomic sources! 🚀

src/Atmos/Model/source.jl
```diff
@@ -9,8 +9,11 @@ end
 function atmos_source!(::Nothing, atmos::AtmosModel, source::Vars, state::Vars, diffusive::Vars, aux::Vars, t::Real)
 end
 # sources are applied additively
-function atmos_source!(stuple::Tuple, atmos::AtmosModel, source::Vars, state::Vars, diffusive::Vars, aux::Vars, t::Real)
-  map(s -> atmos_source!(s, atmos, source, state, diffusive, aux, t), stuple)
```
vchuravy: No opinion on what is cleaner, but you could have done:

```
ntuple(Val(length(stuple)) do i
  Base.@_inline_meta # this is just defensive programming on my part
  s = @inbounds stuple[i]
  atmos_source!(s, atmos, source, state, diffusive, aux, t)
  nothing
end
```

simonbyrne: probably good if we can do without `@generated`.

simonbyrne:
@akshaysridhar can you try this out?

akshaysridhar:
Trying this out right now in `as/bomex-experiment-unrollsources`

akshaysridhar:
`error` at the same point in the code with this

simonbyrne:
@akshaysridhar There appear to be some undefined uses of `FT` in the sources. I have to run now, but can look at it more tomorrow.

akshaysridhar:
> @akshaysridhar There appear to be some undefined uses of `FT` in the sources. I have to run now, but can look at it more tomorrow.
Sorry : here is the latest version with these FT uses accounted for in the sources
https://gist.github.com/akshaysridhar/0e175d6c7b28994918ada0fd26985e39 is the latest updated `gist` against the appropriate branch

simonbyrne:
@vchuravy's solution didn't seem to work, so I switched back to the `@generated function`.

akshaysridhar:
`generated functions` solution worked :D !!! Thank you :D !!

PR #802, opened 2020-03-04
Add `ParameterSets`

charleskawczynski:
# Description

Adds `ParameterSets`, for flexibly handling
 - Planet constants
 - Free parameters

This should allow us to compute planet constants, or free parameters, in a flexible way. Specifically, we aim to be able to overload methods such as
 - `grav(::AbstractParameterSet)` (a planet constant) or
 - `C_smag(::AbstractParameterSet)` (a free parameter)

in the driver.


src/Common/ParameterSets/ParameterSets.jl
```diff
@@ -0,0 +1,21 @@
+"""
+    ParameterSets
+
+Module containing sets of parameters.
+"""
+module ParameterSets
```
simonbyrne: Lets call this module `Parameters` for now.

src/Common/ParameterSets/ParameterSets.jl
```diff
@@ -0,0 +1,21 @@
...
+
+export AbstractParameterSet
+abstract type AbstractParameterSet{FT<:AbstractFloat} end
+
+module Planet
+
+include("Atmos.jl")
```
simonbyrne: these shouldn't be inside the `Planet` module`

src/Common/ParameterSets/PlanetParameters.jl
```diff
@@ -0,0 +1,83 @@
+# Properties of dry air
```
simonbyrne:
```suggestion
module Planet
# Properties of dry air
```

parameters/Parameters.jl
```diff
@@ -0,0 +1,55 @@
+# NOTE: This file will soon be automatically generated
+module Parameters
```
simonbyrne:
```suggestion
```
we can leave off the `module`

parameters/Parameters.jl
```diff
@@ -0,0 +1,55 @@
...
+using CLIMA.ParameterSets
+import CLIMA.ParameterSets.Planet
```
simonbyrne:
```suggestion
import CLIMA
```

parameters/Parameters.jl
```diff
@@ -0,0 +1,55 @@
...
+
+export EarthParameters
+struct EarthParameters{FT} <: AbstractParameterSet{FT} end
```
simonbyrne:
```suggestion
struct ParameterSet{FT} <: AbstractParameterSet end
```

parameters/Parameters.jl
```diff
@@ -0,0 +1,55 @@
...
+
+# Properties of dry air
+Planet.molmass_dryair(ps::EarthParameters{FT}) where {FT} = FT(28.97e-3)
```
simonbyrne:
```suggestion
CLIMA.Parameters.Planet.molmass_dryair(ps::EarthParameters{FT}) where {FT} = FT(28.97e-3)
```
and move all these below to `parameters/Planet.jl`

experiments/Atmos_LES/dycoms.jl
```diff
@@ -344,7 +348,8 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
                            radiation=radiation,
                               source=source,
                    boundarycondition=bc,
-                          init_state=ics)
+                          init_state=ics,
+                       parameter_set=ParameterSet)
```
vchuravy:
```suggestion
                       parameter_set=ParameterSet{FT}())
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -86,7 +87,8 @@ function AtmosModel{FT}(::Type{AtmosLESConfiguration};
                          # TODO: Probably want to have different bc for state and diffusion...
                          boundarycondition::BC=NoFluxBC(),
                          init_state::IS=nothing,
-                         data_config::DC=nothing) where {FT<:AbstractFloat,O,RS,T,M,P,R,S,BC,IS,DC}
+                         data_config::DC=nothing,
+                         parameter_set::PS=nothing) where {FT<:AbstractFloat,O,RS,T,M,P,R,S,BC,IS,DC,PS}
```
vchuravy: Why is this `nothing`? Shouldn't it default to the default ParameterSet?

src/Common/Parameters/AtmosParameters.jl
```diff
@@ -0,0 +1,13 @@
...
+
+module Turbulence
+
+""" Smagorinsky Coefficient [dimensionless] """
+function C_smag end
+
+""" Drag coefficient """
+function C_drag end
+
+end
```
vchuravy:
```suggestion
end  # module Turbulence
```

src/Common/Parameters/AtmosParameters.jl
```diff
@@ -0,0 +1,13 @@
...
+
+end
```
vchuravy:
```suggestion
end # module Atmos
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -59,6 +59,7 @@ struct AtmosModel{FT,O,RS,T,M,P,R,S,BC,IS,DC} <: BalanceLaw
   boundarycondition::BC
   init_state::IS
   data_config::DC
+  parameter_set::PS
```
simonbyrne: lets make it the first element of the struct

src/Atmos/Model/AtmosModel.jl
simonbyrne: `ParameterSet` isn't defined yet.

```suggestion
                         parameter_set::PS) where {FT<:AbstractFloat,O,RS,T,M,P,R,S,BC,IS,DC,PS}
```

experiments/Atmos_LES/dycoms.jl
charleskawczynski: Hah, just caught that myself!

src/Common/Parameters/Parameters.jl
```diff
@@ -0,0 +1,17 @@
...
+    Parameters
+
+Module containing sets of parameters.
+"""
+module Parameters
+
+export AbstractParameterSet
+abstract type AbstractParameterSet{FT<:AbstractFloat} end
+
+include("PlanetParameters.jl")
```
simonbyrne:
```suggestion
include("Planet.jl")
```
is fine, since it's in the `Parameters` directory.

simonbyrne: Also the files under `CLIMA/parameters` and `CLIMA/src/Common/Parameters` should mirror each other (same names, same order etc)

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Yes, but since this is included at the driver level, it basically must be given. I'll add an error checking--still working on this. This is going to be a big PR, since `parameter_set` will be needed virtually everywhere 😬

parameters/Planet.jl
```diff
@@ -0,0 +1,53 @@
+# NOTE: This file will soon be automatically generated
+
+using CLIMA.Parameters
```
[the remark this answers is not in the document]
charleskawczynski: But we need `AbstractParameterSet` in `struct ParameterSet{FT} <: AbstractParameterSet{FT} end`

parameters/Planet.jl
```diff
@@ -0,0 +1,53 @@
...
+import CLIMA
+
+export ParameterSet
+struct ParameterSet{FT} <: AbstractParameterSet{FT} end
```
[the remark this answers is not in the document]
charleskawczynski: Ah, ok

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I guess there's no simple way to ease into this.

charleskawczynski: Actually, I'd like to split this PR into two steps/commits: 
 1) make the struct available everywhere that's needed (requires many interface changes, but guarantees no behavioral changes), also, passing in `parameter_set = nothing` is acceptable until step 2
 2) actually use the available data struct, and remove dependence on `PlanetParameters.jl`

simonbyrne: seems reasonable.

experiments/Atmos_LES/dycoms.jl
```diff
@@ -15,6 +16,9 @@ using CLIMA.MoistThermodynamics
 using CLIMA.PlanetParameters
 using CLIMA.VariableTemplates
 
+const clima_dir = dirname(Base.active_project())
```
vchuravy: If you want the path to `CLIMA/src` you probably want `dirname(pathof(CLIMA))`

```
 julia> using BenchmarkTools
[ Info: Precompiling BenchmarkTools [6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf]

julia> pathof(BenchmarkTools)
"/home/vchuravy/src/BenchmarkTools/src/BenchmarkTools.jl"

julia> dirname(pathof(BenchmarkTools))
"/home/vchuravy/src/BenchmarkTools/src"
```

tapios:
I prefer that the planet constants remain global constants, for two reasons:

1) It discourages defining local parameters for different models. (This is a problem in many climate models, e.g., specific heats of water in ice, land, ocean, and atmosphere models differ, making it impossible to ensure global conservation properties.) Global constants in my view made it clear that these really are meant to be global constants. 

2) I find having to pass `ps` to all thermodynamic functions unnecessarily cumbersome. 

What is the advantage of treating planet constants and free parameters symmetrically? (Conceptually they are very different.)

kpamnany:
@tapios: our understanding was that we want to allow the ability to override planetary constants in order to model climate on other planets (as `fms` does). If so, it simply makes sense to use the same mechanism to support both capabilities.

tapios:
Thanks, @kpamnany. My concern is in part that the code is getting more complicated when having to pass the parameter sets to each function. I had wanted to use, e.g., the moist thermodynamics code for other purposes (simple models, teaching etc.), and this change is making it harder to use. If there is no other way of doing it, fine. But I'd encourage you to think about whether this may be simplified.

kpamnany:
We can absolutely simplify it, but at a cost.

For instance, we can have a global `DefaultParameterSet = ParameterSet{FT}()` and make `ps` the last parameter in all these functions, with a default value of `DefaultParameterSet`. Then you don't have to specify `ps`.

But this entirely defeats one of the goals of this design, which is to avoid the possibility of inconsistent use -- if you want to override a planet parameter but you forget to pass in the override in one of your calls into CLIMA, the default will be used and now your code is suddenly using two different values for that parameter.

Forcing people to specify the parameter set to use doesn't make it impossible to make such a mistake, but it does make it harder.

If we treat this as an absolute requirement, then the only way to do this I can see is the C/Fortran way -- load all parameters from various files into various global structs and then explicitly copy those onto the GPUs. This runs us into other problems -- we can't easily treat parameters loaded this way as constants, so some performance optimizations are lost to us. Then, do we always copy all parameters? If not, how do we figure out the necessary ones? And how do we make it fast?

This is why we felt that using functions for these is the best way -- we leverage the power of Julia's JIT compilation and dispatch capabilities.

So given the rather complicated set of requirements for this feature, I think we have the best design possible. If you'd like to reconsider the requirements or re-order the importance of them, we should meet to discuss.

PS: the goal of making code run well on both CPUs and GPUs transparently is a really lofty one! You see some of the challenges involved in this one feature... :)

@vchuravy, @simonbyrne and @charleskawczynski: please jump in if I'm missing anything.

charleskawczynski:
> I had wanted to use, e.g., the moist thermodynamics code for other purposes (simple models, teaching etc.), and this change is making it harder to use. 

I know that these changes to moist thermo may seem wide-spread, however, they're actually quite a nice improvement. With these changes, we can more easily make moist thermo's test suite independent of planet parameters (which has been a task for a while). Also, from an interface perspective, we've only added one argument to the constructors:

```
ts = PhaseEquil.(ps, e_int, ρ, q_tot)
T = air_temperature(ts)
```

experiments/Atmos_LES/dycoms.jl
charleskawczynski: Nice, I didn't know of `pathof`. Perhaps that's more what I need (I actually need `CLIMA/`, though).

src/Atmos/Model/AtmosModel.jl
```diff
@@ -47,7 +48,8 @@ A `BalanceLaw` for atmosphere modeling.
                boundarycondition, init_state)
 
 """
-struct AtmosModel{FT,O,RS,T,M,P,R,S,BC,IS,DC} <: BalanceLaw
+struct AtmosModel{FT,PS,O,RS,T,M,P,R,S,BC,IS,DC} <: BalanceLaw
```
tapios: Use a more distinctive variable name than ps/PS for parameter set throughout. Otherwise this invites confusion with surface pressure.

tapios:
> > I had wanted to use, e.g., the moist thermodynamics code for other purposes (simple models, teaching etc.), and this change is making it harder to use.
> 
> I know that these changes to moist thermo may seem wide-spread, however, they're actually quite a nice improvement. With these changes, we can more easily make moist thermo's test suite independent of planet parameters (which has been a task for a while). Also, from an interface perspective, we've only added one argument to the constructors:
> 
> ```
> ts = PhaseEquil.(ps, e_int, ρ, q_tot)
> T = air_temperature(ts)
> ```

Let's talk about these issues. I see that this makes it easier to design tests. But it makes it harder for users (e.g., students in classes) to use the code, and the latter is more important. If there really is no other way of doing this, that's ok. But I'd like for you to explore what simplifications are possible.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Is `param_set` okay?

experiments/Atmos_LES/dycoms.jl
```diff
@@ -15,6 +16,9 @@ using CLIMA.MoistThermodynamics
 using CLIMA.PlanetParameters
 using CLIMA.VariableTemplates
 
+const clima_dir = dirname(CLIMA)
+include(joinpath(clima_dir, "..", "Parameters", "Parameters.jl"))
```
simonbyrne: At the moment we can just specify the relative path to the current file:
```suggestion
include(joinpath("..", "..", "Parameters", "Parameters.jl"))
```
Once we build the infrastructure to do it from toml files, we can add that as an option to the driver.

charleskawczynski: Going to stick with this to avoid a variable number of ".."'s for each driver that needs this.

charleskawczynski:
bors try

charleskawczynski:
bors try

simonbyrne:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
@vchuravy @kpamnany @simonbyrne On this branch, hash `fd23b93`, the following CPU test passes:
```julia
ENV["CLIMA_GPU"] = "false" # run on CPU
include(joinpath("test", "DGmethods", "courant.jl"))
```
And the following GPU test fails:
```julia
# GPU run:
include(joinpath("test", "DGmethods", "courant.jl"))
```
giving the following stack trace:

```
InvalidIRError: compiling knl_local_courant!(Cassette.Context{nametype(Ctx),Nothing,Nothing,GPUifyLoops.var"##PassType#422",Nothing,Cassette.DisableHooks}, typeof(CLIMA.DGmethods.knl_local_courant!), AtmosModel{Float64,ParameterSet{Float64},FlatOrientation,NoReferenceState,ConstantViscosityWithDivergence{Float64},DryModel,NoPrecipitation,NoRadiation,Gravity,PeriodicBC,typeof(initialcondition!),Nothing}, Val{2}, Val{4}, CUDAnative.CuDeviceArray{Float64,2,CUDAnative.AS.Global}, typeof(nondiffusive_courant), CUDAnative.CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CUDAnative.CuDeviceArray{Float64,3,CUDAnative.AS.Global}, CUDAnative.CuDeviceArray{Float64,3,CUDAnative.AS.Global}, UnitRange{Int64}, HorizontalDirection, Float64) resulted in invalid LLVM IR
  Reason: unsupported dynamic function invocation (call to overdub)
  Stacktrace:
   [1] air_temperature at C:\Users\kawcz\Dropbox\Caltech\work\dev\ClimateMachine\CLIMA\src\Common\MoistThermodynamics\MoistThermodynamics.jl:261
   [2] soundspeed_air at C:\Users\kawcz\Dropbox\Caltech\work\dev\ClimateMachine\CLIMA\src\Common\MoistThermodynamics\MoistThermodynamics.jl:377
   [3] soundspeed at C:\Users\kawcz\Dropbox\Caltech\work\dev\ClimateMachine\CLIMA\src\Atmos\Model\moisture.jl:33
   [4] nondiffusive_courant at C:\Users\kawcz\Dropbox\Caltech\work\dev\ClimateMachine\CLIMA\src\Atmos\Model\courant.jl:43
   [5] knl_local_courant! at C:\Users\kawcz\Dropbox\Caltech\work\dev\ClimateMachine\CLIMA\src\DGmethods\DGmodel_kernels.jl:1938
   [6] overdub at C:\Users\kawcz\.julia\packages\Cassette\7OymZ\src\overdub.jl:0
  Reason: unsupported dynamic function invocation (call to overdub)
```

This is [plaguing several, but not all, of the GPU tests](https://gist.github.com/climabot/c23150b427e161a9546411c02a60562c). The stack trace appears informative, but I see nothing wrong and I would expect to see the same error on the CPU.

This is a hard block, at the moment, and I cannot merge this PR until the problem is resolved.

I suspect that this may be a missing edge case in Cassette/GPUifyLoops. The potentially complicated logic of [storing a copy of ParameterSet{FT}()](https://github.com/climate-machine/CLIMA/blob/fd23b936208c8545b1b3e168383dd4209edcbb02/src/Common/MoistThermodynamics/states.jl#L71-L72) to simplify later moist thermo function calls seems like it could lead to many edge cases for Cassette/GPUifyLoops to handle. I'll try to test removing this copy tomorrow, but I'm still puzzled by why this would only affect some of the drivers.

Also, in case it helps:
```julia
julia> versioninfo()
Julia Version 1.3.1
Commit 2d5741174c (2019-12-30 21:36 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_PKG_DEVDIR = C:/Users/kawcz/Dropbox/Caltech/work/dev/julia
```

simonbyrne:
@charleskawczynski where is `e_int_v0` defined?
https://github.com/climate-machine/CLIMA/blob/fd23b936208c8545b1b3e168383dd4209edcbb02/src/Common/MoistThermodynamics/MoistThermodynamics.jl#L249-L251

charleskawczynski:
> @charleskawczynski where is e_int_v0 defined?

`e_int_v0` is defined in the [original `PlanetParameters`](https://github.com/climate-machine/CLIMA/blob/ck/ParameterSets/src/Common/MoistThermodynamics/MoistThermodynamics.jl#L14) (the values taken are unchanged). At the moment, no methods defined for `ParameterSet` (e.g., `grav`, etc.) is being used.

Once this PR passes, I'll replace the referenced values in `PlanetParameters` with method calls defined for `ParameterSet` in a subsequent PR.

charleskawczynski:
Thanks for the help on this @vchuravy!

charleskawczynski:
bors try

charleskawczynski:
bors try-

PR #813, opened 2020-03-06
improve formatting of simtime and endtime in logging

simonbyrne:
# Description

Switches to using a fixed format for printing simtime and endtime, and print denominator in callback.

Old:
```
┌ Info: Update
│ simtime = 2.6159163716067161e+02
│ runtime = 00:01:59
└ norm(Q) = 3.0116687421160989e+09
```

New:
```
┌ Info: Update
│ simtime =    15.34 /  1000.00
│ runtime = 00:28:31
└ norm(Q) = 3.0115880375677762e+09
```

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -348,10 +348,11 @@ function invoke!(solver_config::SolverConfiguration;
                                        Dates.dateformat"HH:MM:SS")
                 energy = norm(solver_config.Q)
                 @info @sprintf("""Update
-                               simtime = %.16e
+                               simtime = %8.2f / %8.2f
                                runtime = %s
                                norm(Q) = %.16e""",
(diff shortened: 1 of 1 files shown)


bischtob:
Can we add a lot more information about the grid as well?

kpamnany:
> Can we add a lot more information about the grid as well?

Presumably a one-time dump rather than here. That's done [here](https://github.com/climate-machine/CLIMA/blob/master/src/Driver/Configurations.jl#L91-L100) for Atmos LES, and [here](https://github.com/climate-machine/CLIMA/blob/master/src/Driver/Configurations.jl#L136-L145) for Atmos GCM. What else would you like to output?

Eventually we're going to add another type of detailed output that can be used to repeat the run; that will include every free parameter as well. This is just for interactive use right now, but if it can be made more useful, then sure.

PR #816, opened 2020-03-06
Add viscous sponge layer to Held-Suarez

bischtob:
# Description

This PR adds a viscous sponge layer to the Held-Suarez experiment so that waves and flow are damped near the top of the domain without altering the angular momentum balance of the planet.

The general goal for these kinds of physics-based stabilization techniques is to allow us to run to long-term steady state without issues.

- [x] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions

### DIFF
--- a/src/Atmos/Model/turbulence.jl
+++ b/src/Atmos/Model/turbulence.jl
@@ -19,7 +19,7 @@
 
 using DocStringExtensions
 using CLIMAParameters.Atmos.SubgridScale: inv_Pr_turb
-export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss
+export ConstantViscosityWithDivergence, SmagorinskyLilly, Vreman, AnisoMinDiss, ConstantViscousSponge
 export turbulence_tensors
 
 # ### Abstract Type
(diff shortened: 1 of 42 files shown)


src/Atmos/Model/turbulence.jl
```diff
@@ -106,12 +106,64 @@ end
...
   ν = m.ρν / state.ρ
   τ = (-2*ν) * S + (2*ν/3)*tr(S) * I
   return ν, τ
 end
 
+"""
+    ConstantViscousSponge <: TurbulenceClosure
+
+Turbulence with constant dynamic viscosity (`ρν`).
+Divergence terms are included in the momentum flux tensor.
```
tapios: The diffusion in sponge layers is usually just horizontal, in which case you do not want to include the divergence terms.

src/Atmos/Model/turbulence.jl
```diff
@@ -106,12 +106,64 @@ end
...
+end
+
+function turbulence_tensors(m::ConstantViscousSponge,
+    state::Vars, diffusive::Vars, aux::Vars, t::Real)
+
+  FT = eltype(state)
+  S = diffusive.turbulence.S
+  z = altitude(atmos.orientation,aux)
+  ν = m.ρν / state.ρ
+  if altitude >= FT(m.z_sponge)
```
tapios: Should this be 
```suggestion
  if z >= FT(m.z_sponge)
```?

File: src/Atmos/Model/turbulence.jl [hunk 3]
```diff
@@ -106,12 +106,64 @@ end
...
+    r = (altitude - FT(m.z_sponge))/(s.m.z_max-m.z_sponge)
```
tapios: Same here?
```suggestion
    r = (z - FT(m.z_sponge))/(s.m.z_max-m.z_sponge)
```

src/Atmos/Model/turbulence.jl
```diff
@@ -106,12 +106,64 @@ end
...
+    β_sponge = s.α_max * sinpi(r/2)^s.γ
+    ν += β_sponge*ν
+  end
+  τ = (-2*ν) * S + (2*ν/3)*tr(S) * I
```
tapios: Omit divergence and make horizontal only. (If there's vertical diffusion, you form a boundary layer at the model top even with a free-slip condition. This can lead to severe distortions of the circulation. See http://climate-dynamics.org/wp-content/uploads/2015/08/2004GL022304.pdf). Numerically, you need horizontal viscosities > 10^5 m^2/s in the sponge for it to damp fast waves effectively. If these (unrealistically) are isotropic viscosities and act also in the vertical, and if we deal with the diffusion explicitly, you would be limited to time steps of order 10^{-2} s. We don't want that.

tapios: We also need to be able to combine various turbulence models, e.g., `SmagorinskyLilly` with `ConstantViscousSponge`. Can this be achieved by using 
```suggestion
  τ += (-2*ν) * S 
```

bischtob:
@tapios, no need to review yet. This is an evolving draft with many parts that are being worked on. I am aware of the horizontal vs vertical issues. I will request reviews when bugs are ironed out.

src/Atmos/Model/turbulence.jl
akshaysridhar: There are two ways of doing this - we can allow this particular model (`ViscousSponge`) to be available as a `Source` (preferred perhaps), in which case we can add this in addition to (but structurally decoupled from) existing turbulence models with the current infrastructure. Interaction through `TurbulenceClosures` requires use of `map` functions to allow composition similar to `Source`

bischtob: @akshaysridhar, I like this idea. I am currently testing if we can simply replace the current Rayleigh sponge with this viscous sponge and still have a stable simulation.

bischtob:
Will open a new PR

PR #806, opened 2020-03-05
More consistent and general Courant calculations

jkozdon:
# Description

- The courant functions were atmos specific. This adds a stub module `Courant` then lets `AtmosModel` override them.
- The `calculate_dt` function did not use the initial condition, thus the courant number may be larger for the given initial condition (this lead to a [weird test in `driver_test.jl`](https://github.com/climate-machine/CLIMA/blob/b7515251d1aeec89e69c8997c31ddb09991025c7/test/Driver/driver_test.jl#L82-L87). Since the initial condition is known before the time step is calculated, this can be used to calculate a consistent dt

The new `calculate_dt` is only used for the atmos model and the ocean model remains unchanged (cc @blallen)

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [x] CLIMA developers with relevant expertise have been assigned to review this submission
- [x] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions
- [x] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #822, opened 2020-03-10
Default CFL_max added for various time-stepping schemes for dycoms, r…

skandalaCLIMA:
…isingbubble and surfacebubble

cases in experiments/Atmos_LES/ based on experimental runs.

# Description


tapios:
Thanks for working on this, @skandalaCLIMA. What we wanted here are defaults for each time stepping scheme, not a separate setting for each case. Can't we get universal defaults for the various schemes (say, the minimum CFL numbers of those you listed), get them from the time stepper directly, and use those to set time steps? I want to avoid separately specifying Courant numbers for each configuration.

jkozdon:
@skandalaCLIMA once #806 is in, you might want to recheck your numbers.

experiments/Atmos_LES/dycoms.jl
```diff
@@ -375,7 +380,17 @@ function main()
...
-    solver_config = CLIMA.setup_solver(t0, timeend, driver_config, init_on_cpu=true)
+
+    # Courant number
+    if driver_config.solver_type.solver_method == LSRK144NiegemannDiehlBusch
+        CFL_max = FT(2.4)
+    elseif driver_config.solver_type.solver_method == LSRK54CarpenterKennedy
+        CFL_max = FT(0.6)
+    else
+        CFL_max = FT(0.5)
+    end
```
thomasgibson: There is a lot of unnecessary code repetition introduced here. Let's try something a bit different. I think this check here can be part of the constructor for whatever `AbstractSolverType` comes in. For example, consider something like this in `src/Driver/Configurations.jl`:
```julia
struct ExplicitSolverType <: AbstractSolverType
    solver_method::Function
    courant_number::FT

    function ExplicitSolverType(;solver_method  = LSRK54CarpenterKennedy,
                                 courant_number = nothing)
        # User has specified a specific Courant number to use
        if courant_number !== nothing
            return new(solver_method, courant_number)
        # Otherwise, estimate a stable Courant number
        else
            courant_number = solver_method.estimated_CFL
            return new(solver_method, courant_number)
    end
end
```
You'll need to attach an `estimated_CFL` property to the desired ode solver. Similarly, you can update `IMEXSolverType` in the same way. Not sure if this is the best way, but hopefully the idea is clear.

Now we no longer need to duplicate this piece of code to configure the timestepper. And it allows users to bypass this check entirely and just specify their own Courant number. Now you can change lines 383--391 here to simply
```julia
CFL_max = driver_config.solver_type.courant_number
```
Just my thoughts.

jkozdon: If the rising bubble is fixed (e.g, so the advection velocity is not so high) default will likely work better. Right now DYCOMS with LSRK14 can run with CFL of 2.4 whereas rising bubble and surface bubble are 1.7. I believe that this is why @skandalaCLIMA opted to have a different CFL for each.

jkozdon: @thomasgibson and @fxgiraldo: Does surface bubble have the same high advection velocity problem that the current version of rising bubble does?

thomasgibson: Not sure about the surface bubble.

thomasgibson: > If the rising bubble is fixed (e.g, so the advection velocity is not so high) default will likely work better. Right now DYCOMS with LSRK14 can run with CFL of 2.4 whereas rising bubble and surface bubble are 1.7. I believe that this is why @skandalaCLIMA opted to have a different CFL for each.

Oh, I agree! I was just proposing a possible alternative. And there may be instances where we know that we can push the Courant number out further than the default (acoustic) Courant (multirate, for example).

experiments/Atmos_LES/dycoms.jl
```diff
@@ -346,9 +346,14 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
...
 
+    ode_solver = CLIMA.DefaultSolverType() # imex_solver
+    #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK54CarpenterKennedy)
+    #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
+
+
     config = CLIMA.Atmos_LES_Configuration("DYCOMS", N, resolution, xmax, ymax, zmax,
                                            init_dycoms!,
-                                           solver_type=CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch),
+                                           solver_type=ode_solver,
```
kpamnany: If you're using the default solver type, there's no need to specify it.

experiments/Atmos_LES/dycoms.jl
kpamnany: For there to be a `solver_method.estimated_CFL` is how I imagined this -- agree with this suggestion.

smarras79: > If the rising bubble is fixed (e.g, so the advection velocity is not so high) default will likely work better. Right now DYCOMS with LSRK14 can run with CFL of 2.4 whereas rising bubble and surface bubble are 1.7. I believe that this is why @skandalaCLIMA opted to have a different CFL for each.

@jkozdon with what code version or PR were you able to run dycoms at Courant = 2.4? I can only keep it stable for up to 1.8 with LSRK14. 
Let me know and I'll try that. Thanks

jkozdon: > @jkozdon with what code version or PR were you able to run dycoms at Courant = 2.4?

I have no idea. This is just what @skandalaCLIMA put into the driver.

skandalaCLIMA: Hello @smarras79, I was able to run the simulation in experiments/Atmos_LES/dycoms.jl to completion (timeend = 100; dt = 9.78474e-02) with a CFL of 2.4 for LSRK14. Would you recommend a different timeend for this simulation?

skandalaCLIMA:
> Thanks for working on this, @skandalaCLIMA. What we wanted here are defaults for each time stepping scheme, not a separate setting for each case. Can't we get universal defaults for the various schemes (say, the minimum CFL numbers of those you listed), get them from the time stepper directly, and use those to set time steps? I want to avoid separately specifying Courant numbers for each example.

Default CFL_max values set based on data for the rising bubble case, which is conservative.

skandalaCLIMA:
> @skandalaCLIMA once #806 is in, you might want to recheck your numbers.

Yes. I will repeat the tests after #806 is merged.

experiments/Atmos_LES/dycoms.jl
```diff
@@ -346,9 +346,14 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
                    boundarycondition=bc,
                           init_state=ics)
 
+    ode_solver = CLIMA.DefaultSolverType() # imex_solver
+    #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK54CarpenterKennedy)
+    #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
```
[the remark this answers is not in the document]
skandalaCLIMA: Done.

experiments/Atmos_LES/surfacebubble.jl
```diff
@@ -150,8 +150,9 @@ function config_surfacebubble(FT, N, resolution, xmax, ymax, zmax)
 
   C_smag = FT(0.23)
 
-  imex_solver = CLIMA.DefaultSolverType()
-  explicit_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
+  ode_solver = CLIMA.DefaultSolverType() # imex_solver
+  #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK54CarpenterKennedy)
+  #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
```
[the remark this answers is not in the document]
skandalaCLIMA: Done.

experiments/Atmos_LES/surfacebubble.jl
```diff
@@ -182,9 +183,19 @@ function main()
   t0 = FT(0)
   timeend = FT(2000)
 
-  CFL_max = FT(0.4)
+#  CFL_max = FT(0.4)
```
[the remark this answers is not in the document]
skandalaCLIMA: Done.

experiments/Atmos_LES/risingbubble.jl
```diff
@@ -72,8 +72,11 @@ function config_risingbubble(FT, N, resolution, xmax, ymax, zmax)
   bc = NoFluxBC()
 
   # Choose explicit solver
-  ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
-
+  #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK144NiegemannDiehlBusch)
+  #ode_solver = CLIMA.ExplicitSolverType(solver_method=LSRK54CarpenterKennedy)
```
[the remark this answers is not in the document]
skandalaCLIMA: Done.

experiments/AtmosLES/risingbubble.jl
```diff
@@ -11,7 +11,7 @@ using CLIMA.Mesh.Filters
 using CLIMA.MoistThermodynamics
 using CLIMA.PlanetParameters
 using CLIMA.VariableTemplates
-
+using CLIMA.ColumnwiseLUSolver: ManyColumnLU
```
simonbyrne: This isn't used anywhere, so can be removed.

simonbyrne:
This seems like a reasonable stop gap until we can figure out something more robust. However it seems like this would make more sense to go in `src/Driver/solver_configs.jl`? @kpamnany what do you think?

kpamnany:
Squash first?

simonbyrne:
Yes please squash

PR #823, opened 2020-03-10
Diagnostics: refactor and add interpolated diagnostics for Held Suarez

kpamnany:
# Description

Introduce `DiagnosticsGroup`s, which share an interval and an interpolation. Eventually, these will evolve to support grouping arbitrary individual diagnostics. Currently, all the diagnostics previously developed for DYCOMS are in the "AtmosDefault" `DiagnosticsGroup`. A "DumpStateAndAux" `DiagnosticsGroup` has been introduced that uses interpolation on the bare state and aux variables as required by Held Suarez.

Needs further testing. Will evolve further in other PRs.

- [X] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions


kpamnany:
I'm still debugging a bit. Will ping folks to review when I think this is ready.

src/Diagnostics/dump_state_and_aux.jl
```diff
@@ -0,0 +1,41 @@
...
+    interpolate_local!(diagrp.interpol, dg.auxstate.data, iaux,   project = diagrp.project)
+
+    # filename (may also want to take out)
+    nprefix = @sprintf("%s_%s-%f-step%04d",
+                       diagrp.out_prefix, diagrp.name, Settings.starttime, diagrp.step)
+    statefilename = joinpath(Settings.output_dir, "$(nprefix).nc")
+    auxfilename = joinpath(Settings.output_dir, "$(nprefix)_aux.nc")
+
+    #statenames = flattenednames(vars_state(bl, FT))
+    statenames = ("rho", "rhou", "rhov", "rhow", "rhoe")
```
blallen: Can you add a method for the user to overwrite the state names if necessary and leave `flattenednames(vars_state(bl, FT))` as the default? This way `dump_state_and_aux` would be generic to any model by default.

kpamnany: I've switched to using the `flattenednames` for now. These files will have to be post-processed anyway, so it should be possible to work with the Unicode. If not, I'll figure out a way to allow overriding the names.

kpamnany:
@skandalaCLIMA : please review. In particular, can you ensure that [this code](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L15-L40) sets up interpolation properly?

@smarras79 : could you please check this branch out, run DYCOMS with `--enable-diagnostics`, and ensure that the generated diagnostics still look fine? Note that the output filename has changed.

@bischtob / @LenkaNovak :  ditto please, with Held Suarez? The NC files generated by `DumpStateAndAux` for Held Suarez are enormous. Maybe I have [the resolution](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/experiments/AtmosGCM/heldsuarez.jl#L169) wrong?

skandalaCLIMA:
> @skandalaCLIMA : please review. In particular, can you ensure that [this code](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L15-L40) sets up interpolation properly?

Looks good to me.

kpamnany:
> > @skandalaCLIMA : please review. In particular, can you ensure that [this code](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L15-L40) sets up interpolation properly?
> 
> Looks good to me.

In particular, [this](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L36) `vert_range` is built from the `domain_height` and the `nelem_vert` originally used for setting up the grid, and the `nelem_horz` is also from setup. This is correct? So what the grid is interpolated _to_ is only specified by [`axes`](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L37), yes?

skandalaCLIMA:
> > > @skandalaCLIMA : please review. In particular, can you ensure that [this code](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L15-L40) sets up interpolation properly?
> > 
> > 
> > Looks good to me.
> 
> In particular, [this](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L36) `vert_range` is built from the `domain_height` and the `nelem_vert` originally used for setting up the grid, and the `nelem_horz` is also from setup. This is correct? So what the grid is interpolated _to_ is only specified by [`axes`](https://github.com/climate-machine/CLIMA/blob/kp/gcm-diagnostics/src/Driver/diagnostics_configs.jl#L37), 

Yes. This is correct.

smarras79:
I ran a dycoms test and plotted the results at 2000s and things look good.

experiments/AtmosLES/dycoms.jl
```diff
@@ -360,6 +361,12 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
     return config
 end
 
+function config_diagnostics(driver_config)
+    interval = 10000 # in time steps
+    dgngrp = setup_atmos_default_diagnostics(interval, driver_config.name)
```
smarras79: Great that the output file name contains basic information on the test. 
It would be useful if it also contained info on the number of elements in each direction and even the total number of time steps contained in the file.

smarras79: Also also the SGS model name. E.g. "SGS_SMAGO" or "SGS_Vreman" etc.

smarras79: This being said, a simple ascii log file with the basic information of the simulation setup would be extremely useful for repeatability.

akshaysridhar: I think this generally refers to making the properties of `AtmosModel` available in some way ?

kpamnany: The Parameters module is the first step towards achieving exactly the goal of repeatability. There is some ways to go though...

akshaysridhar: Are we able to tie in the `config_diagnostics` interval to the `simtime` instead of timestep count ? The timestep count is useful for de-bugging, but for diagnostics retrieval the simulation time is perhaps more intuitive ?

kpamnany: As we're using the generic callbacks mechanism, we can use any of `EveryXWallTimeSeconds`, `EveryXSimulationTime`, and `EveryXSimulationSteps`. Currently the driver uses `EveryXSimulationSteps`. We can either change that, or add the ability to choose any of the mechanisms.

akshaysridhar: Yeah I see https://github.com/climate-machine/CLIMA/blob/fc4f374dc5612298c93fc3cf95f257b6311045f0/src/ODESolvers/GenericCallbacks.jl#L54 the commented out section is the `simtime` callback hook we would need

PR #834, opened 2020-03-11
Diagnostics update

kpamnany:
# Description

Abstract output format for diagnostics -- support NetCDF or JLD2. Fix broken "DumpStateAndAux" caused by removal of `Interpolation.write_interpolated_data()`. Also add capability for _not_ using interpolation. Other small updates.


kpamnany:
@smarras79 : currently, the "AtmosDefault" diagnostics are written into a JLD2 file indexed on the (solver) time at which the diagnostics are collected. I propose changing this structure as a first step to moving to NetCDF such that there is one JLD2 file for each time of collection. Each JLD2 file will contain (like a NetCDF file), the dimension names, the axes, and values indexed by variable name, as implemented [here](https://github.com/climate-machine/CLIMA/pull/834/files#diff-bf11fd2d3da7e2bcfaaf148a64f5b1dfR18-R28).

Let's discuss this to ensure it makes sense.

kpamnany:
@smarras79: This is ready for review. Please let me know if you have any questions about the new structure of the diagnostics output.

In brief:
- There is now one JLD2 output file per diagnostics collection. The name of the file has `stepXXXX` to indicate which collection step.
- Access the data with (e.g.):
```
using JLD2
ds = jldopen("output/DYCOMS_AtmosDefault-2020-03-24T15.22.24.309-step0001.jld2", "r")
ds["z"]
ds["q_tot"]
```
With the current DYCOMS setup, there will be 155 elements in each array (`Nqk * nvertelem`). To plot over time, you will have to open each of the diagnostics output files and accumulate what you need.

szy21:
@akshaysridhar would you be able to run this branch to see whether the netcdf output makes sense? Once you review it we can merge it into the master.

src/Diagnostics/Diagnostics.jl
```diff
@@ -154,6 +160,7 @@ TODO: this will be refactored soon.
 function setup_dump_state_and_aux_diagnostics(
     interval::Int,
     out_prefix::String;
+    writer = NetCDFWriter(),
     interpol = nothing,
```
akshaysridhar:
```ERROR: LoadError: type Nothing has no field Npl
Stacktrace:
 [1] getproperty(::Nothing, ::Symbol) at ./Base.jl:20
 [2] dump_state_and_aux_collect(::DiagnosticsGroup, ::Float32) at /central/home/asridhar/CLIMA/src/Diagnostics/dump_state_and_aux.jl:12 ```

**Comment by akshaysridhar** [comment]
> @smarras79: This is ready for review. Please let me know if you have any questions about the new structure of the diagnostics output.
> 
> In brief:
> 
> * There is now one JLD2 output file per diagnostics collection. The name of the file has `stepXXXX` to indicate which collection step.
> * Access the data with (e.g.):
> 
> ```
> using JLD2
> ds = jldopen("output/DYCOMS_AtmosDefault-2020-03-24T15.22.24.309-step0001.jld2", "r")
> ds["z"]
> ds["q_tot"]
> ```
> 
> With the current DYCOMS setup, there will be 155 elements in each array (`Nqk * nvertelem`). To plot over time, you will have to open each of the diagnostics output files and accumulate what you need.

@kpamnany @smarras79 

JLD2 output (`AtmosDefault` configuration) produces correct output. I see that there are now n-JLD2 for n timesteps. Is the simulation time accessible through `keys(data)` (?)

**Comment by kpamnany** [comment]
> JLD2 output (`AtmosDefault` configuration) produces correct output. I see that there are now n-JLD2 for n timesteps. Is the `simulation time` accessible through `keys(data)` ? (right now the filename is tagged with the output step index) .

I will add simulation time as a key in both formats.

>     1. Key `dim_1` points to "z" - is this necessary ? (Key "z" already contains the altitude values)

This is to mirror NetCDF functionality, which supports arbitrarily named dimensions separate from the variables. There's no actual structural difference in JLD2, but in case you rename the dimensions (after interpolation for instance), you can find out the dimension names in `dim_X` where `1<=X<=ndims`, and from this determine the dimension units and extent.

> NETCDF writer does not work with the writer `kwarg` in the diagnostic configuration.

Will debug. Thanks for reviewing and verifying!

**Comment by akshaysridhar** [comment]
> > JLD2 output (`AtmosDefault` configuration) produces correct output. I see that there are now n-JLD2 for n timesteps. Is the `simulation time` accessible through `keys(data)` ? (right now the filename is tagged with the output step index) .
> 
> I will add simulation time as a key in both formats.
> 
> > ```
> > 1. Key `dim_1` points to "z" - is this necessary ? (Key "z" already contains the altitude values)
> > ```
> 
> This is to mirror NetCDF functionality, which supports arbitrarily named dimensions separate from the variables. There's no actual structural difference in JLD2, but in case you rename the dimensions (after interpolation for instance), you can find out the dimension names in `dim_X` where `1<=X<=ndims`, and from this determine the dimension units and extent.
> 
> > NETCDF writer does not work with the writer `kwarg` in the diagnostic configuration.
> 
> Will debug. Thanks for reviewing and verifying!

Two things appear to be causing issues : 
1 ) Interaction with the `interpol` `kwarg`
2) Specifying `NetCDFWriter()` as the required writer option in the `AtmosDefault` configuration

```
asridhar@hpc-23-28 CLIMA (as/test-writers) $ julia --project experiments/AtmosLES/bomex.jl --enable-diagnostics --diagnostics-interval=1 --output-dir=OUTTEST
... Removed WARNING dump for clarity 
┌ Info: Establishing Atmos LES configuration for BOMEX
│ precision        = Float32
│ polynomial order = 4
│ domain           = 6400.00x6400.00x3000.00
│ resolution       = 100x100x40
└ MPI ranks        = 1
[ Info: Initializing BOMEX
┌ Info: Starting BOMEX
│ dt              = 7.63165e-02
│ timeend         =  1800.00
│ number of steps = 23586
└ norm(Q)         = 1.8650791936000000e+10
┌ Info: Diagnostics: AtmosDefault
└ initializing at 0.0
┌ Info: Diagnostics: AtmosDefault
└ collecting at 0.07631646
dgngrp.writer = CLIMA.Writers.NetCDFWriter()
┌ Info: Diagnostics: AtmosDefault
``` The resulting output files are still JLD2. 
If you want to check out the setup I have that in `as/test-writers`

kpamnany:
> 1 ) Interaction with the interpol kwarg

The trouble with "DumpStateAndAux" is that it:
- assumes interpolation (already fixed), and
- uses `write_interpolated_data()` from `Mesh.Interpolation` which has been removed.

I'll fix these.

akshaysridhar:
> > 1 ) Interaction with the interpol kwarg
> 
> The trouble with "DumpStateAndAux" is that it:
> 
> * assumes interpolation (already fixed), and
> * uses `write_interpolated_data()` from `Mesh.Interpolation` which has been removed.
> 
> I'll fix these.

Awesome thanks ! cc:@bischtob who came across this issue in `DumpStateandAux` diagnostics in the GCM configuration as well

kpamnany:
> The resulting output files are still JLD2.

Turns out that the extension was hard-coded. They were actually NetCDF files. Fixed this too.

kpamnany:
I have not added support for non-interpolated "DumpStateAndAux" in this PR. Interpolation reduces across ranks so that there is a single output file. Without interpolation there would be one output file per rank. The behavior should be consistent, so will defer this.

kpamnany:
While "fixing" the filename extension, I found a weird bug: https://github.com/JuliaIO/JLD2.jl/issues/183 :-(

PR #840, opened 2020-03-12
Add `intensity` ENV to test suite

charleskawczynski:
# Description

## Background/motivation

When working on PRs that modify code in `src/`, one can easily test their code with a single driver and be relatively confident that other drivers will work.

However, working on PRs that touches virtually all of the drivers, as was the case in #802, getting feedback from the compiler (that the code is bug-free) is extremely inefficient. If we are going to recover agile development, we must fix this and I would like to provide additional analysis/motivation:

Screen Shot 2020-03-12 at 3 08 39 PM

I propose that we have some sort of switch to go from `P_i` with normal runtime to `P_i` with significantly reduced runtime.

In this PR, I introduce an environment variable `intensity` with `low` and `normal` options and some convenient functions to allow running the test suite with drastically minimal resolution (e.g., 1/2 time step(s), 3 grid points of resolution etc.). This will allow us to reduce the runtime for our test suite while retaining high coverage. Once this is done, we can push changes up to Azure/Slurm and run the test suite with normal intensity.

## Data
To support this claim, I applied this solution to moist thermodynamics, which tests a 3D space (e_int, ρ, q_tot), once, for all thermodynamic functions. I can see a 2x speedup from these changes, and I expect significantly higher returns on other drivers/tests where timesteps are involved.


kpamnany:
I'm not sure this is the right solution to this concern. Let's talk about it?

charleskawczynski:
#854 is turning out to be too tightly integrated to decompose. I will continue with this PR, and expand its use to other drivers/tests, and include slurm tests using "low" intensity, so that I can more efficiently work on #846. We can revert this PR once #846 is merged.

cc: @simonbyrne @kpamnany

PR #841, opened 2020-03-12
Reduce aux usage for Ocean Model

blallen:
# Description

I realized instead of storing the values for viscosity and diffusivity we could just use a function to calculate them pointwise. I would like to do something similar with the Coriolis `f`, wind stress `τ`, and sea surface temperature `θʳ` variables, but these require accessing either `geom` or `coords` in `source!`, `numerical_boundary_flux_*!` and `boundary_state!`.

### DIFF
--- a/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
+++ b/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
@@ -1,7 +1,8 @@
 module HydrostaticBoussinesq
 
 export HydrostaticBoussinesqModel, AbstractHydrostaticBoussinesqProblem,
-       LinearHBModel, calculate_dt
+       LinearHBModel, calculate_dt,
+       SimpleBoxProblem, HomogeneousBox, OceanGyre
 
 using StaticArrays
 using LinearAlgebra: I, dot, Diagonal, norm
@@ -48,6 +49,17 @@ A `BalanceLaw` for ocean modeling.
 
 write out the equations here
 
+ρₒ = reference density of sea water
+cʰ = maximum horizontal wave speed
+cᶻ = maximum vertical wave speed
+αᵀ = thermal expansitivity coefficient
+νʰ = horizontal viscosity
+νᶻ = vertical viscosity
+κʰ = horizontal diffusivity
+κᶻ = vertical diffusivity
+fₒ = first coriolis parameter (constant term)
+β  = second coriolis parameter (linear term)
+
 # Usage
 
     HydrostaticBoussinesqModel(problem)
@@ -55,6 +67,7 @@ write out the equations here
 """
 struct HydrostaticBoussinesqModel{P,T} <: BalanceLaw
   problem::P
+  ρₒ::T
   cʰ::T
   cᶻ::T
   αᵀ::T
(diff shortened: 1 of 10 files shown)


blallen:
@jkozdon and @lcw, this is same topic I just emailed you on, but we can just have the discussion here instead

blallen:
bors try

lcw:
If you need the coordinates then you will need to store them in `aux`.

blallen:
bors try

blallen:
> If you need the coordinates then you will need to store them in `aux`.

Thanks, managed to replace those three auxiliary variables with just one for `y`.

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
```diff
@@ -583,10 +528,9 @@ function update_aux_diffusive!(dg::DGModel, m::HBModel, Q::MPIStateArray, t::Rea
   # update vertical diffusivity for convective adjustment
   function f!(m::HBModel, Q, A, D, t)
     @inbounds begin
-      A.w = -(D.∇u[1,1] + D.∇u[2,2])
+      # probably need to be smarter about factoring out viscosity
+      A.w = -(D.ν∇u[1,1] + D.ν∇u[2,2]) / m.νʰ
```
blallen: @glwagner so right now I can get away with this because we have a diagonal viscosity tensor. Is there a reliable way to do this for an arbitrary viscosity tensor? Combining viscosity and `∇u` into a single variable is very convenient for the boundary condition specification.

blallen:
bors ping

blallen:
bors try-

blallen:
bors try

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
glwagner: Not sure I completely understand --- what is `A.w`? Is the dilemma between using the velocity gradient, versus the diffusive flux components, as prognostic variables in the first-order formulation?

PS: spaces after commas!

blallen: Ah, so `w = ∫(-∇⋅u)dz`. This is storing in `-(∂u/∂x + ∂v/∂y)` to be used as the integrand for the calculation. I just switched from storing just `∇u` to storing `ν∇u` and I'm trying to recover `∂u/∂x` and `∂v/∂y`. This is currently easy because `ν = Diagonal([νʰ,νʰ,νᶻ])`, so I can just divide out `νʰ` from `ν∇u[1,1]` and `ν∇u[2,2]` to get `∂u/∂x` and `∂v/∂y`. 

I'm wondering if there's an easy way to future proof this or are we just doing to need to do something like `∇u = ν \ ν∇u` with a temporary variable and then use the components of `∇u`. (I might have just answered my own question with that)

Also for the purposes of this, you can ignore everything before the `.` in a variable as that's just specifying where in memory it's stored.

glwagner: Got it! As you say I do think that in general you will wind up having to invert a matrix to obtain the velocity components from the components of the viscous momentum flux.

In general, the viscosities will not only be tensorial, but will be nonlinear functions of the velocity and density field.

Also, velocity gradients may be used elsewhere --- mostly in the calculation of the nonlinear viscosities and diffusivities themselves (in parameterizations), and probably also for post-processing.

The general need to access velocity gradients in places other than this line might warrant formulating the model in terms of velocity gradients themselves rather than viscous momentum flux, depending, of course, on how complicated that makes implementing boundary conditions.

blallen: As we discussed on slack, we'll assume we can properly pass the velocity gradients into the parameterizations and continue using the momentum fluxes here.

PR #847, opened 2020-03-14
Try setting Azure PR trigger

simonbyrne:
# Description

Attempt to fix issue where Azure CI is not building on PRs

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,6 +6,9 @@ trigger:
     - staging
     - trying
 
+pr:
+- master
+
 jobs:
 - job: Linux
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,23 +38,8 @@ jobs:
       set -o xtrace
       ./julia-$(JULIA_VERSION)/bin/julia -e 'using InteractiveUtils; versioninfo()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.instantiate()'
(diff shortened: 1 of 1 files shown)


simonbyrne:
Disabling codecov reduces linux tests from 2h 35m 30s to 59 mins

blallen:
Can we run the codecov check as a separate job? So have linux, macOS, windows, and codecov jobs launched on Azure?

kpamnany:
> Can we run the codecov check as a separate job? So have linux, macOS, windows, and codecov jobs launched on Azure?

Probably, but what's the advantage? It has to run the entire test suite to determine coverage anyway.

blallen:
> Probably, but what's the advantage? It has to run the entire test suite to determine coverage anyway.

Ah, not much at that point. But you'd know your code is bug-free an hour earlier, since you can see the status of individual Azure jobs.

jkozdon:
> Ah, not much at that point. But you'd know your code is bug-free an hour earlier, since you can see the status of individual Azure jobs.

Running the tests that run on Azure locally is fairly easy if you just want to see if there is a good chance Azure will pass:

```
julia --project=@. -e 'using Pkg; Pkg.test()'
```

At one point we had scripts to allow running slurm tests locally as well, but those went away when the slurm system was rewritten.

kpamnany:
> At one point we had scripts to allow running slurm tests locally as well, but those went away when the slurm system was rewritten.

It shoud be pretty straightforward to add this capability back into SlurmCI. I'll open an issue there to remind myself to get to it.

PR #850, opened 2020-03-16
Checkpoint/restart capability

### LINKED ISSUE #50: Checkpointing/restart
We need a way to checkpoint the Dycore for restarts (likely requires wrapping MPI I/O stuff).

kpamnany:
# Description

Adds checkpoint and restart capability. Supersedes #576. Closes #50, except:
- Writes to JLD2 rather than NetCDF for now. Will fix this in a future PR.
- Only supports `--checkpoint-walltime`; support for `--checkpoint-simtime` can be added once [this code](https://github.com/climate-machine/CLIMA/blob/master/src/ODESolvers/GenericCallbacks.jl#L52-L95) is enabled.

This is WIP as I need to add tests.

### DIFF
--- a/src/DGmethods/DGmodel.jl
+++ b/src/DGmethods/DGmodel.jl
@@ -517,6 +517,30 @@ function init_ode_state(dg::DGModel, args...; init_on_cpu = false)
     return state
 end
 
+function restart_ode_state(
+    dg::DGModel,
+    state_data;
+    init_on_cpu = false,
+    commtag = 888,
+)
+    bl = dg.balancelaw
+    grid = dg.grid
+
+    state = create_state(bl, grid, commtag)
+    state .= state_data
+
+    MPIStateArrays.start_ghost_exchange!(state)
+    MPIStateArrays.finish_ghost_exchange!(state)
+
+    return state
+end
+
+function restart_auxstate(bl, grid, aux_data)
+    auxstate = create_auxstate(bl, grid)
+    auxstate .= aux_data
+    return auxstate
+end
+
 # fallback
 function update_aux!(
     dg::DGModel,
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -1,6 +1,8 @@
 using ArgParse
 using CUDAapi
 using Dates
+using FileIO
+using JLD2
 using LinearAlgebra
 using Logging
 using MPI
@@ -42,6 +44,11 @@ Base.@kwdef mutable struct CLIMA_Settings
     wall_clock_interval::Int = 10000
     monitor_courant_numbers::Bool = false
     monitor_courant_interval::Int = 10
+    checkpoint_walltime::Int = -1
+    checkpoint_keep_one::Bool = true
+    checkpoint_at_end::Bool = false
+    checkpoint_dir::String = "checkpoint"
+    restart_from_num::Int = -1
     log_level::String = "INFO"
     output_dir::String = "output"
     integration_testing::Bool = false
@@ -136,6 +143,24 @@ function parse_commandline()
         help = "interval in Courant number calculations"
         arg_type = Int
         default = 10
+        "--checkpoint-walltime"
+        help = "interval in seconds at which to create a checkpoint"
+        arg_type = Int
+        default = -1
+        "--checkpoint-keep-all"
+        help = "keep all checkpoints (instead of just the most recent)"
+        action = :store_true
+        "--checkpoint-at-end"
+        help = "create a checkpoint at the end of the simulation"
+        action = :store_true
+        "--checkpoint-dir"
+        help = "the directory in which to store checkpoints"
+        arg_type = String
+        default = "checkpoint"
+        "--restart-from-num"
+        help = "checkpoint number from which to restart (in <checkpoint-dir>)"
+        arg_type = Int
+        default = -1
         "--log-level"
         help = "set the log level to one of debug/info/warn/error"
         arg_type = String
@@ -186,6 +211,11 @@ function init(; disable_gpu = false)
             parsed_args["monitor-courant-numbers"]
         Settings.monitor_courant_interval =
             parsed_args["monitor-courant-interval"]
+        Settings.checkpoint_walltime = parsed_args["checkpoint-walltime"]
+        Settings.checkpoint_keep_one = !parsed_args["checkpoint-keep-all"]
+        Settings.checkpoint_at_end = parsed_args["checkpoint-at-end"]
+        Settings.checkpoint_dir = parsed_args["checkpoint-dir"]
+        Settings.restart_from_num = parsed_args["restart-from-num"]
         Settings.integration_testing = parsed_args["integration-testing"]
         Settings.log_level = uppercase(parsed_args["log-level"])
     catch
@@ -206,6 +236,9 @@ function init(; disable_gpu = false)
     # create the output directory if needed
     if Settings.enable_diagnostics || Settings.enable_vtk
         mkpath(Settings.output_dir)
+    end
+    if Settings.checkpoint_walltime > 0 || Settings.checkpoint_at_end
+        mkpath(Settings.checkpoint_dir)
     end
 
     # set up logging
@@ -288,12 +321,13 @@ function invoke!(
                 )
                 energy = norm(Q)
                 @info @sprintf(
-                    """Update
-                    simtime = %8.2f / %8.2f
-                    runtime = %s
-                    norm(Q) = %.16e""",
+                    """
+         Update
+             simtime = %8.2f / %8.2f
+             runtime = %s
+             norm(Q) = %.16e""",
                     ODESolvers.gettime(solver),
-                    solver_config.timeend,
+                    timeend,
                     runtime,
                     energy
                 )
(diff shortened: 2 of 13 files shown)


src/Driver/Driver.jl
```diff
@@ -381,69 +384,39 @@ function invoke!(
...
-                return nothing
+                show_courant(solver_config)
+                nothing
             end
         callbacks = (callbacks..., cbcfl)
     end
 
+    if Settings.checkpoint_walltime > 0
+        step = Ref(1)
+        cbcheckpoint = GenericCallbacks.EveryXWallTimeSeconds(
```
akshaysridhar: Could we enable the option to use `SimulationTime` callbacks here ?

src/Driver/driver_configs.jl
```diff
@@ -130,6 +130,19 @@ struct DriverConfiguration{FT}
...
+    msg = "AtmosModel composition\n"
+    for key in fieldnames(typeof(model))
+        msg =
+            msg * @sprintf(
+                "    %s = %s\n",
+                string(key),
+                string((getproperty(model, key)))
+            )
+    end
+    @info msg
```
akshaysridhar: Nice, thanks for putting this back in ; One `Info` tag for the entire composition makes sense

test/Driver/mms_solution_generated.jl
```diff
@@ -0,0 +1,408 @@
...
+        sin(pi * x) *
+        cos(pi * t)^2 *
+        cos(pi * x) *
+        cos(pi * y)^2 *
+        cos(pi * z)^2 +
+        42000 * sin(pi * x) * cos(pi * t) * cos(pi * y) * cos(pi * z) -
+        3 * pi * sin(pi * z)^2 * cos(pi * t)^2 * cos(pi * x)^2 * cos(pi * y)^2 -
+        7 * pi * cos(pi * t)^2 * cos(pi * x)^2 * cos(pi * y)^2 * cos(pi * z)^2 +
+        42000 * cos(pi * t) * cos(pi * x) * cos(pi * y) * cos(pi * z)
+    ) / 300
```
akshaysridhar: This appears to be a formatting change

test/Driver/mms3.jl
```diff
@@ -0,0 +1,197 @@
...
+    )
+
+    CLIMA.invoke!(solver_config)
+
+    Qe = init_ode_state(dg, timeend)
+    result = euclidean_distance(solver_config.Q, Qe)
+    @test result ≈ expected_result
+end
+
+main()
```
akshaysridhar: This test is identical to the existing MMS test right (for a specific mesh resolution) ?

src/DGmethods/DGmodel.jl
```diff
@@ -398,6 +398,30 @@ function init_ode_state(
     return state
 end
 
+function restart_ode_state(
+    dg::DGModel,
+    s_state;
```
charleskawczynski: What is `s_state` here?

src/Driver/Driver.jl
```diff
@@ -179,6 +204,11 @@ function init(; disable_gpu = false)
             parsed_args["monitor-courant-numbers"]
         Settings.monitor_courant_interval =
             parsed_args["monitor-courant-interval"]
+        Settings.checkpoint_walltime = parsed_args["checkpoint-walltime"]
+        Settings.checkpoint_keep_one = !parsed_args["checkpoint-keep-all"]
+        Settings.checkpoint_at_end = parsed_args["checkpoint-at-end"]
+        Settings.checkpoint_dir = parsed_args["checkpoint-dir"]
+        Settings.restart_from = parsed_args["restart-from"]
```
charleskawczynski: It might make sense to call this `restart_from_nstep` or something. This way users don't need to look at the docs to see whether it's the step number or time.

test/Driver/cr_unit_tests.jl
```diff
@@ -0,0 +1,143 @@
...
+    turbulence = ConstantViscosityWithDivergence(FT(0))
+    model = AtmosModel{FT}(
+        AtmosGCMConfigType;
+        orientation = orientation,
+        ref_state = ref_state,
+        turbulence = turbulence,
+        moisture = DryModel(),
+        source = Gravity(),
+        init_state = setup,
+        param_set = CLIMA.ParameterSet{FT}(),
```
charleskawczynski:
```suggestion
        param_set = param_set,
```

test/Driver/cr_unit_tests.jl
```diff
@@ -0,0 +1,143 @@
+using CLIMA
+using CLIMA.Atmos
+using CLIMA.ConfigTypes
+using CLIMA.PlanetParameters
+using CLIMA.MoistThermodynamics
+using CLIMA.PlanetParameters
+using CLIMA.VariableTemplates
+using CLIMA.Grids
+using CLIMA.ODESolvers
+
```
charleskawczynski:
```suggestion
using CLIMA.Parameters
using CLIMA.UniversalConstants
const clima_dir = dirname(pathof(CLIMA))
include(joinpath(clima_dir, "..", "Parameters", "Parameters.jl"))
using CLIMA.Parameters.Planet
param_set = ParameterSet()
```

test/Driver/cr_unit_tests.jl
```diff
@@ -0,0 +1,143 @@
...
+    φ = latitude(bl.orientation, aux)
+    z = altitude(bl.orientation, aux)
+
+    β = min(FT(1), setup.α * acos(cos(φ) * cos(λ)))
+    f = (1 + cos(FT(π) * β)) / 2
+    g = sin(setup.nv * FT(π) * z / setup.domain_height)
+    Δp = setup.γ * f * g
+    p = aux.ref_state.p + Δp
+
+    ts = PhaseDry_given_pT(p, setup.T_ref)
```
charleskawczynski:
```suggestion
    ts = PhaseDry_given_pT(p, setup.T_ref, param_set)
```

test/Driver/mms3.jl
```diff
@@ -0,0 +1,197 @@
...
+using CLIMA.Atmos
+using CLIMA.ConfigTypes
+using CLIMA.DGmethods
+using CLIMA.DGmethods.NumericalFluxes
+using CLIMA.Mesh.Grids
+using CLIMA.Mesh.Topologies
+using CLIMA.MoistThermodynamics
+using CLIMA.MPIStateArrays
+using CLIMA.ODESolvers
+using CLIMA.PlanetParameters
```
charleskawczynski: Can we use ParameterSet in here as I suggested in `cr_unit_tests.jl`?

test/Driver/mms3.jl
kpamnany: Yes, and also 64 bit only. I'm not terribly happy about replicating this. What we probably need is a generalized MMS test that can be triggered with different options by different tests.

src/DGmethods/DGmodel.jl
kpamnany: `s_state` is the `Q` that was saved. Strictly speaking, it is `Q.realdata`. I'll update the name to be clearer.

test/Driver/mms_solution_generated.jl
kpamnany: Yes. Not sure if you want me to do anything about it, and not sure there's anything I can do...

src/Driver/Driver.jl
kpamnany: Since that callback implementation is commented, I think it needs tests when we re-enable it. I'll do that in a separate PR.

kpamnany:
There's an issue here about 'step's. The VTK callback originally had code to count the number of times it was invoked; it used this count in the generated filename and called it 'step'. This 'step' has nothing to do with the timestep. The other callbacks have sort of inherited this mechanism, but it is confusing.

Either we ignore this ambiguity, or we rename 'step' to something else, or we update the ODE solver callback mechanism to provide the actual timestep to callbacks. Thoughts?

blallen:
> we update the ODE solver callback mechanism to provide the actual timestep to callbacks. Thoughts?

I like this idea

src/DGmethods/DGmodel.jl
charleskawczynski: Maybe include the type, too?

kpamnany:
> > we update the ODE solver callback mechanism to provide the actual timestep to callbacks. Thoughts?
> 
> I like this idea

Note that this will mean that the filenames for VTK/diagnostics output will have non-contiguous step numbers as the callbacks don't run at every timestep. Is that fine? I'll go ahead with this if so.

blallen:
> > > we update the ODE solver callback mechanism to provide the actual timestep to callbacks. Thoughts?
> > 
> > 
> > I like this idea
> 
> Note that this will mean that the filenames for VTK/diagnostics output will have non-contiguous step numbers as the callbacks don't run at every timestep. Is that fine? I'll go ahead with this if so.

I realized that after posting... and that basically makes paraview useless since it assumes sequential numbering. maybe we just rename step for now and figure out a better long term solution once we have better visualization tools?

test/Driver/cr_unit_tests.jl
kpamnany: Using `param_set` here causes an error on the GPU. Is that expected? Does that mean this must run on CPU then?

charleskawczynski: It's fine to leave this out for now, then

test/Driver/mms_solution_generated.jl
akshaysridhar: Nope, just making sure

kpamnany:
`step` is now `num`.

src/Driver/Driver.jl
kpamnany: Opened #909 to track this.

test/Driver/mms3.jl
kpamnany: Opened #910 to track this.

kpamnany:
This is complete, but the verification tests are adding too much time to CI. :-(  Needs some thought.

PR #852, opened 2020-03-17
WIP Overlapping Communcation and Compuation

lcw:


lcw:
bors try

src/Arrays/MPIStateArrays.jl
```diff
@@ -121,7 +118,10 @@ struct MPIStateArray{
         DAI1 = typeof(vmaprecv)
         Buf = typeof(send_buffer)
         new{FT, typeof(data), DAI1, DAV, Buf}(
-            mpicomm,
+            # Make sure that each MPIStateArray has its own MPI context.  This
+            # allows multiple MPIStateArrays to be communicating asynchronously
+            # at the same time without having to explicitly manage tags.
+            MPI.Comm_dup(mpicomm),
```
vchuravy: Fantastic!

src/Arrays/MPIStateArrays.jl
```diff
@@ -420,49 +382,127 @@ function finish_ghost_send!(Q::MPIStateArray)
...
+"""
+    ghost_exchange!(Q::MPIStateArray; dependencies=nothing)
+
+Asynchronous call to perform the exchange of ghost degrees-of-freedom.  The
+returned `KernelAbstractions.Event` can be waited on to know when the ghost
+exchange is finished.
+"""
+function ghost_exchange!(Q::MPIStateArray; dependencies = nothing)
+    event = Event(__Irecv!, Q; dependencies = dependencies)
+    event = Event(__testall!, Q.recvreq; dependencies = event)
```
vchuravy: This means we need to make sure that MPI is using the thread safe interface, right now this interface is free to execute on any thread. https://github.com/JuliaGPU/KernelAbstractions.jl/blob/8a95e72e286d77be15de75971c4f52eaa4b8a599/src/backends/cpu.jl#L9-L15

@kpamnany do you have a feeling for what the performance impacts are these days for `MPI_Init_thread` with `MPI_THREAD_MULTIPLE`? We will also need to add support for that in MPI.jl

src/Arrays/MPIStateArrays.jl
```diff
@@ -420,49 +382,127 @@ function finish_ghost_send!(Q::MPIStateArray)
...
+    event = fillsendbuf!(
+        get_stage(Q.send_buffer),
+        Q.data,
+        Q.vmapsend;
+        dependencies = event,
+    )
+    event = prepare_transfer!(Q.send_buffer; dependencies = event)
+    # Here we don't wait on the send using a `KernelAbstractions.Event`.  The
+    # wait happens above at the MPI level via `MPI.Testall!`.
+    Event(__Isend!, Q; dependencies = event)
```
vchuravy: Dangling dependency edges like this make me worry. If nothing in the DAG waits on this there is no pressure to execute this task. So I would rather return a `MultiEvent` containing the `send` and `recv` event. Or return a tuple and thread this back into the dependency chain.

lcw:
> Same here, I find it tricky to reason about the implicit synchronization rules so we should lay out the reasoning in the comments.

I find it tricky too.  We have many places where we do this type of synchronization in the code.  Do you think we should put the same comment everywhere?

src/Arrays/MPIStateArrays.jl
lcw: Let's hope it works.

src/Arrays/MPIStateArrays.jl
lcw: Do we want to have an option to `Event` to run in the main thread?

simonbyrne:
Err, why did bors close this?

simonbyrne:
Also, I know we probably want an `Isend`/`Irecv` fallback, but would it be worth trying out the asynchronous neighbourhood collective operations for this? It seems that what we are essentially doing is an [`MPI_Ineighbor_alltoallv`](https://www.open-mpi.org/doc/current/man3/MPI_Ineighbor_alltoallv.3.php).

vchuravy:
> Err, why did bors close this?

I targeted the PR to the branch that had the original KA work, when that was merged bors deleted that branch and apparently deemed this PR not save worthy.

simonbyrne:
ah.

lcw:
When bors closes a pr it really closes it.  Since I can't reopen this one I have started a new one <https://github.com/climate-machine/CLIMA/pull/855>.

lcw:
> Also, I know we probably want an `Isend`/`Irecv` fallback, but would it be worth trying out the asynchronous neighbourhood collective operations for this? It seems that what we are essentially doing is an [`MPI_Ineighbor_alltoallv`](https://www.open-mpi.org/doc/current/man3/MPI_Ineighbor_alltoallv.3.php).

What is the benefit of using `MPI_Ineighbor_alltoallv`?  Usually I prefer just `MPI_Irecv` and `MPI_Isend`.

simonbyrne:
Theoretically it should give the MPI library more information to optimize communication: my understanding is that MPI libraries have put a fair bit of work into the graph topology functionality. It should also simplify the logic considerably (since we will only have one event that we need to wait on).

That said, we would need to do benchmarking, and we're not yet running at the scale where it would make much difference.

src/Arrays/MPIStateArrays.jl
simonbyrne: I've added support for `MPI.Init_thread` in https://github.com/JuliaParallel/MPI.jl/pull/363

PR #846, opened 2020-03-14
Remove `PlanetParameters`

charleskawczynski:
# Description

 - Removes PlanetParameters from moist thermo.
 - Fixes a typo in `EarthParameters` / `Parameters.Planet`
 - Comments tests in `PlanetParameters` due to namespace conflict with `Parameters.Planet` (`PlanetParameters` is in the process of being removed anyway)

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [x] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [x] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #854, opened 2020-03-17
Bug fixes, use ParameterSet, add Microphysics

charleskawczynski:
# Description

#846 is getting large, seems to be diverging, and still needs many iterations of testing. Since the tests are taking several hours to finish (including the queue time), I'm breaking this into smaller PRs. The downside here is that we'll intermediately depend on a mixture of `PlanetParameters` and `ParameterSet`s.


charleskawczynski:
Removing ParameterSet's from the bottom up is too entangled with the drivers, so we're now removing them from the top down.

PR #855, opened 2020-03-17
Multi-GPU Performance Updates

lcw:
This set of changes:
  - adds a wall-clock per time-step callback to `Driver`;
  - removes tag numbers from `MPIStateArray`s;
  - enables CUDA-aware MPI;
  - enables overlapping kernels for non-CUDA-aware MPI;
  - splits the DG face kernels to expose more computation to overlap with
    communication; and
  - removes the communication of the auxiliary state.

The changes were tested on an Amazon EC2 instance of type `p3.16xlarge` with 8
Titan V100s using `experiments/AtmosLES/dycoms.jl` with the following changes
to increase the number of elements in the simulation.

```diff
diff --git a/experiments/AtmosLES/dycoms.jl b/experiments/AtmosLES/dycoms.jl
index 20488bbc..9c764049 100644
--- a/experiments/AtmosLES/dycoms.jl
+++ b/experiments/AtmosLES/dycoms.jl
@@ -349,16 +349,16 @@ function main()
     N = 4
-
     # Domain resolution and size
-    Δh = FT(40)
-    Δv = FT(20)
+    Δh = FT(35)
+    Δv = FT(5)
     resolution = (Δh, Δh, Δv)
-
-    xmax = FT(1000)
-    ymax = FT(1000)
+    xmax = FT(3820)
+    ymax = FT(3820)
     zmax = FT(2500)
-
     t0 = FT(0)
-    timeend = FT(100)
+    timeend = FT(1)
-
     driver_config = config_dycoms(FT, N, resolution, xmax, ymax, zmax)
     solver_config = CLIMA.SolverConfiguration(
```

A simple strong scaling study was performed keeping the problem the same size
the same as the number of GPUs increased.  The results of which are given in
the following table.

                        Wall-Clock Time per Time-Step (seconds)

                                              Number of GPUs
                                          1      2      3      4
    before changes non-CUDA-aware MPI    0.848  0.863  0.858  0.615
    after  changes non-CUDA-aware MPI    0.838  0.482  0.479  0.405
    after  changes     CUDA-aware MPI    0.838  0.461  0.325  0.245

OpenMPI 4.0.1 was used for the non-CUDA-aware MPI runs and OpenMPI 4.0.2 with
UCX was used for the CUDA-aware MPI run.  Note, the before changes runs were ran
on a different instantiation of the AWS instance than the after changes runs.


lcw:
bors try

lcw:
bors try

src/DGmethods/DGmodel.jl
```diff
@@ -83,19 +83,30 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment = false)
...
+                auxstate;
+                dependencies = event,
+            )
         end
     end
 
     if nviscstate > 0 || nhyperviscstate > 0
 
-        event = Event(device)
         event = volumeviscterms!(device, workgroups_volume)(
```
vchuravy:
```suggestion
        volumevisc = volumeviscterms!(device, workgroups_volume)(
```

src/DGmethods/DGmodel.jl
```diff
@@ -142,23 +155,36 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment = false)
             hypervisc_indexmap,
             topology.realelems;
             ndrange = ndrange_surface,
-            dependencies = (event,),
+            dependencies = (event, exchange_Q, exchange_auxstate),
```
vchuravy:
```suggestion
            dependencies = (volumevisc, exchange_Q, exchange_auxstate),
```

vchuravy: Makes it easier to understand, ok the dependencies here are the volume computation and the exchange of Q and exchange of aux

lcw: Okay.  The reason I went with a generic event name was because sometimes kernels are called and sometimes they are not.  Using a generic name let me avoid a bunch of if statements.  Were I don't need a generic name I can do what you suggested.

vchuravy: Yeah if it is a `or` condition going with the generic event is totally fine. I just found myself reading through the code and going up and down to check which event we are waiting upon

lcw:
bors try

lcw:
bors try

lcw:
bors try

lcw:
bors try

lcw:
bors try-

lcw:
bors try

lcw:
bors try

lcw:
bors try

lcw:
bors try-

lcw:
bors try

lcw:
bors try

lcw:
bors try-

src/Arrays/MPIStateArrays.jl
```diff
@@ -138,8 +142,14 @@ struct MPIStateArray{
             nabrtovmaprecv,
             nabrtovmapsend,
             weights,
-            commtag,
         )
+        # Make sure that we have finished all outstanding data for halo
+        # exchanges before the MPIStateArray is finalied.
```
vchuravy, line 147:
```suggestion
        # exchanges before the MPIStateArray is finalized.
```

src/DGmethods/DGmodel.jl
```diff
@@ -140,25 +166,30 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment = false)
...
         end
 
         if nviscstate > 0
-            aux_comm = update_aux_diffusive!(dg, bl, Q, t)
-            @assert typeof(aux_comm) == Bool
-        end
-
-        if aux_comm
-            MPIStateArrays.start_ghost_exchange!(auxstate)
+            wait(device, comp_stream)
```
vchuravy, line 190:
```suggestion
            # update_aux_diffusive may start asynchronous work on the compute device
            # and we synchronize those here through a device event.
            wait(device, comp_stream)
```

src/DGmethods/DGmodel.jl
```diff
@@ -322,33 +434,25 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment = false)
...
         grid.vmap⁺,
         grid.elemtobndy,
-        topology.realelems;
-        ndrange = ndrange_surface,
-        dependencies = (event,),
+        grid.exteriorelems;
+        ndrange = ndrange_exterior_surface,
+        dependencies = (
+            comp_stream,
+            exchange_Q,
```
vchuravy, line 441: I was surprised by these waits, so adding a comment above would be good.

src/Arrays/MPIStateArrays.jl
```diff
@@ -334,149 +334,192 @@ end
...
+    __Irecv!(Q)
 
     # wait on (prior) MPI sends
-    finish_ghost_send!(Q)
+    @tic mpi_sendwait
+    MPI.Waitall!(Q.sendreq)
+    fill!(Q.sendreq, MPI.REQUEST_NULL)
+    @toc mpi_sendwait
 
     @tic mpi_sendcopy
```
vchuravy, line 371: This `tic` no longer measures anything but the time it takes to launch two kernels.

src/DGmethods/DGmodel_kernels.jl
```diff
@@ -462,15 +462,23 @@ See [`odefun!`](@ref) for usage.
         l_F = MArray{Tuple{nstate}, FT}(undef)
     end
 
-    e = @index(Group, Linear)
+    _e = @index(Group, Linear)
```
vchuravy, line 465:
```suggestion
    eI = @index(Group, Linear) # element index
```

vchuravy: Or something else, but a variable name starting with `_` is asking for a better name.

src/DGmethods/DGmodel.jl
```diff
@@ -433,9 +547,9 @@ function indefinite_stack_integral!(
     FT = eltype(Q)
 
     # do integrals
-    nelem = length(topology.elems)
+    nelem = length(elems)
     nvertelem = topology.stacksize
-    nhorzelem = div(nelem, nvertelem)
+    horzelems = fld1(first(elems), nvertelem):fld1(last(elems), nvertelem)
```
kpamnany, line 552: Ah, this is a pattern that's been copied in a number of places and might be worth `grep`ping for to change in all...

src/Mesh/Grids.jl
```diff
@@ -159,6 +159,15 @@ struct DiscontinuousSpectralElementGrid{
     "An array of ranges in `vmapsend` to send to each neighbor"
     nabrtovmapsend
 
+    "Array of real elements that do not have a ghost element as a neighbor"
+    interiorelems
+
+    "Array of real elements that have at least on ghost element as a neighbor"
```
kpamnany, line 165: "at least _one_"...

src/Arrays/MPIStateArrays.jl
```diff
@@ -334,149 +334,192 @@ end
...
-            recvbuf,
-            vmaprecv,
-            length(vmaprecv);
-            ndrange = length(vmaprecv),
-            dependencies = (event,),
+    for n in 1:nnabr
+        Q.sendreq[n] = MPI.Isend(
+            (@view transfer[:, Q.nabrtovmapsend[n]]),
+            Q.nabrtorank[n],
+            666,
```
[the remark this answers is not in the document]
lcw: Yes, you still need to pass a tag to MPI Isend and Irecv.  We just don't need a different one for each MPIStateArray since they each have their own dup'd communicator.

Manifest.toml
```diff
@@ -319,7 +319,9 @@ uuid = "56ddb016-857b-54e1-b83d-db4d58db5568"
 
 [[MPI]]
 deps = ["Distributed", "DocStringExtensions", "Libdl", "Random", "Requires", "Serialization", "Sockets"]
-git-tree-sha1 = "a01349600200a1cb2157fd91e6d19e38642273ed"
+git-tree-sha1 = "e820e194a479146da0c04be6f16c900488427f40"
+repo-rev = "master"
```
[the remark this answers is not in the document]
vchuravy: The tag just landed. https://github.com/JuliaRegistries/General/pull/12306

src/Arrays/MPIStateArrays.jl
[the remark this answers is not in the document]
simonbyrne: would it be better to use `MPI.MPI_ANY_TAG`?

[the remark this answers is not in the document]
vchuravy: No since you still want to make sure that we get the update that is meant for the IRecv/ISend pair, while that should be the only traffic on the communicator (since it is now split), someone might add more ISend/IRecv later on.

src/Arrays/MPIStateArrays.jl
```diff
@@ -334,149 +334,192 @@ end
...
+            Q.mpicomm,
         )
-        wait(device(buf), event)
     end
 end
 
+function __yield(comm)
```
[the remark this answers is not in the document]
lcw: sure

src/DGmethods/DGmodel.jl
```diff
@@ -111,19 +115,41 @@ function (dg::DGModel)(dQdt, Q, ::Nothing, t; increment = false)
             hypervisc_indexmap,
             topology.realelems,
             ndrange = ndrange_volume,
-            dependencies = (event,),
+            dependencies = (comp_stream,),
+        )
+
```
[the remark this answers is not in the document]
lcw: I don't think it is needed.  Both kernels launch asynchronously and the main thread should end up in an `MPI_Waitall()`

src/DGmethods/DGmodel.jl
lcw: There are corresponding changes that need to be applied to the kernels to match these changes.  I would instead suggest (in a different pr) remove the `elems` argument from kernels that don't use them.

src/Mesh/Grids.jl
lcw: thanks!

src/Arrays/MPIStateArrays.jl
lcw: Doesn't measure the time in `MPI.Waitall!()`?  I think all of the calls to `@tic` and `@toc` have this issue.  I suggest that this be addressed in a future pr that makes `@tic` and `@toc` KernelAbstractions and MPI aware.

src/DGmethods/DGmodel.jl
lcw: Added some comments to hopefully make this a little more clear.

src/DGmethods/DGmodel.jl
lcw: I added comments to the waits.  Hopefully this helps.

src/Arrays/MPIStateArrays.jl
lcw: Thanks

src/DGmethods/DGmodel_kernels.jl
lcw: Changed

lcw:
> I'm not sure why the new Callbacks module -- are you aiming at abstracting the details from the Driver module?

That is right.  I was hoping that the callback would be available for some of the test / examples that do not use `Driver`.
 
> Do we need to switch SlurmCI to using `OpenMPI 4.0.2`?

I haven't done enough tests to know for sure but I think the main difference is if your OpenMPI is compiled with UCX.  It just so happened that my versions were different.

lcw:
bors try

lcw:
@kpamnany and @simonbyrne I don't understand why bors is failing to merge this.  It passes slrm CI https://gist.github.com/climabot/697563e848e0ca10cc24206b1f0bb297 and all of the above azure runs.  What do I need to do?

lcw:
Oh I see, bors also runs on azure https://github.com/climate-machine/CLIMA/runs/568004297

PR #858, opened 2020-03-18
Prototype of test suite revamp

charleskawczynski:
# Description

**This is a work in progress**

I think a dictionary, whose keys are the files, is the most efficient way to organize the tests. The "values" of the dictionary are the configuration details for each test. As such, I've tried moving all tests into a TOML file, to be read in and parsed to create a comprehensive test suite.

I'll add in defaults, and remove the default fields in the TOML, once it's clear what they'll be (I'm still wrangling all of the tests).

I think we should be able to run this locally and through slurm.


test/suite.toml
```diff
@@ -0,0 +1,1357 @@
+["test/Mesh/BrickMesh.jl"]
```
vchuravy: I would move the filename to it's own field and name the test `BrickMesh`

charleskawczynski: I thought about that, but then we can have name clashes (many tests are called `runtests.jl`), and we'll need to include the full filename anyway.

simonbyrne:
From discussion with @charleskawczynski, I was thinking something like this:
```toml
# a single job
[ParametersType]
type = "serial"
filename = "test/Utilities/ParametersType/runtests.jl"
frequency = "continuously"


# a matrix of jobs
[held-suarez]
type = "parallel"
filename = "test/experiments/AtmosGCM/heldsuarez.jl"

	[held-suarez.matrix.cpu-1]
	frequency = "continuous"
	script = "slurm/cpu.sh"
	options = ["--nprocs=1", "--time=00:30:00"]
	args = ["--timeend=30"]
	
	[held-suarez.matrix.gpu-full]
	frequency = "nightly"
	script = "slurm/gpu.sh"
	options = ["--nprocs=4", "--time=02:00:00"]
	args = ["--timeend=3000", "--output-vtk=out"]

# a matrix of jobs
[poisson]
frequency = "continuously"
filename = "test/LinearSolvers/poisson.jl"
type = "serial"

       [poisson.matrix.standard]
       type = "direct" # override type, ignore script
       [poisson.matrix.cpu-1]
       script = "slurm/cpu.sh"
       options = ["--nprocs=1"]
       [poisson.matrix.cpu-2]
       script = "slurm/cpu.sh"
       options = ["--nprocs=2"]
       [poisson.matrix.gpu-1]
       script = "slurm/gpu.sh"
       options = ["--nprocs=1"]
       [poisson.matrix.gpu-2]
       script = "slurm/gpu.sh"
       options = ["--nprocs=2"]

```

test/suite.toml
charleskawczynski: Spoke with Simon about this, we can use `basename(dirname(file))` when `basename(file) == runtests.jl`, so long this doesn't result in name clashes.

charleskawczynski: Ugh, already hitting an issue with this: `advection_sphere_ssp33` is tested in `DGmethods` and `DGmethods_old`. Just adding `_old` to the header

charleskawczynski:
> From discussion with @charleskawczynski, I was thinking something like this:

I'm fine with including the fields @simonbyrne suggests, except, there is a key feature of this file that I strongly suggest we preserve: multi-line editability.

Being able to multi-line edit this file would be _incredibly_ valuable, especially since it will grow over time. It's already very close and, if it lends itself well to multi-line editing (which means it must be sufficiently structured), we'll be able to search for, for example, `Atmos/Parameterizations`, and configure all of the tests to be CPU only (which is the case), without having to manually search/edit through the 100+ tests.

I would love to preserve this feature, and I think it will save many people a lot of time. Please, if this is important to you, chime in.

Perhaps this is why I think it would be good to specify the number of mpi ranks as an array, it lends itself better to multi-line editing.

test/suite.toml
jkozdon: If someone finished out #491 this issue would go away.

charleskawczynski: Yep, @trontrytel and I have been working on the microphysics drivers

charleskawczynski:
Values for `slurm` were updated from the `slurmci-test.toml`.

simonbyrne:
I still disagree with making `mpiranks` special: there are only a handful of tests where this is used, and in most cases we would probably want to vary other things with the number of ranks (e.g. resolution, runtime, etc.)

charleskawczynski:
Have you seen the recent changes? They're all arrays now, which allows for the flexibility we want.

kpamnany:
I think this way to specify test parameters is far too heavyweight. At the very least, it should not be necessary to specify empty keys (e.g. `include_only = []`).

More generally, I disagree with the idea of specifying the CI mechanism in the test specification -- what if we change our CI infrastructure? Why should we have to change our test suite?

I think there's virtue to keeping this simple. The test suite should specify what should be run, when, and what the expected results are. I think it's fine to specify the "when" with separate files -- one for merge tests, one for daily tests and one for weekly tests -- either the tests themselves will be different or the parameters will be.

I don't think adding a 2000 line test specification file is the right way forward.

charleskawczynski:
> I think this way to specify test parameters is far too heavyweight. At the very least, it should not be necessary to specify empty keys (e.g. include_only = []).

I know this has quite a bit of boilerplate, but making this file structured will allow us to more easily navigate and manage it.

> More generally, I disagree with the idea of specifying the CI mechanism in the test specification -- what if we change our CI infrastructure? Why should we have to change our test suite?

That's a good point, and I agree with this. However, I don't know another way of specifying all of the tests in a single place while allowing us to specify different combinations of parameters used depending on the CI. IIRC, specifying all of the tests in a single place was something that we wanted.

> I don't think adding a 2000 line test specification file is the right way forward.

I don't think this solution is elegant, but it's simple, easy to navigate/manage, and scalable. Also, I think that simplicity will be very helpful as more scientists contribute to the code base.

kpamnany:
> I know this has quite a bit of boilerplate, but making this file structured will allow us to more easily navigate and manage it.

I disagree that this is easy to navigate or manage. And much smaller and simpler files can also be structured.

> That's a good point, and I agree with this. However, I don't know another way of specifying all of the tests in a single place while allowing us to specify different combinations of parameters used depending on the CI. IIRC, specifying all of the tests in a single place was something that we wanted.

Again, IMO our test specifications should be orthogonal to the testing infrastructure. Why do we want to specify parameters differently depending on the CI? The true dependency should be "short" or "long" and we run the appropriate ones on the appropriate CI.

I agree that specifying tests in a single place is good, but I don't see any issue with separating test specifications based on something like "short" and "long", or "merge" and "daily".

> I don't think this solution is elegant, but it's simple, easy to navigate/manage, and scalable. Also, I think that simplicity will be very helpful as more scientists contribute to the code base.

I don't think any 2000 line file is easy to navigate or manage.

charleskawczynski:
@lcw @jkozdon can you help figure out what the issue is with the failure in [test/DGmethods_old/util/integral_test_sphere.jl](https://dev.azure.com/climate-machine/CLIMA/_build/results?buildId=2330&view=results)? I didn't modify this file. Is this a flaky test?

charleskawczynski:
Ah, nevermind, that test is not run on master. I wish we had coverage on our tests!

jakebolewski:
closed by #1234

PR #859, opened 2020-03-18
Functions for using the Non-Equil-Thermodynamics in atmos moiture model

slifer50:
# Description

Modifications were made to moistur.jl and source.jl to enable work to be done using NonEquilMoist thermodynamics.

I have tested these functions using what's present in PR #601. @charleskawczynski This has been made as a separate PR so that this can be merged into master independently of the squall-line driver while that continues to be worked on. @trontrytel Any suggestions for a quick and simple quantitative test of these functionalities?

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -275,7 +275,7 @@ function gradvariables!(
     transform.u = ρinv * state.ρu
     transform.h_tot = total_specific_enthalpy(atmos, atmos.moisture, state, aux)
 
-    gradvariables!(atmos.moisture, transform, state, aux, t)
+    gradvariables!(atmos.moisture, atmos, transform, state, aux, t)
     gradvariables!(atmos.turbulence, transform, state, aux, t)
 end
 
--- a/src/Atmos/Model/moisture.jl
+++ b/src/Atmos/Model/moisture.jl
@@ -187,6 +187,7 @@ end
 
 function gradvariables!(
     moist::EquilMoist,
+    atmos::AtmosModel,
     transform::Vars,
     state::Vars,
     aux::Vars,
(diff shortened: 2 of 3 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -275,7 +275,7 @@ function gradvariables!(
     transform.u = ρinv * state.ρu
     transform.h_tot = total_specific_enthalpy(atmos, atmos.moisture, state, aux)
 
-    gradvariables!(atmos.moisture, transform, state, aux, t)
+    gradvariables!(atmos.moisture, atmos, transform, state, aux, t)
```
akshaysridhar: Perhaps set leading argument as `atmos` , followed by subcomponents ?

src/Atmos/Model/moisture.jl
```diff
@@ -187,6 +187,7 @@ end
 
 function gradvariables!(
     moist::EquilMoist,
+    atmos::AtmosModel,
```
akshaysridhar: same as for AtmosModel.jl `278`

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
     flux.ρu += d_q_tot .* state.ρu'
     flux.moisture.ρq_tot += d_q_tot * state.ρ
 end
+"""
+  NonEquilMoist
+
+  Assumes that there isn't any thermodynamic equilibrium
```
akshaysridhar: Include comment on additional state variables in description

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    aux::Vars,
+    t::Real,
+)
+    ρinv = 1 / state.ρ
+    transform.moisture.q_tot = state.moisture.ρq_tot * ρinv
+    phase = thermo_state(moist, atmos.orientation, state, aux)
+    R_m = gas_constant_air(phase)
+    T = air_temperature(phase)
+    e_tot = state.ρe * ρinv
+    transform.moisture.h_tot = e_tot + R_m * T
```
akshaysridhar: Computed at `AtmosModel` level, doesn't need to be carried within the moisture model  (?)

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+  NonEquilMoist
+
+  Assumes that there isn't any thermodynamic equilibrium
+"""
+
+struct NonEquilMoist <: MoistureModel end
+
+vars_state(::NonEquilMoist, FT) = @vars(ρq_tot::FT, ρq_liq::FT, ρq_ice::FT)
+vars_gradient(::NonEquilMoist, FT) =
+    @vars(q_tot::FT, q_liq::FT, q_ice::FT, h_tot::FT)
```
akshaysridhar: `h_tot` is already at the `AtmosModel` level right?

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+@inline function atmos_nodal_update_aux!(
+    moist::NonEquilMoist,
+    atmos::AtmosModel,
+    state::Vars,
+    aux::Vars,
+    t::Real,
+)
+    TS = thermo_state(moist, atmos.orientation, state, aux)
+    aux.moisture.temperature = air_temperature(TS)
+    q_p_sat = PhasePartition_equil(
```
akshaysridhar: Is this correct ? If we impose non-equilibrium thermodynamics here, is the `PhasePartition_equil` the right wrapper to use ? I recall seeing `NonEquil` or similar wrappers in `MoistThermodynamics.jl` and `states.jl`

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    moist::NonEquilMoist,
+    atmos::AtmosModel,
+    flux::Grad,
+    state::Vars,
+    aux::Vars,
+    t::Real,
+)
+    ρ = state.ρ
+    u = state.ρu / ρ
+    #z = altitude(atmos.orientation, aux)
```
akshaysridhar:
```suggestion

```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    #usub = subsidence_velocity(atmos.subsidence, z)
```
akshaysridhar:
```suggestion

```

src/Atmos/Model/AtmosModel.jl
slifer50: Sure, I'll do that.

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    #ẑ = vertical_unit_vector(atmos.orientation, aux)
```
akshaysridhar:
```suggestion

```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,131 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    u_tot = u #.- usub * ẑ
```
akshaysridhar: We have a separate `subsidence` source now ; please us this here

src/Atmos/Model/moisture.jl
slifer50: This should be correct since we are comparing the current thermodynamic state with what the state should be at equilibrium in order to calculate how much vapor becomes part of the cloud phase.

akshaysridhar: I suppose the equil state is interpreted as a reference state here, as you mentioned

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I think the original reason `atmos` trailed is because `atmos.moisture` is used for dispatch

src/Atmos/Model/moisture.jl
trontrytel: Yes that is correct. The Equil calculation gives you the state to which you are relaxing to. But because of the assumed relaxation timescale you don't reach Equil state

src/Atmos/Model/moisture.jl
trontrytel: Maybe it's better to say something like: Assumes relaxation to equilibrium state with constant relaxation timescale?

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    TS = thermo_state(moist, atmos.orientation, state, aux)
+    aux.moisture.temperature = air_temperature(TS)
+    q_p_sat = PhasePartition_equil(
+        air_temperature(TS),
+        state.ρ,
+        state.moisture.ρq_tot,
+    )
+    aux.moisture.src_qliq = conv_q_vap_to_q_liq(q_p_sat, PhasePartition(TS))
+    aux.moisture.θ_v = virtual_pottemp(TS)
+    aux.moisture.q_liq = PhasePartition(TS).liq
```
trontrytel: I'm not sure about here though. Your new value of `q_liq` and `q_ice` should be coming from the `conv_q_vap_to_q_liq` source function and not from `PhasePartition`. Since its a `NonEquil` setup we can't diagnose it anymore. We have to calculate it based on state and sources

slifer50: You're right. But here aux.moisture.q_liq is not a state variable, it's likely that line 283 isn't really necessary but is nice when processing outputs. The update to q_liq is done through source.jl at : https://github.com/climate-machine/CLIMA/blob/18b64dc37507cbbb69577f8ac074a18ef01983e2/src/Atmos/Model/source.jl#L173

slifer50: essentially 281 calculates the source for the next time step.
283 sets aux.moisture.q_liq from the current time step which is nice to have as an output since the state variable is state.moisture.ρq_liq ( we can have these both in our output files and not have to divide by ρ in post) Though I understand how this might be confusing.

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
     flux.ρu += d_q_tot .* state.ρu'
     flux.moisture.ρq_tot += d_q_tot * state.ρ
 end
+"""
+  NonEquilMoist
+
+  Assumes that there isn't any thermodynamic equilibrium
```
tapios:
```suggestion
  Assumes local thermal equilibrium but phases may be out of equilibrium and relax toward equilibrium over a finite time.
```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+  
+  Adds addtional moisture variable for cloud water ρq_liq and cloud ice ρq_ice::FT that are transported in the same way as ρq_tot
```
tapios:
```suggestion
  Adds addtional moisture variables for cloud liquid water ρq_liq and cloud ice ρq_ice, which are transported in the same way as total water ρq_tot.
```

trontrytel:
@slifer50 - is there anything here that is still needed? Can we close this one?

The NonEquilMoisture model was added here: https://github.com/CliMA/ClimateMachine.jl/pull/1531
And here are the docs: https://github.com/CliMA/ClimateMachine.jl/pull/1574

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+
+  The equilibrium phase partition is calculated and used as a reference to compute the source term for the cloud phase.
+
+"""
+
+struct NonEquilMoist <: MoistureModel end
+
+vars_state(::NonEquilMoist, FT) = @vars(ρq_tot::FT, ρq_liq::FT, ρq_ice::FT)
```
charleskawczynski, line 256:
```suggestion
vars_state(::NonEquilMoist, ::Prognostic, FT) = @vars(ρq_tot::FT, ρq_liq::FT, ρq_ice::FT)
```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+vars_gradient(::NonEquilMoist, FT) =
```
charleskawczynski, line 257:
```suggestion
vars_state(::NonEquilMoist, ::Gradient, FT) =
```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    @vars(q_tot::FT, q_liq::FT, q_ice::FT, h_tot::FT)
+vars_diffusive(::NonEquilMoist, FT) = @vars(
```
charleskawczynski, line 259:
```suggestion
vars_state(::NonEquilMoist, ::GradientFlux, FT) = @vars(
```

src/Atmos/Model/moisture.jl
```diff
@@ -239,3 +240,128 @@ function flux_diffusive!(moist::EquilMoist, flux::Grad, state::Vars, d_q_tot)
...
+    ∇q_tot::SVector{3, FT},
+    ∇q_liq::SVector{3, FT},
+    ∇q_ice::SVector{3, FT}
+)
+vars_aux(::NonEquilMoist, FT) =
+    @vars(temperature::FT, θ_v::FT, q_liq::FT, q_ice::FT, p::FT, src_qliq::FT)
```
charleskawczynski, line 265:
```suggestion
vars_state(::NonEquilMoist, ::Auxiliary, FT) =
    @vars(src_qliq::FT)
```
We really shouldn't be adding more fields to the aux state.

PR #867, opened 2020-03-19
Fix git pre-commit hook for file removal

jkozdon:
# Description

The git pre-commit hook for the `JuliaFormatter` should ignore files that have been deleted.

### DIFF
--- a/.dev/hooks/pre-commit
+++ b/.dev/hooks/pre-commit
@@ -18,8 +18,10 @@ include(joinpath(toplevel_directory, ".dev", "clima_formatter_options.jl"))
 
 needs_format = false
 
-for filename in readlines(`git diff --name-only --cached`)
-    endswith(filename, ".jl") || continue
+for diffoutput in split.(readlines(`git diff --name-status --cached`))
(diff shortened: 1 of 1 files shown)


.dev/hooks/pre-commit
```diff
@@ -19,7 +19,7 @@ include(joinpath(toplevel_directory, ".dev", "clima_formatter_options.jl"))
 needs_format = false
 
 for filename in readlines(`git diff --name-only --cached`)
-    endswith(filename, ".jl") || continue
+    (isfile(filename) && endswith(filename, ".jl")) || continue
```
lcw: We may want to use `git diff --name-status --cached` to determine if a file is deleted or not.  Because a file can be removed from the index but still exist.

jkozdon: Good call. Will update

charleskawczynski: Thanks for fixing this!

lcw: Update looks good!

Issue #690, opened 2020-02-08
MPIStateArray revamp


simonbyrne:
Currently `MPIStateArray` is somewhere in between a "pure" array structure and a specialized backing datastore for values on a `DiscontinuousSpectralElementGrid`. It has a bunch of compromises and assumptions that makes it a bit cumbersome to use.

My current thinking is that we could define a `Field` object (perhaps there is a better name) that represents a set of numerical quantities that are defined over a spatial domain. In other words, this would be equivalent to a current `MPIStateArray` + `Grid` + `Vars`. 

- Move some "grid-aware" functionality out of MPIStateArray (e.g. `weights` and the weighted norm/dot) into the `Field` object
- Interpolation, integration, gradients, etc. could be defined directly on the `Field` object, and not require any external `DGModel` object
   - e.g. interpolation could return another `Field` backed by a different `Grid` type (e.g. `LatLongGrid` or `RegularCartesianGrid`).
- For splitting schemes (e.g. Stormer-Verlet #645), the DG tendencies only require an subset of input and output variables (which can be different). This could be handled by passing in Fields with a subset of `Vars` variables.
- We could overload `getproperty` on the whole `Field` which could give you a subset of the vars e.g. a scalar or vector field.

I haven't touched on issues like communication or ghost points, and I know others have opinions, so please chime in with your wishlist below.

blallen:
Need to think a bit more about it, but I love the idea of splitting out functionality into a `Field`. This is really the base object we are working with and `MPIStateArray` is just how the data for it is stored.

jkozdon:
@lcw and I have been thinking some about this over the past few months.

It seem that we have too many ideas in MPIStateArrays and need to split some out.

It would also be nice to be semi-agnostic about data layout, e.g., do we want state to be fastest, which degree of freedom first (i, j, k), element order, etc.

We've been thinking that a series of wrapper array types could do this for us. Something along the lines of the following

GhostedArray:
- Array that has a real view and a ghost view

StructArray / VarsArray
- Allows Q.a returns a view into Q
- Q.a should return another VarsArray
- Q[I,j] should return a struct layed out like vars specifies. If the Vars only has one field, it should return whatever the type of the vars field is
- This is what lead to the imperfect modification of MPIStateArrays in PR #731 
- I've played a bit around with this (29f32f02d51c9e4a6fc2f63d803b108afd53f385), but still needs a lot of work

WeightedArray
- Allows norm to be computed correctly with integration

[...]

PR #884, opened 2020-03-23
Allow passing MPIStateArrays to kernels

mwarusz:
# Description

This PR enables passing `MPIStateArray`s directly to GPU kernels by using `Adapt.jl`. This is convenient and cleans up a lot of code in `ODESolvers` that was creating views just for that purpose. 

I have

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -20,6 +20,13 @@ Base.similar(::Type{A}, ::Type{FT}, dims...) where {A <: Array, FT} =
     Base.similar(::Type{A}, ::Type{FT}, dims...) where {A <: CuArray, FT} =
         similar(CuArray{FT}, dims...)
 end
+@init @require CUDAnative = "be33ccc6-a3ff-5ff2-a52e-74243cff1e17" begin
+    using .CUDAnative
+    using Adapt
+    # allows passing MPIStateArrays to GPU kernels
+    Adapt.adapt_storage(to::CUDAnative.Adaptor, a::MPIStateArray) =
+        adapt(to, a.data)
+end
 
 include("CMBuffers.jl")
 using .CMBuffers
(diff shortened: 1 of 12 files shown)


mwarusz:
bors try

src/Arrays/MPIStateArrays.jl
```diff
@@ -20,6 +20,13 @@ Base.similar(::Type{A}, ::Type{FT}, dims...) where {A <: Array, FT} =
     Base.similar(::Type{A}, ::Type{FT}, dims...) where {A <: CuArray, FT} =
         similar(CuArray{FT}, dims...)
 end
+@init @require CUDAnative = "be33ccc6-a3ff-5ff2-a52e-74243cff1e17" begin
+    using .CUDAnative
+    using Adapt
+    # allows passing MPIStateArrays to GPU kernels
+    Adapt.adapt_storage(to::CUDAnative.Adaptor, a::MPIStateArray) =
+        adapt(to, a.data)
```
jkozdon: Should this really be the realdata or all the data?
```suggestion
        adapt(to, a.realdata)
```

mwarusz: I think that in general kernels may need access to all the data, e.g. `facerhs!` ?

jkozdon: OK. Just wanted to check.

mwarusz:
bors try

src/Arrays/MPIStateArrays.jl
vchuravy: So the goal is to unpack the MPIStateArray and not to provide a variant that works on the GPU?

I would have expected something like
```
Adapt.adapt_structure(to, x::MPIStateArray) = MPIStateArray(adapt(to, a.data))
```

Which we should probably still define to make it easier to copy a MPIStateArray to and from the GPU with `adapt(CuArray, MPIStateArray)`.

The way you are currently doing it will lead to CPU and GPU seeing different datastructures.

mwarusz: The goal is to be able to pass an `MPIStateArray` to a kernel, unpacking it was just the easiest way to achieve this. Of course, having a variant that works on the GPU would be ideal. However, since `MPIStateArray` contains not only data but a bunch of other stuff, most of which looks like it's not GPU-compatible (MPI communicator, `Arrays` of MPI requests, buffers, etc.) I don't think it is as easy as doing `MPIStateArray(adapt(to, a.data))`. In practice, we only use the indexing functionality inside kernels. In the long run it seems like #690 may address this.

mwarusz: I guess having an implicit conversion from the `MPIStateArray` to the underlying data array would also be a solution to the annoyances that inspired this PR. What this PR does is half of the implicit conversion, really.

vchuravy: Yeah we could define an KernelAbstraction `Adaptor` and use that to do the unpacking/transforming it into a `StateArray`

mwarusz:
Closing this PR since it got stale and is mostly about a convenience issue that probably would be better addressed by separating concerns in `MPIStateArrays` and making hypothetical `StateArrays` work on the GPU.

PR #898, opened 2020-03-27
Replace DataDeps with Artifacts in MoistThermo

charleskawczynski:
# Description

Replaces the use of DataDeps with Pkg.Artifacts in the MoistThermodynamics test suite for data used to test thermodynamic constructors.

I have


test/Common/MoistThermodynamics/artifacts_wrapper.jl
```diff
@@ -0,0 +1,91 @@
...
+            end
+        end
+
+        # Now bind that hash within our `Artifacts.toml`.  `force = true` means that if it already exists,
+        # just overwrite with the new content-hash.  Unless the source files change, we do not expect
+        # the content hash to change, so this should not cause unnecessary version control churn.
+        bind_artifact!(art_wrap.artifact_toml, art_wrap.data_name, data_hash, force = true)
+    end
+
+    # Get the path of the dycoms dataset, either newly created or previously generated.
```
kpamnany: This file looks pretty much completely general, except for the reference to 'dycoms' here. I'm wondering why it is in Moist Thermo tests -- presumably it will be useful in a lot of places? If this is likely to be used by CLIMA itself, perhaps this whole thing should be a module in `src/InputOutput/`?

charleskawczynski: Oh yeah, forgot to change that dycoms reference, thanks! And yes, I agree about potentially putting this in the src/ directory. I wasn't whether the best practice was to have a single Artifacts.toml or to have multiple. Either way, this can be moved to src, I'll do that.

src/CLIMA.jl
```diff
@@ -1,6 +1,7 @@
 module CLIMA
 
 include(joinpath("Utilities", "TicToc", "TicToc.jl"))
+include(joinpath("Utilities", "ArtifactsWrappers", "ArtifactsWrappers.jl"))
```
kpamnany: Only one niggling comment -- I understand why you used this name, but `ArtifactWrappers` is better English! ;)

charleskawczynski: Updated!

charleskawczynski:
@kpamnany okay to merge? (I'll first apply the formatter)

kpamnany:
Make it so!

simonbyrne:
There was a discussion on the Julia Slack about generating artifacts, and someone pointe dto this: 
https://github.com/JuliaGeo/Proj4.jl/blob/master/gen/create_artifact.jl

charleskawczynski:
Oh, that looks great. So, it hooks in the commit/versions? We should definitely do this.

PR #900, opened 2020-03-28
Update KernelAbstractions to 0.2

vchuravy:
bors try


vchuravy:
bors try

vchuravy:
Before:
```
CPU: Completed tests for ODESolvers, 3014 seconds elapsed
GPU: Completed tests for ODESolvers, 3498 seconds elapsed
```

After:
```
CPU: Completed tests for ODESolvers, 769 seconds elapsed
GPU: Completed tests for ODESolvers, 1495 seconds elapsed
```

bors r+

fxgiraldo:
> Before KA:
> 
> ```
> Completed tests for ODESolvers, 81 seconds elapsed
> Completed tests for ODESolvers, 1090 seconds elapsed
> ```
> 
> Before this PR:
> 
> ```
> CPU: Completed tests for ODESolvers, 3014 seconds elapsed
> GPU: Completed tests for ODESolvers, 3498 seconds elapsed
> ```
> 
> After:
> 
> ```
> CPU: Completed tests for ODESolvers, 769 seconds elapsed
> GPU: Completed tests for ODESolvers, 1495 seconds elapsed
> ```
> 
> The CPU test finish under an hour again and the GPU tests are at ~30minutes.
> 
> bors r+

Hi @vchuravy,  Just want to make sure I understand.  We are faster now but thought that KA would only improve CPU performance via multi-threading.  Also, what was the test you quoted with the "Before KA" heading?

thanks

vchuravy:
The specific test here is the ODESolver tests. It turns out what we tests there besides the correctness of our ODESolvers is the overhead of launching many kernels. When I asked how many kernels we are running I was quoted `O(100000)`. So the timings `Before KA` are the time it costs to launch many many kernels with GPUifyLoops on the CPU, since GPUifyLoops didn't attempt to handle multi-threading and used serial execution on the CPU its overhead on starting many kernels was very low. 

Before work by @mwarusz and me, KernelAbstractions had a launch overhead of about `10μs` now the launch overhead of ` 145ns`. To add an additional wrinkle KernelAbstractions uses asynchronous semantics for kernel launches (to homogenies between GPU and CPU semantics), so the typical use looks like:

```
using KernelAbstractions

@kernel function update_ka!(dQ, Q, rka, rkb, dt)
  i = @index(Global, Linear)
  @inbounds begin
    Q[i] += rkb * dt * dQ[i]
    dQ[i] *= rka
  end
end

dQ = zeros(1)
Q = rand(1)
rka = rand()
rkb = rand()
dt = rand()

kernel! = update_ka!(CPU(), 256, 1)
event = kernel!(dQ, Q, rka, rkb, dt) # Kernel launch
wait(event) # Wait on Kernel
```

This `wait(event)` costs about `1μs` in the ODESolvers test we always launch upon a kernel and them immediately wait upon it.
This is the reason why we are not back at 81s for the time duration of those tests. In general all of those tests are very small and they don't benefit from the parallelism
exposed by KernelAbstractions.

Hope that explanation helps!

fxgiraldo:
Thanks @vchuravy, that helps indeed!

PR #917, opened 2020-04-02
Remove TurbConv framework (which moved to SingleColumnModels.jl)

charleskawczynski:
# Description

Removes the stand-alone turbulence convection framework (and its tests), which was moved to SingleColumnModels.jl

This is possible now that CLIMAParameters and MoistThermodynamics are registered packages and we can share these parameters/functions between SingleColumnModels.jl and CLIMA. We'll still need to implement a form of what is being removed, but it will need to be implemented in the `BalanceLaw` framework. I think a draft of this was started in #555, but there were still missing functionality at the time. The only other thing to note is that the subdomain decomposition machinery (in `StateVecs/`) might be something we want to share between SingleColumnModels.jl and CLIMA. Perhaps we can split this off into a package, but that may require refactoring since I think it's a bit coupled to `StateVec`.

I have


kpamnany:
Need to also add SingleColumnModels.jl to Project.toml and some `using` statements?

charleskawczynski:
SingleColumnModels.jl is a stand-alone framework at the moment. It was originally added to stay in sync with MoistThermo, but since that's now a separate package, both CLIMA and SingleColumnModels.jl can use it.

PR #923, opened 2020-04-03
Add instructions for precompiling JuliaFormatter

jkozdon:
Add procedure for using the [`PackageCompier`](https://github.com/JuliaLang/PackageCompiler.jl) to speeds up the `JuliaFormatter`.

### DIFF
--- a/.dev/hooks/pre-commit.sysimage
+++ b/.dev/hooks/pre-commit.sysimage
@@ -1,0 +1,41 @@
+#!/usr/bin/env -S julia -J.git/hooks/JuliaFormatterSysimage.so
+#
+# Called by git-commit with no arguments.  This checks to make sure that all
+# .jl files are indented correctly before a commit is made.
+#
+# To enable this hook, make this file executable and copy it in
+# $GIT_DIR/hooks.
+
+toplevel_directory = chomp(read(`git rev-parse --show-toplevel`, String))
+
+using Pkg
+Pkg.activate(joinpath(toplevel_directory, ".dev"))
+Pkg.instantiate()
+
+using JuliaFormatter
+
+include(joinpath(toplevel_directory, ".dev", "clima_formatter_options.jl"))
+
+needs_format = false
+
+for diffoutput in split.(readlines(`git diff --name-status --cached`))
+    status = diffoutput[1]
+    filename = diffoutput[end]
+    (!startswith(status, "D") && endswith(filename, ".jl")) || continue
+
+    a = read(`git show :$filename`, String)
+    b = format_text(a; clima_formatter_options...)
+
+    if a != b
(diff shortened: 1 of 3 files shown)


.dev/hooks/pre-commit.sysimage
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env julia -J.git/hooks/JuliaFormatterSysimage.so
```
lcw, line 1:
```suggestion
#!/usr/bin/env -S julia -J.git/hooks/JuliaFormatterSysimage.so
```

The `-S` flag should be used when passing arguments.

.dev/hooks/pre-commit.sysimage
```diff
@@ -0,0 +1,41 @@
...
+    (!startswith(status, "D") && endswith(filename, ".jl")) || continue
+
+    a = read(`git show :$filename`, String)
+    b = format_text(a; clima_formatter_options...)
+
+    if a != b
+        fullfilename = joinpath(toplevel_directory, filename)
+
+        @error """File $filename needs to be indented with:
+            julia -J$(joinpath(toplevel_directory, ".git/hooks", "JuliaFormatterSysimage.so")) $fullfilename
```
lcw, line 33:
```suggestion
            julia -J$(joinpath(toplevel_directory, ".git/hooks", "JuliaFormatterSysimage.so")) $(joinpath(toplevel_directory, ".dev", "climaformat.jl"))  $fullfilename
```

.dev/hooks/pre-commit.sysimage
jkozdon: Thanks.

.dev/hooks/pre-commit.sysimage
jkozdon: Thanks.

jkozdon:
@simonbyrne @lcw @vchuravy: Do you understand this error in the [CI](https://gist.github.com/climabot/45fd52357cf0f13835ba33aaf2623a0c#file-out_9616081-L1113-L1120)? Googling suggests its something related to cuda aware MPI(?).

```
--------------------------------------------------------------------------
The call to cuIpcCloseMemHandle failed. This is a warning and the program
will continue to run.
  cuIpcCloseMemHandle return value:   1
  address: 0x2aac22fd0000
Check the cuda.h file for what the return value means. Perhaps a reboot
of the node will clear the problem.
--------------------------------------------------------------------------
```

lcw:
I don't understand exactly what is going on but https://github.com/horovod/horovod/issues/82 suggests a workaround of passing `--mca btl_smcuda_use_cuda_ipc 0` to `mpirun` which disables using cuda-aware via PCI.  This will be okay if one is using cuda-aware MPI via NVlink.  Not sure what the CI cluster nodes are using.

PR #934, opened 2020-04-06
Fix CoastlineFreeSlip boundary conditions for HydrostaticBoussinesqModel

blallen:
# Description

The current implementation didn't properly zero out the normal component of velocity. Updated these to follow equations 1.26 and 1.28 from the numerics section of the design docs. 

I have

### DIFF
--- a/experiments/OceanBoxGCM/ocean_gyre.jl
+++ b/experiments/OceanBoxGCM/ocean_gyre.jl
@@ -1,32 +1,34 @@
 using Test
 using CLIMA
-using CLIMA.HydrostaticBoussinesq
 using CLIMA.GenericCallbacks
 using CLIMA.ODESolvers
 using CLIMA.Mesh.Filters
-
 using CLIMA.VariableTemplates
 using CLIMA.Mesh.Grids: polynomialorder
+using CLIMA.HydrostaticBoussinesq
 
 using CLIMAParameters
 using CLIMAParameters.Planet: grav
 struct EarthParameterSet <: AbstractEarthParameterSet end
 const param_set = EarthParameterSet()
 
-function config_simple_box(FT, N, resolution, dimensions)
-    prob = OceanGyre{FT}(dimensions...)
+function config_simple_box(FT, N, resolution, dimensions; BC = nothing)
+    if BC == nothing
+        problem = OceanGyre{FT}(dimensions...)
+    else
+        problem = OceanGyre{FT}(dimensions...; BC = BC)
+    end
 
     _grav::FT = grav(param_set)
-
-    cʰ = sqrt(_grav * prob.H) # m/s
-    model = HydrostaticBoussinesqModel{FT}(param_set, prob, cʰ = cʰ)
+    cʰ = sqrt(_grav * problem.H) # m/s
+    model = HydrostaticBoussinesqModel{FT}(param_set, problem, cʰ = cʰ)
 
     config = CLIMA.OceanBoxGCMConfiguration("ocean_gyre", N, resolution, model)
 
     return config
 end
 
-function run_ocean_gyre(; imex::Bool = false)
+function run_ocean_gyre(; imex::Bool = false, BC = nothing)
     CLIMA.init()
 
     FT = Float64
@@ -58,7 +60,7 @@ function run_ocean_gyre(; imex::Bool = false)
             CLIMA.ExplicitSolverType(solver_method = LSRK144NiegemannDiehlBusch)
     end
 
-    driver_config = config_simple_box(FT, N, resolution, dimensions)
+    driver_config = config_simple_box(FT, N, resolution, dimensions; BC = BC)
 
     grid = driver_config.grid
     vert_filter = CutoffFilter(grid, polynomialorder(grid) - 1)
@@ -71,7 +73,7 @@ function run_ocean_gyre(; imex::Bool = false)
         driver_config,
         init_on_cpu = true,
         # ode_dt = dt,
-        Courant_number = 0.25,
+        Courant_number = 0.4,
         ode_solver_type = solver_type,
         modeldata = modeldata,
     )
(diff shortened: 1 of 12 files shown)


src/Ocean/HydrostaticBoussinesq/OceanBoundaryConditions.jl
```diff
@@ -34,10 +33,39 @@ nothing needed to do since these are neumann BCs and no gradients here
...
 @inline function ocean_boundary_state!(::HBModel, ::CoastlineFreeSlip,
-                                       ::Union{Rusanov,
-                                               CentralNumericalFluxGradient},
+                                       ::Rusanov,
                                        Q⁺, A⁺, n⁻, Q⁻, A⁻, t)
-  return nothing
+    u⁻ = Q⁻.u
+    n = @SVector [n⁻[1], n⁻[2]]
+    
+    Q⁺.u = u⁻ - 2 * (n∘u⁻) * n
```
jkozdon, line 41: Do you really want to use `∘` here for dot product? `∘` is a julia function for function composition.

blallen: Ah, I was using it instead of · (\cdot) because it's easier to read. Didn't realize it already had a syntactical meaning. I can switch back to using \circ.

blallen:
> This is what I would expect for zeroing out the normal velocity.
> 
> Can you put in a test?

The only real test we have right now would be running `test_divergence_free.jl` with FreeSlip instead of NoSlip boundary conditions. That's a good first check, but I'm not sure if it will give the level of testing needed. Any thoughts @jm-c ?

jm-c:
The ShallowWater problem with either the Stommel or the Munk solution
(depending on FreeSlip/NoSlip BC on the side and on the bottom) should be 
a nice test, at least for Coastline BC.

blallen:
> The ShallowWater problem with either the Stommel or the Munk solution (depending on FreeSlip/NoSlip BC on the side and on the bottom) should be a nice test, at least for Coastline BC.

Can we do this in 3D? That's the only way to test the BC for this model.

Also, I just pushed a bunch of preliminary work to enable easily swapping between FreeSlip and NoSlip

jm-c:
@blallen  OK, I see your point regarding ShallowWater problem.

blallen:
> Looks good to me. The Rusanov CoastlineFreeSlip BC is fixed for good (similar to sLinearDrag, in shallow_boundary_state).
> Regarding the CentralNumericalFluxGradient BC, it make sense to me but would be nice if @sandreza
> could also agree.
> And I re-run ocean_gyre.jl after switching to CoastlineFreeSlip and it looks reasonable.

For these tests, did you run with OceanFloorFreeSlip or OceanFloorNoSlip? I want to know so I can replicate the setup in the test files.

jm-c:
I've just run keeping OceanFloorNoSlip.
This is different from what I run with MITgcm (to produce the little movie I sent the other day) 
which has FreeSlip @ bottom but noSlip on the coast.

blallen:
> I've just run keeping OceanFloorNoSlip. This is different from what I run with MITgcm (to produce the little movie I sent the other day) which has FreeSlip @ bottom but noSlip on the coast.

Okay, I'm setting up the Azure tests to run just switching Coastline between FreeSlip and NoSlip, but will test all permutations of the boundary conditions on slurm CI.

experiments/OceanBoxGCM/ocean_gyre.jl
```diff
@@ -88,4 +90,51 @@ function run_ocean_gyre(; imex::Bool = false)
...
+            CLIMA.HydrostaticBoussinesq.CoastlineFreeSlip(),
+            CLIMA.HydrostaticBoussinesq.OceanFloorFreeSlip(),
+            CLIMA.HydrostaticBoussinesq.OceanSurfaceNoStressForcing(),
+        ),
+    ]
+
+    for BC in boundary_conditions
+        run_ocean_gyre(imex = false, BC = BC)
+    end
+end
```
blallen, line 140: @simonbyrne @kpamnany @charleskawczynski , do you guys have any ideas on how to easily switch between running all of these options for CI and only running a single setup interactively? 

Currently, this is the only place we can test that all the different boundary conditions are working... besides `test_ocean_gyre.jl` but I fear that running all of these on Azure would take too long and that the integration time is too short to really test the physics anyways

charleskawczynski: Hm, I would suggest testing BCs with a coarse mesh and for a single timestep (assuming they're time-independent BCs)--if it's BCs that you're testing. At first glance, this looks like an entire simulation, in which case I would suggest breaking each case out into individual tests/experiments and running them in parallel on slurm. That said, we definitely need some sort of unit tests for BCs in general.

blallen: But the boundary condition can run for a single time step successfully but not produce correct results if run for a longer length of time.  We need some way to catch these errors that are small per timestep but gradually accumulate.

blallen:
bors try

experiments/OceanBoxGCM/ocean_gyre.jl
charleskawczynski: If the errors depend on the initial state, maybe we should supply an artificial initial state to check the BCs. Or apply a divergence-free random IC. So long the BCs are not time-dependent, a single step/BC application is all that should be needed to test the BCs.

kpamnany: It'd be good to have two tests -- a basic and quick test that provides some confidence that things are working, and a detailed test that assures that things are working. The first will go in merge tests, the second in nightly (which is not up yet).

kpamnany:
Do all the commits pass CI or can you squash?

blallen:
Each commit should pass CI individually. 

aa4f07d7a327307f5417866ee5c7a11bbf15e856, e7109e68a00e735f8498a2e3625f383354e553dc, 439fb1ca0cf838932eca53234b9bfaafb2621c0b aren't functionally; they just rename things or move them around

5dcc80a5ce11b4688b3d8315159b3b029381e975, 2008614eb92b07eb9ee45d93443e9b874d5bde17, 43f0ade96d10b15459c2ca66dd8c2693c8602344 are all atomic changes towards the end goal:
- 5dcc80a5ce11b4688b3d8315159b3b029381e975 fixes the BC
- 2008614eb92b07eb9ee45d93443e9b874d5bde17 enables easily switching BCs but doesn't switch in tests
- 43f0ade96d10b15459c2ca66dd8c2693c8602344 actually tests multiple BCs in one test

blallen:
Welp, didn't see that bors had started and decided to squash after all, here we go again

charleskawczynski:
LGTM!

blallen:
Bors r+

PR #936, opened 2020-04-06
Move coverage CI  to GitHub Actions

charleskawczynski:
# Description

This PR:
 - [x] Adds a coverage GitHub Actions CI (GitHub hosted)
 - [x] Removes coverage from the azure pipeline YML

I have

### DIFF
--- a/.github/workflows/Coverage.yaml
+++ b/.github/workflows/Coverage.yaml
@@ -1,0 +1,52 @@
+name: Coverage
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # Run at 2am every day:
+    - cron:  '0 2 * * *'
+
+jobs:
+  coverage:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ['1.3']
+        os: [ubuntu-latest]
+        project: ['CLIMA']
+
+    steps:
+    - uses: actions/checkout@v1.0.0
+    - name: "Set up Julia"
+      uses: julia-actions/setup-julia@v1
+      with:
+        version: ${{ matrix.julia-version }}
+    - name: Install deps
+      run: |
+        set -o xtrace
+        sudo apt-get update
+        sudo apt-get install mpich libmpich-dev
+    - name: Test with coverage
+      env:
+        JULIA_PROJECT: "@."
+      run: |
+        julia --project=@. -e 'using Pkg;
+                               Pkg.instantiate()'
+        julia --project=@. -e 'using Pkg;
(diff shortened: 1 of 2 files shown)


kpamnany:
Where does the coverage information get posted, if it's not to a PR thread?

vchuravy:
> Where does the coverage information get posted, if it's not to a PR thread?

https://codecov.io/gh/climate-machine/CLIMA

One thing we could do is to actually run it on PRs but to not gate bors on it.

charleskawczynski:
> One thing we could do is to actually run it on PRs but to not gate bors on it.

True, but the CI will still get backlogged by coverage runs unless we trigger them under a different azure organization (which we could do). The other problem with that is that triggering CI with GHA requires [Personal Access Tokens](https://github.com/Azure/pipelines#sample-workflow) which seem to have a maximum [90-day lifetime](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page).

vchuravy:
> True, but the CI will still get backlogged by coverage runs unless we trigger them under a different azure organization (which we could do). 

That should be alleviated by running on Github actions?

> The other problem with that is that triggering CI with GHA requires Personal Access Tokens which seem to have a maximum 90-day lifetime.

Why do we need a PAT? It should be triggered from the PR?

charleskawczynski:
> Why do we need a PAT? It should be triggered from the PR?

Ah, sorry I didn't specify, that's if we want the CI to still run on Azure. We could run coverage no problem with GitHub hosted (by cron or PR)

> That should be alleviated by running on Github actions?

Yes, if we use GitHub hosted. No if we use Azure hosted

charleskawczynski:
Should we just make coverage run on GH resources every PR? That'd be pretty simple I think.

charleskawczynski:
Oh, the only potential problem with that is if we're timing out. But I guess we're moving more things out of `runtests.jl` anyway, so I think this is a reasonable long term solution.

charleskawczynski:
I'll update the cron jobs to be on each PR, since what I have right now is using GH hosted anyway.

charleskawczynski:
@kpamnany @vchuravy looks like this is running, feel free to merge if it looks okay! Hopefully, this will speed up Azure since Azure will no longer be running with coverage!

simonbyrne:
I think a nightly cron on master is fine for now.

vchuravy:
> Hmm, looks like something broke with KernelAbstractions, cc: @vchuravy

Seems like you are running with Julia 1.4, best to use
https://github.com/julia-actions/setup-julia to choose the right version of Julia.

PR #945, opened 2020-04-08
Change order of user callbacks 

akshaysridhar:
# Description

Callbacks are done sequentially so any user-defined callbacks must appear before diagnostic callbacks. Currently this affects filter application.

I have

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -187,7 +187,8 @@ function init(; disable_gpu = false)
 
     # set up the array type appropriately depending on whether we're using GPUs
     if !Settings.disable_gpu &&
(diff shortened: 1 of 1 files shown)


akshaysridhar:
Are these even evaluated in order of args ?

akshaysridhar:
This was meant to go into an issue rather than a PR

charleskawczynski:
I definitely agree with this PR, I think diagnostics should occur last in the callback list, the diagnostics reporting should use the state of the equations after a full step (including applied filters). That said, it might be interesting, for debugging/interpreting purposes, to have diagnostic callbacks before and after the filter is applied.

PR #953, opened 2020-04-09
Move documentation to GitHub Actions, enable preview builds

simonbyrne:
Should make it easier to edit and review docs changes.


simonbyrne:
Now available at https://climate-machine.github.io/CLIMA/previews/PR953/

charleskawczynski:
Hm, building on every push might quickly flood our GitHub resources. Maybe this is fine for now, but it might be nice if we migrate to a [comment-trigger](https://github.community/t5/GitHub-Actions/Accessing-commit-message-in-pull-request-event/td-p/40856). Ideally, users could comment `[build-docs]`, or something, to explicitly trigger the doc preview build since not every push will have documentation changes.

simonbyrne:
Hmm, it seems to be building on every pull action (comments, etc), so need to figure this out.

.github/workflows/Documenter.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1.3
+      - name: Install MPI
+        run: |
+          sudo apt-get update
+          sudo apt-get install mpich libmpich-dev
+      - name: Install dependencies
+        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```
fredrikekre, line 23:
```suggestion
        run: julia --project=docs/ -e 'using Pkg; Pkg.instantiate()'
```

fredrikekre: Since you commit the manifest you can pre-`dev` the `CLIMA` package as a relative path.

simonbyrne: ah, figured that out.

charleskawczynski:
> Hmm, it seems to be building on every pull action (comments, etc), so need to figure this out.

I didn't notice that because I only tested with one push. Glad you caught that.

simonbyrne:
Actually maybe not. I’m not quite sure what events trigger it

fredrikekre:
Seems to work, thanks for testing.

kpamnany:
These last few are failing because of ECC errors on hpc-22-34, not just the close handle failure. No point in retrying -- that node has to be taken out of service. I suggest opening a ticket with IMSS.

charleskawczynski:
Ok, opened a ticket with IMSS.

charleskawczynski:
Thanks for pointing that out @kpamnany

PR #951, opened 2020-04-09
Add back `EveryXSimulationTime` callback mechanism

kpamnany:
# Description

Was not enabled. Closes #878. Also adds tests for `GenericCallbacks`.

I have

- [X] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [X] Followed all necessary [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/format.jl`

# For review by CLIMA developers

- [X] There are no open pull requests for this already
- [X] CLIMA developers with relevant expertise have been assigned to review this submission
- [X] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [X] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

Issue #909, opened 2020-04-01
Support `--checkpoint-simtime`


kpamnany:
# Description

Remaining features from #850:
- Add support for `--checkpoint-simtime` in addition to `--checkpoint-walltime`.
- Write checkpoints in NetCDF.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

PR #958, opened 2020-04-10
Rework command line arguments

### LINKED ISSUE #877: Command line arguments for experiments
# Description

An experiment should be able to get experiment-specific command line arguments.

# For CLIMA Developers

kpamnany:
# Description

Closes #877 and #909.

Most of the command line arguments relate to whether or not something should occur, and then specifying the interval at which it occurs. These intervals are (somewhat randomly) either in wall-clock time or simulation steps.

After #951, specifying simulation time is now also possible. So, rather than have, for instance, `--checkpoint-walltime` and `--checkpoint-simtime` and `--checkpoint-simsteps`, this allows intervals to be specified as:
- `2hours` or `10mins` or `30secs` => wall-clock time,
- `9.5smonths` or `3.3sdays` or `1.5shours` => simulation time,
- `1000steps` => simulation steps,
- `never` => disable,
- `default` => use any pre-defined interval (only present for diagnostics groups at the moment).

And now there is just `--checkpoint <interval>`. All command line arguments are reworked to conform to this.

This also adds the ability for experiments to have their own command line arguments. To do this, create an [`ArgParse.ArgParseSettings`](https://carlobaldassi.github.io/ArgParse.jl/stable/settings/) that defines an [argument group](https://carlobaldassi.github.io/ArgParse.jl/stable/arg_table/#Argument-groups-1) with an argument table containing the experiment's argument definitions. Then:
```
parsed_args = CLIMA.init(arg_settings = experiment_arg_settings)
```
And `parsed_args` is a `Dict` keyed by the arguments in `experiment_arg_settings`. An example will be added in a later PR.

I have

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -1,8 +1,6 @@
 using ArgParse
 using CUDAapi
 using Dates
-using FileIO
-using JLD2
 using LinearAlgebra
 using Logging
 using MPI
@@ -14,12 +12,10 @@ using ..Atmos
 using ..Callbacks
 using ..ColumnwiseLUSolver
 using ..ConfigTypes
-using ..Courant
 using ..Diagnostics
 using ..DGmethods
-using ..DGmethods: vars_state, vars_aux, update_aux!
+using ..DGmethods: update_aux!
 using ..DGmethods.NumericalFluxes
-using ..GenericCallbacks
 using ..HydrostaticBoussinesq
 using ..Mesh.Grids
 using ..Mesh.Topologies
@@ -28,53 +24,11 @@ using ..MPIStateArrays
 using ..ODESolvers
 using ..TicToc
 using ..VariableTemplates
-using ..VTK
-
-# Note that the initial values specified here are overwritten by the
-# command line argument defaults in `parse_commandline()`.
-Base.@kwdef mutable struct CLIMA_Settings
-    disable_gpu::Bool = false
-    show_updates::Bool = true
-    update_interval::Int = 60
-    enable_diagnostics::Bool = false
-    diagnostics_interval::Int = 10000
-    enable_vtk::Bool = false
-    vtk_interval::Int = 10000
-    enable_wall_clock::Bool = false
-    wall_clock_interval::Int = 10000
-    monitor_courant_numbers::Bool = false
-    monitor_courant_interval::Int = 10
-    checkpoint_walltime::Int = -1
-    checkpoint_keep_one::Bool = true
-    checkpoint_at_end::Bool = false
-    checkpoint_dir::String = "checkpoint"
-    restart_from_num::Int = -1
-    log_level::String = "INFO"
-    output_dir::String = "output"
-    integration_testing::Bool = false
-    array_type
-end
-
-const Settings = CLIMA_Settings(array_type = Array)
-
-array_type() = Settings.array_type
-
-const cuarray_pkgid =
-    Base.PkgId(Base.UUID("3a865a2d-5b23-5a0f-bc46-62713ec82fae"), "CuArrays")
-
-@init begin
-    if get(ENV, "CLIMA_GPU", "") != "false" && CUDAapi.has_cuda_gpu()
-        CuArrays = Base.require(cuarray_pkgid)
-    end
-end
-
-function _init_array(::Type{Array})
-    return nothing
-end
 
 @init @require CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
     using .CuArrays, .CuArrays.CUDAdrv, .CuArrays.CUDAnative
-    function _init_array(::Type{CuArray})
+
+    @eval function _init_array(::Type{CuArray})
         comm = MPI.COMM_WORLD
         # allocate GPUs among MPI ranks
         local_comm = MPI.Comm_split_type(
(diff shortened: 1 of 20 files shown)


kpamnany:
@blallen: I haven't tested this argument parsing capability for experiments.

@lcw: I've followed your example of moving the callback implementations into the new `Callbacks` module. I've also taken the liberty of renaming your `wall_clock_time_per_time_step` callback to `monitor_timestep_duration` for consistency with `monitor_courant_numbers`. If you don't like this, I can change it back.

tapios:
Thanks, @kpamnany. This looks very useful. We will need simulation output every x simulated hours, or y simulated months. Instead of just having `ticks`, could we have `shours`, `stays`, and `smonths` as time units for simulated time? (You can assume everything in the model is in SI units, hence time is in seconds.)

src/Driver/Driver.jl
```diff
@@ -89,64 +43,76 @@ end
...
         Base.loaded_modules[CLIMA.cuarray_pkgid].allowscalar(val)
     end
     return nothing
 end
 
+# Note that the initial values specified here are overwritten by the
+# command line argument defaults in `parse_commandline()`.
+Base.@kwdef mutable struct CLIMA_Settings
+    disable_gpu::Bool = false
+    show_updates::String = "60secs"
```
charleskawczynski, line 61: I'm not really a fan of having the number and units combined into one string (now we have to parse this and remember the format).

src/Driver/Driver.jl
```diff
@@ -89,64 +43,76 @@ end
...
 function gpu_allowscalar(val)
     if haskey(Base.loaded_modules, CLIMA.cuarray_pkgid)
         Base.loaded_modules[CLIMA.cuarray_pkgid].allowscalar(val)
     end
     return nothing
 end
 
+# Note that the initial values specified here are overwritten by the
+# command line argument defaults in `parse_commandline()`.
+Base.@kwdef mutable struct CLIMA_Settings
```
charleskawczynski, line 59: Can we add docs / comments to this struct and include valid string options?

charleskawczynski: Oh, maybe just point to options in `parse_commandline`

src/Driver/Driver.jl
kpamnany: I see your point. But command line arguments are tricky. `ArgParse.jl` cannot, I think, handle multiple arguments (e.g. `--checkpoint 60 secs`) so you would have to pass `--checkpoint "60 secs"` which is a bit ugly to me.

charleskawczynski: What about
`show_update_frequency::Int`
`show_update_units::String`
?

kpamnany: So, avoiding the need to have two or more command line arguments for each option is one of the reasons for this PR -- it is already pretty busy. I think if we have a consistent definition for these intervals it should be fine. Let's discuss further at the next meeting.

tapios:
> > Thanks, @kpamnany. This looks very useful. We will need simulation output every x simulated hours, or y simulated months. Instead of just having `ticks`, could we have `shours`, `stays`, and `smonths` as time units for simulated time? (You can assume everything in the model is in SI units, hence time is in seconds.)
> 
> To clarify, @tapios, this will mean: `2smonths` is `2 * 30.0 * 24.0 * 60.0 * 60.0` simulated seconds, `7sdays` is `7 * 24.0 * 60.0 * 60.0` simulated seconds and `4shours` is `4 * 60.0 * 60.0`. Correct?

Yes, that's basically good for now. However, please use the length of day parameter from planet parameters, in case anyone wants to change what we mean by a day. At some point, we'll also need to have a calendar, and then a month is not a fixed time period. But we can do that later.

src/Driver/Driver.jl
kpamnany: @simonbyrne: any thoughts on this?

src/Utilities/Callbacks/Callbacks.jl
```diff
@@ -57,7 +225,182 @@ function wall_clock_time_per_time_step(interval, array_type, comm)
...
+            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
+    elseif endswith(interval, "mins")
+        secs = 60 * parse(Int, interval[1:(end - 4)])
+        return (func) ->
+            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
+    elseif endswith(interval, "secs")
+        secs = parse(Int, interval[1:(end - 4)])
+        return (func) ->
+            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
+    elseif endswith(interval, "smonths")
```
simonbyrne, line 370: what's an `smonth`?

src/Utilities/Callbacks/Callbacks.jl
```diff
@@ -57,7 +225,182 @@ function wall_clock_time_per_time_step(interval, array_type, comm)
...
+constructor. If `interval` is "default", then use `default` as the interval.
+"""
+function CB_constructor(interval::String, solver_config, default = nothing)
+    mpicomm = solver_config.mpicomm
+    solver = solver_config.solver
+    dg = solver_config.dg
+    bl = dg.balancelaw
+    secs_per_day = day(bl.param_set)
+
+    if endswith(interval, "hours")
```
simonbyrne, line 358: Would it be better to use regex here, e.g.
```julia
    if (m = match(r"^([0-9]+)(hours|mins|secs|smonths|sdays|shours|steps)$",interval); m !== nothing)
        ninterval = parse(Int, m[1])
        if m[2] == "hours"
            ...
 

```

src/Utilities/Callbacks/Callbacks.jl
simonbyrne: oh, simulated. I'd prefer `simmonths`

src/Driver/Driver.jl
simonbyrne: I agree with @kpamnany that it would be good to minimise the number of arguments.

src/Utilities/Callbacks/Callbacks.jl
kpamnany: Sim time is Float64 so it would be a bit more complicated. Not sure it's worth it.

PR #961, opened 2020-04-13
Update function and variable names in DGmethods

blallen:
# Description

This PR updates the naming in `src/DGmethods` to improve clarity and match the numerics design doc.There should be no functional changes from this code, although function names in other areas of the code have been changed for consistency.  

Julia formatted has NOT been applied yet. I will do that once everything is finalized and I squash the commits.

Names related to hyperdiffusion internals and the specifics of the grid have not been touched in this PR. I will address those with separate PRs after this one, as leaving invasive PRs like this one open too long result in many headaches with merge conflicts.

I have

### DIFF
--- a/src/Atmos/Model/hyperdiffusion.jl
+++ b/src/Atmos/Model/hyperdiffusion.jl
@@ -4,11 +4,11 @@ using LinearAlgebra
 export HyperDiffusion, NoHyperDiffusion, StandardHyperDiffusion
 
 abstract type HyperDiffusion end
-vars_state(::HyperDiffusion, FT) = @vars()
-vars_aux(::HyperDiffusion, FT) = @vars()
-vars_gradient(::HyperDiffusion, FT) = @vars()
+vars_state_conservative(::HyperDiffusion, FT) = @vars()
+vars_state_auxiliary(::HyperDiffusion, FT) = @vars()
+vars_state_gradient(::HyperDiffusion, FT) = @vars()
 vars_gradient_laplacian(::HyperDiffusion, FT) = @vars()
-vars_diffusive(::HyperDiffusion, FT) = @vars()
+vars_state_gradient_flux(::HyperDiffusion, FT) = @vars()
 vars_hyperdiffusive(::HyperDiffusion, FT) = @vars()
 function atmos_init_aux!(
     ::HyperDiffusion,
@@ -16,21 +16,21 @@ function atmos_init_aux!(
     aux::Vars,
     geom::LocalGeometry,
 ) end
-function atmos_nodal_update_aux!(
+function atmos_nodal_update_auxiliary_state!(
     ::HyperDiffusion,
     ::AtmosModel,
     state::Vars,
     aux::Vars,
     t::Real,
 ) end
-function gradvariables!(
+function compute_gradient_argument!(
     ::HyperDiffusion,
     transform::Vars,
     state::Vars,
     aux::Vars,
     t::Real,
 ) end
-function hyperdiffusive!(
+function transform_post_gradient_laplacian!(
     h::HyperDiffusion,
     hyperdiffusive::Vars,
     gradvars::Grad,
@@ -38,7 +38,7 @@ function hyperdiffusive!(
     aux::Vars,
     t::Real,
 ) end
-function flux_diffusive!(
+function flux_second_order!(
     h::HyperDiffusion,
     flux::Grad,
     state::Vars,
@@ -47,7 +47,7 @@ function flux_diffusive!(
     aux::Vars,
     t::Real,
 ) end
-function diffusive!(
+function compute_gradient_flux!(
     h::HyperDiffusion,
     diffusive::Vars,
     ∇transform::Grad,
@@ -58,22 +58,22 @@ function diffusive!(
 
 """
   NoHyperDiffusion <: HyperDiffusion
-Defines a default hyperdiffusion model with zero diffusive fluxes. 
+Defines a default hyperdiffusion model with zero diffusive fluxes.
 """
 struct NoHyperDiffusion <: HyperDiffusion end
 
 """
   StandardHyperDiffusion{FT} <: HyperDiffusion
 Horizontal hyperdiffusion methods for application in GCM and LES settings
-Timescales are prescribed by the user while the diffusion coefficient is 
+Timescales are prescribed by the user while the diffusion coefficient is
 computed as a function of the grid lengthscale.
 """
 struct StandardHyperDiffusion{FT} <: HyperDiffusion
     τ_timescale::FT
 end
 
-vars_aux(::StandardHyperDiffusion, FT) = @vars(Δ::FT)
-vars_gradient(::StandardHyperDiffusion, FT) =
+vars_state_auxiliary(::StandardHyperDiffusion, FT) = @vars(Δ::FT)
+vars_state_gradient(::StandardHyperDiffusion, FT) =
     @vars(u::SVector{3, FT}, h_tot::FT)
 vars_gradient_laplacian(::StandardHyperDiffusion, FT) =
     @vars(u::SVector{3, FT}, h_tot::FT)
@@ -89,7 +89,7 @@ function atmos_init_aux!(
     aux.hyperdiffusion.Δ = lengthscale(geom)
 end
 
-function gradvariables!(
+function compute_gradient_argument!(
     h::StandardHyperDiffusion,
     transform::Vars,
     state::Vars,
@@ -101,7 +101,7 @@ function gradvariables!(
     transform.hyperdiffusion.h_tot = transform.h_tot
 end
 
-function hyperdiffusive!(
+function transform_post_gradient_laplacian!(
     h::StandardHyperDiffusion,
     hyperdiffusive::Vars,
     hypertransform::Grad,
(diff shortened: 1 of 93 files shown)


blallen:
Ouch, apparently a lot has changed since just Thursday. I will rebase on top of master and start fixing these conflicts.

blallen:
bors try

blallen:
bors cancel

src/Atmos/Model/hyperdiffusion.jl
```diff
@@ -58,25 +58,25 @@ function diffusive!(
...
 end
 
-vars_aux(::StandardHyperDiffusion, FT) = @vars(Δ::FT)
-vars_gradient(::StandardHyperDiffusion, FT) =
+vars_state_auxiliary(::StandardHyperDiffusion, FT) = @vars(Δ::FT)
+vars_state_transform(::StandardHyperDiffusion, FT) =
     @vars(u::SVector{3, FT}, h_tot::FT)
 vars_gradient_laplacian(::StandardHyperDiffusion, FT) =
-    @vars(u::SVector{3, FT}, h_tot::FT)
+    @vars(Δu::SVector{3, FT}, Δh_tot::FT)
```
charleskawczynski, line 79: Is this to compute gradients of the Laplacian?

blallen: yeah, the input to the hyperdiffusive variables is `∇Δstate`. 

The notation here is a little confusing because this variable is both where the Laplacian is stored and where the gradient is stored after taking the Laplacian. Same thing with the old`vars_gradient`/`vars_transform`. It initially holds the thing you want to take the derivative of, filled in `gradvariables!`/`transform_pre_gradient!`, and then holds the derivative of it when being called in `diffusive!`/`transform_post_gradient!`. 
Because of this ambiguity/dual role, it is a bit hard to come up with a good name for these things, and it's only worse in the case of hyperdiffusion, as you have even more reuse of variables for storage purposes. 

For second-order/diffusive terms, I've decided to go with calling the thing before the gradient the transform state and the variables actually called in `flux_second_order!` the gradient state. Ideally, we will come up with some similar notation for hyperdiffusion... but that's a separate PR.

blallen:
bors try

blallen:
> * `vars_state_conservative` all state variables are conserved, so think I'd prefer `vars_state` 🤔

I mostly did this for easy sed'ing and grep'ing. And while all state variables are conserved, we also call things the "auxiliary state" or "gradient/diffusive state" where the variables aren't necessarily conserved. I'm open for suggestions on this as "conservative state" is a mouthful.

> * `vars_state_auxiliary` while "auxiliary" is descriptive, I think it's a bit wordy 🤔.

Same thing here. Technically some of these aren't even state variables, so I think I'd be more okay with `vars_auxiliary` and dropping `state` from the function names versus shortening things back to `aux`. While we all understand `aux` at this point, it is completely meaningless to someone looking at the code for the first time and carries as much information as me labeling them `A`.

If we're going for verbose names, we should actually have the names be descriptive and as self-documenting as possible versus some compromise between short, non-descriptive names and long, descriptive names that end up being medium, non-descriptive names with the disadvantages of both and benefits of neither.

bischtob:
> Same thing here. Technically some of these aren't even state variables, so I think I'd be more okay with `vars_auxiliary` and dropping `state` from the function names versus shortening things back to `aux`. While we all understand `aux` at this point, it is completely meaningless to someone looking at the code for the first time and carries as much information as me labeling them `A`.
> 
> If we're going for verbose names, we should actually have the names be descriptive and as self-documenting as possible versus some compromise between short, non-descriptive names and long, descriptive names that end up being medium, non-descriptive names with the disadvantages of both and benefits of neither.

I second that. I like the new names better than before. I think that so far we were erring on the side of too short and not descriptive enough, so I think that this is a good move.

examples/Microphysics/ex_1_saturation_adjustment.jl
```diff
@@ -68,7 +68,7 @@ function init_kinematic_eddy!(eddy_model, state, aux, (x, y, z), t)
     return nothing
 end
 
-function kinematic_model_nodal_update_aux!(
+function kinematic_model_nodal_update_auxiliary_state!(
```
vchuravy, line 71: Could this be just `nodal_update_auxiliary_state!`? The first argument is KinematicModel.

charleskawczynski: I don't think it would be good because this follows a pattern where the function [is passed](https://github.com/climate-machine/CLIMA/blob/master/src/Atmos/Model/AtmosModel.jl#L508-L518) to `nodal_update_aux!`, and while we might be able to use the same name, it might also be very confusing.

blallen: I just made this change to extend the name, @trontrytel decided to append kinematic_model in front of the original function name. I decided to make these changes as minimal as possible and just update the names to make the changes I made.

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try-

blallen:
bors try

blallen:
bors try

blallen:
bors try-

blallen:
bors try

blallen:
bors try

src/Atmos/Model/hyperdiffusion.jl
blallen: This seems to be causing an issue @mwarusz. Do these need to have the exact same name as the `vars_state_gradient` in the main model?

blallen:
@akshaysridhar @mwarusz can you guys take a look at the rising thermal bubble test failing on slurm? this most recent time it got a NaN error and on my laptop, it attempted to take the sqrt of a negative number. https://gist.github.com/climabot/ce1ab217836af7bb956e3132e90cd1b0#file-out_9696433

This PR wasn't supposed to have any functional changes, so I'm not sure what happened. My guess is it's something with hyperdiffusion, based on previous error messages. I just reverted a change where I renamed `vars_state_gradient` and `vars_state_gradient_laplacian` to have a `Δ` symbol in front of them, in case that is the cause. 

I'd really appreciate some help with this, since this is the only test left and I'd like to merge this before needing to do another round of rebase and merge.

src/Atmos/Model/hyperdiffusion.jl
mwarusz: The only have to match for a given model, i.e. `vars_gradient_laplacian(m)` has to be a subset of `vars_state_gradient(m)`.

src/Atmos/Model/hyperdiffusion.jl
```diff
@@ -89,7 +89,7 @@ function atmos_init_aux!(
     aux.hyperdiffusion.Δ = lengthscale(geom)
 end
```
mwarusz, line 90: Shouldn't `atmos_init_aux!` above be renamed to ` init_state_auxiliary!` ?

mwarusz: Nevermind ...

mwarusz:
The rising thermal bubble test fails because `update_aux!`
on line 21 in `src/Atmos/Model/remainder.jl` was not renamed to `update_auxiliary_state!`.

Other missing renames that I found:

`auxstate` on line 115 in `src/Driver/solver_configs.jl`
`dg.auxstate` on line 144 in `src/Utilities/Callbacks/Callbacks.jl`

src/Atmos/Model/hyperdiffusion.jl
blallen: and is this just for the hyperdiffusion model or does it also have to match the main atmos model?

blallen:
> The rising thermal bubble test fails because `update_aux!`
> on line 21 in `src/Atmos/Model/remainder.jl` was not renamed to `update_auxiliary_state!`.
> 
> Other missing renames that I found:
> 
> `auxstate` on line 115 in `src/Driver/solver_configs.jl`
> `dg.auxstate` on line 144 in `src/Utilities/Callbacks/Callbacks.jl`

Thanks! Just fixed these. Currently rebasing ontop of master again

src/Atmos/Model/hyperdiffusion.jl
mwarusz: They way that hyperdiffusion is implemented is in the `Atmos` model the names should only need to match the hyperdiffusion model. However, this means that we are recomputing gradients if hyperdiffusion is used. Not sure if there was a specific reason to do that @akshaysridhar ?

blallen:
bors try

blallen:
bors try-

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try

charleskawczynski:
Should we also rename `vars_state` [here](https://github.com/climate-machine/CLIMA/blob/master/test/Utilities/VariableTemplates/varsindex.jl#L10)?

blallen:
> Should we also rename `vars_state` [here](https://github.com/climate-machine/CLIMA/blob/master/test/Utilities/VariableTemplates/varsindex.jl#L10)?

Doing that caused some weird import-related failure. That test doesn't actually use the real `AtmosModel` or even a real `BalanceLaw`, so I think it's okay to leave it as `vars_state`. I think the purpose is just to check the functionality of recursive `VariableTemplates`.

PR #966, opened 2020-04-14
Simplify docs environment

charleskawczynski:
# Description

Decompose `docs/` environment (using a simple form of [environment stacks](https://docs.julialang.org/en/v1/manual/code-loading/#Environment-stacks-1)?) into

 - `src`
 - `Plots`
 - `Docs`

This way, as we add packages to CLIMA, we don't need to update the `docs/` environment, too.

This also ignores the non-`src` Manifests, and fixes the logic of the `generate_tutorials` `Bool` in `docs/make.jl`.

I have


charleskawczynski:
Maybe we don't want to track the `Plots` manifest, though

simonbyrne:
I'm not sure I understand the problem here: is it that the `docs/Manifest.toml` is getting out of date when a new package is added?

charleskawczynski:
Yes

charleskawczynski:
That and potentially the docs Project.toml

simonbyrne:
`docs/Project.toml` should be fine, as it already includes CLIMA.

docs/make.jl
```diff
@@ -1,5 +1,9 @@
 Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
 
+push!(LOAD_PATH, joinpath(@__DIR__, ".."))               # add CLIMA env
```
charleskawczynski, line 3: @simonbyrne are you saying that we should just remove this line since we already depend on CLIMA?

charleskawczynski:
> Since this is the docs build environment, can we test the idea of removing the Manifest and just using compat bounds in the Project file?

That's kind of what I had in mind. That said, it turns out that we need the Manifest while CLIMA is not registered:

```julia
julia> include("docs/make.jl")
ERROR: LoadError: ArgumentError: Package CLIMA [777c4786-024f-11e9-21a3-85d5d4106250] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.

Stacktrace:
 [1] _require(::Base.PkgId) at ./loading.jl:993
 [2] require(::Base.PkgId) at ./loading.jl:922
...

(docs) pkg> instantiate
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
ERROR: Package CLIMA [777c4786-024f-11e9-21a3-85d5d4106250] not found in a registry.
```

> Apart from that, this looks fine to me but I'm not fully clear on how these environments work.

I'm a bit fuzzy on them, too. @vchuravy is this what you had in mind during out Manifest in vs. out discussions? And, not to sound like a broken record but, is this in the spirit of how stack environments is set out to work? Or, does this look reasonable to you?

simonbyrne:
Yes, the issue is that it is the Manifest which stores the directory. You can get around this by first doing `]dev ..`

docs/make.jl
charleskawczynski: I removed this, is that what we should do?

charleskawczynski:
> Yes, the issue is that it is the Manifest which stores the directory. You can get around this by first doing ]dev ..

Isn't that what we [already have](https://github.com/climate-machine/CLIMA/blob/d46f329926a3d371223297d6bce419284285391f/docs/Manifest.toml#L65-L69)? If we remove the `docs/Manifest.toml`, this relative path is gone, right? My original thought was that we could get away with not having the `docs/Manifest.toml` if we use `include(LOAD_PATH, joinpath(@__DIR__, ".."))` in the `docs/make.jl`.

charleskawczynski:
Bump @vchuravy @simonbyrne

simonbyrne:
I still don't think this is that much of a problem, is it? Also, why do you want to split out the plots?

charleskawczynski:
> I still don't think this is that much of a problem, is it?

I don't think that it's really a problem either, but doesn't this simplify things a bit? This PR effectively decouples the CLIMA environment and docs environment, no? With this PR, anything we add to the CLIMA environment does not need to be added to the `docs` environment.

> Also, why do you want to split out the plots?

The plots are split out so that we can use `julia --project=env/Plots`. Although, I guess we'll still need `using Plots` to trigger `Requires` to make `Plots` available where needed. So, I don't mind adding this back into the docs environment. I guess I just thought that this is how the environment stacks were meant to be implemented (pick and choose which environments to stack on-top).

charleskawczynski:
@simonbyrne bump. Added

```julia
using Pkg
Pkg.activate(@__DIR__)
Pkg.instantiate(;verbose=true)
Pkg.develop(PackageSpec(path=joinpath(@__DIR__, "..")))
```

The `verbose=true` keyword [prints the package versions](https://julialang.github.io/Pkg.jl/v1/api/index.html#Pkg.instantiate). The other option I saw was [Pkg.status](https://julialang.github.io/Pkg.jl/v1/api/index.html#Pkg.status).

charleskawczynski:
This [failure](https://github.com/CliMA/ClimateMachine.jl/pull/1059/checks?check_run_id=657017590) is an example for why this PR is useful IMO. I don't actually need `RootSolvers.jl` in the docs `env`, but since `AbstractTolerance` is not currently exported, it seems that we must add `RootSolvers.jl` to the docs environment.

charleskawczynski:
This has gotten quite stale, and it feels much less a severe problem than when it was opened. Closing for now.

PR #967, opened 2020-04-15
Add a few inbounds annotations

mwarusz:
Adds a few inbounds annotations


mwarusz:
Closing because those annotations are not needed, since we are indexing StaticArrays. I thought this helped me with some performance issues I saw, but it turns out I was mistaken. Thanks for pointing that to me @jkozdon

Issue #943, opened 2020-04-07
Additional LES diagnostics


szy21:
# Description

Some more diagnostics in the LES would be helpful. Here is the diagnostics we currently have in the LES: https://gist.github.com/akshaysridhar/cc7888d27a93237eee4f8373630a0d6a. Some useful diagnostics include (I checked the ones we already have in current master):

### default

Profiles (t,z):

- [x] u, v ,w, q_tot, thl, e_int
- [x] ρ 
- [x] u'u', v'v', w'w'
- [x] q_tot'q_tot', e_int'e_int', thl'thl'
- [x] w’ρ’, w’u', w’v', w’q_tot', w’thl'
- [x] w'e_int', q_tot'thl', q_tot'e_int' 
- [x] cloud fraction: horizontal area (fraction) where q_liq>0

Timeseries (t):

- [ ] liquid water path: lwp (vertical integral of ρ q_liq)
- [x] cloud cover: fraction of columns where lwp > 0
- [x] cloud top: highest z level where q_liq>0
- [x] cloud base: lowest z level where q_liq>0

### core

Profiles (t,z):

- [x] horizontal area (fraction) where q_liq > 0 and w > 0
- [x] u, v ,w, ρ, q_tot, thl, e_int conditional on q_liq > 0 and w > 0
- [x] q_tot'q_tot', thl'thl', e_int'e_int' conditional on q_liq > 0 and w > 0
- [x] u'u', v'v', w'w' conditional on q_liq > 0 and w > 0
- [x] w'ρ', w'q_tot', w'thl', w'e_int', q_tot'thl', q_tot'e_int' conditional on q_liq > 0 and w > 0

**Optimization: for all the variance and covariance (a'a' and a'b'), we can save aa and ab and a and b, as an alternative to save computation. E.g., we can save the q_tot and q_totq_tot instead of q_tot'q_tot', and w, e_int and we_int instead of w'e_int'.**

# Additional context

This is related to #214 but I decided to submit this as a separate issue because it would be good to have these LES diagnostics for the GCP runs in 1-2 weeks.

# For CLIMA Developers

- [x] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

szy21:
[...]

PR #972, opened 2020-04-16
update VariableList.md

szy21:
# Description

This PR updates VariableList.md and provides suggested code names, short names, and long names for the diagnostics we need in #943. Hopefully, it also provides guidance for the names of future diagnostics.

I have

### DIFF
--- a/docs/src/VariableList.md
+++ b/docs/src/VariableList.md
@@ -102,6 +102,51 @@ conv_q_liq_to_q_rai_accr = tendency to q_rai due to accretion from q_liq [1/s]
 conv_q_rai_to_q_vap      = tendency to q_vap due to evaporation from q_rai [1/s]
 ```
 
+### 2.7 Diagnostics
+Here are suggested names in the output. Format: short name (code name) = long name
+```
+rho(ρ) = density
+u(u) = x_velocity
+v(v) = y_velocity
+w(w) = z_velocity
+qt(q_tot) = total specific humidity
+et(e_tot) = total specific energy
+qv(q_vap) = water vapor specific humidity
+ql(q_liq) = liquid water specific humidity
+ei(e_int) = specific internal energy
+thd(θ_dry) = potential temperature
+thl(θ_liq_ice) = liquid-ice potential temperature
+thv(θ_vir) = virtual potential temperature
+hm(h_moi) = specific enthalpy
+ht(h_tot) = total specific enthalpy
+
+var_u(u′u′) = variance of x-velocity
+var_v(v′v′) = variance of y-velocity
+var_w(w'w') = variance of z-velocity
+TKE(TKE) = turbulence kinetic energy
+var_qt(q_tot'q_tot') = variance of total specific humidity
+var_ei(e_int'e_int') = variance of specific internal energy
+var_thl(θ_liq_ice'θ_liq_ice') = variance of liquid-ice potential temperature
+cov_qt_thl(q_tot'θ_liq_ice') = covariance of total specific humidity and liquid-ice potential temperature
+cov_qt_ei(q_tot'e_int') = covariance of total specific humidity and specific internal energy
+w3(w′w′w′) = the third moment of z-velocity
+
+cov_w_rho(w′ρ′) = vertical flux of mass
+cov_w_u(w′u′) = vertical flux of x-velocity
+cov_w_v(w′v′) = vertical flux of y-velocity
+cov_w_qt(w'q_tot') = vertical flux of total specific humidity
+cov_w_qv(w′q_vap′) = vertical flux of water vapor specific humidity
+cov_w_ql(w′q_liq′) = vertical flux of liuqid water specific humidity
+cov_w_thd(w′θ_dry′) = vertical flux of potential temperature
+cov_w_thv(w′θ_vir′) = vertical flux of virtual temperature
+cov_w_thl(w′θ_liq_ice′) = vertical flux of liquid-ice potential temperature
+w_qt_sgs(d_q_tot) = vertical sgs flux of total specific humidity
+w_ht_sgs(d_h_tot) = vertical sgs flux of total specific enthalpy
+
+cl(cl) = cloud fraction
+clt(clt) = cloud cover
+```
+
 ### TODO
 ```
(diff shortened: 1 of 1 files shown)


szy21:
@tapios @smarras79 I talked with Akshay and Kiran, and we came up with some 'standard names' in the diagnostics (Note we changed the names of some current diagnostics). Please take a look and let me know if you have any suggestions. Thanks!

docs/src/VariableList.md
```diff
@@ -102,6 +102,50 @@ conv_q_liq_to_q_rai_accr = tendency to q_rai due to accretion from q_liq [1/s]
...
+var_u(u′u′) = variance of x-velocity
+var_v(v′v′) = variance of y-velocity
+var_w(w'w') = variance of z-velocity
+TKE(TKE) = turbulence kinetic energy
+var_qt(q_tot'q_tot') = variance of total specific humidity
+var_ei(e_int'e_int') = variance of specific internal energy
+var_thl(θ_liq_ice'θ_liq_ice') = variance of liquid-ice potential temperature
+cov_qt_thl(q_tot'θ_liq_ice') = covariance of total specific humidity and liquid-ice potential temperature
+cov_qt_ei(q_tot'e_int') = covariance of total specific humidity and specific internal energy
+wskew(w′w′w′) = skewness of w-velocity
```
tapios, line 132: I'd only call it skewness if you divide by variance^(3/2). Otherwise, perhaps `w3` or something like it is better?

docs/src/VariableList.md
```diff
@@ -102,6 +102,50 @@ conv_q_liq_to_q_rai_accr = tendency to q_rai due to accretion from q_liq [1/s]
...
+
+zflux_rho(w′ρ′) = vertical flux of mass
```
tapios, line 134: Why not `w_rho` etc for all of these? Seems easier and clearer to me.

tapios:
Thanks for doing this. Looks good to me. But why not leaving `q_tot` in the output too, rather than `qt`, and similarly for other variables? You probably had a reason, and what you have seems good. But if it's just brevity, maybe the slightly longer names in the code are a little more transparent?

szy21:
> Thanks for doing this. Looks good to me. But why not leaving `q_tot` in the output too, rather than `qt`, and similarly for other variables? You probably had a reason, and what you have seems good. But if it's just brevity, maybe the slightly longer names in the code are a little more transparent?

The reason is that we don't want to end up with cov_q_tot_theta_liq, or potentially longer variable names in the output. But I'm happy to change them back.

tapios:
What you have is fine. It may also be worth considering CMIP variable names (e.g., https://www.dkrz.de/up/services/data-management/projects-and-cooperations/ipcc-data/cmip5-variables), but for the most part they are not very intuitive either. It would just be a way to connect with the broader community.

tapios:
A more exhaustive CMIP6 list: http://clipc-services.ceda.ac.uk/dreq/mipVars.html

szy21:
> What you have is fine. It may also be worth considering CMIP variable names (e.g., https://www.dkrz.de/up/services/data-management/projects-and-cooperations/ipcc-data/cmip5-variables), but for the most part they are not very intuitive either. It would just be a way to connect with the broader community.

We actually talk about this (and thanks to the GCP runs I almost remember every single variable name of CMIP now). None of the models I know of follows this - they postprocess the original output following CMIP conventions. I think we are far away from participating in CMIP so we end up not following it. I think it may be a good idea to follow it for the GCM? They don't have most variables we need in the LES anyway.

docs/src/VariableList.md
szy21: Does cov_w_rho look good to you? I want it to be consistent with the names of other covariances.

szy21:
But I am happy to change everything to follow CMIP standards (when the variables are available), @akshaysridhar what do you think? (huss is definitely not as intuitive as qt.)

szy21:
> Maybe this is more of a code-centric comment but, I thought it might be nice to have a single "in-house" name for all of our variables and then use methods to return translated names to interface outside of our in-house names:
> 
> ```julia
> cmip_name(::Val{:our_var_name}) = :cmip_var_name
> ```
> 
> I did this in RRTMGP.jl [for gases](https://github.com/climate-machine/RRTMGP.jl/blob/master/src/rrtmgp/Gases.jl#L89-L116) (but with types), and I kind of like how it turned out.
> 
> That said, maybe this PR is a bit simpler.

I assume these are names in the code, rather than names in the output?

tapios:
> But I am happy to change everything to follow CMIP standards (when the variables are available), @akshaysridhar what do you think? (hus is definitely not as intuitive as qt.)

It's fine not to worry about CMIP for now at all. It would be one thing if they had intuitive names, but they don't. So we can translate when we need to later.

docs/src/VariableList.md
```diff
@@ -102,6 +102,50 @@ conv_q_liq_to_q_rai_accr = tendency to q_rai due to accretion from q_liq [1/s]
 conv_q_rai_to_q_vap      = tendency to q_vap due to evaporation from q_rai [1/s]
 ```
 
+### 2.7 Diagnostics
+Here are suggested names in the output. Format: short name (code name) = long name
+```
+rho(ρ) = density
+u(u) = x_velocity
+v(v) = y_velocity
+w(w) = w_velocity
```
claresinger, line 111: I think you mean z_velocity :)

szy21:
bors try

akshaysridhar:
> Maybe this is more of a code-centric comment but, I thought it might be nice to have a single "in-house" name for all of our variables and then use methods to return translated names to interface outside of our in-house names:
> 
> ```julia
> cmip_name(::Val{:our_var_name}) = :cmip_var_name
> ```
> 
> I did this in RRTMGP.jl [for gases](https://github.com/climate-machine/RRTMGP.jl/blob/master/src/rrtmgp/Gases.jl#L89-L116) (but with types), and I kind of like how it turned out.
> 
> That said, maybe this PR is a bit simpler.

Longer term this sounds awesome ! I think right now we need a single point of reference for in-house varnames, and maintaining / using this var-list file is a simple solution

Issue #831, opened 2020-03-11
TKE calculation issue : `AtmosDiagnostics.jl`


akshaysridhar:
# Description

https://github.com/climate-machine/CLIMA/blob/2959529adf93d652141d9d5b33f504d4a36e5f86/src/Diagnostics/AtmosDiagnostics.jl#L222

This is not consistent, since `MH` is being accounted for twice. This will result in the vars_diagnostic symbol `:TKE` returning **wrong** stuff. @slifer50 could you please confirm and we can fix this :D  

## Steps to reproduce

1. Open output from any LES in `diagnostics*.jld2`.
2. Plot (`(:uvariance + :vvariance + :wvariance) / 2`) and Plot (`:TKE`) 
3. They are not consistent (i.e. order of magnitude), leading to the point about the discrepancy above. 

## Expected behavior

The components listed in 2. should be the approximately the same.

# For CLIMA developers

- [x] The bug can be reproduced
- [x] Labels have been added and developers assigned
- [x] The bug can be fixed (if not, please explain why not in a comment below)
- [x] There is a timeline to fix the bug
- [ ] The bug has been fixed (Please link the PR)

kpamnany:
This is now here:

https://github.com/climate-machine/CLIMA/blob/8b113f87308304af2a63c912a5f1af38673c0e77/src/Diagnostics/atmos_default.jl#L256

akshaysridhar:
Thanks @kpamnany :D @slifer50 and I can sort this as part of a set of verification tests against the new b.c. (Unify ATMOS and BOMEX are both good to go so I can use that to stress test the diagnostics module)

smarras79:
> This is now here:
> 
> https://github.com/climate-machine/CLIMA/blob/8b113f87308304af2a63c912a5f1af38673c0e77/src/Diagnostics/atmos_default.jl#L256

There are two error in this calculation: 
* MH should be removed
* The calculation must be square rooted

akshaysridhar:
I've seen RMS velocities used as a diagnostic, but sqrt(TKE) would be unusual(?)

smarras79:
> I've seen RMS velocities used as a diagnostic, but `sqrt(TKE)` would be unusual(?)

Sorry, @akshaysridhar yes, forget what I said. I let me remove that comment in full before people get confused.

smarras79:
[...]

PR #974, opened 2020-04-17
More LES diagnostics

### LINKED ISSUE #739: high order moments
https://github.com/climate-machine/CLIMA/blob/0ef45f32ef07c4fce847de2d8461144418f8f79e/src/Diagnostics/AtmosDiagnostics.jl#L213

It would save computation to save the HorizontalMean of products of variables and leave the computation of high order moments to offline computation such that the code will save HM(u) , HM(u^2), HM(u^3), HM(u) , HM(u *v), HM(u *v *w) and so on, such that in general the offline computation will read:

vu_second_moment =  HM(u*v) - HM(u)*HM(u)
uvw_third_moment =  HM(u*v*w) - HM(u*w)*HM(v) - HM(v*w)*HM(u) - HM(u*v)*HM(w) + 2*HM(u)*HM(v)*HM(w)

kpamnany:
# Description

- Adds more of the `default` diagnostics from #943 (excluding LWP and cloud cover).
- Changes the variable names for the diagnostics according to the [variable list](https://climate-machine.github.io/CLIMA/latest/DiagnosticVariables/).
- Changes the computation of the vertical fluxes and variances to use density-weighted means.
- Applies the idea from and closes #739.
- Changed `Writers.write_data()` to allow specifying dimension names for a variable.

I have

### DIFF
--- a/src/Diagnostics/atmos_core.jl
+++ b/src/Diagnostics/atmos_core.jl
@@ -1,0 +1,93 @@
+using ..Atmos
+using ..Atmos: thermo_state, turbulence_tensors
+using ..Mesh.Topologies
+using ..Mesh.Grids
+using ..MoistThermodynamics
+using LinearAlgebra
+
+function atmos_core_init(dgngrp::DiagnosticsGroup, currtime)
+    mpicomm = Settings.mpicomm
+    dg = Settings.dg
+    Q = Settings.Q
+    FT = eltype(Q)
+    grid = dg.grid
+    topology = grid.topology
+    N = polynomialorder(grid)
+    Nq = N + 1
+    Nqk = dimensionality(grid) == 2 ? 1 : Nq
+    nrealelem = length(topology.realelems)
+    nvertelem = topology.stacksize
+    nhorzelem = div(nrealelem, nvertelem)
+
+    if Array ∈ typeof(Q).parameters
+        localvgeo = grid.vgeo
+    else
+        localvgeo = Array(grid.vgeo)
+    end
+
+    @warn """
+        Diagnostics ($dgngrp.name): not yet implemented
+        """
+
+    return nothing
+end
+
+"""
+    atmos_core_collect(bl, currtime)
+
+Perform a global grid traversal to compute various diagnostics.
+"""
+function atmos_core_collect(dgngrp::DiagnosticsGroup, currtime)
+    mpicomm = Settings.mpicomm
+    dg = Settings.dg
+    Q = Settings.Q
+    mpirank = MPI.Comm_rank(mpicomm)
+    current_time = string(currtime)
+
+    # extract grid information
+    bl = dg.balancelaw
+    grid = dg.grid
+    topology = grid.topology
+    N = polynomialorder(grid)
+    Nq = N + 1
+    Nqk = dimensionality(grid) == 2 ? 1 : Nq
+    npoints = Nq * Nq * Nqk
+    nrealelem = length(topology.realelems)
+    nvertelem = topology.stacksize
+    nhorzelem = div(nrealelem, nvertelem)
+
+    # get the state, auxiliary and geo variables onto the host if needed
+    if Array ∈ typeof(Q).parameters
+        localQ = Q.realdata
+        localaux = dg.auxstate.realdata
+        localvgeo = grid.vgeo
+        localdiff = dg.diffstate.realdata
+    else
+        localQ = Array(Q.realdata)
+        localaux = Array(dg.auxstate.realdata)
+        localvgeo = Array(grid.vgeo)
+        localdiff = Array(dg.diffstate.realdata)
+    end
+    FT = eltype(localQ)
+
+    nstate = num_state(bl, FT)
+    nauxstate = num_aux(bl, FT)
+    ndiff = num_diffusive(bl, FT)
+
+    # TODO
+
+    if mpirank == 0
+        dprefix = @sprintf(
+            "%s_%s-%s-num%04d",
+            dgngrp.out_prefix,
+            dgngrp.name,
+            Settings.starttime,
+            dgngrp.num
+        )
+        dfilename = joinpath(Settings.output_dir, dprefix)
+    end
+
+    return nothing
+end # function collect
+
+function atmos_core_fini(dgngrp::DiagnosticsGroup, currtime) end
--- a/src/Diagnostics/atmos_default.jl
+++ b/src/Diagnostics/atmos_default.jl
@@ -1,19 +1,21 @@
 using ..Atmos
-using ..Atmos: thermo_state, turbulence_tensors
+using ..Atmos: MoistureModel, thermo_state, turbulence_tensors
 using ..Mesh.Topologies
 using ..Mesh.Grids
 using ..MoistThermodynamics
-using CLIMAParameters.Atmos.SubgridScale: inv_Pr_turb
 using LinearAlgebra
 
-Base.@kwdef mutable struct AtmosCollectedDiagnostics
+Base.@kwdef mutable struct AtmosDefaultCollectedDiagnostics
     zvals::Union{Nothing, Array} = nothing
     repdvsr::Union{Nothing, Array} = nothing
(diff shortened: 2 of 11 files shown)


src/Diagnostics/atmos_default.jl
```diff
@@ -194,65 +242,87 @@ function compute_diagnosticsums!(
     horzavgs,
     dsums,
 )
-    zvals = CollectedDiagnostics.zvals
     th = thermo_vars(thermoQ[ijk, e])
     ha = horzavg_vars(horzavgs[Nqk * (ev - 1) + k])
     ds = diagnostic_vars(dsums[Nqk * (ev - 1) + k])
 
     u = state.ρu[1] / state.ρ
+    u′ = ha.ρu / ha.ρ
```
szy21, line 250: The prime symbol is confusing to me because it is commonly used for deviation from the mean in fluid dynamics. @akshaysridhar what do you think?

kpamnany: How about I rename all `x′` to `avg_x`?

kpamnany:
The file to focus on for code review is `src/Diagnostics/atmos_default.jl`.

More practically, checking out this branch and running BOMEX or one of the other experiments with `--enable-diagnostics` and then checking the output diagnostics to see if they make sense would be very helpful.

szy21:
@yairchn would you like to take a look at this? I may not be able to provide a full review until later this week.

kpamnany:
P.S.: If #958 merges before you get a chance to test, `--enable-diagnostics` should become `--diagnostics default` (or replace 'default' with an interval specification as described there).

src/Diagnostics/atmos_default.jl
```diff
@@ -3,16 +3,70 @@ using ..Atmos: thermo_state, turbulence_tensors
...
+        var_w::FT               # w′w′
+        var_qt::FT              # q_tot′q_tot′
+        var_ei::FT              # e_int′e_int′
+        var_thl::FT             # thl′thl′
+
+        # skewness
+        w3::FT                  # w′w′w′
+
+        # turbulent kinetic energy
+        tke::FT
```
smarras79, line 60: As it is now there is a potential bug in the calculation of tke. 
Since tke can be obtained from other quantities that are written to file, should be just omit it from output?

src/Diagnostics/atmos_default.jl
```diff
@@ -194,65 +241,87 @@ function compute_diagnosticsums!(
...
+        ds.cov_qt_ei += MH * (q_tot - avg_q_tot) * (e_int - avg_e_int)
+        ds.var_qt += MH * (q_tot - avg_q_tot)^2
     end
 
-    # variances
-    ds.uvariance += MH * (u - ũ)^2
-    ds.vvariance += MH * (v - ṽ)^2
-    ds.wvariance += MH * (w - w̃)^2
-
     # skewness
```
smarras79, line 320: #skewness should be replaced with #w3 since they are not the same quantity. Skewness can be obtained from w3 and var_w so that it's not necessary to calculate it here.

src/Diagnostics/atmos_default.jl
smarras79, line 60: Nevermind my past comment. TKE is calculated and output correctly.

akshaysridhar: This was addressed in issue #831

src/Diagnostics/atmos_default.jl
```diff
@@ -194,65 +241,87 @@ function compute_diagnosticsums!(
...
+    ds.thl += MH * avg_θ_liq_ice
+    ds.thv += MH * avg_θ_v
+    ds.ei += MH * avg_e_int
+    ds.hm += MH * avg_h_moi
+    ds.ht += MH * avg_h_tot
+    ds.qt_sgs += MH * avg_qt_sgs
+    ds.ht_sgs += MH * avg_ht_sgs
+
+    # covariances and variances
+    ds.cov_w_rho += MH * (w - avg_w) * (state.ρ - ha.ρ)
```
szy21, line 291: I have talked with @tapios and we want density averaged density here (yes, it's strange). So you would calculate th.ρρ = state.ρ*state.ρ in `compute_thermo`, and avg_ρ = ha.ρρ / ha.ρ, and use avg_ρ instead of ha.ρ here, and only here.

szy21:
We would need density weighted average for the output. For example, `ds.cov_w_qt += MH * (w - avg_w) * (q_tot - avg_q_tot)`, and for the output we would like to have `avg(ds.cov_w_qt*state.ρ)/ha.ρ`, while right now it's `avg(ds.cov_w_qt)`. Maybe the simplest solution is to have another function for density weighted horizontal means?

kpamnany:
> We would need density weighted average for the output.

This isn't clear to me. Let's talk about it.

kpamnany:
Reworking this per discussion with @szy21 and @akshaysridhar so don't worry about reviewing it yet.

src/Diagnostics/atmos_core.jl
```diff
@@ -0,0 +1,107 @@
...
+    nhorzelem = div(nrealelem, nvertelem)
+
+    if Array ∈ typeof(Q).parameters
+        localvgeo = grid.vgeo
+    else
+        localvgeo = Array(grid.vgeo)
+    end
+
+    @warn """
+Diagnostics ($dgngrp.name): not yet implemented"""
```
vchuravy, line 29: Triple quotes don't need to begin at the beginning of the line.

```
julia>     """
           hello
           """
```
is the same as:
```
julia> """
       hello
       """

**Comment by kpamnany** [comment]
This is ready for review. Please especially look at [this](https://github.com/climate-machine/CLIMA/blob/kp/more-diagnostics/src/Diagnostics/atmos_default.jl#L117-L144) and [this](https://github.com/climate-machine/CLIMA/blob/kp/more-diagnostics/src/Diagnostics/atmos_default.jl#L178-L192).

Also please do run an experiment with these diagnostics enabled and check the output, including cloud fraction, cloud top and cloud base.

**Comment by smarras79** [comment]
> This is ready for review. Please especially look at [this](https://github.com/climate-machine/CLIMA/blob/kp/more-diagnostics/src/Diagnostics/atmos_default.jl#L117-L144) and [this](https://github.com/climate-machine/CLIMA/blob/kp/more-diagnostics/src/Diagnostics/atmos_default.jl#L178-L192). Note that all these (except `avg_rho`) are later divided by `avg_rho`.
> 
> Also please do run an experiment with these diagnostics enabled and check the output, including cloud fraction, cloud top and cloud base.

Fantastic, @kpamnany I'll try it with dycoms this afternoon

**Comment by kpamnany** [comment]
After more discussion with @szy21, @akshaysridhar, @yairchn, and @slifer50, this is being reworked yet again. Please hold on reviewing until I update again tomorrow.

**Comment by kpamnany** [comment]
Only thing remaining is fixing `cld_frac` to be on `z`.

**Comment by szy21** [comment]
@kpamnany Could we have `temp`=temperature too? (not urgent)

File: src/Diagnostics/atmos_default.jl [hunk 6]
```diff
@@ -34,268 +36,261 @@ function atmos_default_init(dgngrp::DiagnosticsGroup, currtime)
...
+    w = state.ρu[3] / state.ρ
+    w′ = w - ha.w
+    e_int′ = thermo.e_int - ha.ei
+    θ_dry′ = thermo.θ_dry - ha.thd
+    θ_vir′ = thermo.θ_vir - ha.thv
+
+    sums.var_u += MH * state.ρu[1] / state.ρ * state.ρu[1]
+    sums.var_v += MH * state.ρu[2] / state.ρ * state.ρu[2]
+    sums.var_w += MH * state.ρu[3] / state.ρ * state.ρu[3]
+    sums.w3 += MH * state.ρu[3] / state.ρ * state.ρu[3] / state.ρ * state.ρu[3]
```
szy21, line 219: Why are these here? (221-224 are the correct definition of var_u, var_v, var_w, and w3.)

PR #983, opened 2020-04-20
Fix moist thermo call

charleskawczynski:
# Description

Undo `maxiter` and `tolerance` change in #980. See [this comment](https://github.com/climate-machine/CLIMA/pull/980#pullrequestreview-395937103). This PR allows saturation adjustment to fail, but will dump warnings with instructions on what to do.

Special thanks to @leios and @vchuravy for `@print` support!

I have


charleskawczynski:
bors try

charleskawczynski:
> Is the only way the solver may not converge if `maxiter` is reached? If so, it may be worth stating in the warning message that `maxiter` is reached. If not, it may be worth being more granular in the reason about it did not converge.

Convergence is reached once the root equation residual is lower than the prescribed tolerance. More iterations increases the runtime ceiling, but the solver exits the loop if the convergence is reached sooner. That said, reaching `maxiter` is effectively equivalent to not converging, so we're basically doing that with this PR's warning. As far as I know, answering "why" the solution did not converge is non-trivial and requires offline numerical analysis.

The other thing I've found, with this PR, is that these warnings severely bloat the output-- the warning are printed (when convergence fails) for every RHS evaluation, at every node, for every timestep. Due to this drastic IO dump, it will probably severely slow down the current Dycoms simulations.

tapios:
> Convergence is reached once the root equation residual is lower than the prescribed tolerance. More iterations increases the runtime ceiling, but the solver exits the loop if the convergence is reached sooner. That said, reaching `maxiter` is effectively equivalent to not converging, so we're basically doing that with this PR's warning. As far as I know, answering "why" the solution did not converge is non-trivial and requires offline numerical analysis.

If the only possible cause of failure is that `maxiter` has been reached, I suggest you state that. (My question was not about why that happened, but rather if there could be other reasons for bailout, e.g., a singular derivative.)

> The other thing I've found, with this PR, is that these warnings severely bloat the output-- the warning are printed (when convergence fails) for every RHS evaluation, at every node, for every timestep. Due to this drastic IO dump, it will probably severely slow down the current Dycoms simulations.

Please make the temperature tolerance generous, so the algorithm converges in one iteration almost always. Then there shouldn’t be an output problem.

charleskawczynski:
bors try

charleskawczynski:
@tapios based on what I found in #1017, these seem to be our currently passing `maxiter` and `tolerance` configurations for `PhaseEquil`:

```julia
    maxiter = 3, tolerance = 2 # Passes
    maxiter = 2, tolerance = 3 # Passes
    maxiter = 1, tolerance = 16 # Passes

    maxiter = 1, tolerance = 15 # Fails
```

So, we cannot continue with `maxiter = 1` and `tolerance = 2` because Dycoms is timing out.

tapios:
> @tapios based on what I found in #1017, these seem to be our currently passing `maxiter` and `tolerance` configurations for `PhaseEquil`:
> 
> ```julia
>     maxiter = 3, tolerance = 2 # Passes
>     maxiter = 2, tolerance = 3 # Passes
>     maxiter = 1, tolerance = 16 # Passes
> 
>     maxiter = 1, tolerance = 15 # Fails
> ```
> 
> So, we cannot continue with `maxiter = 1` and `tolerance = 2` because Dycoms is timing out.

Can you just set `maxiter = 1`, `tolerance = 1` for DYCOMS only? I suspect the problems arise near freezing, which is not an issue for DYCOMS. You can leave the defaults in general at `maxiter = 5, tolerance = 1` (or something like it).

charleskawczynski:
Will update this, based on discussions, to use a residual tolerance once [it's merged in RootSolvers.jl](https://github.com/climate-machine/RootSolvers.jl/pull/13) and a new release is made.

charleskawczynski:
This issue was resolved somewhere else. Right now, in Master, we have
```julia
moisture = EquilMoist{FT}(maxiter = 4, tolerance = FT(1))
```
, which seems fine.

PR #986, opened 2020-04-20
Apply outline skeleton to docs

charleskawczynski:
# Description

Adds docs skeleton. I've split up each section into separate files since I think these may get rather involved. Also, since the Literate examples need additional processing, I've split these off from the How-To-Guides. We can integrate them later, but right now it's a lot easier to keep them separate. I'm not certain that everyone will agree with the restructuring here, so I'd like to hear thoughts!

I have


docs/make.jl
```diff
@@ -2,70 +2,30 @@ Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
...
+generate_examples = true
 
-GENERATED_BL_EXAMPLES = [
-    joinpath("examples", "DGmethods_old", "generated", f)
-    for
-    f in ("ex_001_periodic_advection.md", "ex_002_solid_body_rotation.md")
-]
-GENERATED_BL_EXAMPLES = filter!(x -> isfile(x), GENERATED_BL_EXAMPLES)
-
-# Literate examples created in their own literate_example directory for now
```
leios, line 15: These lines allow us to stick all the literate files into a single directory, which I thought was what we wanted? Right now, all the literate examples go in the `literate_examples` subdirectory and are generated automatically with this.

examples/literate_markdown.jl
```diff
@@ -2,18 +2,7 @@
...
 #
-# For now, all literate examples are held in the `literate_examples` directory, and generated with the following code in `docs/make.jl`:
-
-using Literate
-const examples_directory = pwd() * "/literate_examples/"
-const output_directory = pwd() * "/generated/"
-
-examples = readdir(examples_directory)
-
-for example in examples
```
charleskawczynski, line 13: @leios I don't know that we should expect users to add/modify these lines. I originally thought that contributors will simply add literate examples to the `CLIMA/examples` directory, and the docs will automatically pick them up. The only additional thing that's needed is for the .md files to be referenced, but this could be done automatically as well.

docs/make.jl
leios: Ok, great! I didn't know that `examples` was supposed to be purely literate files, which was why we made the change before.

examples/literate_markdown.jl
leios: You are right, they shouldn't change these lines. I just put them there to explain that everything was being picked up from a single directory. If I'm honest, this file was meant as a placeholder for adding things later.

charleskawczynski:
Preview available [here](https://climate-machine.github.io/CLIMA/previews/PR986/)

docs/list_of_discussions.jl
```diff
@@ -0,0 +1,15 @@
...
+discussions =
+Any[
+"Home" => "Discussions/index.md",
+"Atmos" => Any[
+        "Discussions/Atmos/EDMFEquations.md",
+    ],
+"Ocean" => Any[],
+"Land" => Any[],
+"Common" => Any[],
+"Numerics" => Any[],
```
simonbyrne, line 14: Comment out empty pages until we have some content there so they don't render in HTML:

```suggestion
# "Ocean" => Any[],
# "Land" => Any[],
# "Common" => Any[],
# "Numerics" => Any[],
```

docs/list_of_extending_clima.jl
```diff
@@ -0,0 +1,44 @@
...
+"Home" => "ExtendingCLIMA/Contributing/CONTRIBUTING.md",
+"Arrays" => Any[
+    "Home" => "ExtendingCLIMA/Arrays/index.md"
+],
+"Common" => Any[
+    "SurfaceFluxes" => "ExtendingCLIMA/Common/SurfaceFluxes.md",
+    "MoistThermodynamics" => "ExtendingCLIMA/Common/MoistThermodynamics.md",
+    ],
+"Atmos" => Any[
+        "AtmosModel" => "ExtendingCLIMA/Atmos/AtmosModel.md",
```
simonbyrne, line 16: Could also comment out some of these empty ones as well.

charleskawczynski:
There's still more pieces to separate, but it'll be nice to get this through so that we can start iterating

PR #993, opened 2020-04-21
Add tendencies computation flow chart

charleskawczynski:
# Description

This PR adds a diagram to the docs showing the order of function calls made by the [DG methods functor](https://github.com/climate-machine/CLIMA/blob/master/src/Numerics/DGmethods/DGmodel.jl#L44), which computes all of the PDE tendencies. This flow chart is not intended to explain _everything_ that happens in this functor, only the necessary things to assist users in implementing balance law interface methods.

I have


akshaysridhar:
YES!

charleskawczynski:
Preview available [here](https://climate-machine.github.io/CLIMA/previews/PR993/). Flow chart [here](https://climate-machine.github.io/CLIMA/previews/PR993/FlowChart/)

charleskawczynski:
Weird, I saw the figure when I built the docs locally. We still need to fix something here.

blallen:
Maybe remove the `nothing` at the end of the code block?

charleskawczynski:
> Maybe remove the nothing at the end of the code block?

I don't think that's it because Anna's [Microphysics example](https://climate-machine.github.io/CLIMA/latest/Atmos/Microphysics/#Accretion-1) [has `nothing`](https://github.com/climate-machine/CLIMA/blob/master/docs/src/Atmos/Microphysics.md#accretion) and it shows up fine. I'll give a try in a bit though

simonbyrne:
> Weird, I saw the figure when I built the docs locally. We still need to fix something here.

```
│   value = IOError: could not spawn `lualatex --enable-write18 --output-directory=./jl_Q1YTSX ./jl_Q1YTSX/tendencies_flow_chart.tex`: no such file or directory (ENOENT)
```

Looks like TeX is not installed. Do we want to add this as a build dependency?

simonbyrne:
From https://github.com/JuliaTeX/TikzPictures.jl

> In order to use this library, lualatex must be installed. The texlive and miktex distributions include lualatex. You must also have dvisvgm installed. On Ubuntu, you can get these, if not already present, by running sudo apt-get install texlive-latex-base and sudo apt-get install texlive-binaries.
> 
> You also need pdf2svg. On Ubuntu, you can get this by running sudo apt-get install pdf2svg. On Windows, you can download the binaries from http://www.cityinthesky.co.uk/opensource/pdf2svg/. Be sure to add pdf2svg to your path (and restart).

simonbyrne:
An alternative might be something like [Luxor.jl](http://juliagraphics.github.io/Luxor.jl/stable/), though it has much more basic functionality.

PR #997, opened 2020-04-22
Update moist thermo doc strings

charleskawczynski:
# Description

Updates the moist thermodynamics docs to include `param_set` arguments. Since `param_set` is used in many function arguments, I made the docs for `param_set` for each function simply say:

```
- `param_set` an `AbstractParameterSet`, see the [`MoistThermodynamics`](@ref) for more details
```

At the top of the module, I added a section explaining [more details](https://github.com/climate-machine/CLIMA/pull/997/files#diff-b930a7e34792312cae59cfd0823efe90R7-R24).

I have

### DIFF
--- a/src/Common/MoistThermodynamics/MoistThermodynamics.jl
+++ b/src/Common/MoistThermodynamics/MoistThermodynamics.jl
@@ -4,6 +4,24 @@
 Moist thermodynamic functions, e.g., for air pressure (atmosphere equation
 of state), latent heats of phase transitions, saturation vapor pressures, and
 saturation specific humidities.
+
+
+## AbstractParameterSet's
+
+Many functions defined in this module rely on CLIMAParameters.jl.
+CLIMAParameters.jl defines several functions (e.g., many planet
+parameters). For example, to compute the mole-mass ratio:
+
+```julia
+using CLIMAParameters.Planet: molmass_ratio
+using CLIMAParameters: AbstractEarthParameterSet
+struct EarthParameterSet <: AbstractEarthParameterSet end
+param_set = EarthParameterSet()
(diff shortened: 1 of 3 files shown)


src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -4,6 +4,24 @@
...
+Many functions defined in this module rely on CLIMAParameters.jl.
+CLIMAParameters.jl defines several functions (e.g., many planet
+parameters). For example, to compute the mole-mass ratio:
+
+```julia
+using CLIMAParameters.Planet: molmass_ratio
+using CLIMAParameters: AbstractEarthParameterSet
+struct EarthParameterSet <: AbstractEarthParameterSet end
+param_set = EarthParameterSet()
+_molmass_ratio = molmass_ratio(param_set)
```
kpamnany, line 20: In the systems and C world, an underscore prefix is, by convention, used for internal symbols. Can we rename this suggestion here to something else?

charleskawczynski: I've been using this convention across all `param_set` calls, so we should probably change it everywhere else if we change this (and that would be a big PR). I don't mind changing it, but did you have something in mind? I'd like to keep the original variable name in the alias to avoid confusion, so pretending a `_` seemed like a reasonable option.

kpamnany: Can't think of anything reasonable, and not sure it's important.

charleskawczynski: Ok, I'll leave this for now. Should we raise an issue to change this in another PR?

PR #999, opened 2020-04-23
Improve output of `--help`

### LINKED ISSUE #996: Unclear command line options
# Description
The command line options in the new master have changed but the help list that is thrown to the user is not self-explanatory. In particular, in the list below, the following keywords need explaination:

`DIAGNOSTICS`
`VTK` 
`MONITOR-TIMESTEP-DURATION` 
`MONITOR-COURANT-NUMBERS`
`CHECKPOINT`
do they indicate time intervals? If so, should they be given in time steps or simulation times?
 
```
unrecognized option --enable-diagnostics
usage: dycoms-test-morediags.jl [--disable-gpu]
                        [--show-updates SHOW-UPDATES]
                        [--diagnostics DIAGNOSTICS] [--vtk VTK]
                        [--monitor-timestep-duration MONITOR-TIMESTEP-DURATION]
                                                [--monitor-courant-numbers MONITOR-COURANT-NUMBERS]
                        [--checkpoint CHECKPOINT]
                        [--checkpoint-keep-all] [--checkpoint-at-end]
                        [--checkpoint-dir CHECKPOINT-DIR]
                        [--restart-from-num RESTART-FROM-NUM]
                        [--log-level LOG-LEVEL]
                        [--output-dir OUTPUT-DIR]
                        [--integration-testing]
```
The wiki explanation should also be updated.

## Additional context

# For CLIMA Developers

kpamnany:
# Description

This will output what the WIKI [now shows](https://github.com/climate-machine/CLIMA/wiki/CLIMA-command-line-arguments). Closes #996.

I have

### DIFF
--- a/src/CLIMA.jl
+++ b/src/CLIMA.jl
@@ -1,5 +1,10 @@
 module CLIMA
 
+using Pkg.TOML
+
+const CLIMA_VERSION =
+    VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
+
 include(joinpath("Utilities", "TicToc", "TicToc.jl"))
 include(joinpath("Utilities", "ArtifactWrappers", "ArtifactWrappers.jl"))
 include(joinpath("InputOutput", "Writers", "Writers.jl"))
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -82,7 +82,22 @@ const Settings = CLIMA_Settings(array_type = Array)
 function parse_commandline(custom_settings)
     exc_handler =
         isinteractive() ? ArgParse.debug_handler : ArgParse.default_handler
-    s = ArgParseSettings(exc_handler = exc_handler)
+    s = ArgParseSettings(
+        prog = "CLIMA",
+        description = "Climate Machine: an Earth System Model that automatically learns from data\n",
+        preformatted_description = true,
+        epilog = """
+            Any <interval> unless otherwise stated may be specified as:
+                - 2hours or 10mins or 30secs => wall-clock time
+                - 9.5smonths or 3.3sdays or 1.5shours => simulation time
+                - 1000steps => simulation steps
+                - never => disable
+                - default => use experiment specified interval (only for diagnostics at present)
+            """,
+        preformatted_epilog = true,
+        version = string(CLIMA_VERSION),
+        exc_handler = exc_handler,
+    )
     add_arg_group!(s, "CLIMA")
 
     @add_arg_table! s begin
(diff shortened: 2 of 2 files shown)


src/Driver/Driver.jl
```diff
@@ -223,8 +248,8 @@ function init(; disable_gpu = false, arg_settings = nothing)
     end
 
     # set up the array type appropriately depending on whether we're using GPUs
-    if get(ENV, "CLIMA_GPU", "") != "false" && !Settings.disable_gpu &&
-       CUDAapi.has_cuda_gpu()
+    if get(ENV, "CLIMA_GPU", "") != "false" &&
+       !Settings.disable_gpu && CUDAapi.has_cuda_gpu()
```
charleskawczynski, line 252: Do we still need `atyp = CuArrays.CuArray` (in the next line) when `Settings.disable_gpu=false`?

src/CLIMA.jl
```diff
@@ -1,5 +1,9 @@
 module CLIMA
 
+const CLIMA_MAJOR_VERSION = 0
+const CLIMA_MINOR_VERSION = 0
+const CLIMA_VERSION = VersionNumber(CLIMA_MAJOR_VERSION, CLIMA_MINOR_VERSION)
```
charleskawczynski, line 5: Can we not get this from `Pkg` somehow?

src/Driver/Driver.jl
kpamnany: If we're _not_ disabling the GPU, then the array type should be `CuArrays.CuArray`.

src/CLIMA.jl
kpamnany: I don't know. Can we @simonbyrne?

src/Driver/Driver.jl
charleskawczynski: Not important here, but the logic might be easier to follow without double negatives (i.e., change the name to `enable_gpu` and flip the bools)

kpamnany: Yes, it confuses me too. But we want `--enable-gpu` to be the default. So the command line option needs to be `--disable-gpu`. I can invert it after parsing though so that it is `Settings.enable_gpu`. I'll do that.

src/CLIMA.jl
charleskawczynski: We can always update this later, so long this information is defined in a single place. Also, looks like we'll have an [experimental way](https://julialang.github.io/Pkg.jl/v1/api/#Pkg.project) to get this in 1.4.

simonbyrne: I'm not sure that would even work, since CLIMA could be a dependency of a different project.

simonbyrne:
```suggestion
using Pkg.TOML
const CLIMA_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__),"Project.toml"))["version"])
```

PR #1018, opened 2020-04-28
GCM diagnostics

### LINKED ISSUE #1165: Modify values of variable t and add pressure to LES diagnostics
# Description

Currently, in the LES diagnostics, the variable `t` with dimension (t) has values of 1, and another variable `simtime` with dimension (t) has values of the actual simulation time. It would be good to assign the actual simulation time to the variable `t`. It would also be good to assign the dimensions (t,z) to the variables which currently have the dimension (z). In addition, it would be good to have pressure in the LES diagnostics.

- [ ] Assign the actual simulation time to `t`, and change the dimension of variables to (t,z)
- [ ] Add pressure to LES diagnostics

# For CLIMA Developers

- [x] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)

LenkaNovak:
# Description

This adds a GCM diagnostics group. [These](https://github.com/CliMA/ClimateMachine.jl/blob/ln/new-dgngrp/src/Diagnostics/atmos_gcm_default.jl#L38-L67) diagnostics are currently in this group. Attributes have been added to the NetCDF files to comply with the CF convention. 

Some refactoring of the Diagnostics module accompanies this. Also:
- Includes #1082.
- Closes #1165.

Some remaining things will be addressed in subsequent PRs:
- Density weighting.
- More diagnostics:
  - Variances and covariances
  - Means/covariances based time-window convolution (e.g. Lanczos filtering)
  - Horizontal divergence, streamfunction (from the Laplacian of vorticity), potential vorticity
  - Ability for user-defined diagnostics (general kernels for gradients, etc.) 
- Aggregation of files using unlimited time dimension
- Support for all diagnostics to be collected on GPU (right now GPU used only where necessary, e.g. vorticity calculation)

The Held Suarez experiment now uses this group by default.

I have

### DIFF
--- a/experiments/AtmosGCM/heldsuarez.jl
+++ b/experiments/AtmosGCM/heldsuarez.jl
@@ -151,7 +151,7 @@ function held_suarez_forcing!(
 end
 
 function config_diagnostics(FT, driver_config)
-    interval = "100000steps" # chosen to allow a single diagnostics collection
+    interval = "40000steps" # chosen to allow a single diagnostics collection
 
     _planet_radius = FT(planet_radius(param_set))
 
@@ -167,12 +167,13 @@ function config_diagnostics(FT, driver_config)
         resolution,
     )
 
-    dgngrp = setup_dump_state_and_aux_diagnostics(
+    dgngrp = setup_atmos_default_diagnostics(
+        AtmosGCMConfigType(),
         interval,
         driver_config.name,
         interpol = interpol,
-        project = true,
     )
+
     return ClimateMachine.DiagnosticsConfiguration([dgngrp])
 end
 
@@ -182,7 +183,7 @@ function main()
     poly_order = 5                           # discontinuous Galerkin polynomial order
     n_horz = 5                               # horizontal element number
     n_vert = 5                               # vertical element number
-    n_days = 120                             # experiment day number
+    n_days = 50                              # experiment day number
     timestart = FT(0)                        # start time (s)
     timeend = FT(n_days * day(param_set))    # end time (s)
 
--- a/.dev/climaformat.jl
+++ b/.dev/climaformat.jl
@@ -77,7 +77,7 @@ if haskey(opts, :help)
     exit(0)
 end
 if isempty(ARGS)
-    filenames = readlines(`git diff --name-only --diff-filter=AM HEAD`)
+    filenames = readlines(`git diff --name-only --diff-filter=AM master`)
     filter!(f -> endswith(f, ".jl"), filenames)
 else
     filenames = ARGS
--- a/docs/src/DevDocs/DiagnosticVariables.md
+++ b/docs/src/DevDocs/DiagnosticVariables.md
@@ -1,10 +1,16 @@
-# CliMA Diagnostic Variable List
+# Diagnostic Variable List
 
-This document contains the diagnostic variables in CliMA.
+This is a list of all the diagnostic variables that the `ClimateMachine`
+can produce, partitioned into groups.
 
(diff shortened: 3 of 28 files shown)


bischtob:
@LenkaNovak, @kpamnany, is this ready to get in?

kpamnany:
I should be able to take a look starting Tuesday.

kpamnany:
This is mostly good to go. I don't understand why the variables in the NetCDF file seem to be written with the dimensions in the wrong order; I'll try and figure this out. In the meantime, please check for correctness.

bors try

szy21:
The LES diagnostics looks good to me. I will leave the naming of `internal enthalpy` to @akshaysridhar.

experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -151,13 +151,13 @@ function held_suarez_forcing!(
 end
 
 function config_diagnostics(FT, driver_config)
-    interval = "100000steps" # chosen to allow a single diagnostics collection
+    interval = "40000steps" # chosen to allow a single diagnostics collection
 
-    _planet_radius = FT(planet_radius(param_set))
+    _planet_radius = planet_radius(param_set)
```
charleskawczynski, line 156: In case this variable is used again, wouldn’t it be safer to type-cast once here?

kpamnany:
We seem to be doing it right. Opened an issue: https://github.com/Alexander-Barth/NCDatasets.jl/issues/87.

kpamnany:
@skandalaCLIMA : need your help here. A couple of changes in this PR are causing the interpolation test case to fail.

`u`, `v` and `w` were being mapped incorrectly (to `rad`, `lat` and `long`) in `project_cubed_sphere_CUDA!()` and have been changed in this PR to `long`, `lat` and `rad` [here](https://github.com/CliMA/ClimateMachine.jl/blob/ln/new-dgngrp/src/Numerics/Mesh/Interpolation.jl#L1624-L1626).

Another change to the interpolation code that might be a problem is [here](https://github.com/CliMA/ClimateMachine.jl/blob/ln/new-dgngrp/src/Numerics/Mesh/Interpolation.jl#L1804-L1810), with the order being changed from `rad`, `lat`, `long` to `long`, `lat`, `rad`.

Other changes to this file are mostly formatting related (and one function being added -- `dimensions()`), which should all be irrelevant to the test.

This is how we want interpolation to work, so please push a commit to this PR that fixes the interpolation test case?

kpamnany:
Regarding ordering, see the [explanation here](https://github.com/Alexander-Barth/NCDatasets.jl/issues/87#issuecomment-636098859).

The only solution is for us to explicitly transpose our arrays before writing them, which would be slow. Then they'd have to be re-transposed when reading them in from Julia (or from R, Matlab, etc.). Since we're writing our Viz framework in Julia, I think we just ignore this unless it's causing difficulties?

szy21:
> Regarding ordering, see the [explanation here](https://github.com/Alexander-Barth/NCDatasets.jl/issues/87#issuecomment-636098859).
> 
> The only solution is for us to explicitly transpose our arrays before writing them, which would be slow. Then they'd have to be re-transposed when reading them in from Julia (or from R, Matlab, etc.). Since we're writing our Viz framework in Julia, I think we just ignore this unless it's causing difficulties?

Sounds good to me. The order of dimensions doesn't seem to be a problem I think.

kpamnany:
Thanks to @skandalaCLIMA who pointed me at what to fix.

bors r+

PR #1026, opened 2020-04-29
Use Requires with Interpolation code to avoid loading CUDA packages

simonbyrne:
# Description

This should prevent the warnings, and hopefully speed up code loading when no GPU is present.

I have


simonbyrne:
Ah, this breaks our various code loading. Anyway it seems that CuArrays has removed the warning in the latest release, so we should just use that.

simonbyrne:
I couldn't get this to work correctly. I think the solution is to explicitly load CuArrays and ditch Requires.

PR #1035, opened 2020-04-30
Store the correct `DGmodel` in `solver_config`

kpamnany:
# Description

For Multirate solvers, the `DGModel` being stored in `solver_config` was not the one being updated.

I have

### DIFF
--- a/src/Driver/solver_configs.jl
+++ b/src/Driver/solver_configs.jl
@@ -94,7 +94,8 @@ function SolverConfiguration(
     numerical_flux_second_order = driver_config.numerical_flux_second_order
     numerical_flux_gradient = driver_config.numerical_flux_gradient
 
-    # create DG model, initialize ODE state
+    # Create the DG model and initialize the ODE state. If we're restarting,
+    # use state data from the checkpoint.
     if Settings.restart_from_num > 0
         s_Q, s_aux, t0 = Callbacks.read_checkpoint(
             Settings.checkpoint_dir,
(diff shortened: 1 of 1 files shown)


kpamnany:
This won't work. `RemainderModel has no field param_set`.

kpamnany:
For diagnostics, [this code](https://github.com/climate-machine/CLIMA/blob/master/src/Diagnostics/atmos_default.jl#L76-L78) uses the diffusive flux which it gets from the `DGModel`. When an experiment was switched to using Multirate, this diagnostic variable immediately went to 0.

src/Driver/solver_configs.jl
```diff
@@ -180,18 +182,18 @@ function SolverConfiguration(
...
-        slow_dg = DGModel(
+        dg = DGModel(
             slow_model,
             grid,
             numerical_flux_first_order,
             numerical_flux_second_order,
             numerical_flux_gradient,
-            state_auxiliary = dg.state_auxiliary,
+            state_auxiliary = state_auxiliary,
         )
```
jkozdon, line 195: Would this work?

```julia
dg = DGModel(
    balance_law,
    grid,
    numerical_flux_first_order,
    numerical_flux_second_order,
    numerical_flux_gradient,
    state_auxiliary = state_auxiliary,
)
slow dg = DGModel(
    slow_model,
    grid,
    numerical_flux_first_order,
    numerical_flux_second_order,
    numerical_flux_gradient,
    state_auxiliary = dg.state_auxiliary,
    state_gradient_flux = dg.state_gradient_flux,
    states_higher_order = dg.states_higher_order,
)
```

kpamnany: That would be much neater!

kpamnany: It appears that `fast_dg` is also using `dg.state_auxiliary`. Should these parameters (for `state_gradient_flux` and `states_higher_order` be added there as well? And if so, then should they also be applied to the vertical `DGModel` created for IMEX solvers?

jkozdon: possibly, though in current use cases the fast (linear) models do not have diffusion so it should have no impact.

PR #1056, opened 2020-05-06
Add instructions for creating custom system image

jkozdon:
CLIMA startup times are long, which can make iterative development and
debugging difficult / painful. This adds a process for creating a custom
system image which drastically reduces initialization time.

|                      command                      |  no image |  with image |
|---------------------------------------------------|-----------|-------------|
| experiments/AtmosLES/dycoms.jl --help             |   31.44s  |       2.5s  |
| experiments/AtmosLES/dycoms.jl                    |  244.03s  |    138.64s  |
| test/Numerics/DGmethods/Euler/isentropicvortex.jl |  229.62s  |    144.19s  |

I have

### DIFF
--- a/.dev/systemimage/climate_machine_image.jl
+++ b/.dev/systemimage/climate_machine_image.jl
@@ -1,0 +1,33 @@
+#!/usr/bin/env julia
+#
+# Called with no arguments will create the system image
+#     ClimateMachine.so
+# in the `@__DIR__` directory.
+#
+# Called with a single argument the system image will be placed in the path
+# specified by the argument (relative to the callers path)
+sysimage_path =
+    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+
+using Pkg
+Pkg.add("PackageCompiler")
+using PackageCompiler
+Pkg.activate(joinpath(@__DIR__, "..", ".."))
+Pkg.instantiate()
+pkgs = Pkg.installed()
+delete!(pkgs, "Pkg")
+
+PackageCompiler.create_sysimage(
+    [Symbol(s) for s in keys(pkgs)];
+    sysimage_path = sysimage_path,
+    precompile_execution_file = joinpath(
+        @__DIR__,
+        "..",
+        "..",
+        "test",
+        "Numerics",
+        "DGmethods",
+        "Euler",
+        "isentropicvortex.jl",
+    ),
+)
--- a/.dev/systemimage/climate_machine_image.jl
+++ b/.dev/systemimage/climate_machine_image.jl
@@ -17,17 +17,22 @@ Pkg.instantiate()
 pkgs = Pkg.installed()
 delete!(pkgs, "Pkg")
 
-PackageCompiler.create_sysimage(
-    [Symbol(s) for s in keys(pkgs)];
-    sysimage_path = sysimage_path,
-    precompile_execution_file = joinpath(
-        @__DIR__,
-        "..",
-        "..",
-        "test",
-        "Numerics",
-        "DGmethods",
-        "Euler",
-        "isentropicvortex.jl",
-    ),
-)
+# "TF_BUILD" is check to see if we are running on Azure for CI
+if !haskey(ENV, "TF_BUILD")
+    PackageCompiler.create_sysimage(
+        [Symbol(s) for s in keys(pkgs)];
+        sysimage_path = sysimage_path,
+        precompile_execution_file = joinpath(
+            @__DIR__,
+            "..",
+            "..",
+            "test",
+            "Numerics",
+            "DGmethods",
+            "Euler",
+            "isentropicvortex.jl",
+        ),
+    )
+else
+    return true
+end
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ src/**/Manifest.toml
 *.jpg
 *.jpeg
 *.svg
+
+# Julia System Images
+*.so
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,6 +30,7 @@ jobs:
       ./julia-$(JULIA_VERSION)/bin/julia -e 'using InteractiveUtils; versioninfo()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.instantiate()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.test()'
+      ./julia-$(JULIA_VERSION)/bin/julia -e 'using Test; @testset "system image" begin;@test include(".dev/systemimage/climate_machine_image.jl"); end;'
     displayName: 'Run the tests'
     continueOnError: true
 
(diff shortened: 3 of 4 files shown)


.dev/systemimage/climate_machine_image.jl
```diff
@@ -0,0 +1,27 @@
...
+# specified by the argument (relative to the callers path)
+sysimage_path =
+    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+
+using Pkg
+Pkg.add("PackageCompiler")
+using PackageCompiler
+Pkg.activate(joinpath(@__DIR__, "../../"))
+Pkg.instantiate()
+pkgs = Pkg.installed()
```
jakebolewski, line 17: Should this be constrained to just the whitelisted CliMA project dependencies? (ex. a user installs additional packages into their project environment could lead to divergent system images maybe causing confusion / debugging issues)

jkozdon: The installed only includes packages from the project.

As seen below after the `Pkg.activate` both PackageCompiler and Revise are not listed by `Pkg.installed`

```
❯ julia -e "using Pkg; display(Pkg.installed()); println(); println();Pkg.activate(@__DIR__); println(); display(Pkg.installed());"
Dict{String,Union{Nothing, VersionNumber}} with 2 entries:
  "PackageCompiler" => v"1.1.1"
  "Revise" => v"2.6.4"

Activating environment at `~/codes/CLIMA/Project.toml`

Dict{String,Union{Nothing, VersionNumber}} with 34 entries:
  "Pkg" => nothing
  "Statistics" => nothing
  "ForwardDiff" => v"0.10.10"
  "Coverage" => v"1.0.0"
  "Distributions" => v"0.23.2"
  "MPI" => v"0.13.0"
  "CLIMAParameters" => v"0.1.3"
  "Random" => nothing
  "GaussQuadrature" => v"0.5.4"
  "CUDAdrv" => v"6.0.0"
  "OrderedCollections" => v"1.1.0"
  "Literate" => v"2.3.1"
  "DelimitedFiles" => nothing
  "LinearAlgebra" => nothing
  "DoubleFloats" => v"1.1.8"
  "GPUifyLoops" => v"0.2.9"
  "Combinatorics" => v"1.0.0"
  "CUDAapi" => v"3.1.0"
  "LazyArrays" => v"0.16.5"
  "CuArrays" => v"1.7.3"
  "DocStringExtensions" => v"0.8.1"
  "WriteVTK" => v"1.4.1"
  "CUDAnative" => v"2.10.2"
  "FileIO" => v"1.2.4"
  "StaticArrays" => v"0.12.1"
  "JLD2" => v"0.1.13"
  "NCDatasets" => v"0.10.2"
  "Dates" => nothing
  "SpecialFunctions" => v"0.10.0"
  "Logging" => nothing
  "Printf" => nothing
  "KernelAbstractions" => v"0.2.3"
  "RootSolvers" => v"0.2.0"
  "ArgParse" => v"1.1.0"
```

simonbyrne:
As a future reference point, we should also look at which methods we invalidate: see https://github.com/JuliaLang/julia/pull/35714

jakebolewski:
Was able to reproduce, -> ~2s startup. It doesn't seem like the generated PackageCompiler system image checks if the Julia minor run time versions are the same though in the so file (leads to segfault).

jkozdon:
> Was able to reproduce, -> ~2s startup. It doesn't seem like the generated PackageCompiler system image checks if the Julia minor run time versions are the same though in the generated so file (leads to segfault).

Good to know. Once CLIMA is running on 1.4 we'll need to change this anyway since [`installed` is begin deprecated](https://github.com/JuliaLang/Pkg.jl/blob/127412cd23224a4f8fcdb049d80c4c16486a1e93/src/Pkg.jl#L531-L541) in favor of [`dependencies`](https://julialang.github.io/Pkg.jl/dev/api/#Pkg.dependencies)

jkozdon:
The other thing I do not like about this is that using `isentropic_vortex` is a bit heavy of an initialization file (don't need to run that long). Open to other suggestions.

kpamnany:
IMO this should be in v0.1.

.dev/systemimage/climate_machine_image.jl
```diff
@@ -0,0 +1,27 @@
...
+delete!(pkgs, "Pkg")
+
+PackageCompiler.create_sysimage(
+    [Symbol(s) for s in keys(pkgs)];
+    sysimage_path = sysimage_path,
+    precompile_execution_file = joinpath(
+        @__DIR__,
+        "../../test/Numerics/DGmethods/Euler/isentropicvortex.jl",
```
charleskawczynski, line 25:
```suggestion
        "..","..","test","Numerics","DGmethods","Euler","isentropicvortex.jl",
```

.dev/systemimage/climate_machine_image.jl
```diff
@@ -0,0 +1,27 @@
...
+#
+# Called with a single argument the system image will be placed in the path
+# specified by the argument (relative to the callers path)
+sysimage_path =
+    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+
+using Pkg
+Pkg.add("PackageCompiler")
+using PackageCompiler
+Pkg.activate(joinpath(@__DIR__, "../../"))
```
charleskawczynski, line 15:
```suggestion
Pkg.activate(joinpath(@__DIR__, "..",".."))
```

jakebolewski:
Probably should add a sanity check test as well to make sure we catch instances when this breaks on PR's.

jkozdon:
> IMO this should be in v0.1.

For users (as opposed to folks not hacking the CLIMA package itself), you could include a version that just depends on CLIMA itself, though there you would want to use as a precompile script something that uses the driver.

jkozdon:
> Probably should add a sanity check test as well to make sure we catch instances when this breaks on PR's.

How are you thinking that this could be done?

jakebolewski:
Maybe just a sanity check for now, invoke the command to run the script and make sure it doesn't segfault.

jkozdon:
> Maybe just a sanity check for now, invoke the command / subprocess to run the script and make sure it doesn't segfault.

Not sure what you mean.

jakebolewski:
@jkozdon I guess my question is if we want to ensure that this continues to work going forward (ex. include a test that precompilation is successful on CI), or if it's considered experimental for now.

jkozdon:
> @jkozdon I guess my question is if we want to ensure that this continues to work going forward (ex. include a test that precompilation / creating a custom system image is successful on CI), or if it's considered experimental for now.

Something like https://github.com/CliMA/ClimateMachine.jl/pull/1056/commits/18b4bb5f57ced4b9638309717d23dcd2c35da711?

I don't think we want to actually build the full system image in the CI, but I could be wrong.

jkozdon:
Possible the new MPI killed this.

Investigating more locally...

jkozdon:
> Possible the new MPI killed this.
> 
> Investigating more locally...

nevermind, was a personal issue

bor r+

PR #1059, opened 2020-05-06
Make moist thermo tolerance residual-based

charleskawczynski:
# Description

Changes Moist Thermo's saturation adjustment tolerance to accept the residual tolerance and not the solution tolerance.

 - Some of the tolerances in moist thermo's test suite need to be reviewed

I have

### DIFF
--- a/src/Common/MoistThermodynamics/MoistThermodynamics.jl
+++ b/src/Common/MoistThermodynamics/MoistThermodynamics.jl
@@ -27,6 +27,7 @@ module MoistThermodynamics
 
 using DocStringExtensions
 using RootSolvers
+using RootSolvers: AbstractTolerance
 
 using CLIMAParameters: AbstractParameterSet
 using CLIMAParameters.Planet
--- a/src/Common/MoistThermodynamics/states.jl
+++ b/src/Common/MoistThermodynamics/states.jl
@@ -89,19 +89,25 @@ Moist thermodynamic phase, given
 and, optionally
  - `maxiter` maximum iterations for saturation adjustment
  - `temperature_tol` temperature tolerance for saturation adjustment
- - `sat_adjust` function pointer to particular saturation adjustment function
+ - `sat_adjust` function pointer to particular saturation adjustment method, options include
+    - `saturation_adjustment` uses Newtons method with analytic gradients
+    - `saturation_adjustment_SecantMethod` uses Secant method
 """
 function PhaseEquil(
     param_set::APS,
     e_int::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 3,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 6,
+    temperature_tol::FT = FT(1e-1),
     sat_adjust::Function = saturation_adjustment,
 ) where {FT <: Real}
     # TODO: Remove these safety nets, or at least add warnings
     # waiting on fix: github.com/vchuravy/GPUifyLoops.jl/issues/104
+
+    _cv_d = FT(cv_d(param_set))
+    # Convert temperature tolerance to a convergence criterion on internal energy residuals
+    tol = ResidualTolerance(temperature_tol * _cv_d)
     q_tot_safe = clamp(q_tot, FT(0), FT(1))
     T = sat_adjust(param_set, e_int, ρ, q_tot_safe, maxiter, tol)
     return PhaseEquil{FT, typeof(param_set)}(param_set, e_int, ρ, q_tot_safe, T)
@@ -156,7 +162,7 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
  - `θ_liq_ice` liquid-ice potential temperature
  - `ρ` (moist-)air density
  - `q_tot` total specific humidity
- - `tol` tolerance for saturation adjustment
+ - `temperature_tol` temperature tolerance for saturation adjustment
  - `maxiter` maximum iterations for saturation adjustment
 """
 function LiquidIcePotTempSHumEquil(
@@ -164,9 +170,10 @@ function LiquidIcePotTempSHumEquil(
     θ_liq_ice::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 30,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 36,
+    temperature_tol::FT = FT(1e-1),
 ) where {FT <: Real}
+    tol = ResidualTolerance(temperature_tol)
     T = saturation_adjustment_q_tot_θ_liq_ice(
         param_set,
         θ_liq_ice,
@@ -189,7 +196,7 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
  - `θ_liq_ice` liquid-ice potential temperature
  - `p` pressure
  - `q_tot` total specific humidity
- - `tol` tolerance for saturation adjustment
+ - `temperature_tol` temperature tolerance for saturation adjustment
  - `maxiter` maximum iterations for saturation adjustment
 """
 function LiquidIcePotTempSHumEquil_given_pressure(
@@ -198,8 +205,9 @@ function LiquidIcePotTempSHumEquil_given_pressure(
     p::FT,
     q_tot::FT,
     maxiter::Int = 30,
-    tol::FT = FT(1e-1),
+    temperature_tol::FT = FT(1e-1),
 ) where {FT <: Real}
+    tol = ResidualTolerance(temperature_tol)
     T = saturation_adjustment_q_tot_θ_liq_ice_given_pressure(
         param_set,
         θ_liq_ice,
@@ -280,7 +288,7 @@ Constructs a [`PhaseNonEquil`](@ref) thermodynamic state from:
  - `ρ` (moist-)air density
  - `q_pt` phase partition
 and, optionally
- - `tol` tolerance for non-linear equation solve
+ - `potential_temperature_tol` potential temperature for non-linear equation solve
  - `maxiter` maximum iterations for non-linear equation solve
 """
 function LiquidIcePotTempSHumNonEquil(
@@ -288,9 +296,10 @@ function LiquidIcePotTempSHumNonEquil(
     θ_liq_ice::FT,
     ρ::FT,
     q_pt::PhasePartition{FT},
-    maxiter::Int = 5,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 10,
+    potential_temperature_tol::FT = FT(1e-2),
 ) where {FT <: Real}
+    tol = ResidualTolerance(potential_temperature_tol)
     T = air_temperature_from_liquid_ice_pottemp_non_linear(
         param_set,
         θ_liq_ice,
--- a/test/Common/MoistThermodynamics/runtests.jl
+++ b/test/Common/MoistThermodynamics/runtests.jl
@@ -2,6 +2,7 @@ using Test
 using ClimateMachine.MoistThermodynamics
 using NCDatasets
 using Random
+using RootSolvers
 MT = MoistThermodynamics
 using LinearAlgebra
 
@@ -258,7 +259,7 @@ end
         ρ,
         q_tot,
         10,
-        1e-2,
+        ResidualTolerance(1e-2),
     ) ≈ 300.0
     @test abs(
         MT.saturation_adjustment(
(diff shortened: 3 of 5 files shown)


src/Common/MoistThermodynamics/MoistThermodynamics.jl
```diff
@@ -982,7 +983,7 @@ function saturation_adjustment(
     ρ::FT,
     q_tot::FT,
     maxiter::Int,
-    tol::FT,
+    tol::AbstractTolerance,
```
tapios, line 986: It would be good to make clearer in the documentation above what this tolerance is precisely.

src/Common/MoistThermodynamics/states.jl
```diff
@@ -78,18 +78,36 @@ struct PhaseEquil{FT, PS} <: ThermodynamicState{FT}
...
+ - `sat_adjust` function pointer to particular saturation adjustment function
+"""
 function PhaseEquil(
     param_set::APS,
     e_int::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 3,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 5,
```
tapios, line 102: Do we need 'maxiter' to be 5 in the tests. Or does 3 suffice?

src/Common/MoistThermodynamics/states.jl
```diff
@@ -78,18 +78,36 @@ struct PhaseEquil{FT, PS} <: ThermodynamicState{FT}
...
+    temperature_tol::FT = FT(1e-1),
     sat_adjust::Function = saturation_adjustment,
 ) where {FT <: Real}
     # TODO: Remove these safety nets, or at least add warnings
     # waiting on fix: github.com/vchuravy/GPUifyLoops.jl/issues/104
     q_tot_safe = clamp(q_tot, FT(0), FT(1))
+    _cp_d = FT(cp_d(param_set))
+    tol = ResidualTolerance(temperature_tol*_cp_d)
```
tapios, line 110: Add an explanation here:
```suggestion
    # Convert temperature tolerance to a convergence criterion on internal energy residuals 
    tol = ResidualTolerance(temperature_tol*_cv_d)
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -78,18 +78,36 @@ struct PhaseEquil{FT, PS} <: ThermodynamicState{FT}
...
-    maxiter::Int = 3,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 5,
+    temperature_tol::FT = FT(1e-1),
     sat_adjust::Function = saturation_adjustment,
 ) where {FT <: Real}
     # TODO: Remove these safety nets, or at least add warnings
     # waiting on fix: github.com/vchuravy/GPUifyLoops.jl/issues/104
     q_tot_safe = clamp(q_tot, FT(0), FT(1))
+    _cp_d = FT(cp_d(param_set))
```
tapios, line 109:
```suggestion
    _cv_d = FT(cv_d(param_set))
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -143,17 +161,19 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
  - `θ_liq_ice` liquid-ice potential temperature
  - `ρ` (moist-)air density
  - `q_tot` total specific humidity
- - `tol` tolerance for saturation adjustment
+ - `temperature_tol` estimated temperature tolerance for saturation adjustment
```
tapios, line 164:
```suggestion
 - `temperature_tol` temperature tolerance for saturation adjustment
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -143,17 +161,19 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
...
 """
 function LiquidIcePotTempSHumEquil(
     param_set::APS,
     θ_liq_ice::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 30,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 40,
+    temperature_tol::FT = FT(1e-5),
```
tapios, line 173: This seems unreasonably small. Why? We will never know temperature to this accuracy. Better:
```suggestion
    temperature_tol::FT = FT(1e-2),
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -143,17 +161,19 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
...
  - `maxiter` maximum iterations for saturation adjustment
 """
 function LiquidIcePotTempSHumEquil(
     param_set::APS,
     θ_liq_ice::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 30,
-    tol::FT = FT(1e-1),
+    maxiter::Int = 40,
```
tapios, line 172: Seems much too high:
```suggestion
    maxiter::Int = 10,
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -176,7 +196,7 @@ Constructs a [`PhaseEquil`](@ref) thermodynamic state from:
  - `θ_liq_ice` liquid-ice potential temperature
  - `p` pressure
  - `q_tot` total specific humidity
- - `tol` tolerance for saturation adjustment
+ - `temperature_tol` estimated temperature tolerance for saturation adjustment
```
tapios, line 199:
```suggestion
 - `temperature_tol` temperature tolerance for saturation adjustment
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -185,8 +205,10 @@ function LiquidIcePotTempSHumEquil_given_pressure(
     p::FT,
     q_tot::FT,
     maxiter::Int = 30,
-    tol::FT = FT(1e-1),
+    temperature_tol::FT = FT(1e-3),
```
tapios, line 208:
```suggestion
    potential_temperature_tol::FT = FT(1e-2),
```
(Please do not make this too small. We do not have temperatures or potential temperatures to greater accuracy than 1e-2 K.)

src/Common/MoistThermodynamics/states.jl
```diff
@@ -185,8 +205,10 @@ function LiquidIcePotTempSHumEquil_given_pressure(
...
 ) where {FT <: Real}
+    _cp_d = FT(cp_d(param_set))
```
tapios, line 210:
```suggestion
    _cp_d = FT(cp_d(param_set))
```
(don't need it)

src/Common/MoistThermodynamics/states.jl
```diff
@@ -185,8 +205,10 @@ function LiquidIcePotTempSHumEquil_given_pressure(
...
+    tol = ResidualTolerance(temperature_tol*_cp_d)
```
tapios, line 211:
```suggestion
    tol = ResidualTolerance(potential_temperature_tol)
```
(we minimize potential temperature residuals, right? So no cp/cv factors here)

src/Common/MoistThermodynamics/states.jl
```diff
@@ -267,7 +289,7 @@ Constructs a [`PhaseNonEquil`](@ref) thermodynamic state from:
  - `ρ` (moist-)air density
  - `q_pt` phase partition
 and, optionally
- - `tol` tolerance for non-linear equation solve
+ - `temperature_tol` estimated temperature tolerance for saturation adjustment
```
tapios, line 292: I thought this is non-equilibrium, so why do we need any saturation adjustment? (If we do, as above, the tolerance should be a potential temperature tolerance.)

src/Common/MoistThermodynamics/states.jl
```diff
@@ -276,8 +298,10 @@ function LiquidIcePotTempSHumNonEquil(
     ρ::FT,
     q_pt::PhasePartition{FT},
     maxiter::Int = 5,
-    tol::FT = FT(1e-1),
+    temperature_tol::FT = FT(1e-3),
```
tapios, line 301: I don't know why we need this in non-equilibrium (by definition, saturation adjustment does not apply). If we do for some reason, it should be 
```suggestion
    potential_temperature_tol::FT = FT(1e-2),
```
(or 1e-1).

src/Common/MoistThermodynamics/states.jl
```diff
@@ -276,8 +298,10 @@ function LiquidIcePotTempSHumNonEquil(
...
 ) where {FT <: Real}
+    _cp_d = FT(cp_d(param_set))
```
tapios, line 303:
```suggestion
```

src/Common/MoistThermodynamics/states.jl
```diff
@@ -276,8 +298,10 @@ function LiquidIcePotTempSHumNonEquil(
...
+    tol = ResidualTolerance(temperature_tol*_cp_d)
```
tapios, line 304: Again, we shouldn't need any of this. If we do, it should be 
```suggestion
    tol = ResidualTolerance(potential_temperature_tol)
```

test/Common/MoistThermodynamics/runtests.jl
```diff
@@ -404,8 +405,8 @@ end
                 θ_liq_ice,
                 ρ,
                 q_tot,
-                40,
-                FT(1e-3),
+                100,
+                FT(1e-6),
```
tapios, line 409: Seems like unattainable accuracy in practice. Smaller than 1e-2 or maybe 1e-3 does not make sense in practice.

src/Common/MoistThermodynamics/states.jl
charleskawczynski: This non-linear equation shows up when we need to construct a non-equilibrium phase with `θ_liq_ice, ρ, q_pt` (because we must somehow compte temperature from these arguments). We replaced the particular equation docs with
`T - air_temperature_from_liquid_ice_pottemp_given_pressure(param_set, θ_liq_ice, air_pressure(param_set, T, ρ, q), q) = 0`
This should only ever be used for initialization when [`LiquidIcePotTempSHumNonEquil`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Common/MoistThermodynamics/states.jl#L273) constructor is used. Also, it's used as a [comparison against our approximation in `air_temperature_from_liquid_ice_pottemp`](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Common/MoistThermodynamics/runtests.jl#L615).

src/Common/MoistThermodynamics/states.jl
smarras79: I found by trial and error that a default value of at least 1 helps. @tapios would 1 be too large as a default?

src/Common/MoistThermodynamics/states.jl
tapios: Ok. Saturation adjustment is a misnomer in this case then. We do not adjust the saturation state (it is given by `q`). It is just a nonlinear solve for temperature. It would be good to make this clearer.

src/Common/MoistThermodynamics/states.jl
tapios: @smarras79 We changed what the convergence criterion means. In DYCOMS, `1e-1` should work fine with 1 iteration, from the tests we looked at.

src/Common/MoistThermodynamics/states.jl
charleskawczynski: Turns out we need 6 iterations at `temperature_tol = 1e-1`

src/Common/MoistThermodynamics/states.jl
charleskawczynski: At `temperature_tol=1e-1`, the best we can do is `maxiter=36`

test/Common/MoistThermodynamics/runtests.jl
charleskawczynski: Updated this to something more reasonable

src/Common/MoistThermodynamics/states.jl
charleskawczynski: This shouldn't be a problem since all constructors aside from
 - `PhaseDry`
 - `PhaseEquil`
 - `PhaseNonEquil`
are only used for initialization

src/Common/MoistThermodynamics/states.jl
charleskawczynski: I updated this to `temperature_tol=1e-1`, and there, the best we can do is `maxiter=36`

PR #1062, opened 2020-05-07
Pass named tuples to BalanceLaw functions to simplify function signatures

blallen:
# Description

This places all Vars/Grad Arrays passed to BalanceLaw functions, i.e. `flux_first_order!`, `compute_gradient_argument!`, and `boundary_state!`, into a single NamedTuple, simplifying the function signature. 

There are two main purposes for this: 1) developers don't have to update every single BalanceLaw when they want to change the MPIStateArrays required for a function and 2) make there only be one function signature for `boundary_state!` that just dispatches on the NumericalFlux and BalanceLaw.

This PR is a WIP. I've modified the HydrostaticBoussinesqModel to show how these changes would affect a balance law in practice but there are still some kinks that need to be worked out.

I have


src/Numerics/DGmethods/NumericalFluxes.jl
```diff
@@ -272,39 +239,33 @@ struct CentralNumericalFluxFirstOrder <: NumericalFluxFirstOrder end
...
-) where {S, A}
+)
 
-    FT = eltype(fluxᵀn)
+    FT = eltype(states.flux)
     num_state_conservative = number_state_conservative(balance_law, FT)
-    fluxᵀn = parent(fluxᵀn)
 
-    flux⁻ = similar(fluxᵀn, Size(3, num_state_conservative))
+    flux⁻ = similar(states.flux, Size(3, num_state_conservative))
```
blallen, line 250: The `similar` call here fails because it doesn't properly guess the type of `states.flux`

jkozdon:
> This places all MPIStateArrays passed to BalanceLaw functions, i.e. `flux_first_order!`, `compute_gradient_argument!`, and `boundary_state!`, into a single NamedTuple, simplifying the function signature.

Just for clarification, none of these functions take an MPIStateArray, they all take arrays, but they are not MPIStateArrays

blallen:
> > This places all MPIStateArrays passed to BalanceLaw functions, i.e. `flux_first_order!`, `compute_gradient_argument!`, and `boundary_state!`, into a single NamedTuple, simplifying the function signature.
> 
> Just for clarification, none of these functions take an MPIStateArray, they all take arrays, but they are not MPIStateArrays

Yes, sorry. I normally think of the Vars/Grad arrays at each node as a view of the larger MPIStateArray, but they aren't that exactly. Updated the PR description.

blallen:
> Mixed feelings on this. Seems to obfuscate the function arguments a bit IMO, now I have to jump back through several files to see what I have access to in a function.
> 
> That said, as we have more arguments and balance laws updating these is a pain.
> 
> What about using a struct for each function, so that you know what arguments you have instead of a `NamedTuple`? Then there is a single point of documentation for what needs to be filled for each typeof function you want to call.
> 
> There could be value to putting all the arguments in such a struct, not just the arrays. For my remainder model PR I had to add direction to a few functions, and the current structure of this would not have kept me from having to modify a bunch of files.

If we could make a callable struct for each of these functions, as we did with the DGmodel, that would be great for me. The real point behind this is to make changing the function signatures easier and not require jumping into every BalanceLaw. Simplifying the necessary documentation is a good outcome as well.

blallen:
#1420 is a better attempt at this

PR #1071, opened 2020-05-08
Manipulating adapted MPIStateArrays

leios:
# Description

This is a draft to alow for manipulating MPIStateArrays with wrapper types like PermutedDims and reshapedArrays.

I have an initial test right now for reshaped MPIStateArrays, but cannot quite get the `copy(...)` or `similar(...)` functions to work as expected. Broadcasting seems to work fine, though.

I have


src/Arrays/MPIStateArrays.jl
```diff
@@ -756,9 +791,11 @@ end
 # better names, for example `device` is also defined in CUDAdrv
 
 device(::Union{Array, SArray, MArray}) = CPU()
+device(Q::MPIDestArray) = device(parent(Q))
```
vchuravy, line 794:
```
for (W, _) in Adapt.wrappers
    @eval device(wrapper::$W where {AT <: Any}) = device(parent(wrapper))
end
```

Instead of specialising on the Union.

leios:
bors try

leios:
bors try

leios:
bors try

leios:
bors try-

leios:
bors try

leios:
bors try

src/Arrays/MPIStateArrays.jl
```diff
@@ -300,40 +326,55 @@ function Base.similar(
...
 end
 
+@inline Base.copyto!(
+    dest::MPIDestArray,
+    bc::Broadcasted{<:Broadcast.AbstractArrayStyle{0}},
+) = copyto!(dest, convert(Broadcasted{Nothing}, bc))
+
+# The following `__no_overlap_*` functions exist to support the old DG balance
+# law where we still use GPUifyLoops.  They should NOT be used in new code.
+function __no_overlap_begin_ghost_exchange!(Q::MPIStateArray)
```
vchuravy, line 366: This looks like a rebase error.

src/Arrays/MPIStateArrays.jl
```diff
@@ -735,15 +776,22 @@ function Base.mapreduce(
     MPI.Allreduce(cpuify(locreduce), max, Q.mpicomm)
 end
 
+# `realview` and `array_device` are helpers that enable
```
vchuravy, line 779: comment needs to be partially moved. `realview` is now defined higher-up.

test/Arrays/broadcasting.jl
```diff
@@ -16,8 +16,8 @@ const mpicomm = MPI.COMM_WORLD
         QA = MPIStateArray{Float32}(mpicomm, ArrayType, localsize...)
         QB = similar(QA)
 
-        QA .= A
-        QB .= B
+        copyto!(QA, A)
```
vchuravy, line 19: @lcw @jkozdon I think it is cleaner to make the data-movement explicit instead of having it be specialised on a particular form of broadcast. Heterogenous broadcasts are not supported as so I could see a user get confused about this one special case.

jkozdon: I thought this was originally added for the ODE solver stuff, but maybe @mwarusz knows (he added all the original broadcasting stuff)

leios:
bors try

leios:
So the reshape test is working, which is all we need for the Ocean set-up; however, this PR should also be able to support other adapted types:
```
julia> for wrapper in Adapt.wrappers
           println(wrapper)
       end
:(SubArray{<:Any, <:Any, AT}) => Adapt.var"#1#12"()
:(PermutedDimsArray{<:Any, <:Any, <:Any, <:Any, AT}) => Adapt.var"#2#13"()
:(Base.ReshapedArray{<:Any, <:Any, AT, <:Any}) => Adapt.var"#3#14"()
:(LinearAlgebra.Adjoint{<:Any, AT}) => Adapt.var"#4#15"()
:(LinearAlgebra.Transpose{<:Any, AT}) => Adapt.var"#5#16"()
:(LinearAlgebra.LowerTriangular{<:Any, AT}) => Adapt.var"#6#17"()
:(LinearAlgebra.UnitLowerTriangular{<:Any, AT}) => Adapt.var"#7#18"()
:(LinearAlgebra.UpperTriangular{<:Any, AT}) => Adapt.var"#8#19"()
:(LinearAlgebra.UnitUpperTriangular{<:Any, AT}) => Adapt.var"#9#20"()
:(LinearAlgebra.Diagonal{<:Any, AT}) => Adapt.var"#10#21"()
:(LinearAlgebra.Tridiagonal{<:Any, AT}) => Adapt.var"#11#22"()

```

The problem is that some of these do not work at all with the current version of `MPIStateArrays`

Do we want to fully support all these types? Or just support them one-by-one as they are needed? I could see `PermutedDimsArays` and `SubArrays` being useful right now, but I am not sure about the LinearAlgebra types because we have specific methods for these operations that do not return the adapted versions.

For example, we would define our own method that returns a tridiagonal `MPIStateArray`, not a `LinearAlgebra.Tridiagonal{MPIStateArray...}`, so there is no reason to support an MPIStateArray that has been adapted into a tridiagonal form.

src/Arrays/MPIStateArrays.jl
```diff
@@ -352,7 +353,20 @@ function Base.copyto!(dest::MPIDestArray, src::MPIDestArray)
 end
 
 @inline function Base.copyto!(dest::MPIDestArray, bc::Broadcasted{Nothing})
-    copyto!(realview(dest), transform_broadcasted(bc, dest))
+    # check for the case a .= b, where b is an array
+    if bc.f === identity && bc.args isa Tuple{AbstractArray}
```
vchuravy, line 357: I still think that this is fragile and encouraged the wrong code.

leios: Yeah, I tried to get everything to work out without this functionality over the weekend, but failed. Some initializations must be done on the CPU first (like when random numbers are involved).

I don't really like the CPU->GPU init method, so maybe we look into all reasons why such operations must be initialized on the CPU and fix them so they can be initialized on the GPU instead?

After that, it should be easy to remove this code.

vchuravy: Right, or as elsewhere in the PR `dest .= src` needs to be rewriten as `copyto!(dest, src)`.

leios: This functionality has been adapted into a [commented function in the current PR](https://github.com/CliMA/ClimateMachine.jl/pull/1071/files#diff-0c102a9f3b330c12cac54a2a4821a48dR350). This function calls a [copyto method in GPUArrays](https://github.com/CliMA/ClimateMachine.jl/pull/1071/files#diff-0c102a9f3b330c12cac54a2a4821a48dR350) that copies twice, so it is probably a bit slow to use.

@mwarusz Could we chat to make sure all the copyto functionality is doing what it is intended to?

mwarusz: Sure, thanks for cleaning up my broadcasting code. I believe that the only special functionality that I wanted to support
was to make MPIStateArrays broadcast only over the real elements (although it have never checked if this gives us any measurable performance benefit so maybe we should rethink this ?). I agree with Valentin that `copyto!` is cleaner for just transferring data.

leios:
bors try

leios:
bors try

leios:
bors try

leios:
bors try

leios:
bors try

leios:
bors try

leios:
It looks like all the bors tests are passing now (after I reran the docs build for some reason). I removed the draft label and am happy to hear how we can improve this if anyone has suggestions

src/Arrays/MPIStateArrays.jl
```diff
@@ -281,9 +304,17 @@ Base.setindex!(Q::MPIStateArray, x...; kw...) =
 
 Base.eltype(Q::MPIStateArray, x...; kw...) = eltype(Q.data, x...; kw...)
 
-Base.Array(Q::MPIStateArray) = Array(Q.data)
+Base.Array(Q::MPIDestArray) = Array(rawview(Q))
```
vchuravy, line 307: I think we are overusing the Union `MPIDestArray`. It should only be necessary for the LHS of a broadcast `copyto!`

src/Arrays/MPIStateArrays.jl
```diff
@@ -302,40 +333,43 @@ function Base.similar(
 end
 
 # transform all arguments of `bc` from MPIStateArrays to Arrays
+function transform_broadcasted(bc::Broadcasted, dest::MPIDestArray)
```
vchuravy, line 336:
```suggestion
function transform_broadcasted(bc::Broadcasted, dest)
```

src/Arrays/MPIStateArrays.jl
leios: Is there any down-side for doing it this way, though? I mean, why not generalize this to an MPIDestArray?

I could imagine people just wanting to transfer an MPIDestArray to the CPU and doing it in this way.

PR #1078, opened 2020-05-10
Fix docs, execute ipynb

charleskawczynski:
# Description

 - Fixes broken docs, both in `docs/` and in `tutorials/`
 - Turns back on `Literate.notebook(input, gen_dir, execute = true)`
 - This will allow us to include figures in our Literate tutorials (figs are not generated unless the ipynb executes).
 - Removes `println("Generated files: $(generated_files)")`, since this is very noisy when the ipynb executes.

Fixes ipynb execution for the following tutorials:
 - `topo.jl`
 - `nonnegative.jl`
 - `ex_1_saturation_adjustment.jl`
 - `ex_2_Kessler.jl`

I ran into problems with
 - `heldsuarez` (breaks)
 - `dry_rayleigh_benard` (takes too long, ~1 hr)

As a general note, it might be nice to add a `@test`s in the docs and, if/when we set `strict = true` in `docs/make.jl`, then these tutorials will hopefully never go stale.

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/format.jl`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1029, opened 2020-04-29
Add heat equation tutorial and single stack configuration

charleskawczynski:
# Description

Adds a heat equation tutorial.

~~- @kpamnany there are some pieces in here (see `helper_funcs.jl` / `plotting_funcs.jl`) that we should properly incorporate into source for our a `SingleColumnModel`, or `SingleStackModel`.~~

Also add `ClimateMachine.SingleStackConfiguration()` as well as `ClimateMachine.SingleStackUtils`.

 - @yairchn I think this is a good starting point for extending into an EDMF parameterization, or at least to add advection to start.

I have

- [X] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [X] Followed all necessary [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/format.jl`
- [X] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [X] There are no open pull requests for this already
- [X] CLIMA developers with relevant expertise have been assigned to review this submission
- [X] The code conforms to the [style guidelines](https://climate-machine.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [X] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1086, opened 2020-05-12
Cnh/statestats01

christophernhill:
# Description

Code for quick debugging and quick regression test statistics on MPIStateArrays

I have


docs/Project.toml
```diff
@@ -9,6 +9,7 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
 DoubleFloats = "497a8b3b-efae-58df-a0af-a86822472b78"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
```
simonbyrne, line 12: is this being used anywhere?

christophernhill:
I think its used in tutorials/Diagnostics/Debug/StateCheck.jl.
I got an error without it in docs.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+    iVal = 1
+    iPrec = 2
+    allPass = true
+
+    for row in cb.func.curStats_flat
+        ## Debugging
+        # println(row)
+        # println(refDat[iVal][irow])
+        # println(refDat[iPrec][irow])
+        rowPass = true
```
charleskawczynski, line 446: variable names should be lowercase under-score separated

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+    println("# SC \"F=\"  row pass count      ")
+    println("# SC \"NA=\" row not checked count      ")
+    println("# SC ")
+    println("# SC        Label         Field      min()      max()     mean()      std() ")
+    irow = 1
+    iVal = 1
+    iPrec = 2
+    allPass = true
+
+    for row in cb.func.curStats_flat
```
charleskawczynski, line 441: variable names should be lowercase under-score separated

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+# Imports from ClimateMachine core
+import ClimateMachine.GenericCallbacks: EveryXSimulationSteps
+import ClimateMachine.MPIStateArrays: MPIStateArray
+import ClimateMachine.VariableTemplates: flattenednames
+
+####
+# For testing put a new function signature here!
+# Needs to go in src/Utilities/VariableTemplates/var_names.jl
+# This handles SMatrix case
+flattenednames(::Type{T}; prefix = "") where {T <: SArray} =
```
charleskawczynski, line 35: Can we add proper documentation for what this is doing? What's written will not show up in doc strings.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+    ntuple(i -> "$prefix[$i]", length(T))
+####
+
+struct vstat
```
charleskawczynski, line 39: Can we add documentation here?

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+# ntFreqDef:: default frequency (in time steps) for output.
+# precDef  :: default precision used for formatted output table
+ntFreqDef = 10
+precDef = 15
+
+"""
+ sccreate :: Create a a "state check" call-back for one or more MPIStateArrays  \n
+             that will report basic statistics for the fields in the array.
+
+             Input:
```
charleskawczynski, line 61: This should be un-indented. Use `julia --project ?sccreate` to check formatting. Also, code should be wrapped in single backticks. [This](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Common/MoistThermodynamics/relations.jl) may be a good reference.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+
+# Global functions to expose
+# sccreate - Create a state checker call back
+export sccreate
+export scdocheck
+export scprintref
+
+# ntFreqDef:: default frequency (in time steps) for output.
+# precDef  :: default precision used for formatted output table
+ntFreqDef = 10
```
simonbyrne, line 54: non-constant globals are generally a bad idea.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+            for s in slist
+                if printHead
+                    if MPI.Comm_rank(MPI.COMM_WORLD) == 0
+                        println("# SC Creating state check callback labeled \"$lab\" for symbols")
+                    end
+                    printHead = false
+                end
+                if MPI.Comm_rank(MPI.COMM_WORLD) == 0
+                    println("# SC ", s)
+                end
```
charleskawczynski, line 121:
```suggestion
                if MPI.Comm_rank(MPI.COMM_WORLD) == 0
                    if printHead
                        println("# SC Creating state check callback labeled \"$lab\" for symbols")
                    end
                    println("# SC ", s)
                    printHead = false
                end
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+export scprintref
+
+# ntFreqDef:: default frequency (in time steps) for output.
+# precDef  :: default precision used for formatted output table
+ntFreqDef = 10
+precDef = 15
+
+"""
+ sccreate :: Create a a "state check" call-back for one or more MPIStateArrays  \n
+             that will report basic statistics for the fields in the array.
```
simonbyrne, line 59: The convention is that the first line of the docstring should be the function signature, indented 4 spaces so that it renders as a code block.

Also `::` has special meaning in Julia, so probably not a good idea to use it in other contexts.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+                        $ntFreqDef that sets how freuently (in time-step counts) the
+                        statistics are reported.
+             Named:
+                  prec: A named argument that sets number of decimal places to print for
+                        statistics, defaults to $precDef.
+
+             Return: 
+                sccb: A state checker that can be used in a callback().
+
+             Example:
```
charleskawczynski, line 76: We should surround the code in triple back-ticks, here. Appending the first set with `julia` will trigger Julia formatting, which is a plus :)

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+             julia> Q2=MPIStateArray{Float64,F2}(MPI.COMM_WORLD,ClimateMachine.array_type(),4,6,8);
+             julia> cb=StateCheck.sccreate([(Q1,"My gradients"),(Q2,"My fields")],1; prec=$precDef);
+             julia> cb()
+ ========================================================================================
+"""
+sccreate(
+    fields::Array{<:Tuple{<:MPIStateArray, String}},
+    ntFreq::Int = ntFreqDef;
+    prec = precDef,
+) = (
```
simonbyrne, line 95: inline function syntax should only be used for single-line functions. 
```suggestion
function sccreate(
    fields::Array{<:Tuple{<:MPIStateArray, String}},
    ntFreq::Int = ntFreqDef;
    prec = precDef,
)
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+             Return: 
+                sccb: A state checker that can be used in a callback().
+
+             Example:
+             julia> using ClimateMachine.VariableTemplates
+             julia> using StaticArrays
+             julia> using ClimateMachine.MPIStateArrays
+             julia> using MPI
+             julia> MPI.Init()
+             julia> T=Float64
```
charleskawczynski, line 82:
```suggestion
             julia> FT=Float64
```
We've been using `FT` to indicate the float type. Would be nice to keep this consistent (a big part to avoid ambiguity with temperature `T`)

src/Diagnostics/Debug/StateCheck.jl
simonbyrne: The convention is that types use CamelCase

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+            V = typeof(mArray).parameters[2]
+
+            ## Iterate over fields in each MPIStateArray
+            #  (use ivar to index individual arrays within the MPIStateArray)
+            ivar = 0
+            statsVal_dict = Dict()
+            for i in 1:length(V.names)
+                for n in flattenednames(fieldtype(V, i), prefix = fieldname(V, i))
+                    ivar = ivar + 1
+                    nStr = @sprintf("%9.9s", n)
```
charleskawczynski, line 195: @kpamnany this looks similar to what I've implemented in the heat equation tutorial, [`get_vars_from_stack`](https://github.com/CliMA/ClimateMachine.jl/pull/1029/files#diff-484affb4b21233277ef7c161e4377d59R151). I think that we should unify these, and it would probably be nice to have one that operates on the entire stack as well as per grid point.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+                            statsString[2],
+                            "|",
+                            statsString[3],
+                            "|",
+                            statsString[4],
+                            "|",
+                        )
+                    end
+                    statsVal_dict[n] = statsString[5]
+                    append!(
```
charleskawczynski, line 213: Shouldn't we know the size of `curStats_flat` beforehand? It would be better to allocate an array before-hand and avoid dynamic memory allocation.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+             julia> F1=@vars begin; ν∇u::SMatrix{3, 2, T, 6}; κ∇θ::SVector{3, T}; end
+             julia> F2=@vars begin; u::SVector{2, T}; θ::SVector{1, T}; end
+             julia> Q1=MPIStateArray{Float32,F1}(MPI.COMM_WORLD,ClimateMachine.array_type(),4,9,8);
+             julia> Q2=MPIStateArray{Float64,F2}(MPI.COMM_WORLD,ClimateMachine.array_type(),4,6,8);
+             julia> cb=StateCheck.sccreate([(Q1,"My gradients"),(Q2,"My fields")],1; prec=$precDef);
+             julia> cb()
+ ========================================================================================
+"""
+sccreate(
+    fields::Array{<:Tuple{<:MPIStateArray, String}},
```
jakebolewski, line 92: it's a good idea to parameterize all IO functions with an IO argument, so that IO can be re-directed to a file, or any IO stream that supports the interface (and thread that through):

```
sccreate(io::IOStream, ....)
```

then define this variant where the default IOStream is just passed stdout

```
function screate(....)
    return screate(Base.stdout, ....)
end
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+                        println(
+                            statsString[1],
+                            "|",
+                            statsString[2],
+                            "|",
+                            statsString[3],
+                            "|",
+                            statsString[4],
+                            "|",
+                        )
```
charleskawczynski, line 210:
```suggestion
                        println(join(statsString, "|"))
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+function scstats(V, ivar, nprec)
+
+    # Get number of MPI procs
+    nproc = MPI.Comm_size(V.mpicomm)
+
+    npr = nprec
+    fmt = @sprintf("%%%d.%de", npr + 8, npr)
+
+    # Min
+    phiLoc = minimum(V.realdata[:, ivar, :])
```
charleskawczynski, line 254: It might be nice to write a convenience function for this, since it digs into the internals of the `realdata` memory layout.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+                s3,
+                ",",
+                s4,
+                ",",
+                s5,
+                ",",
+                s6,
+                " ],",
+            )
+        end
```
charleskawczynski, line 374: Can/should we use something like [PrettyTables.jl](https://github.com/ronisbr/PrettyTables.jl) here and below? @simonbyrne @kpamnany

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+end
+
+"""
+ scdocheck :: Compare a current State check call-back set of values with a 
+              reference set and match precision table pair.
+               
+              Input:  cb - StateCheck call-back variables
+                      refDat - table of reference values and match precisions
+
+              Output: true  - comparison passed
```
charleskawczynski, line 421: Can we include what the metric that results in a pass/fail comparison is? It might be good to include an example here

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,556 @@
...
+                      false - comparison failed
+"""
+function scdocheck(cb, refDat)
+
+    if MPI.Comm_rank(MPI.COMM_WORLD) != 0
+        return true
+    end
```
charleskawczynski, line 428:
```suggestion
    MPI.Comm_rank(MPI.COMM_WORLD) == 0 || return true
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,548 @@
...
+               some subset of points at a fixed frequency.
+
+               Functions:
+               StateCheck.sccreate()   - Create a StateCheck call back variable.
+
+               StateCheck.scdocheck()  - Check Statecheck variable values against reference values.
+
+               StateCheck.scprintref() - Print Statecheck variable in format for creating 
+                                         reference values.
+"""
```
simonbyrne, line 16: Please use [markdown](https://docs.julialang.org/en/v1/stdlib/Markdown/):
```suggestion
"""
    StateCheck
    
Module with a minimal set of functions for gettings statistics 
and basic I/O from ClimateMachine DG state arrays (MPIStateArray type). 
Created for regression testing and code change tracking and debugging.
StateCheck functions iterate over named variables in an MPIStateArray,
calculate and report their statistics and/or write values for all or
some subset of points at a fixed frequency.

# Functions
- [`StateCheck.sccreate()`](@ref): Create a StateCheck call back variable.
- [`StateCheck.scdocheck()`](@ref): Check Statecheck variable values against reference values.
- [`StateCheck.scprintref()`](@ref): Print Statecheck variable in format for creating reference values.
"""
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,548 @@
...
+# nt_freq_def:: default frequency (in time steps) for output.
+# prec_def   :: default precision used for formatted output table
+const nt_freq_def = 10
+const prec_def    = 15
+
+"""
+    sccreate (io::IO, fields::Array{<:Tuple{<:MPIStateArray, String}}, nt_freq::Int = $nt_freq_def; prec = $prec_def)
+
+Create a "state check" call-back for one or more `MPIStateArray` variables 
+             that will report basic statistics for the fields in the array.
```
simonbyrne, line 61: remove indentation

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,548 @@
...
+    phi = phi_mean
+    # aveVstr=@sprintf("%23.15e",phi)
+    ave_v_str = sprintf1(fmt, phi)
+
+    # Std
+    phi_loc = ( getByField(V,ivar) .- phi_mean) .^ 2
+    n_val = length(phi_loc) * 1.0
+    phi_sum = MPI.Allreduce(phi_loc, +, Vmcomm)
+    n_val_sum = MPI.Allreduce(n_val, +, Vmcomm)
+    phi_std = (sum(phi_sum) / (n_val_sum - 1))^0.5
```
simonbyrne, line 267:
```suggestion
    phi_std = sqrt(sum(phi_sum) / (n_val_sum - 1))
```

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,548 @@
...
+
+####
+# For testing put a new function signature here!
+# Needs to go in src/Utilities/VariableTemplates/var_names.jl
+# This handles SMatrix case
+flattenednames(::Type{T}; prefix = "") where {T <: SArray} =
+    ntuple(i -> "$prefix[$i]", length(T))
+####
+
+struct vStat
```
simonbyrne, line 39: what is the `v` for? Also, still not CamelCase.

christophernhill:
@simonbyrne the Literate piece for this is not building at moment. @leios is investigating. In a stand-alone repo ( https://github.com/christophernhill/temp-clima-statetools ) with a very simple make file (  https://github.com/christophernhill/temp-clima-statetools/blob/master/docs/make.jl ).

charleskawczynski:
> I think its used in tutorials/Diagnostics/Debug/StateCheck.jl. I got an error without it in docs.

Could be related to stale things in `docs/src/generated`. Planning to fix this in #1097 or #1078

christophernhill:
Hi @simonbyrne @charleskawczynski @jakebolewski @leios does this look OK to undraft for merging ...... maybe ? Thanks, Chris

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,593 @@
...
+                "# SC =======|============|==========|",
+                min_str,
+                "|",
+                max_str,
+                "|",
+                ave_str,
+                "|",
+                std_str,
+                "|",
+            )
```
charleskawczynski, line 190:
```suggestion
        fmt_str = [" min() ", " max() ", " mean() ", " std() "]
        fmt_str = sprintf1.(Ref(h_var_fmt), fmt_str)
        if MPI.Comm_rank(MPI.COMM_WORLD) == 0
            print(io, "# SC =======|============|==========|")
            println(join(fmt_str, "|"))
        end
```
Might want to check that this produces the same format.

charleskawczynski:
> does this look OK to undraft for merging ...... maybe ?

We’ll have to rebase, first. Also, is it worth incorporating PrettyTables.jl first? I think it may help make things more compact

tutorials/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,174 @@
...
+# Set up a basic environment
+using MPI
+using StaticArrays
+using Random
+using ClimateMachine
+using ClimateMachine.VariableTemplates
+using ClimateMachine.MPIStateArrays
+using ClimateMachine.StateCheck
+MPI.Init()
+T = Float64
```
charleskawczynski, line 46: Can we switch this to `FT`?

christophernhill:
Yes

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
bors try

PR #1100, opened 2020-05-14
fixed restart minor bug

slifer50:
# Description

This is a fix for the restart functionality, the create_state function would cause an exit with an error saying the function does not exist, I've replaced it with create_conservative_state and restarting works again.

@simonbyrne @kpamnany @akshaysridhar


kpamnany:
All well with this fix?

slifer50:
@kpamnany can this be r+ed ? looks like all tests pass.

PR #1107, opened 2020-05-15
Add missing sync in vertical volume tendency kernel

mwarusz:
# Description

Fixes a possible race condition in the vertical tendency evaluation.

I have

### DIFF
--- a/src/Numerics/DGmethods/DGmodel_kernels.jl
+++ b/src/Numerics/DGmethods/DGmodel_kernels.jl
@@ -320,6 +320,9 @@ end
             local_MI[k] = vgeo[ijk, _MI, e]
         end
 
+        # ensure D is loaded
+        @synchronize(dim == 3)
+
         @unroll for k in 1:Nqk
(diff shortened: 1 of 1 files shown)


src/Numerics/DGmethods/DGmodel_kernels.jl
```diff
@@ -320,6 +320,9 @@ end
             local_MI[k] = vgeo[ijk, _MI, e]
         end
 
+        # ensure D is loaded
+        @synchronize
```
jkozdon, line 324: To match the call on [line 438](https://github.com/CliMA/ClimateMachine.jl/blob/97cbf3b357422a56697bd682755e6676a4a65fef/src/Numerics/DGmethods/DGmodel_kernels.jl#L438) do we want

```suggestion
        @synchronize(dim == 3)
```

mwarusz: Yes, good catch !

PR #1037, opened 2020-05-01
Reworking the Remainder Model

jkozdon:
This PR reworks the remainder model to achieve three things:

- Move Remainder Model Upstream DG Model Level
- Make the Remainder Model a true Discrete Splitting
- Allow Remainder Model to be Involve Models with Different Directions

### Move Remainder Model Upstream DG Model Level

The Remainder model previously existed in the atmos model even though it was not atmos specific. By moving it upstream it will be usable across CLIMA.

### Make the Remainder Model a true Discrete Splitting

Previously the remainder model defined a wave speed in an ad hoc way since it tried to construct a splitting at the physics level, here we seek to create a true discrete splitting so that
```
remainder tendency + subcomponent tendency == main tendency
```
To do this the numerical flux of each subcomponent is needed (since they might all be different). This is enabled by using a tuple of numerical fluxes. 

This was simplified by introducing an outer `DGModel` constructor for `remainder_DGModel`, which allows the user to compose `DGModel`s to construct a remainder based `DGModel` (as opposed to composing `BalanceLaws`)

### Allow Remainder Model to be Involve Models with Different Directions

Previously the remainder model only support the main model and subcomponents being active in the same direction. This extends things so that if the main model is `EveryDirection` the subcomponents can be either `EveryDirection`, `HorizontalDirection`, or `VerticalDirection`. If the main model is either `HorizontalDirection` or `VerticalDirection`, then the subcomponents must match this.

### Remaining limitations

- The subcomponent models must involve only first order terms (no diffusion or hyper-diffusion)
- The subcomponent models must be defined for the full set of states as the main component model.

PR #1110, opened 2020-05-15
remove Rusanov penalties from the linear model

jkozdon:
Previously the linear model penalized discontinuities for all variables in the state vector, not just those being used in the linear model.

### DIFF
--- a/src/Atmos/Model/linear.jl
+++ b/src/Atmos/Model/linear.jl
@@ -1,5 +1,6 @@
 using CLIMAParameters.Planet: R_d, cv_d, T_0, e_int_v0, e_int_i0
-
+using ..DGmethods.NumericalFluxes: RusanovNumericalFlux
+import ..DGmethods.NumericalFluxes: update_penalty!
 """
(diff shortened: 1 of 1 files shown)


jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

src/Atmos/Model/linear.jl
```diff
@@ -1,4 +1,5 @@
 using CLIMAParameters.Planet: R_d, cv_d, T_0, e_int_v0, e_int_i0
+using ..DGmethods.NumericalFluxes: RusanovNumericalFlux
```
mwarusz, line 3:
```suggestion
import ..DGmethods.NumericalFluxes: update_penalty!
```

lcw: Thanks! You beat me too this.

jkozdon:
This probably doesn't not sense to use / try until #1037 is merged.

src/Atmos/Model/linear.jl
jkozdon: Thanks!

PR #1115, opened 2020-05-18
Init shared path prefix from a single rank before run

### LINKED ISSUE #1114: `mkpath` on multiple MPI ranks
# Description

When testing Julia 1.4, I came across the following error:
```
ERROR: ERROR: LoadError: LoadError: IOError: mkdir: file already exists (EEXIST)
Stacktrace:IOError: mkdir: file already exists (EEXIST)
Stacktrace:
 [1] uv_error at ./libuv.jl:97 [inlined]
 [2] mkdir(::
 [1] uv_error at ./libuv.jl:97 [inlined]
 [2] String; mode::UInt16) at ./file.jl:177
 [3] mkpath(::String; mode::UInt16) at ./file.jl:227
 [4] mkpath at ./file.jl:222 [inlined]
 [5] mkdir(::run(::MPI.Comm, ::Type{T} where T, ::BrickTopology{2,Float64}, ::SwirlingFlow{Int64}, ::Float64, ::Int64, ::Int64, ::Type{T} where T, ::String, ::Float64) at /central/scratchio/esm/slurmci/sources/bbbe534d902b7508badd82202dac58935e8ea459/tutorials/Numerics/DGmethods/nonnegative.jl:163
 [6] top-level scope at /central/scratchio/esm/slurmci/sources/bbbe534d902b7508badd82202dac58935e8ea459/tutorials/Numerics/DGmethods/nonnegative.jl:247
 [7] include(::Module, ::String) at ./Base.jl:377
 [8] exec_options(::Base.JLOptions) at ./client.jl:288
 [9] _start() at ./client.jl:484
in expression starting at /central/scratchio/esm/slurmci/sources/bbbe534d902b7508badd82202dac58935e8ea459/tutorials/Numerics/DGmethods/nonnegative.jl:202
String; mode::UInt16) at ./file.jl:177
 [3] mkpath(::String; mode::UInt16) at ./file.jl:227
 [4] mkpath at ./file.jl:222 [inlined]
 [5] run(::MPI.Comm, ::Type{T} where T, ::BrickTopology{2,Float64}, ::SwirlingFlow{Int64}, ::Float64, ::Int64, ::Int64, ::Type{T} where T, ::String, ::Float64) at /central/scratchio/esm/slurmci/sources/bbbe534d902b7508badd82202dac58935e8ea459/tutorials/Numerics/DGmethods/nonnegative.jl:163
 [6] top-level scope at /central/scratchio/esm/slurmci/sources/bbbe534d902b7508badd82202dac58935e8ea459/tutorials/Numerics/DGmethods/nonnegative.jl:247
 [7] include(::Module, ::String) at ./Base.jl:377
 [8] exec_options(::Base.JLOptions) at ./client.jl:288
 [9] _start() at ./client.jl:484
```

https://gist.github.com/climabot/767c3b19502d5e096969154a211ebc59#file-out_9882097

I'm not exactly sure what is going on here, but it could be that we are hitting a race condition when `mkpath` is called on multiple MPI ranks.

For now the simplest option is to only call `mkdir` on rank 0 + `MPI.Barrier`, but we should look into what https://github.com/hpc/mpifileutils/ does.

# For CLIMA developers

jakebolewski:
# Description
Only create top-level prefix path on rank 0 of a communicator group to prevent output race conditions.  This doesn't impact unit tests because they are run sequentially, but slurm-ci tests have issues if multiple concurrent tests are scheduled to write to a common prefix.  We'll need to also change slurm-ci so that jobs don't write to a common prefix for test output.

Closes #1114 

I have

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -154,7 +154,7 @@ function parse_commandline(custom_settings)
         help = "directory for output data"
         metavar = "<path>"
         arg_type = String
-        default = "output"
+        default = get(ENV, "CLIMATEMACHINE_OUTPUT_DIR", "output")
         "--integration-testing"
         help = "enable integration testing"
         action = :store_true
@@ -253,13 +253,16 @@ function init(; disable_gpu = false, arg_settings = nothing)
     _init_array(atyp)
     Settings.array_type = atyp
 
-    # create the output directory if needed
-    if Settings.diagnostics !== "never" || Settings.vtk !== "never"
-        mkpath(Settings.output_dir)
+    # create the output directory if needed on delegated rank
+    if MPI.Comm_rank(MPI.COMM_WORLD) == 0
+        if Settings.diagnostics != "never" || Settings.vtk != "never"
+            mkpath(Settings.output_dir)
+        end
+        if Settings.checkpoint != "never" || Settings.checkpoint_at_end
+            mkpath(Settings.checkpoint_dir)
+        end
     end
-    if Settings.checkpoint !== "never" || Settings.checkpoint_at_end
-        mkpath(Settings.checkpoint_dir)
-    end
+    MPI.Barrier(MPI.COMM_WORLD)
 
     # set up logging
     loglevel = Settings.log_level == "DEBUG" ? Logging.Debug :
--- a/tutorials/Microphysics/ex_1_saturation_adjustment.jl
+++ b/tutorials/Microphysics/ex_1_saturation_adjustment.jl
@@ -210,19 +210,27 @@ function main()
     mpicomm = MPI.COMM_WORLD
 
     # output for paraview
+
+    # initialize base prefix directory from rank 0
+    vtkdir = abspath(joinpath(ClimateMachine.Settings.output_dir, "vtk"))
+    if MPI.Comm_rank(mpicomm) == 0
+        mkpath(vtkdir)
+    end
+    MPI.Barrier(mpicomm)
+
     model = driver_config.bl
     step = [0]
     cbvtk =
         GenericCallbacks.EveryXSimulationSteps(output_freq) do (init = false)
-            mkpath("vtk/")
-            outprefix = @sprintf(
-                "vtk/new_ex_1_mpirank%04d_step%04d",
+            out_dirname = @sprintf(
+                "new_ex_1_mpirank%04d_step%04d",
                 MPI.Comm_rank(mpicomm),
                 step[1]
             )
-            @info "doing VTK output" outprefix
+            out_path_prefix = joinpath(vtkdir, out_dirname)
+            @info "doing VTK output" out_path_prefix
             writevtk(
-                outprefix,
+                out_path_prefix,
                 solver_config.Q,
                 solver_config.dg,
                 flattenednames(vars_state_conservative(model, FT)),
--- a/tutorials/Microphysics/ex_2_Kessler.jl
+++ b/tutorials/Microphysics/ex_2_Kessler.jl
@@ -350,18 +350,26 @@ function main()
         end
 
     # output for paraview
+
+    # initialize base output prefix directory from rank 0
+    vtkdir = abspath(joinpath(ClimateMachine.Settings.output_dir, "vtk"))
+    if MPI.Comm_rank(mpicomm) == 0
+        mkpath(vtkdir)
+    end
+    MPI.Barrier(mpicomm)
+
     step = [0]
     cb_vtk =
         GenericCallbacks.EveryXSimulationSteps(output_freq) do (init = false)
-            mkpath("vtk/")
-            outprefix = @sprintf(
-                "vtk/new_ex_2_mpirank%04d_step%04d",
+            out_dirname = @sprintf(
+                "new_ex_2_mpirank%04d_step%04d",
                 MPI.Comm_rank(mpicomm),
                 step[1]
             )
-            @info "doing VTK output" outprefix
+            out_path_prefix = joinpath(vtkdir, out_dirname)
+            @info "doing VTK output" out_path_prefix
             writevtk(
-                outprefix,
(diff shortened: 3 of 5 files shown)


src/Driver/Driver.jl
```diff
@@ -253,13 +253,16 @@ function init(; disable_gpu = false, arg_settings = nothing)
...
 
-    # create the output directory if needed
-    if Settings.diagnostics !== "never" || Settings.vtk !== "never"
-        mkpath(Settings.output_dir)
-    end
-    if Settings.checkpoint !== "never" || Settings.checkpoint_at_end
-        mkpath(Settings.checkpoint_dir)
+    # create the output directory if needed on delegated rank
+    if MPI.Comm_rank(MPI.COMM_WORLD) == 0
+        if Settings.diagnostics !== "never" || Settings.vtk !== "never"
```
simonbyrne, line 258: These probably shouldn't be testing for identity:
```suggestion
        if Settings.diagnostics != "never" || Settings.vtk != "never"
```

src/Driver/Driver.jl
```diff
@@ -253,13 +253,16 @@ function init(; disable_gpu = false, arg_settings = nothing)
...
+            mkpath(Settings.output_dir)
+        end
+        if Settings.checkpoint !== "never" || Settings.checkpoint_at_end
```
simonbyrne, line 261:
```suggestion
        if Settings.checkpoint != "never" || Settings.checkpoint_at_end
```

tutorials/Microphysics/ex_1_saturation_adjustment.jl
```diff
@@ -210,19 +210,27 @@ function main()
     mpicomm = MPI.COMM_WORLD
 
     # output for paraview
+
+    # initialize base prefix directory from rank 0
+    vtkdir = abspath(joinpath(ClimateMachine.Settings.output_dir, "vtk"))
```
simonbyrne, line 215: do we need to check if `Settings.vtk != "never"`? Otherwise what will happen if no argument is used?

jakebolewski: what should be the behaviour to be in that case?  error out if the output directory is not defined?

jakebolewski: maybe run the simulation but skip over output generation?

simonbyrne: ah, i was worried that the output directory wasn't defined by default. I think this looks fine.

simonbyrne:
this doesn't fix the issue of multiple jobs writing to the same place. The easiest option for now would be to add `--output-dir=<unique job name>` in slurmci-test.toml

jakebolewski:
Yes this needs something on the slurmci side as well.  In general writing to relative directories on multiple nodes / ranks is not really well defined.  Probably there needs to be some way of overloading the base prefix, which is relative to `--output-dir` and enforcing at setup that the base path prefix is absolute.

jakebolewski:
There seems to only be a handful of tests that do output, and they all run on a single node, so Ill add the relative path construction suggested above for now.

charleskawczynski:
Should we be dumping this much data to disk in the tutorials? I ended up removing `write_data` calls altogether in the heat equation tutorial. The experiments are probably a different story, though.

jakebolewski:
@charleskawczynski I would think not, probably "tests" that generate output that is not checked by a computer are better run periodically

slurmci-test.toml
```diff
@@ -38,7 +38,7 @@ cpu_gpu = [
   { file = "test/Numerics/LinearSolvers/bandedsystem.jl", slurmargs = ["--ntasks=3", "--time=02:00:00"], args = [] },
   { file = "test/Numerics/Mesh/interpolation.jl", slurmargs = ["--ntasks=3", "--time=02:00:00"], args = [] },
   { file = "test/Ocean/ShallowWater/GyreDriver.jl", slurmargs = ["--ntasks=1"], args = [] },
-  { file = "tutorials/Numerics/DGmethods/nonnegative.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "tutorials/Numerics/DGmethods/nonnegative.jl", slurmargs = ["--ntasks=3"], args = ["--output-dir=output/dgmethods_nonnegative"] },
```
simonbyrne, line 41: This might have to be moved to have separate ones on gpu and cpu, since otherwise both jobs will use the same directory.

simonbyrne:
I think you're right about the environment variable being the simplest option. Ideally we would make the default value of `output-dir` be overrideable by an environment variable (e.g. `CLIMATEMACHINE_OUTPUT_DIR`, so we could set it in the slurm script. I'm not sure if this is supported by the ArgParse package.

simonbyrne:
I'm trying this out with https://github.com/CliMA/slurmci/pull/35

tutorials/Microphysics/ex_2_Kessler.jl
```diff
@@ -350,18 +350,25 @@ function main()
         end
 
     # output for paraview
+
+    # initialize base output prefix directory from rank 0
```
simonbyrne, line 354:
```suggestion
    # initialize base prefix directory from rank 0
    vtkdir = abspath(joinpath(ClimateMachine.Settings.output_dir, "vtk"))
```

tutorials/Numerics/DGmethods/nonnegative.jl
```diff
@@ -229,7 +234,10 @@ let
     CFL = 1
     dt = CFL * dx / maxvelocity
 
-    vtkdir = "vtk_nonnegative"
+    vtkdir = abspath(joinpath(
+        ClimateMachine.Settings.output_path,
```
simonbyrne, line 238:
```suggestion
        ClimateMachine.Settings.output_dir,
```

PR #1118, opened 2020-05-18
Add quick debug/regression statistics on raw state

christophernhill:
# Description

The is PR adds a quick look statistics call-back capability aimed primarily at rapid but reliable regression testing. The core feature is the ability to generate and recover in a very 
lightweight and convenient way summary raw statistics of key raw physical/numerical fields 
in the underlying internal prognostic and intermediate fields of any balance law that can be both human read and machine processed for CI. 

The code is being actively used to help check for problems in ocean development and could
potentially have use more broadly.  At this stage we would like to have it included in main repo as
a way to make ocean work easier and to explore more broad use for other contexts not just ocean, but Atmos, land, single-column etc... 

I have


christophernhill:
bors try

christophernhill:
bors try

christophernhill:
bors try

tutorials/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,176 @@
...
+# Now we can create a ```StateCheck``` call-back, _cb_, tied to the ```MPIStateArray```
+# variables _Q1_ and _Q2_. Each ```MPIStateArray``` in the array
+# of ```MPIStateArray``` variables tracked is paired with a label
+# to identify it. The call-back is also given a frequency (in time step numbers) and
+# precision for printing summary tables.
+cb = ClimateMachine.StateCheck.sccreate(
+    [(Q1, "My gradients"), (Q2, "My fields")],
+    1;
+    prec = 15,
+)
```
leios, line 91: This is outputting the following box in the docs:
```
# SC Start: creating state check callback
# SC Creating state check callback labeled "My gradients" for symbols
# SC ν∇u
# SC κ∇θ
# SC Creating state check callback labeled "My fields" for symbols
# SC u
# SC θ
# SC Finish: creating state check callback
```
I don't know if that is intended or not.

tutorials/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,176 @@
...
+# and utilizes some utility functions to create the array and add named
+# persistent state variables.
+# This is usually handled automatically as part of model deefinition in regular
+# ClimateMachine activity.
+
+# Set up a basic environment
+using MPI
+using StaticArrays
+using Random
+using ClimateMachine
```
kpamnany, line 41: If you add `ClimateMachine.init()` here, you don't need the `MPI.Init()` call below. Also, `ClimateMachine.array_type()`, used later on, will not return the correct type unless `ClimateMachine.init()` is called (it checks CUDA availability and `$CLIMATEMACHINE_GPU`).

tutorials/Diagnostics/Debug/StateCheck.jl
christophernhill: It was intended. Not sure if it helps, but I'm inclined to leave it for now to see.

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,604 @@
+"""
+    StateCheck  
+
+ Module with a minimal set of functions for gettings statistics
```
simonbyrne, line 4: This doesn't need to be indented.

simonbyrne:
My overall thoughts: I think this should be split into two parts: 
1. computing relevant summary statistics: there is already some atmos-specific stuff in src/Diagnostics, but we should make a generic fallback that works on any state vector, and computes basic quantities of interest. This work could form the basis of that
2. functions to compare with previously computed value and log the output

src/Diagnostics/Debug/StateCheck.jl
```diff
@@ -0,0 +1,604 @@
...
+        ## Print header
+        nprec = min(max(1, prec), 20)
+        if MPI.Comm_rank(MPI.COMM_WORLD) == 0
+            println(
+                io,
+                "# SC +++++++++++ClimateMachine StateCheck call-back start+++++++++++++++++",
+            )
+            println(
+                io,
+                "# SC  Step  |   Label    |  Field   |                            Stats                       ",
```
kpamnany, line 179: Does the output have to be formatted text? Writing JLD2s or NC files would make both writing the output and reading it back much easier.

skandalaCLIMA: @christophernhill : Can you please fill in the "Description" section for the PR with a summary of the feature being added.

christophernhill: > Does the output have to be formatted text? Writing JLD2s or NC files would make both writing the output and reading it back much easier.

The plain text format is deliberate. Generally for lightweight stuff like this a human readable little fragment can work quite well. For example https://github.com/CliMA/ClimateMachine.jl/blob/dd706a4ac370db3499afc5b3a8459d3d11f70992/test/Numerics/LinearSolvers/poisson.jl#L234 does this and we have been trialing this in https://github.com/christophernhill/temp-clima-statetools/blob/master/test/Ocean/HydrostaticBoussinesq/test_ocean_gyre_refvals.jl and finding it works well. 

Importantly the direct human manipulation format is designed to allow expert knowledge to be edited into the precision match numbers. For different fields and different statistics and different tests the statistical match needs to be set appropriately.

christophernhill:
bors try

charleskawczynski:
>  @charleskawczynski and I have discussed (very briefly) doing something similar for single column work, which doesn't even use MPIStatearray and friends. 

The recent single-column model development in ClimateMachine is using MPIStateArrays, just FYI. SingleColumnModels.jl, which is where I added some utilities to configure similar sorts of comparisons, does not use MPIStateArrays.

charleskawczynski:
> 1. We can probably get the `Base.IOError("stream is closed or unusable", 0)` error messages to go away if we use the following in literate examples:

I still see `Base.IOError("stream is closed or unusable", 0)` error the when I use `generate_tutorials = false`, so I don't think this error is related to Literate.

christophernhill:
> > 1. We can probably get the `Base.IOError("stream is closed or unusable", 0)` error messages to go away if we use the following in literate examples:
> 
> I still see `Base.IOError("stream is closed or unusable", 0)` error the when I use `generate_tutorials = false`, so I don't think this error is related to Literate.

@charleskawczynski that's correct, its more on the Documenter side of Literate/Documenter working together.

simonbyrne:
I'm happy to have this in until we can figure out something more robust. The main thing it needs are some simple unit tests: I know there is a literate example, but my experience is that errors in docs are more difficult to debug.

christophernhill:
> I'm happy to have this in until we can figure out something more robust. The main thing it needs are some simple unit tests: I know there is a literate example, but my experience is that errors in docs are more difficult to debug.

@simonbyrne I can add a unit test based on literate example. Would that work for now?

simonbyrne:
That seems like a good idea.

christophernhill:
bors try

christophernhill:
@skandalaCLIMA did you get a chance to look over things on this. Happy to address any comments. I think all the main review items are done. PR is chillin in bors try/rebase limbo at moment, so its not too late to send any tweaks.

charleskawczynski:
> I still see Base.IOError("stream is closed or unusable", 0) error the when I use generate_tutorials = false, so I don't think this error is related to Literate.

I added some print lines, and it seems that these errors _are_ coming from Literate after all.

PR #1120, opened 2020-05-19
Remove the use of moisture in Atmos linear model

lcw:
With this change the linear model is linearized around a dry rest state.

A benefit of this change is that linear model depends on only air
density, momentum, and total energy density. This reduction in variables
may be exploited for efficiency gains.

### DIFF
--- a/src/Atmos/Model/linear.jl
+++ b/src/Atmos/Model/linear.jl
@@ -1,7 +1,7 @@
-using CLIMAParameters.Planet: R_d, cv_d, T_0, e_int_v0, e_int_i0
+using CLIMAParameters.Planet: R_d, cv_d, T_0
 
 """
-    linearized_air_pressure(ρ, ρe_tot, ρe_pot, ρq_tot=0, ρq_liq=0, ρq_ice=0)
+    linearized_air_pressure(ρ, ρe_tot, ρe_pot)
 
 The air pressure, linearized around a dry rest state, from the equation of state
 (ideal gas law) where:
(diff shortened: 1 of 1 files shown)


lcw:
bors try

tapios:
It's no problem linearizing about a dry state. But neglecting the moisture terms in the linearized pressure (e.g., whether linearized about a dry or moist state) is not strictly correct. Doing so has led to stability problems in other models, especially if gravity waves are included in the linearization. It may be nice to have it as an option to remove these terms, but blanket erasure seems risky.

lcw:
> It's no problem linearizing about a dry state. But neglecting the moisture terms in the linearized pressure (e.g., whether linearized about a dry or moist state) is not strictly correct. Doing so has led to stability problems in other models, especially if gravity waves are included in the linearization. It may be nice to have it as an option to remove these terms, but blanket erasure seems risky.

Thanks for the feedback.  I will think about a different way of reducing the linear system size.

tapios:
> > It's no problem linearizing about a dry state. But neglecting the moisture terms in the linearized pressure (e.g., whether linearized about a dry or moist state) is not strictly correct. Doing so has led to stability problems in other models, especially if gravity waves are included in the linearization. It may be nice to have it as an option to remove these terms, but blanket erasure seems risky.
> 
> Thanks for the feedback. I will think about a different way of reducing the linear system size.

What you can always exploit is that the linearized moisture *tendencies* are zero, i.e., the moisture variables are frozen for the fast dynamics.

lcw:
> What you can always exploit is that the linearized moisture
> *tendencies* are zero, i.e., the moisture variables are frozen for the
> fast dynamics.

True, but right now their is not an easy way to pass the moisture
variables from the Atmos model as auxiliary variables to the linear
model. There is a baked in assumption, into at least the LU linear
solver, that the system size is based on the number of state variables.

PR #1103, opened 2020-05-14
update CuArrays to 2.2

simonbyrne:
# Description

Fixes #1044, supercedes  #1073 and  #1099.

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1123, opened 2020-05-19
Fix pre-compilation test, run unit tests with built system image 

jakebolewski:
Drop test macro, just execute script. move into a sep build stage

I have

### DIFF
--- a/.dev/systemimage/climate_machine_image.jl
+++ b/.dev/systemimage/climate_machine_image.jl
@@ -6,33 +6,54 @@
 #
 # Called with a single argument the system image will be placed in the path
 # specified by the argument (relative to the callers path)
+#
+# Called with a specified systemimg path and `true`, the system image will
+# compile the climate machine package module (useful for CI)
+
 sysimage_path =
-    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : abspath(ARGS[1])
+
+climatemachine_pkg = get(ARGS, 2, "false") == "true"
+
+@info "Creating system image object file at: '$(sysimage_path)'"
+@info "Building ClimateMachine into system image: $(climatemachine_pkg)"
+
+start_time = time()
 
 using Pkg
+Pkg.activate(joinpath(@__DIR__, "..", ".."))
+Pkg.instantiate(verbose = true)
+
+pkgs = Symbol[]
+if climatemachine_pkg
+    push!(pkgs, :ClimateMachine)
+else
+    # TODO: reorg project structure (Project.toml) to use
+    # Pkg.dependencies() with PackageCompiler
+    #if VERSION >= v"1.4"
+    #    append!(pkgs, [Symbol(v.name) for v in values(Pkg.dependencies())])
+    #end
+    append!(pkgs, [Symbol(name) for name in keys(Pkg.installed())])
+end
+
+# use package compiler
 Pkg.add("PackageCompiler")
 using PackageCompiler
-Pkg.activate(joinpath(@__DIR__, "..", ".."))
-Pkg.instantiate()
-pkgs = Pkg.installed()
-delete!(pkgs, "Pkg")
+PackageCompiler.create_sysimage(
+    pkgs,
+    sysimage_path = sysimage_path,
+    precompile_execution_file = joinpath(
+        @__DIR__,
+        "..",
+        "..",
+        "test",
+        "Numerics",
+        "DGMethods",
+        "Euler",
+        "isentropicvortex.jl",
+    ),
+)
 
-# "TF_BUILD" is check to see if we are running on Azure for CI
-if !haskey(ENV, "TF_BUILD")
-    PackageCompiler.create_sysimage(
-        [Symbol(s) for s in keys(pkgs)];
-        sysimage_path = sysimage_path,
-        precompile_execution_file = joinpath(
-            @__DIR__,
-            "..",
-            "..",
-            "test",
-            "Numerics",
-            "DGMethods",
-            "Euler",
-            "isentropicvortex.jl",
-        ),
-    )
-else
-    return true
-end
+tot_secs = Int(floor(time() - start_time))
+@info "Created system image object file at: $(sysimage_path)"
+@info "System image build time: $tot_secs sec"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,24 +15,28 @@ jobs:
 
   strategy:
     matrix:
-      Julia 1.3:
-        JULIA_VERSION: '1.3'
-
+      Julia 1.4:
+        JULIA_VERSION: '1.4'
   steps:
   - bash: |
       set -o xtrace
       wget -nv https://julialang-s3.julialang.org/bin/linux/x64/$(JULIA_VERSION)/julia-$(JULIA_VERSION)-latest-linux-x86_64.tar.gz
       mkdir julia-$(JULIA_VERSION)
       tar zxf julia-$(JULIA_VERSION)-latest-linux-x86_64.tar.gz -C julia-$(JULIA_VERSION) --strip-components 1
-    displayName: 'Download and extract Julia'
+    displayName: 'Linux - Download and extract Julia'
   - bash: |
       set -o xtrace
-      ./julia-$(JULIA_VERSION)/bin/julia -e 'using InteractiveUtils; versioninfo()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.instantiate()'
-      ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.test()'
-      ./julia-$(JULIA_VERSION)/bin/julia -e 'using Test; @testset "system image" begin;@test include(".dev/systemimage/climate_machine_image.jl"); end;'
-    displayName: 'Run the tests'
-    continueOnError: true
+    displayName: 'Linux - Install project packages'
+  - bash: |
+      set -o xtrace
+      ./julia-$(JULIA_VERSION)/bin/julia --project=@. .dev/systemimage/climate_machine_image.jl ./ClimateMachine.so true
+    displayName: 'Linux - Build project system image'
+  - bash: |
+      set -o xtrace
+      ./julia-$(JULIA_VERSION)/bin/julia --project=@. -J ClimateMachine.so -e 'using InteractiveUtils; versioninfo()'
+      ./julia-$(JULIA_VERSION)/bin/julia --project=@. -J ClimateMachine.so -e 'using Pkg; Pkg.test()'
+    displayName: 'Linux - Run the tests'
 
 - job: macOS
 
@@ -43,8 +47,8 @@ jobs:
 
   strategy:
     matrix:
-      Julia 1.3:
-        JULIA_VERSION: '1.3'
+      Julia 1.4:
+        JULIA_VERSION: '1.4'
 
   steps:
   - bash: |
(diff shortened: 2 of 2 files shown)


azure-pipelines.yml
```diff
@@ -30,9 +30,13 @@ jobs:
       ./julia-$(JULIA_VERSION)/bin/julia -e 'using InteractiveUtils; versioninfo()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.instantiate()'
       ./julia-$(JULIA_VERSION)/bin/julia --project=@. -e 'using Pkg; Pkg.test()'
-      ./julia-$(JULIA_VERSION)/bin/julia -e 'using Test; @testset "system image" begin;@test include(".dev/systemimage/climate_machine_image.jl"); end;'
     displayName: 'Run the tests'
     continueOnError: true
+  - bash: |
+      set -o xtrace
+      ./julia-$(JULIA_VERSION)/bin/julia -e 'include(joinpath(".dev", "systemimage", "climate_machine_image.jl"))'
```
simonbyrne, line 37: Should we just call the script?
```suggestion
      ./julia-$(JULIA_VERSION)/bin/julia .dev/systemimage/climate_machine_image.jl
```

jakebolewski: yes probably, didn't know if we want to re-wrap this in a test macro (if this is the issue).

jakebolewski:
didn't escape the execution string correctly, but it seems like this was the issue.  I'll move ahead with getting the system image built and trying to use it for running the unit-tests.

simonbyrne:
Disappointing the PackageCompiler.jl doesn't work on Windows:
```
C:/Users/VssAdministrator/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: <unknown-file>:0: syntax error
Warning: .drectve `-export:ccalllib_C:\julia-1.3\bin\LLVM.dll,data ' unrecognized
C:/Users/VssAdministrator/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Cannot export ccalllib_C:: symbol not found
collect2.exe: error: ld returned 1 exit status
```

Possibly related to https://github.com/JuliaLang/PackageCompiler.jl/issues/377?

I wonder if Julia 1.4 would help?

simonbyrne:
It could be https://github.com/JuliaLang/julia/issues/34680, which is supposed to be fixed in 1.4, or it could be https://github.com/JuliaLang/PackageCompiler.jl/issues/365 / https://github.com/JuliaLang/julia/pull/35039.

charleskawczynski:
I’m a bit confused why we’re calling the sys image script in the azure yml. Isn’t this creating a sys image for a specific experiment file?

jakebolewski:
@charleskawczynski Simon asked if this was possible to start introducing into the pipeline, what it's actually pre-compiling at this point I don't know is that important (any code that exercises a decent amount of CliMA code will be sufficient, we can re-target what it's exactly pre-compiling to reduce overall average latency at a later point with some profiling).

@simonbyrne it looks like Julia 1.4 works on Windows! which is good news.

simonbyrne:
@charleskawczynski The script used in the precompilation is just used as a template so it know what function signatures to build into the system image. We might want to do something more like use https://github.com/timholy/SnoopCompile.jl to generate a list of all statements we want to build into the tests. At the moment, we might want to use a script trim down the script used (we don't actually need it to do anything other than trigger the compilation path.

Lets switch to 1.4 on Windows for now, as that will unblock #1103 as well, then we can figure out the Slurm CI slowdowns later.

simonbyrne:
before:
```
┌ Info: Running MPI test...
│   file = "/home/vsts/work/1/s/test/Numerics/Mesh/mpi_centroid.jl"
└   ntasks = 3
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
 54.380740 seconds (49.87 k allocations: 2.478 MiB)
```

after:
```
┌ Info: Running MPI test...
│   file = "/home/vsts/work/1/s/test/Numerics/Mesh/mpi_centroid.jl"
└   ntasks = 3
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
[ Info: CUDAdrv.jl failed to initialize, GPU functionality unavailable (set JULIA_CUDA_SILENT or JULIA_CUDA_VERBOSE to silence or expand this message)
  5.450521 seconds (49.87 k allocations: 2.477 MiB)
```

jkozdon:
> At the moment, we might want to use a script trim down the script used (we don't actually need it to do anything other than trigger the compilation path

Yeah what I did was a quick hack.

Also, if we are using it for testing, we may want to include ClimateMachine itself in the image. I didn't do this in what I did since I wanted to develop ClimateMachine itself.

.dev/systemimage/climate_machine_image.jl
```diff
@@ -11,28 +11,26 @@ sysimage_path =
...
-            "DGmethods",
-            "Euler",
-            "isentropicvortex.jl",
-        ),
-    )
-else
-    return true
-end
+PackageCompiler.create_sysimage(
+    [Symbol(s) for s in keys(pkgs)];
```
simonbyrne, line 22: From @jkozdon  in https://github.com/CliMA/ClimateMachine.jl/pull/1056#issuecomment-624845226
> For users (as opposed to folks not hacking the CLIMA package itself), you could include a version that just depends on CLIMA itself, though there you would want to use as a precompile script something that uses the driver.

We might want to do this for the tests. From what I understand, this would just be:
```suggestion
    [:ClimateMachine];
```
but we probably want an option to do both.

.dev/systemimage/climate_machine_image.jl
```diff
@@ -11,28 +11,26 @@ sysimage_path =
 
 using Pkg
 Pkg.add("PackageCompiler")
+
 using PackageCompiler
 Pkg.activate(joinpath(@__DIR__, "..", ".."))
 Pkg.instantiate()
 pkgs = Pkg.installed()
```
simonbyrne, line 18:
```suggestion
pkgs = VERSION < v"1.4" ? keys(Pkg.installed()) : [v.name for v in values(Pkg.dependencies())]
```
then change the use of `keys(pkgs)` below to just `pkgs`.

jkozdon, line 18: When we go to Julia 1.4 we will want to use [`dependencies`](https://julialang.github.io/Pkg.jl/dev/api/#Pkg.dependencies) here since [`installed` is deprecated](https://github.com/JuliaLang/Pkg.jl/blob/127412cd23224a4f8fcdb049d80c4c16486a1e93/src/Pkg.jl#L531-L541)

jkozdon: See @simonbyrne already caught this.

.dev/systemimage/climate_machine_image.jl
jkozdon: I agree. We want both behavior.

simonbyrne: add an option that can be checked by ARGS?

jakebolewski: added something simple for now (define path and add `true`)

jakebolewski:
Updated

.dev/systemimage/climate_machine_image.jl
```diff
@@ -6,33 +6,47 @@
 #
 # Called with a single argument the system image will be placed in the path
 # specified by the argument (relative to the callers path)
+#
+# Called with a specified systemimg path and `true`, the system image will
+# compile the climate machine package module (useful for CI)
+
 sysimage_path =
-    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+    len(ARGS) < 1 ? joinpath(@__DIR__, "ClimateMachine.so") : abspath(ARGS[1])
```
simonbyrne, line 14: Still writing Python?
```suggestion
    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : abspath(ARGS[1])
```

jakebolewski: yes :disappointed:

.dev/systemimage/climate_machine_image.jl
```diff
@@ -6,33 +6,47 @@
...
 # specified by the argument (relative to the callers path)
+#
+# Called with a specified systemimg path and `true`, the system image will
+# compile the climate machine package module (useful for CI)
+
 sysimage_path =
-    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : ARGS[1]
+    isempty(ARGS) ? joinpath(@__DIR__, "ClimateMachine.so") : abspath(ARGS[1])
+
+climatemachine_pkg = len(ARGS) > 1 && ARGS[2] == "true" ? true : false
```
simonbyrne, line 16:
```suggestion
climatemachine_pkg = get(ARGS,2,"false") == "true"
```

.dev/systemimage/climate_machine_image.jl
```diff
@@ -6,33 +6,47 @@
...
+    append!(pkgs, :ClimateMachine)
 else
-    return true
+    if VERSION < v"1.4"
+        append!(pkgs, keys(Pkg.installed()))
+    else
+        append!(pkgs, [v.name for v in values(Pkg.dependencies())])
+    end
 end
+delete!(pkgs, "Pkg")
```
charleskawczynski, line 35: Why remove `Pkg`?

.dev/systemimage/climate_machine_image.jl
```diff
@@ -6,33 +6,47 @@
...
-            "..",
-            "..",
-            "test",
-            "Numerics",
-            "DGmethods",
-            "Euler",
-            "isentropicvortex.jl",
-        ),
-    )
+pkgs::Vector{Symbol} = []
```
simonbyrne, line 25:
```suggestion
pkgs = Symbol[]
```

.dev/systemimage/climate_machine_image.jl
```diff
@@ -6,33 +6,54 @@
...
-            "Numerics",
-            "DGmethods",
-            "Euler",
-            "isentropicvortex.jl",
-        ),
-    )
+
+pkgs = Symbol[]
+if climatemachine_pkg
+    append!(pkgs, :ClimateMachine)
```
simonbyrne, line 32:
```suggestion
    push!(pkgs, :ClimateMachine)
```

jakebolewski:
It seems like there are issues with Pkg.dependencies() and our current Project.toml / Manifest file structure.  I'm still getting used to how things work so maybe we should switch to defining a project structure that doesn't cause `PkgCompiler` to complain about missing dependencies.   Pkg.installed() still works for collecting project packages for building a sys image that doesn't include `ClimateMachine`

jakebolewski:
Seems to be working so maybe this is ready (finally?).  The test suite doesn't really work on my 4G NVIDIA laptop, but there seems to be some LLVM IR issues on that side with precompilation so this may have to be CPU only until those get sorted out.

jakebolewski:
@simonbyrne  is this ready to merge?

jakebolewski:
Needs to be rebased on  #1185

PR #1124, opened 2020-05-19
Use reference state moisture in Atmos linear model

lcw:
The reference state moisture is used in the linear model to avoid needing
the moisture variables in the state vector.  This enables the
removal of the moist variables from the linear model state.
   
Co-authored-by: Jeremy E Kozdon <jekozdon@nps.edu>

### DIFF
--- a/src/Atmos/Model/linear.jl
+++ b/src/Atmos/Model/linear.jl
@@ -58,7 +58,7 @@ end
         state.ρ,
         state.ρe,
         ρe_pot,
-        state.moisture.ρq_tot,
+        aux.ref_state.ρq_tot,
     )
 end
 


lcw:
bors try

lcw:
I think we need to copy `ρq_tot` from the state to the reference state for the multi-rate time-stepping.  I am looking into added a callback to do this after every slow time-step to see if it fixes it.

jkozdon:
> I think we need to copy `ρq_tot` from the state to the reference state for the multi-rate time-stepping. I am looking into added a callback to do this after every slow time-step to see if it fixes it.

Why is that? Won't that mean the linear system is changing every step? (I thought we wanted to avoid that)

lcw:
> Why is that? Won't that mean the linear system is changing every step? (I thought we wanted to avoid that)

Yep.  Maybe until we have fast system building we should leave moisture in the linear state.

jkozdon:
> Yep. Maybe until we have fast system building we should leave moisture in the linear state.

I still like the idea of freezing it if we can since otherwise using an iterative solver and a direct solver are technically working with different linear systems (which seems weird to me and I could see leading to future confusion).

jkozdon:
> > Yep. Maybe until we have fast system building we should leave moisture in the linear state.
> 
> I still like the idea of freezing it if we can since otherwise using an iterative solver and a direct solver are technically working with different linear systems (which seems weird to me and I could see leading to future confusion).

I see that I was wrong, it's a linear variable.

lcw:
After thinking about it.  I think it is best to leave moisture in the state variables.  Then the linear system doesn't depend on the moisture.  Thus, if a direct method is being used to implicitly time-step the linear system we don't need to refactor the matrix at each time-step.

lcw:
kkk

PR #1125, opened 2020-05-20
Driver update: Solver types for explicit, IMEX, multirate, and MIS methods

thomasgibson:
# Description

This PR does the following:

- It moves current time-stepping solver types into their own self-contained directory, but with the intention of being used specifically with the driver. The rationale behind this: the stream of `if elseif elseif elseif...` is not sustainable in the long term and one can EASILY imagine this getting way out of hand when adding more time-stepping wrappers.
- I've added discrete splitting types to be used with the solver types: (1): `SlowFastSplitting()` which is designed to be simple splitting of fast (acoustic in all directions) and slow (everything else). And (2): `HEVISplitting()` - this splitting will split the acoustic part into horizontal and vertical parts. The vertical part is implicitly treated while all other tendencies are explicit.
- It extends the current multirate solver type to now support IMEX fast solvers. This is done by specifying `splitting_type = `HEVISplitting()` and passing an IMEX ARK method as the fast solver.
- I've also added a new solver type for the Multirate Infinitesimal Step methods in Clima (A generalization of split-explicit RK methods). At the moment, it only supports purely explicit integration by setting `splitting_type =SlowFastSplitting()`

If anyone sees something strange or thinks it can be done in a better (more Julian?) way, I'd be very keen to learn! Also, I struggled with thinking of designing a set of unit tests for this. One idea is to purposefully use these solver types in the unit tests in `test/Driver` (I've started this).

If others have better ideas, I'd appreciate the feedback.

I have

### DIFF
--- a/src/Driver/SolverTypes/IMEXSolverType.jl
+++ b/src/Driver/SolverTypes/IMEXSolverType.jl
@@ -1,0 +1,172 @@
+
+export IMEXSolverType
+
+"""
+# Description
+    IMEXSolverType(;
+        splitting_type = HEVISplitting(),
+        implicit_model = AtmosAcousticGravityLinearModel,
+        implicit_solver = ManyColumnLU,
+        implicit_solver_adjustable = false,
+        solver_method = ARK2GiraldoKellyConstantinescu,
+        solver_storage_variant = LowStorageVariant(),
+    )
+
+This solver type constructs a solver for ODEs with the
+additively-partitioned form:
+
+```math
+  \\dot{Q} = [l(Q, t)] + [f(Q, t) - l(Q, t)]
+```
+
+where `Q` is the state, `f` is the full tendency and `l` is the chosen
+implicit operator.
+
+# Arguments
+- `splitting_type` (DiscreteSplittingType): The type of discrete
+    splitting to apply to the right-hand side.
+    Default: `HEVISplitting()`
+- `implicit_model` (Type): The model describing dynamics to be
+    treated implicitly.
+    Default: `AtmosAcousticGravityLinearModel`
+- `implicit_solver` (Type): A solver for inverting the
+    implicit system of equations.
+    Default: `ManyColumnLU`
+- `implicit_solver_adjustable` (Bool): A flag identifying whether
+    or not the `implicit_solver` can be updated as the time-step
+    size changes.
+    Default: `false`
+- `solver_method` (Function): Function defining the particular additive
+    Runge-Kutta method to be used for the IMEX method.
+    Default: `ARK2GiraldoKellyConstantinescu`
+- `solver_storage_variant` (Type): Storage type for the additive
+    Runge-Kutta method.
+    Default: `LowStorageVariant()`
+
+### References
+    @article{giraldo2013implicit,
+      title={Implicit-explicit formulations of a three-dimensional
+             nonhydrostatic unified model of the atmosphere ({NUMA})},
+      author={Giraldo, Francis X and Kelly, James F and Constantinescu, Emil M},
+      journal={SIAM Journal on Scientific Computing},
+      volume={35},
+      number={5},
+      pages={B1162--B1194},
+      year={2013},
+      publisher={SIAM}
+    }
+"""
+struct IMEXSolverType{DS, ST} <: AbstractSolverType
+    # The type of discrete splitting to apply to the right-hand side
+    splitting_type::DS
+    # The implicit model
+    implicit_model::Type
+    # Choice of implicit solver
+    implicit_solver::Type
+    # Can the implicit solver be updated with changing dt?
+    implicit_solver_adjustable::Bool
+    # Function for the IMEX method
+    solver_method::Function
+    # Storage type for the ARK scheme
+    solver_storage_variant::ST
+
+    function IMEXSolverType(;
+        splitting_type = HEVISplitting(),
+        implicit_model = AtmosAcousticGravityLinearModel,
+        implicit_solver = ManyColumnLU,
+        implicit_solver_adjustable = false,
+        solver_method = ARK2GiraldoKellyConstantinescu,
+        solver_storage_variant = LowStorageVariant(),
+    )
+
+        DS = typeof(splitting_type)
+        ST = typeof(solver_storage_variant)
+
+        return new{DS, ST}(
+            splitting_type,
+            implicit_model,
+            implicit_solver,
+            implicit_solver_adjustable,
+            solver_method,
+            solver_storage_variant,
+        )
+    end
+end
+
+"""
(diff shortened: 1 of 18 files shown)


thomasgibson:
BTW: I will not be squashing any commits until the code is approved for merging

src/Driver/solver_configs.jl
```diff
@@ -146,11 +146,21 @@ function SolverConfiguration(
     end
```
kpamnany, line 146: Maybe move this `dtmodel` determination below, where it's used?

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,160 @@
...
+      pages={B1162--B1194},
+      year={2013},
+      publisher={SIAM}
+    }
+"""
+struct IMEXSolverType{DS, ST} <: AbstractSolverType
+    # The type of discrete splitting to apply to the right-hand side
+    splitting_type::DS
+    # Linear model describing fast dynamics
+    linear_model::Type
```
mwarusz, line 62: Maybe `implicit_model` would be better to be consistent with `implicit_solver` and to cover future generalizations.

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,160 @@
...
+  \\dot{Q} = [l(Q, t)] + [f(Q, t) - l(Q, t)]
+```
+
+where `Q` is the state, `f` is the full tendency and `l` is the chosen
+implicit operator.
+
+# Arguments
+- `splitting_type` (DiscreteSplittingType): The type of discrete
+    splitting to apply to the right-hand side.
+    Default: `SlowFastSplitting()`
```
mwarusz, line 28: Doesn't match the actual default.

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,160 @@
...
+in the additive Runge-Kutta method.
+
+# Comments:
+Currently, the only HEVI-type splitting ClimateMachine can currently
+do only involves splitting the acoustic processes; it is not currently
+possible to perform more fine-grained separation of tendencies
+(for example, including vertical advection or diffusion in the 1-D implicit problem)
+"""
+function solversetup(
+    ode_solver::IMEXSolverType{DS},
```
mwarusz, line 119: Here and in other places I think you can just do `IMEXSolverType{HEVISplitting}` and get rid of the `where` clause since the splitting type is concrete.

src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -0,0 +1,155 @@
...
+        number={5},
+        pages={2067--2081},
+        year={2014}
+    }
+"""
+struct MISSolverType{DS} <: AbstractSolverType
+    # The type of discrete splitting to apply to the right-hand side
+    splitting_type::DS
+    # Linear model describing fast dynamics
+    linear_model::Type
```
mwarusz, line 57: `fast_model` ? Doesn't have to be linear. Similarly for other multi-rate solvers.

src/Driver/SolverTypes/IMEXSolverType.jl
kpamnany: Just a reminder -- if you change this, you'll have to change some tests, experiments and tutorials that are constructing `IMEXSolverType`s.

src/Driver/SolverTypes/IMEXSolverType.jl
thomasgibson: Yes that was my bad! I was copy-pasting parts of docstrings I previously wrote and wasn't careful here!

src/Driver/SolverTypes/MISSolverType.jl
thomasgibson: Really good point here, in fact, when we go full HEVI, `fast_model` will most certainly be non-linear.

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,171 @@
...
+```
+
+where `Q` is the state, `f` is the full tendency and `l` is the chosen
+implicit operator.
+
+# Arguments
+- `splitting_type` (DiscreteSplittingType): The type of discrete
+    splitting to apply to the right-hand side.
+    Default: `HEVISplitting()`
+- `implicit_model` (Type): The linear model describing fast dynamics.
```
mwarusz, line 29: Change linear to implicit in the model description

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,171 @@
...
+    Default: `AtmosAcousticGravityLinearModel`
+- `implicit_solver` (Type): A linear solver for inverting the
```
mwarusz, line 31: Change linear to implicit in the solver description

src/Driver/SolverTypes/IMEXSolverType.jl
```diff
@@ -0,0 +1,171 @@
...
+      number={5},
+      pages={B1162--B1194},
+      year={2013},
+      publisher={SIAM}
+    }
+"""
+struct IMEXSolverType{DS, ST} <: AbstractSolverType
+    # The type of discrete splitting to apply to the right-hand side
+    splitting_type::DS
+    # Linear model describing fast dynamics
```
mwarusz, line 61: Implicit model

src/Driver/SolverTypes/MultirateSolverType.jl
```diff
@@ -0,0 +1,351 @@
...
+function solversetup(
+    ode_solver::MultirateSolverType{SlowFastSplitting},
+    dg,
+    Q,
+    dt,
+    t0,
+    diffusion_direction,
+)
+
+    # Extract linear model and define a DG model
```
mwarusz, line 170: Fast model

src/Driver/SolverTypes/MultirateSolverType.jl
```diff
@@ -0,0 +1,351 @@
...
+function solversetup(
+    ode_solver::MultirateSolverType{HEVISplitting},
+    dg,
+    Q,
+    dt,
+    t0,
+    diffusion_direction,
+)
+
+    # Extract linear model and define a DG model
```
mwarusz, line 275: Fast model

kpamnany:
Go ahead and merge this if you're happy with it?

thomasgibson:
bors try

thomasgibson:
bors try

thomasgibson:
bors try

PR #1131, opened 2020-05-20
Method dispatch on `liquid_frac` based on thermodynamic state type

charleskawczynski:
# Description

Since `liquid_frac` really depends on whether the phase is in equilibrium or not, we need to pass this information down from the constructors. In this PR, `ThermodynamicState`'s are passed down to inform which state is being constructed.

Co-authored by: @kpamnany, @trontrytel and @tapios 

I have

### DIFF
--- a/src/Common/MoistThermodynamics/relations.jl
+++ b/src/Common/MoistThermodynamics/relations.jl
@@ -338,7 +338,7 @@ The internal energy per unit mass, given
 end
 
 """
-    internal_energy_sat(param_set, T, ρ, q_tot)
+    internal_energy_sat(param_set, T, ρ, q_tot, phase_type)
 
 The internal energy per unit mass in thermodynamic equilibrium at saturation where
 
@@ -346,17 +346,19 @@ The internal energy per unit mass in thermodynamic equilibrium at saturation whe
  - `T` temperature
  - `ρ` (moist-)air density
  - `q_tot` total specific humidity
+ - `phase_type` a thermodynamic state type
 """
 function internal_energy_sat(
     param_set::APS,
     T::FT,
     ρ::FT,
     q_tot::FT,
+    phase_type::Type{<:ThermodynamicState},
 ) where {FT <: Real}
     return internal_energy(
         param_set,
         T,
-        PhasePartition_equil(param_set, T, ρ, q_tot),
+        PhasePartition_equil(param_set, T, ρ, q_tot, phase_type),
     )
 end
 
@@ -372,6 +374,7 @@ internal_energy_sat(ts::ThermodynamicState) = internal_energy_sat(
     air_temperature(ts),
     air_density(ts),
     total_specific_humidity(ts),
+    typeof(ts),
 )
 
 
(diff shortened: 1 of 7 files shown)


charleskawczynski:
@kpamnany might it be easier/simpler to just use

 - `PhaseDry`
 - `PhaseEquil`
 - `PhaseNonEquil`

here instead of adding indicator/dispatch types:
```julia
abstract type PhaseType end
struct DryPhase <: PhaseType end
struct EquilPhase <: PhaseType end
struct NonEquilPhase <: PhaseType end
get_phase_type(ts::PhaseDry) = DryPhase()
get_phase_type(ts::PhaseEquil) = EquilPhase()
get_phase_type(ts::PhaseNonEquil) = NonEquilPhase()
```
? Originally, it didn't seem that I had the flexibility that I wanted. For example, using the `PhaseType` we only need one constructor for

```julia
saturation_excess(ts::ThermodynamicState) = saturation_excess(
    ts.param_set,
    air_temperature(ts),
    air_density(ts),
    get_phase_type(ts),
    PhasePartition(ts),
)
```

charleskawczynski:
@trontrytel, @tapios's recommended `liquid_fraction` for non-equilibrium thermodynamic states is causing the [microphysics tutorials](https://gist.github.com/climabot/bf7a3e83c1aa2a4fe7d83a2a9fc89801) to fail. Any idea how to fix this? Maybe just a smaller timestep?

trontrytel:
> @trontrytel, @tapios's recommended `liquid_fraction` for non-equilibrium thermodynamic states is causing the [microphysics tutorials](https://gist.github.com/climabot/bf7a3e83c1aa2a4fe7d83a2a9fc89801) to fail. Any idea how to fix this? Maybe just a smaller timestep?

The quick thoughts that I have:
- The saturation-adjustment run that does not use non-equil thermodynamics is also failing
- The tests are failing because the function returns NaNs
- The time-step in those tests is already quite small, especially in the Kessler run. I would be hesitant to make it even smaller.

Can you show me what exactly was changed?

charleskawczynski:
This PR changes `liquid_fraction` from 
```julia
function liquid_fraction(
    param_set::APS,
    T::FT,
    q::PhasePartition{FT} = q_pt_0(FT),
) where {FT <: Real}
    _T_freeze::FT = T_freeze(param_set)
    q_c = q.liq + q.ice     # condensate specific humidity
    if q_c > 0
        return q.liq / q_c
    else
        # For now: Heaviside function for partitioning into liquid and ice: all liquid
        # for T > T_freeze; all ice for T <= T_freeze
        return FT(T > _T_freeze)
    end
end
```
to
```julia
function liquid_fraction(
    param_set::APS,
    T::FT,
    phase_type::Type{<:ThermodynamicState},
    q::PhasePartition{FT} = q_pt_0(FT),
) where {FT <: Real}
    _T_freeze::FT = T_freeze(param_set)
    return FT(T > _T_freeze)
end
function liquid_fraction(
    param_set::APS,
    T::FT,
    phase_type::Type{<:PhaseNonEquil},
    q::PhasePartition{FT} = q_pt_0(FT),
) where {FT <: Real}
    q_c = q.liq + q.ice     # condensate specific humidity
    return q.liq / q_c
end
```
So, the way `liquid_fraction` is computed is now statically determined by the thermodynamic state rather than the instantaneous `PhasePartition`.

PR #1132, opened 2020-05-20
Reduce number of tasks for some tests

jkozdon:
Reduce the number of MPI ranks on the three longest standalone tests,
all of which are too small to benefit from multiple GPUs.


jkozdon:
bors try

jkozdon:
As compared with https://github.com/CliMA/ClimateMachine.jl/commit/dd706a4ac370db3499afc5b3a8459d3d11f70992 (see [slurmci](https://gist.github.com/climabot/5cdde58247f29d63928a8fcf3c647846)) the last [try](https://gist.github.com/climabot/650ddcc687908474fc7f4a42b5ccef94) gives:

| test | old | new |
|---|---|---|
| bomex | 00:42:40 | 00:19:29 |
| heldsuarez | 00:51:04 | 00:18:51 |
| Raleigh Benard | 01:00:35 | 00:21:26|

jkozdon:
Biggest offender still remains though, `scripts/test-cpu-tests.sh` at 01:06:40 (see [gist](https://gist.github.com/climabot/650ddcc687908474fc7f4a42b5ccef94) and [output](https://gist.githubusercontent.com/climabot/650ddcc687908474fc7f4a42b5ccef94/raw/297eaeb1e10d13e62100b4e5ed48ac7bba12bf3e/out_9912690))

Not sure what to do about that one...

jkozdon:
Any reason not to merge this?

jakebolewski:
No, sounds good to me if the test is still meaningful with a reduced number of ranks.

simonbyrne:
> Biggest offender still remains though, `scripts/test-cpu-tests.sh` at 01:06:40 (see [gist](https://gist.github.com/climabot/650ddcc687908474fc7f4a42b5ccef94) and [output](https://gist.githubusercontent.com/climabot/650ddcc687908474fc7f4a42b5ccef94/raw/297eaeb1e10d13e62100b4e5ed48ac7bba12bf3e/out_9912690))

We could just not run it, and rely on Azure?

jakebolewski:
The test suite being run here is just to check that the unit-tests also pass on the cluster env. (centos) in addition to Azure (ubuntu?).

jkozdon:
> The test suite being run here is just to check that the unit-tests also pass on the cluster env. (centos) in addition to Azure (ubuntu?).

Will #1123 help here?

jakebolewski:
@jkozdon it will help, but realistically I think it's pretty safe to not re-run the unit-tests on the cluster in addition to Azure pipelines.  Any env issues would show up in the other slurm-ci test jobs (in addition to the unit tests) so it feels redundant.

jkozdon:
> @jkozdon it will help, but realistically I think it's pretty safe to not re-run the unit-tests on the cluster in addition to Azure pipelines. Any env issues would show up in the other slurm-ci test jobs (in addition to the unit tests) so it feels redundant.

Very true.

PR #1134, opened 2020-05-20
Remove use of MoistThermo internals

charleskawczynski:
# Description

While fixing #1131, I realized that there are several calls to moist thermo's internals, and removing these calls will greatly simplify things.

This PR

 - Removes several calls to MoistThermo's internal functions
 - Renames `TemperatureSHumEquil` -> `TemperatureSHumEquil_given_pressure`
 - Adds a constructor for `TemperatureSHumEquil` (which accepts density). As a result, I also trimmed the `Extending` section in the docs since we actually needed what was suggested as an extension.
 - Adds a `PhaseDry_given_ρT` constructor

This PR should not be behavior-changing.

I have


charleskawczynski:
@trontrytel can you take a look at the microphysics changes here?

trontrytel:
> @trontrytel can you take a look at the microphysics changes here?

I'll take a look, but we don't have a good test for the precipitation model yet. So it's hard to test. Thanks for letting me know. Especially changes from \rho to p can be tricky to spot...

A quick comment: I thought we were supposed to remove the Shums from function names? Everywhere else in the code specific humidity is denoted as `q`, even in the docstrings and the variable names of the functions here. 

And in general I don't follow the logic in the naming of those functions. For example:

`TemperatureSHumEquil(param_set, T, ρ, q_tot)` 

has a docstring of 

`Constructs a [`PhaseEquil`](@ref) thermodynamic state from temperature`

Is the name based on function arguments only? Wouldn't it be neater to have just `ThermoState(...)` and dispatch based on arguments and equil type? (similar to dispatching over liquid and ice?)

charleskawczynski:
> A quick comment: I thought we were supposed to remove the Shums from function names?

Perhaps, I think we can rename things in a separate PR. Ideally, all the state constructors would be named
- `PhaseDry`
- `PhaseEquil`
- `PhaseNonEquil`

And accept kwargs for density, pressure etc., but this is currently an issue for GPU performance

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -233,7 +233,8 @@ function conv_q_rai_to_q_vap(
     _MP_n_0::FT = MP_n_0(param_set)
 
     if qr > FT(0)
-        qv_sat = q_vap_saturation(param_set, T, ρ, q)
+        ts_neq = TemperatureSHumNonEquil(param_set, T, ρ, q)
+        qv_sat = q_vap_saturation(ts_neq)
         q_v = q.tot - q.liq - q.ice
```
trontrytel, line 238: I just need saturation vapour specific humidity here. Why is it better to construct the whole thermo state to get just that one number?

test/Atmos/Parameterizations/Microphysics/runtests.jl
```diff
@@ -121,7 +121,8 @@ end
         ρ::FT,
     ) where {FT <: Real}
 
-        q_sat = q_vap_saturation(param_set, T, ρ, q)
+        ts_neq = TemperatureSHumNonEquil(param_set, T, ρ, q)
+        q_sat = q_vap_saturation(ts_neq)
         q_vap = q.tot - q.liq
```
trontrytel, line 126: Same as above. Why do I now need to write two lines of code to get one number?

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
charleskawczynski: Because we need to change the interface of `q_vap_saturation(param_set, T, ρ, q)`, and we'd also like to continue being able to change the interface. The idea is that the interface of the thermodynamic states should be stable. Otherwise, this line will read

```julia
phase_type = MoistThermodynamics.PhaseNonEquilType() # internal to moist thermodynamics
qv_sat = q_vap_saturation(param_set, T, ρ, phase_type, q)
```

And this may further change when we add other thermodynamic states (for ocean, land, etc.)

test/Atmos/Parameterizations/Microphysics/runtests.jl
charleskawczynski: Same reason as above

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
trontrytel: Is it optimised by the compiler? If not it's an overhead that will be done in every grid point every time I compute RHS.

charleskawczynski: I don't think performance is an issue here. The non-equil states do not need an iterative solver so it's only a few extra flops (which will be hidden by MPI communication). The main advantage of leveraging the high-level interface is that it can be more extensively tested for correctness and consistency. There are fewer guarantees on correctness/consistency when using moist thermo's internals (since users will then need to provide these guarantees).

PR #1153, opened 2020-05-23
Speed up 1D IMEX for Atmos with passive tracers

lcw:
This set of changes removes the tracers from the Atmos linear model making
the 1D IMEX linear systems smaller and speeding up their direct solve.  For
example, `experiments/AtmosGCM/heldsuarez.jl` with 32 passive tracers is ~19x
faster.  For more detail, see the table below of timings on a V100.

```
# Tracers   Wall-Time Per Step (s)
            before    after
----------------------------------
  1         5.86e-03   5.73e-03
  2         6.81e-03   6.06e-03
  4         8.26e-03   6.89e-03
  8         1.23e-02   8.70e-03
 16         1.33e-01   1.45e-02
 32         5.19e-01   2.65e-02
```

Unfortunately, removing the tracers from the linear model required changing
the numerical flux (in the full model moisture and tracer variables are now
penalized with the advective wave speed) so results might be slightly
different after these changes are applied.


lcw:
bors try

src/Numerics/DGmethods/DGmodel.jl
```diff
@@ -79,6 +83,10 @@ function (dg::DGModel)(
     ndrange_interior_surface = Nfp * length(grid.interiorelems)
     ndrange_exterior_surface = Nfp * length(grid.exteriorelems)
 
+    if !increment && num_state_conservative < num_state_tendency
+        tendency .= -zero(FT)
+    end
```
jkozdon, line 88: What would need to change so that we didn't do this zeroing? Just wondering since this could be a lot of excess memory writes (since we only have 5 or 6 conservative variables and 10s of tracers).

Is it the remainder model and IMEX solvers that don't know about the difference between conservative and tendency?

Though probably not a huge deal since filling an array should be fairly efficient compared to everything else we do in the kernels below.

bischtob:
I like it.

bischtob:
@lcw, is this the smallest system we can get away with in general for 1D IMEX?

src/Numerics/DGmethods/DGmodel.jl
lcw: Yea, I think the issue is we are sharing vectors between the models.  We would need to define what it means to add a vector with 3 states to a vector with 10 states.

lcw:
> @lcw, is this the smallest system we can get away with in general for 1D IMEX?

I think we want to include the moisture variables to keep from having to refactor the matrix during the simulation.  Currently building the LU factors is slow.

We could combine the velocity into one.  @jkozdon said it would require some changes to the numerical flux that he has been thinking about.  It would probably take a couple of weeks.

One of the main problems of the current implementation is that the forward and backward solves do not strong scale.  I could look at improving this if desired.  I am not sure if it will be possible so it would be speculative work.  We could also look into speeding up building the systems and factorizing them.  This may be desired if we want to use LU as a preconditioner in HEVI.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -474,7 +474,20 @@ end
...
     u = ρinv * state.ρu
-    return abs(dot(nM, u)) + soundspeed(m, m.moisture, state, aux)
+    uN = abs(dot(nM, u))
+    ss = soundspeed(m, m.moisture, state, aux)
+
+    FT = typeof(state.ρ)
+    ws = -@MVector(zeros(FT, number_state_conservative(m, FT)))
+    vars_ws = Vars{vars_state_conservative(m, FT)}(ws)
+
+    ws .= uN
```
mwarusz, line 484: Not sure if it makes any difference, but you could do `ws = fill(uN, MVector{number_state_conservative(m, FT), FT})`
to create a vector filled with the value `uN`.

src/Numerics/DGmethods/DGmodel.jl
blallen: > Yea, I think the issue is we are sharing vectors between the models. We would need to define what it means to add a vector with 3 states to a vector with 10 states.

Can this be done with vars_index / vars-aware MPIStateArrays? Being to add vectors of different sizes but with overlapping variable names could be very useful elsewhere as well

jkozdon: > Can this be done with vars_index / vars-aware MPIStateArrays?

Possibly. I would say we should get this in, and then revisit it in a future PR. Maybe open an issue?

src/Atmos/Model/AtmosModel.jl
lcw: Looks good.  I will update it as I try to debug.

lcw:
bors try

lcw:
bors try-

lcw:
bors try

lcw:
bors try

src/Atmos/Model/remainder.jl
```diff
@@ -99,9 +99,51 @@ compute_gradient_flux!(
...
+            wavespeed(sub, nM, state, aux, t)
+        for sub in Base.tail(rem.subs)
+            rs[1:number_state_conservative(sub, FT)] .=
+                rs[1:number_state_conservative(sub, FT)] .+
+                wavespeed(sub, nM, state, aux, t)
+        end
+        ws .-= rs
+    end
+
+    return ws
```
lcw, line 146: @vchuravy and @jkozdon: Do you guys have a suggestion on how to clean this code up?  I am struggling to get a clean version of the loop to compile on the GPU.

lcw:
bors try

src/Atmos/Model/remainder.jl
lcw: Thanks to @jkozdon and StaticNumbers.jl I think I have a fix

lcw:
bors try

lcw:
bors try

lcw:
@simonbyrne I believe this pr is ready for your review.

PR #1157, opened 2020-05-26
1) Revert BOMEX to use IMEX timestepper

akshaysridhar:
Rolls back BOMEX driver to use 1D-IMEX method.	
```modified:   experiments/AtmosLES/bomex.jl```
While pending issues with MRRK interactions with physics and boundaries are addressed, this is the stable solution. 

### DIFF
--- a/experiments/AtmosLES/bomex.jl
+++ b/experiments/AtmosLES/bomex.jl
@@ -419,7 +419,7 @@ function config_bomex(FT, N, resolution, xmax, ymax, zmax)
     )
 
     # Choose default IMEX solver
-    ode_solver_type = ClimateMachine.IMEXSolverType();
+    ode_solver_type = ClimateMachine.IMEXSolverType()
 
     # Assemble model components
     model = AtmosModel{FT}(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by akshaysridhar** [comment]
bors try

File: experiments/AtmosLES/bomex.jl [hunk 1]
```diff
@@ -514,19 +510,33 @@ function main()
...
-            Δρ_ratio = (norm_ρ - norm_ρ_Q₀) / norm_ρ_Q₀
-            @test isapprox(Δρ_ratio, FT(0); atol = 1e-3)
+    # State variable
+    Q = solver_config.Q
+    # Volume geometry information
+    vgeo = driver_config.grid.vgeo
+    M = vgeo[:,Grids._M,:]
+    # Unpack prognostic vars
+    ρ₀ = Q.ρ
+    ρe₀ = Q.ρe
```
kpamnany, line 520: This won't work here -- `Q` is an `MPIStateArray` (like `vgeo` above). Are you looking for all `ρ` for the entire grid?

akshaysridhar: Yes

akshaysridhar:
> Note #1125 -- the way an `IMEXSolverType` is created is going to change. Not sure which PR will merge first.

I see - this uses the `default` IMEXSolverType defined in the driver_configs, would this be affected by the 1125 changes ?

jkozdon:
> Rolls back BOMEX driver to use 1D-IMEX method.

Can you clarify why this is being done? (I don't have a problem with it, just wondering)

kpamnany:
> I see - this uses the `default` IMEXSolverType defined in the driver_configs, would this be affected by the 1125 changes ?

There are some new defaults specified [here](https://github.com/CliMA/ClimateMachine.jl/blob/thg/solver-types/src/Driver/SolverTypes/IMEXSolverType.jl#L73-L80). Not sure if you would want to override any of them.

akshaysridhar:
> > Rolls back BOMEX driver to use 1D-IMEX method.
> 
> Can you clarify why this is being done? (I don't have a problem with it, just wondering)

I think there are still some issues with the MRRK w.r.t top-boundary issues that we'd like to understand - so for the time being I want to roll this back.

jkozdon:
I think there are still some issues with the MRRK w.r.t top-boundary interaction that we'd like to understand - so for the time being I want to roll this back.

Thanks for the info. Was just wondering.

simonbyrne:
i'll backport it once it's merged

Issue #276, opened 2019-06-18
example2 : Kessler 


trontrytel:
The Kessler microphysics example doesn't work when you try to advect rain with rain velocity. Possible suspects are:
 - wrong boundary condition at the top of the domain
 - ripples in the solution for q_rain
 - not enough diffusion to make things stable. 

This should be solved before we try to use this microphysics in LES setups. 

I tag: @trontrytel , @lcw and @smarras79

charleskawczynski:
@trontrytel is this still an issue?

Issue #275, opened 2019-06-18
rain velocity rho/rho_ground factor


trontrytel:
Should rain terminal velocity be multiplied by (rho/rho_surface)^(1/2)?

See for example https://doi.org/10.1175/1520-0493(1996)124<0487:TTLSLM>2.0.CO;2

Assigned to me (@trontrytel)

tapios:
I see no physical reason why this would make sense. It's non-causal. Why should a falling raindrop know about the density at the ground before hitting it? It suspect this factor comes from empirical formulae used to fit radar observations.

trontrytel:
I agree. I just want to track the paper with an explanation. So that we can for sure decide not to do it. Its just a reminder that I should check it

mwarusz:
It seems like this factor was introduced by Kessler in https://doi.org/10.1007/978-1-935704-36-2_1 (eq. 7.1 on page 20). His earlier papers don't have it, compare eq. 37 on page 23 in

http://www.atmos.albany.edu/facstaff/rfovell/ATM562/kessler-1967.pdf

I don't see much of an explanation, but I will let @trontrytel  figure it out.

trontrytel:
I think the issue is that they assume a constant coefficient multiplying the square root of drop radius:

v(r) = K sqrt(2r)

But from the force balance when you consider also buoyancy, there should be a  square root(1/air_density) factor there (see terminal velocity in our docs). So thats the argument for adding back the sqrt(1/air density). I have no explanation for air density at the ground. 

Assuming air density at the ground = 1.2kg/m3 we would get the coefficient used in Kessler paper sqrt(2 K rho_ground) = 201. For our formulation it is sqrt(8/3/C_drag (\rho_water - rho_air) g). For C_drag=0.55 we get 218. So I think it's all consistent and we don't need to add it.

Issue #274, opened 2019-06-17
print CFL at run time


smarras79:
Print the CFL number to screen at run time to keep an eye on code stability.

kpamnany:
I believe #806 added this capability (`--monitor-courant-numbers`).

Issue #273, opened 2019-06-17
Float32 tests


simonbyrne:
We should test Float32 by default everywhere.

lcw:
If possible yes.  If Float32 works that is an easy way of getting a 2x performance increase.

charleskawczynski:
@asraero and I have talked about doing this for checking for type-stability also.

PR #1162, opened 2020-05-27
DifferentialEquations.jl-Based ODE Solvers

ChrisRackauckas:
This sets up "two" ODE solvers: DiffEqJLSolver and DiffEqJLIMEXSolver. These solvers allow taking in a `DEAlgorithm` which can be any DiffEq common interface ODE solver that satisfies the common integrator interface, which includes OrdinaryDiffEq, Sundials, and a few others. 

The only kernel calls are done through fused broadcasts, so in theory there shouldn't be any overhead on large problems as long as we are only using the right number of kernels. In practice, since this is our first super larger scale application since we plan to track, there may be an extra kernel call somewhere, and so I assume that after we get attempt 1 done here we may need to make an upstream change. The standard RK methods (including the SSP methods, which includes some methods with enhanced stability specifically for DG discretizations) should all be optimized already. I think we may need to do something special for the low-storage RK methods.

As for implicit and IMEX methods, I setup the interface for how to hook into them, though we may need to setup a linear solver for DiffEq. That can be done via https://docs.sciml.ai/latest/features/linear_nonlinear/#Linear-Solvers:-linsolve-Specification-1 , so we can just make the solvers call your linear solvers.

## Why?

There are a few things we're looking at for how this will benefit both our own research and Clima. To list a few:

- Our methods for [numerical uncertainty quantification](https://docs.sciml.ai/latest/analysis/uncertainty_quantification/) will allow for understanding the error induced by the time stepping methods. There's other (undocumented) UQ methods being published soon which could be useful too.
- [Checkpointed adjoint sensitivity analysis](https://docs.sciml.ai/latest/analysis/sensitivity/) for parameter fitting and UQ is probably a big feature Clima would gain from this. More work might be required to complete the integration of this, but I think there could be a lot of uses.
- Stochastic differential equation models would be pretty accessible from this form, so if there are reasonable extensions to SPDEs they can be considered and high order adaptive methods can be applied.
- [Generalized Physics-Informed Learning](https://youtu.be/SEhMWkgcTOI): you can see from this talk that I want to start taking this to climate models for developing surrogates.
- Tons of new methods. Specifically:
  - [A bunch of SSP methods](https://docs.sciml.ai/latest/solvers/ode_solve/#Explicit-Strong-Stability-Preserving-Runge-Kutta-Methods-for-Hyperbolic-PDEs-(Conservation-Laws)-1)
  - [A bunch of low-storage RK methods](https://docs.sciml.ai/latest/solvers/ode_solve/#Low-Storage-Methods-1)
  - [A bunch of IMEX methods](https://docs.sciml.ai/latest/solvers/split_ode_solve/#OrdinaryDiffEq.jl-1)
  - Whatever we build in the future.

The benefits to SciML are:

- We see that there's been some recent work in MRI methods. It would be nice to capture that momentum as something that can be added directly to OrdinaryDiffEq (or even just in SimpleDiffEq.jl or another common interface algorithm package) so that they can be used on an SplitODEProblem. This would enhance the Julia ecosystem beyond Clima and be something that can then be demonstrated as a benefit of the Clima project to other projects down the line. By creating this common interface ODE solver algorithm for Clima, I hope we can accelerate this effort because now, if the MRI methods are moved to a DiffEq interface compatible package, they should be usable in Clima without overhead which should be a good incentive to do so!
- We can start to use Clima as one of our canonical benchmarks, and hopefully overtime improve algorithm performance on Clima's type of problems and use it as a source of real-world problems in publications and grants (possibly we could look for a joint grant in this area).
- This serves as a good testing ground for MPI and GPU based array usage. It won't be perfect until we have a real test, so let's get a test!

So in short, this should help Clima get a few new features while improving SciML for large models to help out other PDE-based modeling projects beyond Clima.

## Current Issues

There are a few small issues noticed in this PR:

- What test case should I be using? I just grabbed what was in the ODESolver tests, but what's a good demo climate simulation I can start using to test out CPU usage, GPU usage, and MPI+GPU usage? I'd like to have these three cases for development and to add to [DiffEqBenchmarks.jl](https://github.com/SciML/DiffEqBenchmarks.jl).
- When is `p` defined? In the DiffEq solvers, we want `p` at the time of the construction of the `ODEProblem` so we can concretely type the integrator. We at least need to know and have something of its type. The test just has `nothing` parameters, so that works, but I wonder if this will be an issue. I added `p` as a keyword argument to the solver's constructor.
- The methods from Sundials and other C++/Fortran libraries are "technically" supported, but in reality they need the array type to be an `Array`, which I assume will almost never be the case with Clima. Sundials does have some good MRI methods that would be interesting to try out here, but making them work would require making NVector wrappers which, it might just be easier to make Julia versions of all of that (and that is a summer project for students).
- Is there any more structure in the ODEs that can be exploited? Semilinearity for exponential integrators or anything of that sort? Symplecticness or partitioning of the states? 
- Dependencies: are you open to a DiffEqBase dependency? That's a whole lot smaller than OrdinaryDiffEq or DifferentialEquations.jl. A user would need to `using DifferentialEquations` (or just `using OrdinaryDiffEq`) to actually get solver methods.

## What's Next?

If this is setup, then I would like to have some of our GSoC and MLH students optimize this interfacing and demonstrate results using it. That would then set the stage for less technical more mathematical students to just do pure algorithm development in OrdinaryDiffEq knowing that this would be efficient. I hope we can demonstrate that this connection ends up calling the optimal number of kernels by the end of the summer.

In 18.337 Scientific Machine Learning and Parallel Computing in the next fall, each student has final projects to do and, if this is all setup, I would like to point a few students towards implementing some new methods and demonstrating the results on Clima, so this is where new IMEX, MRI, etc. methods can get developed, improved, and tested.

# Checklist


ChrisRackauckas:
My current testing script is:

```julia
using Test
using ClimateMachine
using LinearAlgebra
ClimateMachine.init()
const ArrayType = ClimateMachine.array_type()
using ClimateMachine.ODESolvers
using ClimateMachine.LinearSolvers
using OrdinaryDiffEq

a = 100
b = 1
c = 1 / 100
Δ = sqrt(4 * a * c - b^2)

α1, α2 = 1 / 4, 3 / 4
β1, β2, β3 = 1 / 3, 3 / 6, 1 / 6

function rhs!(dQ, Q, ::Nothing, t; increment)
    if increment
        @. dQ += $cos(t) * (a + b * Q + c * Q^2)
    else
        @. dQ = $cos(t) * (a + b * Q + c * Q^2)
    end
end
function rhs_linear!(dQ, Q, ::Nothing, t; increment)
    if increment
        @. dQ += $cos(t) * b * Q
    else
        @. dQ = $cos(t) * b * Q
    end
end

function exactsolution(t, q0, t0)
    k = @. 2 * atan((2 * c * q0 + b) / Δ) / Δ - sin(t0)
    solution = @. (Δ * tan((k + sin(t)) * Δ / 2) - b) / (2 * c)
    return ArrayType(solution)
end

q0 = ArrayType === Array ? [1.0] : range(-1.0, 1.0, length = 303)
t0 = 0.1
finaltime = 1.2

Qinit = exactsolution(t0, q0, t0)
Q = similar(Qinit)
Qexact = exactsolution(finaltime, q0, t0)

dts = [0.125, 0.0625]
errors = similar(dts)

dt = dts[1]
Q .= Qinit
solver = DiffEqJLSolver(rhs!, Tsit5(), Q; dt = dt, t0 = t0, adaptive=false)
ClimateMachine.ODESolvers.solve!(Q, solver; timeend = finaltime)
norm(Q - Qexact) #0.00011413923949684727

dt = dts[2]
Q .= Qinit
solver = DiffEqJLSolver(rhs!, Tsit5(), Q; dt = dt, t0 = t0, adaptive=false)
ClimateMachine.ODESolvers.solve!(Q, solver; timeend = finaltime)
norm(Q - Qexact) #3.0659664173199496e-7
```

ChrisRackauckas:
CC: @kanav99 @yingboma

ChrisRackauckas:
@vchuravy is there a good way for us to easily track the total number of kernels called for unit testing? For example, could there be an array type which has a counter on every time it's broadcasted, so that way we can unit test that only X broadcasts are called per step? If we can do something like that, then we can make sure to keep peak performance for these applications while not requiring some fancy MPI timing in tests.

src/Numerics/ODESolvers/DifferentialEquations.jl
```diff
@@ -0,0 +1,115 @@
...
+"""
+mutable struct DiffEqJLIMEXSolver{I} <: AbstractDiffEqJLSolver
+    integ::I
+
+    function DiffEqJLIMEXSolver(rhs!,rhs_implicit!,alg,Q,args...;
+                            t0 = 0,
+                            kwargs...)
+        prob = SplitODEProblem(
+                          (du,u,p,t)->rhs_implicit!(du,u,p,t; increment=false),
+                          (du,u,p,t)->rhs!(du,u,p,t; increment=false),
```
simonbyrne, line 55:
```suggestion
                          rhs_implicit!,
                          rhs!,
```
should just work?

ChrisRackauckas: I thought it was a required keyword argument? If it has a default then yes it will work.

src/Numerics/ODESolvers/DifferentialEquations.jl
```diff
@@ -0,0 +1,115 @@
...
+  \\dot{Q} = f_I(Q, t) + f_E(Q, t)
+```
+
+via a DifferentialEquations.jl DEAlgorithm, which includes support
+for OrdinaryDiffEq.jl, Sundials.jl, and more.
+"""
+mutable struct DiffEqJLIMEXSolver{I} <: AbstractDiffEqJLSolver
+    integ::I
+
+    function DiffEqJLIMEXSolver(rhs!,rhs_implicit!,alg,Q,args...;
```
simonbyrne, line 50: it probably would make sense to switch all our solvers so that the implicit part is first. this is probably going to be more important once we want multirate.

src/Numerics/ODESolvers/DifferentialEquations.jl
simonbyrne: It does:
https://github.com/CliMA/ClimateMachine.jl/blob/de2c373bbcec3e686b0d57b65794131ae9e7de90/src/Numerics/DGmethods/DGmodel.jl#L45-L51

simonbyrne:
Nice, this seems like a step in the right direction. 

The simplest "real problem" to try is the rising thermal bubble, see:
https://clima.github.io/ClimateMachine.jl/latest/generated/Atmos/risingbubble/ 
or
https://github.com/CliMA/ClimateMachine.jl/blob/master/tutorials/Atmos/risingbubble.jl

src/Numerics/ODESolvers/DifferentialEquations.jl
ChrisRackauckas: The test problems have it as a required argument: https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Numerics/ODESolvers/ode_tests_basic.jl#L18

src/Numerics/ODESolvers/DifferentialEquations.jl
```diff
@@ -0,0 +1,115 @@
...
+        integ = DiffEqBase.init(prob,alg,args...;kwargs...)
+        new{typeof(integ)}(integ)
+    end
+end
+
+gettime(solver::AbstractDiffEqJLSolver) = solver.integ.t
+getdt(solver::AbstractDiffEqJLSolver) = solver.integ.dt
+updatedt!(solver::AbstractDiffEqJLSolver, dt) = DiffEqBase.set_proposed_dt!(solver.integ,dt)
+updatetime!(solver::AbstractDiffEqJLSolver, t) = DiffEqBase.set_t!(solver.integ,t)
+isadjustable(solver::AbstractDiffEqJLSolver) = true # Is this isadaptive? Or something different?
```
thomasgibson, line 67: `isadjustable` here mainly deals with ode solvers with an implicit component (IMEX RK for example). We currently don't support adaptive time-stepping. It has less to do with the overall `ODESolver` and more to do with the linear solver for the implicit part (you can see how it's used in `src/ODESolvers/BackwardEulerSolvers.jl`.

It's basically a flag determining whether or not the solver can/should be "rebuilt" for changing `dt`. An example would be using an IMEX Runge-Kutta method with different stage time-steps. If you're using, for example, and LU factorization to solve the implicit part, then `isadjustable` will help decide whether or not the LU factors are built only once or during each stage update.

I think for now, what you have here is fine. This is something we need to look more carefully at and try to clean up. IIRC, a solver being adjustable is determined by the `LinearBackwardEulerSolver` passed to the IMEX method.

@jkozdon can probably correct me if I am wrong on this.

src/Numerics/ODESolvers/ODESolvers.jl
```diff
@@ -43,7 +43,7 @@ function general_dostep!(
     timeend::Real;
     adjustfinalstep::Bool,
 )
-    time, dt = solver.t, solver.dt
+    time, dt = gettime(solver), getdt(solver)
```
thomasgibson, line 46: Thank you for cleaning lines 46, 59, and 62!

src/Numerics/ODESolvers/DifferentialEquations.jl
```diff
@@ -0,0 +1,115 @@
...
+    solver::AbstractDiffEqJLSolver,
+    p,
+    time,
+    slow_δ = nothing,
+    slow_rv_dQ = nothing,
+    in_slow_scaling = nothing,
+)
+
+    integ = solver.integ
+    integ.p = p # Can this change?
```
thomasgibson, line 104: Sorry for the confusion! Sadly, `p` is never really documented or used by most ODESolvers. It's only ever used in special circumstances where we are nesting RK methods within, say, a multirate or multirate infinitesimal-step (only the GARK solvers right now) time-stepping method. It's more or less a context manager for passing information through nested ODESolvers.

So in answer to your question on whether it can change: it may certainly change if this particular ODESolver becomes part of a multi-rate method. However, I am not sure what this would look like using the differential evolution algorithm (I assume this is what you call the DE algorithm correct?).

My experience with multirate is purely within the Runge-Kutta framework.

ChrisRackauckas:
@kanav99 did https://github.com/SciML/OrdinaryDiffEq.jl/pull/1153 which should make it so the LSRK methods have the optimal number of kernel calls per time step.

src/Numerics/ODESolvers/DifferentialEquations.jl
ChrisRackauckas: Okay. Hmm, if it doesn't have the parameters information then we won't be able to do adjoint sensitivities. I'll want to talk to someone about that, but that can be the next thing to do.

src/Numerics/ODESolvers/DifferentialEquations.jl
ChrisRackauckas: >It's basically a flag determining whether or not the solver can/should be "rebuilt" for changing dt. 

Oh, DiffEq is all safe with changing `dt`, so we can just set it to `false`. LU-factorization reduction via Jacobian reuse is one thing we'll need to look into though.

src/Numerics/ODESolvers/DifferentialEquations.jl
simonbyrne: we should change that!

kanav99:
Hi! I have been trying the Atmos Rising Bubble example and I can't seem to debug a problem. For some reason, I get negative temperature and which causes the formula `c = √(γRT/m)` to fail. Can I get some help where in the code T can get a negative value? Might help me debugging why it passes on the inbuilt solvers but fail on DifferentialEquations solvers.

charleskawczynski:
>Hi! I have been trying the Atmos Rising Bubble example and I can't seem to debug a problem. For some reason, I get negative temperature and which causes the formula c = √(γRT/m) to fail. Can I get some help where in the code T can get a negative value? Might help me debugging why it passes on the inbuilt solvers but fail on DifferentialEquations solvers.

Hi @kanav99, @akshaysridhar may know better but, my first guess is to try using a smaller timestep / Courant number in case this is a stability issue.

kanav99:
Thanks @charleskawczynski I have just changed the https://github.com/CliMA/ClimateMachine.jl/blob/master/tutorials/Atmos/risingbubble.jl#L215 line to use the new Solver instead. So I suppose the time step should be the same for both. I haven't checked Courant Number, I will try to see if that's same for both cases or not

kanav99:
I used the exact same method used in the tutorial (called `NDBLSRK144` in OrdinaryDiffEq) and it works fine until it errors with
```
Fatal error in MPI_Comm_dup: Other MPI error, error stack:
MPI_Comm_dup(179)...................: MPI_Comm_dup(comm=0xc4000005, new_comm=0x1068d68b0) failed
MPI_Comm_dup(164)...................: 
MPIR_Comm_dup_impl(57)..............: 
MPII_Comm_copy(571).................: 
MPIR_Get_contextid_sparse_group(602): Too many communicators (0/2048 free on this process; ignore_id=0)
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=337723919
:```
I think this is just come limit set my MPI and would run fine on good machines. 
Where should I add this tutorial as a test? @simonbyrne

**Comment by simonbyrne** [comment]
@kanav99 do you have a script that can reproduce this error?

**Comment by kanav99** [comment]
Here is the script - https://gist.github.com/kanav99/051da5a338c8c52ff23d9556867e92ed.  I just changed the solver to the DiffEqJLSolver.
Also what all steps should I take to complete this project?

**Comment by simonbyrne** [comment]
@kanav99 OrdinaryDiffEq isn't in the Manifest:
```
ERROR: LoadError: ArgumentError: Package OrdinaryDiffEq not found in current path:
- Run `import Pkg; Pkg.add("OrdinaryDiffEq")` to install the OrdinaryDiffEq package.
```

**Comment by kanav99** [comment]
Does that cause some problem in ClimateMachine? I mean, is OrdinaryDiffEq needed to be added as a Project dependency? Can't we let it use from our environment?

**Comment by simonbyrne** [comment]
Adding that in then gives the error
```
[ Info: Initializing DryRisingBubble
ERROR: LoadError: MethodError: no method matching OrdinaryDiffEq.LowStorageRK2NCache(::ClimateMachine.MPIStateArrays.MPIStateArray{Float64,NamedTuple{(:ρ, :ρu, :ρe, :turbulence, :hyperdiffusion, :moisture, :radiation, :tracers),Tuple{Float64,SArray{Tuple{3},Float64,1,3},Float64,NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(:ρχ,),Tuple{SArray{Tuple{4},Float64,1,4}}}}},Array{Float64,3},Array{Int64,1},SubArray{Float64,3,Array{Float64,3},Tuple{Base.Slice{Base.OneTo{Int64}},Base.Slice{Base.OneTo{Int64}},UnitRange{Int64}},true},ClimateMachine.MPIStateArrays.CMBuffers.CMBuffer{Float64,Array{Float64,2},Nothing}}, ::ClimateMachine.MPIStateArrays.MPIStateArray{Float64,NamedTuple{(:ρ, :ρu, :ρe, :turbulence, :hyperdiffusion, :moisture, :radiation, :tracers),Tuple{Float64,SArray{Tuple{3},Float64,1,3},Float64,NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(:ρχ,),Tuple{SArray{Tuple{4},Float64,1,4}}}}},Array{Float64,3},Array{Int64,1},SubArray{Float64,3,Array{Float64,3},Tuple{Base.Slice{Base.OneTo{Int64}},Base.Slice{Base.OneTo{Int64}},UnitRange{Int64}},true},ClimateMachine.MPIStateArrays.CMBuffers.CMBuffer{Float64,Array{Float64,2},Nothing}}, ::Array{Float64,3}, ::Array{Float64,3}, ::OrdinaryDiffEq.LowStorageRK2NConstantCache{13,Float64,Float64}, ::Nothing, ::Bool)
Closest candidates are:
  OrdinaryDiffEq.LowStorageRK2NCache(::#273#uType, ::#273#uType, ::#274#rateType, !Matched::#273#uType, ::#275#TabType, ::#276#WrapperType, ::Bool) where {#273#uType, #274#rateType, #275#TabType, #276#WrapperType} at /Users/simon/.julia/packages/OrdinaryDiffEq/RYCSN/src/caches/low_storage_rk_caches.jl:4
```
@kanav99: if you can create your own branch that works, I can try to debug your issue.

**Comment by simonbyrne** [comment]
> Does that cause some problem in ClimateMachine? I mean, is OrdinaryDiffEq needed to be added as a Project dependency? Can't we let it use from our environment?

Yes, it needs to be added to the environment you run it in. Do you have the Project and Manifest you used?

**Comment by kanav99** [comment]
> if you can create your own branch that works, I can try to debug your issue.  
Oh yes sorry, I forgot that I did some changes in the OrdinaryDiffEq code (`zero(MPIStateArray)` returns a normal array for some reason).  Use `OrdinaryDiffEq#kg/lskr-similar` please.
> Do you have the Project and Manifest you used?  
I ran the code in the ClimateMachine environment, it picked up OrdinaryDiffEq from my main julia environment.

**Comment by simonbyrne** [comment]
@kanav99 can you just create your own branch with these changes? Then I can just clone it and see what is happening.

**Comment by simonbyrne** [comment]
Also, I think https://github.com/SciML/OrdinaryDiffEq.jl/pull/1153#issuecomment-637268143 would need to be addressed as well.

**Comment by kanav99** [comment]
Adding the OrdinaryDiffEq dependency give's me this error:
```
ERROR: LoadError: LoadError: UndefVarError: MP_n_0 not defined
Stacktrace:
 [1] include(::Module, ::String) at ./Base.jl:377
 [2] include(::String) at /Users/kanav/.julia/dev/ClimateMachine.jl/src/ClimateMachine.jl:1
 [3] top-level scope at /Users/kanav/.julia/dev/ClimateMachine.jl/src/ClimateMachine.jl:15
 [4] include(::Module, ::String) at ./Base.jl:377
 [5] top-level scope at none:2
 [6] eval at ./boot.jl:331 [inlined]
 [7] eval(::Expr) at ./client.jl:449
 [8] top-level scope at ./none:3
in expression starting at /Users/kanav/.julia/dev/ClimateMachine.jl/src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl:16
in expression starting at /Users/kanav/.julia/dev/ClimateMachine.jl/src/ClimateMachine.jl:15
```
Without the addition, the code runs fine

**Comment by simonbyrne** [comment]
`] add CLIMAParameters@0.1.4`, or merge into the current ClimatMachine master (where we cap the package).

**Comment by kanav99** [comment]
@simonbyrne https://github.com/kanav99/ClimateMachine.jl/tree/lsrk this is my branch

**Comment by simonbyrne** [comment]
@kanav99 can you add in the version of OrdinaryDiffEq you were using?

**Comment by kanav99** [comment]
I have checked out #kg/lskr-similar branch of OrdinaryDiffEq, that is reflected in the Manifest.toml

**Comment by simonbyrne** [comment]
OrdinaryDiffEq isn't in https://github.com/kanav99/ClimateMachine.jl/blob/lsrk/Manifest.toml

**Comment by kanav99** [comment]
Sorry again, I updated my ClimateMachine branch, I forgot to commit the manifest and project files.

**Comment by simonbyrne** [comment]
I'm not exactly sure what is happening, but its clear you're creating lots of MPIStateArrays, and for some reason they're not being garbage collected.

**Comment by simonbyrne** [comment]
My first thought was that by calling `MPI_comm_dup` iteratively, MPICH itself wasn't cleaning up internally, but `MPI.REFCOUNT[]` is also increasing and manually calling `GC.gc()` doesn't fix it.

There must be something that is holding references to all the `MPIStateArray`s that is preventing garbage collection.

**Comment by ChrisRackauckas** [comment]
@kanav99 make sure you're running it with settings like `save_everystep=false`.

**Comment by simonbyrne** [comment]
yup, it looks like that
```
using ClimateMachine, MPI
ClimateMachine.init()

using ClimateMachine.MPIStateArrays

a = 100
b = 1
c = 1 / 100
Δ = sqrt(4 * a * c - b^2)


function rhs!(dQ, Q, ::Nothing, t; increment = false)
    if increment
        @. dQ += $cos(t) * (a + b * Q + c * Q^2)
    else
        @. dQ = $cos(t) * (a + b * Q + c * Q^2)
    end
end

function exactsolution(t, q0, t0)
    k = @. 2 * atan((2 * c * q0 + b) / Δ) / Δ - sin(t0)
    solution = @. (Δ * tan((k + sin(t)) * Δ / 2) - b) / (2 * c)
    return Array(solution)
end

q0 = MPIStateArray{Float64}(MPI.COMM_WORLD, Array, 1,1,1)
Q = MPIStateArray{Float64}(MPI.COMM_WORLD, Array, 1,1,1)

fill!(q0,-1)
t0 = 0.1
finaltime = 1.2

Qinit = exactsolution(t0, q0, t0)

dt = 2.0^-6
using ClimateMachine.ODESolvers, OrdinaryDiffEq
solver = DiffEqJLSolver(rhs!, NDBLSRK144(;williamson_condition=false), Q; t0 = t0, dt = dt, adaptive = false, calck = false)

ODESolvers.solve!(Qinit, solver; timeend = finaltime)
MPI.REFCOUNT[]

size(solver.integ.sol.k)
```
seems to be saving all the intermediate arrays.

One thing we would need is assurance that we don't allocate _any_ storage during the `solve!` (allocating work arrays during setup is fine). This is especially important on the GPU.

**Comment by ChrisRackauckas** [comment]
So then `save_everystep=false,save_start=false,save_end=false` should be setup by default. That will make the solution empty and just use the stepper.

**Comment by kanav99** [comment]
I checked with the Atmos example and they worked now! As I can't push to this branch, I will suggest some changes.
Apart from that, to work with OrdinaryDiffEq, we also need to have an overload of `zero` for `MPIStateArray`, because right now, `zero(mpisa)` gives me `Array{Float64,3}`. So I had to manually replace all the zero calls to similar and fill.

File: src/Numerics/ODESolvers/DifferentialEquations.jl [hunk 6]
```diff
@@ -0,0 +1,115 @@
...
+
+    function DiffEqJLIMEXSolver(rhs!,rhs_implicit!,alg,Q,args...;
+                            t0 = 0,
+                            kwargs...)
+        prob = SplitODEProblem(
+                          (du,u,p,t)->rhs_implicit!(du,u,p,t; increment=false),
+                          (du,u,p,t)->rhs!(du,u,p,t; increment=false),
+                           Q, (float(t0),typemax(typeof(float(t0)))),
+                           p)
+        integ = DiffEqBase.init(prob,alg,args...;kwargs...)
```
kanav99, line 58:
```suggestion
        integ = DiffEqBase.init(prob,alg,args...;adaptive = false, save_everystep = false, save_start = false, save_end = false, kwargs...)

```

src/Numerics/ODESolvers/DifferentialEquations.jl
```diff
@@ -0,0 +1,115 @@
...
+equation given by the right-hand-side function `f` with the state `Q`, i.e.,
+
+```math
+  \\dot{Q} = f_I(Q, t) + f_E(Q, t)
+```
+
+via a DifferentialEquations.jl DEAlgorithm, which includes support
+for OrdinaryDiffEq.jl, Sundials.jl, and more.
+"""
+mutable struct DiffEqJLIMEXSolver{I} <: AbstractDiffEqJLSolver
```
kanav99, line 47: We would also need a constructor which takes in a OrdinaryDiffEq algorithm and outputs a function like we have for LSRK144NiegemannDiehlBusch - https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/ODESolvers/LowStorageRungeKuttaMethod.jl#L331

kanav99:
@simonbyrne This PR should be good, it now runs on OrdinaryDiffEq master, we need to do the changes on the OrdinaryDiffEq side, if you have any changes in mind, let me know!

kanav99:
Should we add OrdinaryDiffEq to test dependencies or normal ones?

simonbyrne:
> Should we add OrdinaryDiffEq to test dependencies or normal ones?

Just add it to the regular dependencies.

kanav99:
What else do we need in this to be merged @charleskawczynski ?

charleskawczynski:
> What else do we need in this to be merged @charleskawczynski ?

Looks like we need to rebase. Are we finished with implementation and tests? If so I think we can merge after resolving conflicts+rebase.

simonbyrne:
This looks great, thanks to @kanav99 and @charleskawczynski for getting this cleaned up.

bors r+

jkozdon:
@simonbyrne: Remove WIP in PR title before merge?

PR #1163, opened 2020-05-27
Mountain gravity waves: linear hydrostatic and non-hydrostatic

smarras79:
# Description

This draft PR is for:

1) linear hydrsotatic gravity waves: Case 6 in [1] 
2)non-linear hyrostatuc gravity waves: Section 5.2.5. Nonlinear Hydrostatic Flow in [2]
3)  removed experiments/AtmosLES/equilibrium.jl --> there will be a PR for this alone.

From this tutorial you will also see how a mountain is added by looking for the function `warp_agnesi` in tutorial/Atmos/agnesi_*

**References:**
[1] Giraldo and Restelli (2007) A study of spectral element and discontinuous Galerkin methods for the Navier–Stokes equations in nonhydrostatic mesoscale atmospheric modeling: Equation sets and test cases, J. Comput. Phys. 227

[2] Restelli PhD thesis

I have

### DIFF
--- a/docs/list_of_tutorials.jl
+++ b/docs/list_of_tutorials.jl
@@ -23,6 +23,10 @@ if generate_tutorials
         "Atmos" => [
             "Dry Idealized GCM" => "Atmos/heldsuarez.jl",
             "Flow in Single Element Stack" => "Atmos/burgers_single_stack.jl",
+            "Linear Hydrostatic Mountain (Topography)" =>
+                "Atmos/agnesi_hs_lin.jl",
+            "Linear Non-Hydrostatic Mountain (Topography)" =>
+                "Atmos/agnesi_nh_lin.jl",
         ],
         "Ocean" => [],
         "Land" => ["Heat" => ["Heat Equation" => "Land/Heat/heat_equation.jl"]],
--- a/tutorials/Atmos/agnesi_hs_lin.jl
+++ b/tutorials/Atmos/agnesi_hs_lin.jl
@@ -1,0 +1,320 @@
+# # [Linear HS mountain waves (Topography)](@id EX-LIN_HS-docs)
+#
+# ## Description of experiment
+# 1) Dry linear Hydrostatic Mountain Waves
+# The atmosphere is dry and the flow impinges against a witch of Agnesi mountain of heigh `hm=1 m`
+# and base parameter `a=10,000` m and centered in `xc = 120 km` in a 2D domain
+# `\Omega = 240 km\times 50 km`. The mountain is defined as
+#
+# ``
+#  z = \frac{hm}{1 + \frac{x - xc}{a}}
+# ``
+# The 2D problem is setup in 3D by using 1 element in the y direction.
+# To damp the upward moving gravity waves, a Reyleigh absorbing layer is added at `z = 15,000 m`.
+#
+# The initial atmosphere is defined such that it has a stability frequency `N=g/\sqrt{c_p T_0}`, where
+#
+# ``T_0 = 250 K``
+# so that
+#
+# ``
+# \theta = \theta_0 = T_0
+# ``
+# ``
+# \pi = 1 + \frac{g^2}{c_p \theta_0 N^2}\left(\exp\left(\frac{-N^2 z}{g} \right)\right)
+#
+# where `theta_0 = T_0 K`.
+# ``
+# so that
+#
+# ``
+# ρ = \frac{p_{sfc}}{R_{gas}\theta}pi^{c_v/R_{gas}}
+# ``
+# and
+# ``
+# \theta \pi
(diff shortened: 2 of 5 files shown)


smarras79:
bors try

smarras79:
bors try

smarras79:
I added both agnesi_hs_lin.jl and agnesi_hs_nl.jl to tutorials/Atmos but I wonder whether one of these two should go into experiments instead since the setups are similar.

smarras79:
I set timeend to a smaller timee than requested by the nominal test for 2 resons:
1) so that it can be run among the CI tests
2) because we need lateral sponges (other the top sponge which is already there) to run to completion to the nominal simulation time of several hours. @slifer50 added lateral sponges to his TC test so that feel free to pull this branch and add them this simple 2D(1/2) test case whenever you have a chance.

smarras79:
bors try

smarras79:
bors try

smarras79:
bors try

tutorials/Atmos/agnesi_hs_lin.jl
```diff
@@ -0,0 +1,306 @@
...
+    cbtmarfilter = GenericCallbacks.EveryXSimulationSteps(1) do (init = false)
+        Filters.apply!(solver_config.Q, 6, solver_config.dg.grid, TMARFilter())
+        nothing
+    end
+
+    # Invoke solver (calls `solve!` function for time-integrator), pass the driver, solver and diagnostic config
+    # information.
+    result = ClimateMachine.invoke!(
+        solver_config;
+        #user_callbacks = (cbtmarfilter,),
```
thomasgibson, line 298: Was this intentional? Just making sure the filter was not meant to be applied.

tutorials/Atmos/agnesi_hs_lin.jl
```diff
@@ -0,0 +1,306 @@
...
+
+    ##SR LSRK144
+    ode_solver = ClimateMachine.ExplicitSolverType(
+        solver_method = LSRK144NiegemannDiehlBusch,
+    )
+
+    source = (Gravity(), rayleigh_sponge)
+
+    temp_profile_ref = DecayingTemperatureProfile{FT}(param_set)
+    ref_state = HydrostaticState(temp_profile_ref)
```
thomasgibson, line 229: Please correct me if I'm wrong, but shouldn't this test case use an isothermal background profile?

tutorials/Atmos/agnesi_hs_lin.jl
smarras79: yes, no need for filter here. It doesn't hurt to have it on but it should not be necessary.

tutorials/Atmos/agnesi_hs_lin.jl
smarras79: theta should be uniform in z, but T will still decrease. My question is whether DecayingTemperatureProfile is consistent with the initial background though.

akshaysridhar:
Screen Shot 2020-06-09 at 7 12 21 PM

At 7980 seconds, we have a stable solution, will continue running for the required 15000 seconds to verify. @mwarusz thanks for the filter updates .

smarras79:
bors try

smarras79:
bors try

docs/list_of_tutorials.jl
```diff
@@ -64,6 +66,8 @@ if generate_tutorials
     skip_execute = [
         "Atmos/dry_rayleigh_benard.jl",               # takes too long
         "Atmos/risingbubble.jl",                      # broken
+        "Atmos/agnesi_hs_lin.jl",                     # broken
```
charleskawczynski, line 69: Sorry for flip-flopping. If these tutorials are not generating valid markdown as indicated, then let's remove them from `skip_execute` and `tutorials` above. The `heat_equation.jl` and `heldsuarez.jl` tutorials generate valid markdown, for reference, if you'd like to make the necessary changes and if you think they would make good tutorials.

In particular, the main issue with not generating valid markdown is that, inside functions, double #'s must be used for comments to tell Literate to not break the function into multiple example blocks (`heldsuarez.jl` uses this extensively).

smarras79: thanks @charleskawczynski let me take a look at how that is done. However, this may take me some time because I am not on this full time. What do you suggest in the meantime so that these drivers are kept up to date with master? These are important tests to always makes sure things are not breaking. 
Thanks

tapios:
@charleskawczynski Could you please work with @smarras79 to get this merged? Either by fixing the docs here to be the way they ought to be, or by establishing some "parking lot" where this can be merged before it's moved to its final destination.

smarras79:
bors try

PR #1170, opened 2020-05-28
Combine LinearSolvers into single module

charleskawczynski:
# Description

This PR combines the linear solvers into a single module, just as we did with the ODESolvers.jl module. This is convenient for a few reasons:

 - Less boilerplate
 - Fewer modules to keep track of for `using`/`import`
 - Easier to cross-reference different linear solvers in the docs

The only drawback I see from this is that this does combine the namespace. We could, instead, make each of these modules submodels of LinearSolvers, but I'm not sure it's worth the added structure.

I have


blallen:
If we're doing this, should we name the module `ImplicitSolvers`? Some things like GMRES can be used on non-linear problems.

charleskawczynski:
Good point. I kind of don't like the name `ImplicitSolvers`, though. How about `IterativeSolvers`? I haven't dug through carefully, are any of them direct solvers?

blallen:
> Good point. I kind of don't like the name `ImplicitSolvers`, though. How about `IterativeSolvers`? I haven't dug through carefully, are any of them direct solvers?

The LU solvers are direct solvers. Not sure if there are any others

charleskawczynski:
`AxEqualsBSolvers` 😄

charleskawczynski:
Maybe we should split this into `IterativeSolvers` and `DirectSolvers`. I think this is a common way to separate things.

blallen:
> Maybe we should split this into `IterativeSolvers` and `DirectSolvers`. I think this is a common way to separate things.

@sandreza and @thomasgibson, what are your thoughts?

jkozdon:
`SystemSolvers`?

charleskawczynski:
I do like that we could keep them all together with `SystemSolvers`!

docs/src/APIs/Numerics/SystemSolvers/LinearSolvers.md
```diff
@@ -0,0 +1,34 @@
+# System Solvers
```
blallen, line 1: might want to update the name of this file as well

docs/src/HowToGuides/Numerics/ODESolvers/Timestepping.md
```diff
@@ -90,8 +90,8 @@ For example, consider the following:
 ```julia
 export MyIMEXMethod
 
-using ..LinearSolvers
-const LS = LinearSolvers
+using ..SystemSolvers
+const LS = SystemSolvers
```
blallen, line 94: maybe do `LS` -> `SS`, just so future people don't get confused, but not entirely necessary

thomasgibson:
What's wrong with `LinearSolver`? If you have a nonlinear solver, then you have a `NonlinearSolver`. A `LinearSolver` could be used as part of a `NonlinearSolver`, i.e. Newton-Krylov methods for solving PDEs.

ATM there isn't any support for nonlinear solvers (yet!) and I think throwing them in this PR is premature; we don't want to bulldoze something in without working out a proper software abstraction (refactoring LinearSolvers into a single module while maintaining current functionality/abstraction is totally fine I think).

I think `LinearSolver` is acceptable. Distinguishing between `IterativeSolver` and `DirectSolver` is certainly an option. This, however, may change when @sandreza and I start bringing in preconditioning.

For example, one can view computing an LU factorization as a preconditioner. Consider the (left) preconditioned residual of the linear system (assuming a linear preconditioner):
```
0 = P(b - Ax) = P(b) - P(Ax)
```
Whenever `P` is applying the inverse of the LU factors, or a Cholesky factorization for symmetric systems, then we get `0 = A^{-1}b - Ix`, where `I` is the identity matrix. In this sense, all `DirectMethods` can actually be recast as a preconditioner. Moreover, we can have a bunch of Krylov/iterative methods (columwise/batched, monolithic, etc.), which take preconditioners.
In the case I'm illustrating, ANY iterative method (based on the minimization of residuals) acting on the linear system here should converge in ONE iteration. In other numerical linear/nonlinear algebra libraries like PETSc, direct methods are simply treated as 'exact' preconditioners with a linear solver (think iterative method) that does one thing: apply the preconditioner (`KSPType Preonly` for the PETSc gurus)

We should try to emulate the mathematical abstraction summarized in Chapter 4 of https://www.mcs.anl.gov/petsc/petsc-current/docs/manual.pdf

Why do I think this? One word: composability. I think this will save us all a lot of trouble later on down the line if we can come up with a robust and composable software abstraction. Right now, we don't have preconditioners and we have two disjoint linear solver types (monolithic GMRES, GCR, and batched methds like bgmres and the conjugate grandient method in master now). This suggests that preconditioning may need to take special care when dealing with the different types of iterative methods we have now (columwise/batched preconditioning, for example). This is also why we should very carefully think about how LinearSolvers should be updated.

TLDR; combining LinearSolvers into a single module is great. Let's not rush into changing LinearSolvers too much though, otherwise, it might just make engineering a "better" way to do linear/nonlinear solvers more cumbersome and difficult.

src/Numerics/SystemSolvers/batched_generalized_minimal_residual_solver.jl
```diff
@@ -44,7 +39,7 @@ Solving n linear systems iteratively
 - Needs to perform a transpose of original data structure into current data structure: Could perhaps do a transpose free version, but the code gets a bit clunkier and the memory would no longer be coalesced for the heavy operations
 """
 struct BatchedGeneralizedMinimalResidual{FT, IT, VT, AT, TT1, TT2} <:
-       LS.AbstractIterativeLinearSolver
+       AbstractIterativeLinearSolver
```
blallen, line 42: Maybe switch this to `AbstractIterativeSystemSolver`

src/Numerics/SystemSolvers/columnwise_lu_solver.jl
```diff
@@ -1,20 +1,7 @@
...
-    BalanceLaw, DGModel, number_state_conservative, number_state_gradient_flux
-using ..LinearSolvers
-const LS = LinearSolvers
-using ..MPIStateArrays
-using LinearAlgebra
-using KernelAbstractions
-using Adapt
-import ..Mesh.Grids: polynomialorder, dimensionality
-
 abstract type AbstractColumnLUSolver <: AbstractLinearSolver end
```
blallen, line 5: Maybe replace this with `AbstractSystemSolver`

charleskawczynski:
> What's wrong with LinearSolver? .... ATM there isn't any support for nonlinear solvers (yet!) 

As @blallen's noted:

> Some things like GMRES can be used on non-linear problems.

Since GMRES is currently in `LinearSolvers`, I suggest we move forward with changing the name to `SystemSolvers`. AFAIU, this shouldn't impede any functionality, comparability, or abstractions.

docs/src/HowToGuides/Numerics/ODESolvers/Timestepping.md
charleskawczynski: We can actually just delete this since everything is in one module.

Issue #1169, opened 2020-05-28
Clarification for reference state setup and custom reference state definitions


slifer50:
# Description
@smarras79 and I have been  working on a balanced flow test for the flow in a box with topography where  if the initial  state is hydrostatically balanced we should see everything  be still and w = 1e-16.  Currently we've seen the  following: 
1- Using initial conditions that are hydrostatically balanced but do not correspond to a the reference state used yield largely non zero w values as shown in the image below:
![image (2)](https://user-images.githubusercontent.com/54990670/83160751-95355500-a0d5-11ea-81a6-2ca9a7243e6d.png)
2- Using The same  initial conditions while specifying NoReferenceState yields w =1e-7:
![image (3)](https://user-images.githubusercontent.com/54990670/83160848-b433e700-a0d5-11ea-8c4f-2b796e616062.png)
3-Using Initial conditions set to be the same as those in  the reference state yield w = 1e-13.
![image (4)](https://user-images.githubusercontent.com/54990670/83160939-cdd52e80-a0d5-11ea-985a-eab5ab178e39.png).
The first point is that because a default reference state is always set, it's important that the user has to know that when using initial conditions not based off of it, they need to either turn it off or set a corresponding reference state and this should be documented in some way. The second point is that when using sounding to define initial conditions, there is no guarantee that the temperature profiles currently defined are able to define an adequate reference state for some given sounding which would result in unexpected non-hydrostatic behavior, this of particular importance since the multi-rate/IMEX solvers require a reference state and using the NoReferenceState setting is not an option when using them. As such it seems to me that we will need to be able to set up custom reference states based on sounding data. 

# Additional context

This will help avoid confusion among new users and allow for less issues when setting up sounding based cases. 

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

akshaysridhar:
In `AtmosModel.jl` we have 
```
    if m.ref_state isa HydrostaticState
        flux.ρu += (p - aux.ref_state.p) * I
    else
        flux.ρu += p * I
    end 
```
The following change should address the sounding issue - this does not address the default driver issue and interaction with timesteppers. 
```
    if m.ref_state isa NoReferenceState
        flux.ρu += p * I
    else
[...]

PR #1171, opened 2020-05-28
Update Atmos.md to add small warning about reference state use-cases

slifer50:
# Description

Documentation change for reference state see issue #1169 

I have added a brief warning to the documentation for driver configuration defaults. This seeks to warn new users about the necessity to adjust from the default reference state setting based on the solver or initial conditions used.

### DIFF
--- a/docs/src/GettingStarted/Atmos.md
+++ b/docs/src/GettingStarted/Atmos.md
@@ -34,6 +34,15 @@ possible options for each subcomponent.
     configurations.  Equation sets are written in vector-invariant form and
     solved in Cartesian coordinates.  The component `orientation` determines
     whether the problem is solved in a `box (LES)` or a `sphere (GCM)`)
+    
+!!! warning
+
+    The above defaults do not consider any potential interaction with the time stepping
+    components. The user must take appropriate care when setting up the driver configurations.
+    The default hydrostatic reference state may not be suitable to any initial conditions and 
+    it recommended to use a reference state corresponding to the initial condition's background.
+    It is also recommended to use the NoReferenceState type when using a single rate explicit 
+    time stepper.
 
 
 ### [GCM Configuration](@id GCMConfig)(with defaults)


docs/src/GettingStarted/Atmos.md
```diff
@@ -34,6 +34,15 @@ possible options for each subcomponent.
     configurations.  Equation sets are written in vector-invariant form and
     solved in Cartesian coordinates.  The component `orientation` determines
     whether the problem is solved in a `box (LES)` or a `sphere (GCM)`)
+    
+!!! warning
+
+    The above defaults do not consider any potential interaction with the time stepping
+    components. The user must take appropriate care when setting up the driver configurations.
+    The default hydrostatic reference state may not be suitable to any initial conditions and
```
kpamnany, line 42:
```suggestion
    The default hydrostatic reference state may not be suitable to every initial condition and 
```

docs/src/GettingStarted/Atmos.md
```diff
@@ -34,6 +34,15 @@ possible options for each subcomponent.
     configurations.  Equation sets are written in vector-invariant form and
     solved in Cartesian coordinates.  The component `orientation` determines
     whether the problem is solved in a `box (LES)` or a `sphere (GCM)`)
+    
+!!! warning
+
+    The above defaults do not consider any potential interaction with the time stepping
+    components. The user must take appropriate care when setting up the driver configurations.
+    The default hydrostatic reference state may not be suitable to any initial conditions and 
+    it recommended to use a reference state corresponding to the initial condition's background.
```
kpamnany, line 43:
```suggestion
    it is recommended to use a reference state corresponding to the initial condition's background.
```

docs/src/GettingStarted/Atmos.md
```diff
@@ -34,6 +34,15 @@ possible options for each subcomponent.
...
+    It is also recommended to use the NoReferenceState type when using a single rate explicit
```
kpamnany, line 44:
```suggestion
    It is also recommended to use the `NoReferenceState` type when using a single rate explicit 
```

smarras79:
> Rather than adding a warning, my preference would be to resolve this issue. We should be able to specify any initial state with any hydrostatic reference state--they really should not interact, except in questions of numerical cancelations in the upper atmosphere because of finite-precision arithmetic and low densities.

I agree with @tapios 
This afternoon @akshaysridhar @bischtob and I were on a phone a while to discuss the way forward. A somewhat detailed description of the issue is on the way and we can discuss options based on that.

Issue #947, opened 2020-04-08
Namelist for simplified user experience


smarras79:
# Description
_Terminology: in the following, "driver" indicates any setup file contained in `experiments` and not the `Driver.jl` file._

I'd like to go back to the open discussion from last summer and fall about the use of a namelist file that contains the main and only informations needed to execute CLIMA.

As of now, although the drivers have been greatly cleaned up and minimized in the past few months, they still require a user to be very familiar with Julia code; not only, the user should also be familiar with the guts of CLIMA if he really wants to be able to modify the driver without breaking the code or without removing relevant information from it.

To avoid this issue, it is standard practice in the atmospheric (WRF, CM1, Pycles, NUMA,  etc.) and CFD (OpenFOAM, Nektar++, etc.) communities to use a `namelist` file (often simply called input file) that does not contain any code but. A namelist is nothing else than only a list of parameters that the code will simply read and do what it should. 
These namelist can be as simple as a sequential list of parameters, or more hierarchical such as the one used by Pycles which is a JSON file.

The most simple (and most used) type of namelist is the sequential list of parameters. For example, here is the one used by Pycles:

`{
    "conditional_stats": {
        "classes": [
            "Spectra"
        ],
        "frequency": 600.0,
        "stats_dir": "cond_stats"
    },
    "damping": {
        "Rayleigh": {
            "gamma_r": 0.002,
            "z_d": 500.0
        },
        "scheme": "Rayleigh"
    },
    "diffusion": {
        "qt_entropy_source": false
    },
    "fields_io": {
        "diagnostic_fields": [
            "ql",
            "temperature",
            "buoyancy_frequency",
            "viscosity"
        ],
        "fields_dir": "fields",
        "frequency": 3600.0
    },
    "grid": {
        "dims": 3,
        "dx": 35.0,
        "dy": 35.0,
        "dz": 5.0,
        "gw": 5,
        "nx": 96,
        "ny": 96,
        "nz": 300
    },
    "microphysics": {
        "ccn": 100000000.0,
        "cloud_sedimentation": false,
        "phase_partitioning": "liquid_only",
        "scheme": "None_SA"
    },
    "momentum_transport": {
        "order": 7
    },
  
    "output": {
        "output_root": "./"
    },
    "restart": {
        "frequency": 600.0,
        "init_from": false,
        "input_path": "./",
        "output": true
    },
    "sgs": {
        "scheme": "Smagorinsky"
    },
    
    "thermodynamics": {
[...]

PR #1172, opened 2020-05-29
Separate cli arg parsing from initialization

### LINKED ISSUE #1167: ClimateMachine.init() should not parse process ARGS by default?
# Description

Currently ClimateMachine parses command line arguments by default upon initialization: https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Driver.jl#L212.

This has the side effect of exclusively claiming process global `ARGS`, which doesn't compose well when invoked in a programmatic / sub-procedure context (ex. it makes creating an alternative CLI program that invokes ClimateMachine awkward to write).  One example of this behaviour is that it breaks running a subset of tests with the current test harness: 

`julia --project=@. -e 'using Pkg; Pkg.test(test_args=["Numerics"])`

```
Starting tests for Numerics
Starting tests for Mesh
┌ Warning: Incompatibility detected between CUDA and LLVM 8.0+; disabling debug info emission for CUDA kernels
└ @ CUDAnative ~/.julia/packages/CUDAnative/hfulr/src/CUDAnative.jl:114
too many arguments
usage: <PROGRAM> [--disable-gpu] [--show-updates <interval>]
                 [--diagnostics <interval>] [--vtk <interval>]
                 [--monitor-timestep-duration <interval>]
                 [--monitor-courant-numbers <interval>]
                 [--checkpoint <interval>] [--checkpoint-keep-all]
                 [--checkpoint-at-end] [--checkpoint-dir <path>]
                 [--restart-from-num <number>] [--fix-rng-seed]
                 [--log-level <level>] [--output-dir <path>]
                 [--integration-testing]
ERROR: Package ClimateMachine errored during testing
```

### Proposal

`ClimateMachine.int()` should be configured through keyword arguments and the public API should not depend on passing ArgParse data structures. Methods with overloads for simple cases like passing a dictionary of values could be added (ex. json / TOML processing).  `ClimateMachine.cli()` would exhibit the current behaviour of parsing process `ARGS` and then internally invoking `ClimateMachine.init()`.

# For CLIMA Developers

jakebolewski:
# Description

Separates out cli argument processing from `ClimateMachine.init()` by defining a new method `ClimateMachine.cli()` that parses arguments and initializes ClimateMachine runtime.  `ClimateMachine.init()` is preserved to allow for programmatic initialization when command line parsing is not needed (tests / notebooks / scripts).  `ClimateMachine.cli` and `ClimateMachine.init` now take keyword arguments to allow for easy overloading for `ClimateMachine.Settings` at configuration / runtime.  The experiment files have been updated to use `cli`, the tutorials I kept the same. 

Consistent overloading fallback behaviour is also defined: 

1. cli arguments (if applicable)
2.  keyword argument overloads
3. `CLIMATEMACHINE_SETTINGS_<VALUE_NAME>` process environment definitions
4. `ClimateMachine.Settings` default values. 

This allows for easy overloading of settings in batch scripts / container environments without having to define configuration files or change the driver file source code.  We were currently doing this in kind of an add-hoc fashion.

Closes #1167 

I have

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -46,11 +46,12 @@ _init_array(::Type{Array}) = nothing
 
 const cuarray_pkgid =
     Base.PkgId(Base.UUID("3a865a2d-5b23-5a0f-bc46-62713ec82fae"), "CuArrays")
-function gpu_allowscalar(val)
+
+function gpu_allowscalar(val::Bool)
     if haskey(Base.loaded_modules, ClimateMachine.cuarray_pkgid)
         Base.loaded_modules[ClimateMachine.cuarray_pkgid].allowscalar(val)
     end
-    return nothing
+    return
 end
 
 # Note that the initial values specified here are overwritten by the
@@ -67,20 +68,79 @@ Base.@kwdef mutable struct ClimateMachine_Settings
     checkpoint_at_end::Bool = false
     checkpoint_dir::String = "checkpoint"
     restart_from_num::Int = -1
+    fix_rng_seed::Bool = false
     log_level::String = "INFO"
     output_dir::String = "output"
     integration_testing::Bool = false
-    array_type
+    array_type::Type = Array
 end
-const Settings = ClimateMachine_Settings(array_type = Array)
+
+const Settings = ClimateMachine_Settings()
 
 
 """
-    parse_commandline()
+    ClimateMachine.array_type()
+
+Return the array type used by ClimateMachine. This defaults to (CPU-based) `Array`
+and is only correctly set (based on choice from the command line, from
+an environment variable, or from experiment code) after `ClimateMachine.init()`
+is called.
 """
-function parse_commandline(custom_settings)
-    exc_handler =
-        isinteractive() ? ArgParse.debug_handler : ArgParse.default_handler
+array_type() = Settings.array_type
+
+
+"""
+    get_setting(setting_name::Symbol, settings, defaults)
+
+Define fallback behavior for driver settings, first accessing overloaded `settings`
+if defined, followed by constructed global ENV variable `CLIMATEMACHINE_SETTINGS_<SETTING_NAME>`,
+then (global) defaults.
+
+Returns setting value.
+"""
+function get_setting(setting_name::Symbol, settings, defaults)
+    if !haskey(defaults, setting_name)
+        error("setting $setting_name is not defined in `defaults`")
+    end
+    setting_type = typeof(defaults[setting_name])
+    setting_env = "CLIMATEMACHINE_SETTINGS_" * uppercase(String(setting_name))
+    if haskey(settings, setting_name)
+        return convert(setting_type, settings[setting_name])
+    elseif haskey(ENV, setting_env)
+        env_val = ENV[setting_env]
+        v = tryparse(setting_type, env_val)
+        if isnothing(v)
+            error("Cannot parse ENV $setting_env value $env_val, to setting type $setting_type")
+        end
+        return v
+    elseif haskey(defaults, setting_name)
+        return defaults[setting_name]
+    else
+        error("setting $setting_name is not contained in either settings or defaults")
+    end
+end
+
+"""
+    parse_commandline(defaults::Union{Nothing, Dict{Symbol,Any}),
+                      custom_settings::Union{Nothing,ArgParseSettings}=nothing)
+
+Parse process command line ARGS values.
+If `defaults` is not nothing, override default values for cli argument defaults.
+If `custom_settings` arg is not nothing, add ArgParseSettings to the parsing step.
+
+Returns a `Dict` containing parsed process ARGS values.
+"""
+function parse_commandline(
+    defaults::Union{Nothing, Dict{Symbol, Any}} = nothing,
+    custom_settings::Union{Nothing, ArgParseSettings} = nothing,
+)
+    if isnothing(defaults)
+        defaults = Dict{Symbol, Any}()
+    end
+    exc_handler = ArgParse.default_handler
+    if Base.isinteractive()
+        exc_handler = ArgParse.debug_handler
+    end
     s = ArgParseSettings(
         prog = PROGRAM_FILE,
         description = "Climate Machine: an Earth System Model that automatically learns from data\n",
@@ -96,8 +156,13 @@ function parse_commandline(custom_settings)
         preformatted_epilog = true,
         version = string(CLIMATEMACHINE_VERSION),
         exc_handler = exc_handler,
+        autofix_names = true,  # switches --flag-name to 'flag_name'
     )
     add_arg_group!(s, "ClimateMachine")
+
+    global_defaults = Dict{Symbol, Any}(
+        (n, getproperty(Settings, n)) for n in propertynames(Settings)
+    )
 
     @add_arg_table! s begin
         "--disable-gpu"
@@ -107,32 +172,34 @@ function parse_commandline(custom_settings)
         help = "interval at which to show simulation updates"
         metavar = "<interval>"
         arg_type = String
-        default = "60secs"
+        default = get_setting(:show_updates, defaults, global_defaults)
         "--diagnostics"
         help = "interval at which to collect diagnostics"
         metavar = "<interval>"
         arg_type = String
-        default = "never"
+        default = get_setting(:diagnostics, defaults, global_defaults)
         "--vtk"
         help = "interval at which to output VTK"
         metavar = "<interval>"
         arg_type = String
-        default = "never"
+        default = get_setting(:vtk, defaults, global_defaults)
         "--monitor-timestep-duration"
         help = "interval in time-steps at which to output wall-clock time per time-step"
         metavar = "<interval>"
         arg_type = String
-        default = "never"
+        default =
+            get_setting(:monitor_timestep_duration, defaults, global_defaults)
         "--monitor-courant-numbers"
         help = "interval at which to output acoustic, advective, and diffusive Courant numbers"
         metavar = "<interval>"
         arg_type = String
-        default = "never"
+        default =
+            get_setting(:monitor_courant_numbers, defaults, global_defaults)
         "--checkpoint"
         help = "interval at which to create a checkpoint"
         metavar = "<interval>"
         arg_type = String
-        default = "never"
+        default = get_setting(:checkpoint, defaults, global_defaults)
         "--checkpoint-keep-all"
         help = "keep all checkpoints (instead of just the most recent)"
         action = :store_true
@@ -143,12 +210,12 @@ function parse_commandline(custom_settings)
         help = "the directory in which to store checkpoints"
         metavar = "<path>"
         arg_type = String
-        default = "checkpoint"
+        default = get_setting(:checkpoint_dir, defaults, global_defaults)
         "--restart-from-num"
         help = "checkpoint number from which to restart (in <checkpoint-dir>)"
         metavar = "<number>"
         arg_type = Int
-        default = -1
+        default = get_setting(:restart_from_num, defaults, global_defaults)
         "--fix-rng-seed"
         help = "set RNG seed to a fixed value for reproducibility"
         action = :store_true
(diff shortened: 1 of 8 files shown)


src/Driver/Driver.jl
```diff
@@ -107,32 +137,64 @@ function parse_commandline(custom_settings)
         help = "interval at which to show simulation updates"
         metavar = "<interval>"
         arg_type = String
-        default = "60secs"
+        default = if haskey(defaults, :show_updates)
+            defaults[:show_updates]
+        else
+            get(ENV, "CLIMATEMACHINE_SETTINGS_SHOW_UPDATES", Settings.show_updates)
+        end
```
kpamnany, line 144: Maybe something like the following? A function might be more compact...
```
macro get_default(setting, defaults_dict, defaults_struct)
    quote
        let setting = $(esc(setting)),
            defaults_dict = $(esc(defaults_dict)),
            defaults_struct = $(esc(defaults_struct)),
            setting_sym = Symbol(setting),
            setting_upper = uppercase(setting),
            env_var = string("CLIMATEMACHINE_SETTINGS_", setting_upper)
        if haskey(defaults_dict, setting_sym)
            defaults_dict[setting_sym]
        else
            get(ENV, env_var, Base.getproperty(defaults_struct, setting_sym))
        end
        end
    end
end
```

src/Driver/Driver.jl
```diff
@@ -143,53 +205,108 @@ function parse_commandline(custom_settings)
...
-    ClimateMachine.array_type()
+    ClimateMachine.cli(;arg_settings=nothing, init_driver=true, kwargs...)
 
-Return the array type used by ClimateMachine. This defaults to (CPU-based) `Array`
-and is only correctly set (based on choice from the command line, from
-an environment variable, or from experiment code) after `ClimateMachine.init()`
-is called.
+Initialize the ClimateMachine runtime with cli argument parsing.
+
+- Additional `ArgParseSetting` behavior can be injected into the default
```
kpamnany, line 274:
```suggestion
- Additional `ArgParseSettings` behavior can be injected into the default
```

src/Driver/Driver.jl
```diff
@@ -143,53 +205,108 @@ function parse_commandline(custom_settings)
...
+ClimateMachine `ArgParseSetting` cofiguration by setting the `arg_settings`
```
kpamnany, line 275:
```suggestion
ClimateMachine `ArgParseSettings` cofiguration by setting the `custom_settings`
```

src/Driver/Driver.jl
```diff
@@ -143,53 +205,108 @@ function parse_commandline(custom_settings)
...
+keyword value.
+
+- Setting `init_driver=false` will set the ClimateMachine.Setting singleton
```
kpamnany, line 278:
```suggestion
- Setting `init_driver = false` will set the ClimateMachine.Settings singleton
```

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
-function init(; disable_gpu = false, arg_settings = nothing)
-    # set up timing mechanism
-    tictoc()
+function init(; init_driver::Bool = true, kwargs...)
+    # init global setting values
+    # TODO: add validation for initialization values
+    if haskey(kwargs, :disable_gpu)
+        Settings.disable_gpu = kwargs[:disable_gpu]
+    elseif haskey(ENV, "CLIMATEMACHINE_GPU")
+        #TODO: included for backwards compat, move to shared prefix scheme
```
kpamnany, line 366:
```suggestion
        @warn("CLIMATEMACHINE_GPU will be deprecated; use CLIMATEMACHINE_SETTINGS_DISABLE_GPU")
```

src/Driver/Driver.jl
kpamnany: Also some similar below.

src/Driver/Driver.jl
```diff
@@ -143,53 +205,108 @@ function parse_commandline(custom_settings)
...
+function cli(;
+    custom_settings::Union{Nothing, ArgParseSettings} = nothing,
+    init_driver::Bool = true,
+    kwargs...,
+)
+    kw_defaults = Dict{Symbol, Any}(kwargs)
+    parsed_args = parse_commandline(kw_defaults, custom_settings)
+    # we need to munge the parsed arg dict a bit as parsed arg keys
+    # and climatemachine initialization keywords are not 1:1
+    parsed_args["checkpoint_keep_one"] = !parsed_args["checkpoint_keep_all"]
```
kpamnany, line 298: No idea why I did this. Let's change it so there isn't such an exception? In this file and in `Utilities/Callbacks/Callbacks.jl`.

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
-        Settings.vtk = parsed_args["vtk"]
-        delete!(parsed_args, "vtk")
-        Settings.monitor_timestep_duration =
-            parsed_args["monitor-timestep-duration"]
-        delete!(parsed_args, "monitor-timestep-duration")
+    if haskey(kwargs, :vtk)
+        Settings.vtk = kwargs[:vtk]
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_VTK")
+        Settings.vtk = get(ENV, "CLIMATEMACHINE_SETTINGS_VTK")
+    end
```
kpamnany, line 389: Another macro, as above, or a function?

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
+    if haskey(kwargs, :restart_from_num)
+        Settings.restart_from_num = kwargs[:restart_from_num]
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_RESTART_FROM_NUM")
+        Settings.restart_from_num =
+            parse(Int, get(ENV, "CLIMATEMACHINE_SETTINGS_RESTART_FROM_NUM"))
     end
 
+    if haskey(kwargs, :fix_rng_seed)
+        Settings.fix_rng_seed = kwargs[:fix_rng_seed]
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_FIX_RNG")
```
kpamnany, line 442:
```suggestion
    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED")
```

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
+        Settings.fix_rng_seed =
+            parse(Bool, get(ENV, "CLIMATEMACHINE_SETTINGS_FIX_RNG"))
```
kpamnany, line 444:
```suggestion
            parse(Bool, get(ENV, "CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED"))
```

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
+        Settings.log_level = uppercase(kwargs[:log_level])
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_LOG_LEVEL")
+        Settings.log_level =
+            uppercase(get(ENV, "CLIMATEMACHINE_SETTINGS_LOG_LEVEL"))
+    end
+
+    if haskey(kwargs, :output_dir)
+        Settings.output_dir = kwargs[:output_dir]
+    elseif haskey(ENV, "CLIMATEMACHINE_OUTPUT_DIR")
+        Settins.output_dir = get(ENV, "CLIMATEMACHINE_OUTPUT_DIR")
```
kpamnany, line 457:
```suggestion
    elseif haskey(ENV, "CLIMATEMACHINE_OUTPUT_DIR")
        @warn("CLIMATEMACHINE_OUTPUT_DIR will be deprecated; use CLIMATEMACHINE_SETTINGS_OUTPUT_DIR")
        Settings.output_dir = get(ENV, "CLIMATEMACHINE_OUTPUT_DIR")
```

src/Driver/Driver.jl
```diff
@@ -200,68 +317,181 @@ it will be imported into ClimateMachine's settings.
...
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_OUTPUT_DIR")
+        Settings.output_dir = get(ENV, "CLIMATEMACHINE_SETTINGS_OUTPUT_DIR")
+    end
+
+    if haskey(kwargs, :integration_testing)
+        Settings.integration_testing = kwargs[:integration_testing]
+    elseif haskey(ENV, "CLIMATEMACHINE_SETTINGS_INTEGRATION_TESTING")
+        Settings.integration_testing =
+            get(ENV, "CLIMATEMACHINE_INTEGRATION_TESTING")
```
kpamnany, line 466:
```suggestion
            get(ENV, "CLIMATEMACHINE_SETTINGS_INTEGRATION_TESTING")
```

kpamnany:
Worth checking SlurmCI results, since a couple of different command line arguments are used.

bors try

src/Driver/Driver.jl
```diff
@@ -67,20 +68,48 @@ Base.@kwdef mutable struct ClimateMachine_Settings
...
+If `defaults` is not nothing, override default values for cli argument defaults.
+If `custom_settings` arg is not nothing, add ArgParseSettings to the parsing step.
+
+Returns a `Dict` containing parsed process ARGS values.
+"""
+function parse_commandline(
+    defaults::Union{Nothing, Dict{Symbol, Any}} = nothing,
+    custom_settings::Union{Nothing, ArgParseSettings} = nothing,
+)
+    if defaults == nothing
```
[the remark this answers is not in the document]
jakebolewski: good point, `val == nothing` is just less efficient

src/Driver/Driver.jl
jakebolewski: In general I loathe macros, I feel they are overused :) A function would be better. There is some custom processing / validation that need to happen for each setting so I don't know if it's worth it to consolidate, so the loc gain would be minimal at the moment.  maybe something to do later when there are more settings depending on how this evolves?

jakebolewski:
I went with `CLIMATEMACHINE_SETTINGS_` to make it easier to grep / aggregate the logs.  Invariably there will probably be more ENV variables that change different behaviours so probably there should be an easy way to group them.

jakebolewski:
bors try

blallen:
Will this let us use command-line arguments when running CI tests on azure and/or slurm? If so, that will be a great boon to simplifying the number of different test files we have.

src/Driver/Driver.jl
kpamnany: Ah, I see the `parse(Int, ...)` and the `uppercase()...`, yeah. Messy, but no way out. :-/

blallen:
> @blallen do you have some examples? I think maybe (the need for this was born out of decoupling some of the test runners from overloading argument parsing).

For the ocean model, `experiments/OceanBoxGCM/ocean_gyre.jl` and `test/Ocean/HydrostaticBoussinesq/test_ocean_gyre.jl` are basically the same file, but with different numbers of elements, integration times, and sets of boundary conditions. If this works on both slurm and azure, we can only keep the `experiments` version and change the differences between the two using command-line arguments. It will also make the experiment file more user friendly, as currently, you have to modify the file to actually run just a single experiment offline instead of running 8 variations as we do for the slurm CI test.

kpamnany:
I think this is what the `--integration-testing` argument is for Brandon...?

jakebolewski:
bors try

src/Driver/Driver.jl
charleskawczynski: What about

```julia
function get_default(Settings, defaults, sym)
    default = if haskey(defaults, sym)
        defaults[sym]
    else
        get(
            ENV,
            "CLIMATEMACHINE_SETTINGS_"*upper(String(sym)),
            getproperty(Settings, sym),
        )
    end
end
```

jakebolewski: I think we will need to move to a parameter framework that just generalizes everything so you can support things like running with a config / env overloads/ cli generation / and run time configuration that would encapsulate a lot of these details.  Maybe something like this already exists in the Julia world that we can re-use.  I don't want to do that right now though, or if it's worth it until other details get sorted out.

charleskawczynski: I guess the question is, how many custom cases do we need to account for? If it's fewer than 5, I think it's worth writing some helper scripts / using existing Julia tools to simplify things. But I'm fine tackling this in a later PR.

kpamnany: > I think we will need to move to a parameter framework that just generalizes everything so you can support things like running with a config / env overloads/ cli generation / and run time configuration that would encapsulate a lot of these details. Maybe something like this already exists in the Julia world that we can re-use. I don't want to do that right now though, or if it's worth it until other details get sorted out.

Absolutely. 💯  See #947 for where we need to take this entire layer at some point.

jakebolewski:
@charleskawczynski I consolidated a bit with about a 50 loc reduction so not nothing, if we get rid of the special cases then we probably can just use reflection while things are simple / easy.  

I think eventually the special cases and need for better errors will probably dictate something more sophisticated.

blallen:
> I think this is what the `--integration-testing` argument is for Brandon...?

Ah, didn't know this existed. This will cover making the default behavior for the users nice, but not running multiple tests out of one script. But I should probably make a PR using the `--integration-testing`

kpamnany:
> Ah, didn't know this existed. This will cover making the default behavior for the users nice, but not running multiple tests out of one script. But I should probably make a PR using the `--integration-testing`

A number of the `Numerics/` tests are using it, if you'd like to see examples.

As for the multiple tests from one script, you can use the Driver code as a template for making an `ArgParse.ArgParseSettings`, pass it into `ClimateMachine.cli()` and get your CLI arguments back.

Or you can parse `ARGS` in whatever way you choose in your experiment file and use `ClimateMachine.init() to send the general `ClimateMachine` arguments into the Driver, but this will break what Jake's trying to do here maybe?

src/Driver/Driver.jl
```diff
@@ -143,53 +210,84 @@ function parse_commandline(custom_settings)
...
+- ClimateMachine.init key value pairs can be supplied to overload
+default system defaults at runtime, default values will be merged with the
+parsed argument settings, with parsed cli argument values taking precedent
+over runtime defined default values.
+
+
+Returns a `Dict` containing parsed process ARGS values.
 """
-array_type() = Settings.array_type
+function cli(;
```
simonbyrne, line 270: should we call this `init_cli`? or `init_args`?

jakebolewski:
bors try

src/Driver/Driver.jl
```diff
@@ -143,53 +210,84 @@ function parse_commandline(custom_settings)
...
         metavar = "<path>"
         arg_type = String
-        default = get(ENV, "CLIMATEMACHINE_OUTPUT_DIR", "output")
+        default = if haskey(defaults, :output_dir)
+            defaults[:output_dir]
+        elseif haskey(ENV, "CLIMATEMACHINE_OUTPUT_DIR")
+            ENV["CLIMATEACHINE_OUTPUT_DIR"]
+        else
+            get(ENV, "CLIMATEMACHINE_SETTINGS_OUTPUT_DIR", Settings.output_dir)
+        end
```
simonbyrne, line 237:
```suggestion
        default = get(defaults, :output_dir) do
            get(ENV, "CLIMATEMACHINE_OUTPUT_DIR") do
                get(ENV, "CLIMATEMACHINE_SETTINGS_OUTPUT_DIR", Settings.output_dir)
            end
        end
```

jakebolewski:
bors try

jakebolewski:
bors try

src/Driver/Driver.jl
```diff
@@ -67,20 +68,79 @@ Base.@kwdef mutable struct ClimateMachine_Settings
...
+    if haskey(settings, setting_name)
+        return convert(setting_type, settings[setting_name])
+    elseif haskey(ENV, setting_env)
+        env_val = ENV[setting_env]
+        v = tryparse(setting_type, env_val)
+        if isnothing(v)
+            error("Cannot parse ENV $setting_env value $env_val, to setting type $setting_type")
+        end
+        return v
+    elseif haskey(defaults, setting_name)
```
kpamnany, line 116:
```suggestion
    else
```

src/Driver/Driver.jl
```diff
@@ -67,20 +68,79 @@ Base.@kwdef mutable struct ClimateMachine_Settings
...
+        return defaults[setting_name]
+    else
+        error("setting $setting_name is not contained in either settings or defaults")
+    end
```
kpamnany, line 120:
```suggestion
    end
```

PR #1177, opened 2020-05-29
Move remaining Mesh tests to runmpi structure

### LINKED ISSUE #1168: Use runmpi abstraction for all tests that depend on MPI
# Description

Move all the tests that use MPI (including those that call `ClimateMachine,init()`) to be run wrapped by `runmpi`

Ref: https://github.com/CliMA/ClimateMachine.jl/issues/1167#issuecomment-634987358

# For CLIMA Developers

jakebolewski:
# Description

Move remaining Numerics / Mesh tests to use `runmpi` test harness.  Should be sped up by #1123.

closes #1168 

I have


test/testhelpers.jl
```diff
@@ -1,19 +1,20 @@
...
 
 function runmpi(file; ntasks = 1)
-    MPI.Initialized() &&
-    !MPI.Finalized() &&
-    error(
-        "runmpi does not work if MPI has been " *
-        "Initialized but not Finalized",
-    )
-
+    if MPI.Initialized() && !MPI.Finalized()
```
simonbyrne, line 4: should we just change this to error if `MPI.Initialized()`?

jakebolewski: that makes sense, I think you need both in the case of callbacks registered during finalization IIRC which we don't use.

jakebolewski: I don't really understand why this check is needed at all given that run test harness is run in a separate process, is there a reason this check is included?

simonbyrne: Not sure either: perhaps there were problems if you tried to launch an MPI process from within another one? 

We can probably get rid of it for now.

PR #1182, opened 2020-06-01
Add filter targets and filter perturbations in Held-Suarez

### LINKED ISSUE #1161: Filters need to be applied correctly to state variables
# Description

When running a **solid-body** rotation test case, the flow does not remain at rest and a **hydrostatic state** shows spurious warming and cooling at the top of the domain and at the element edges. The problem is more severe when diffusion or exponential filters are applied and the problem gets more severe as the filter exponent is increased. The problem occurs for explicit and IMEX-type solvers alike, but the 1DIMEX is what we need to use for the GCM for now. In addition, right now, on the sphere, the simulations will no stably integrate without exponential filters. The problem occurs in `Float32` and `Float64`.

This most likely causes significant issues in the **Held-Suarez** experiment case.

## Steps to reproduce

1. Run the solid-body test from [here](https://github.com/CliMA/ClimateMachine.jl/blob/tb/viscous-sponge/experiments/AtmosGCM/solidbody.jl).
2. Use VizCLIMA to genereate plots using this script [here](https://github.com/CliMA/VizCLIMA.jl/blob/tb/hs-testing/src/scripts/default_dry_gcm.jl).

## Expected behavior

In a solid-body test case we should expect the state variables to remain unchanged to high precision. This does not seem to be the case.

## Screenshots
**Example 1**: No sponges, No forcing, No diffusion, No hyperdiffusion, Isothermal reference and initial temperature profile, exponential filter exponent: 64

The figure shows vertical slices of the difference between the final state at 2hrs and the initial state. **Top-left:** zonal-mean zonal velocity difference, **top-right:** zonal-mean meridional velocity difference, **bottom-left:** zonal-mean vertical velocity difference, **bottom-right:** zonal-mean temperature difference. All fields are evaluated at the equator of the planet.

zonal_mean_change_vertical_slice

**Example 2**: No sponges, No forcing, No diffusion, No hyperdiffusion, Isothermal reference and initial temperature profile, exponential filter exponent: 16

The figure shows vertical slices of the difference between the final state at 2hrs and the initial state. **Top-left:** zonal-mean zonal velocity difference, **top-right:** zonal-mean meridional velocity difference, **bottom-left:** zonal-mean vertical velocity difference, **bottom-right:** zonal-mean temperature difference. All fields are evaluated at the equator of the planet.

zonal_mean_change_vertical_slice

**Example 3**: No sponges, No forcing, No diffusion, No hyperdiffusion, Stably stratified reference and initial temperature profile, exponential filter exponent: 64

The figure shows vertical slices of the final state at 1hr at various locations on the sphere. **Top-left:** zonal-mean zonal velocity, **top-right:** zonal-mean meridional velocity, **bottom-left:** zonal-mean vertical velocity, **bottom-right:** zonal-mean temperature. Note the impact on the temperature field at the top of the domain (the small kink).

time_zonal_mean_vertical_slice

## Environment and error messages

No error messages, just physical inconsistencies.

## Additional context

# For CLIMA developers

- [x] The bug can be reproduced
- [x] Labels have been added and developers assigned

mwarusz:
# Description

This adds hooks to modify filtered variables before application of the filter, which is
needed for filtering only the perturbations from the reference state.

Closes #1161

I have

### DIFF
--- a/src/Numerics/Mesh/Filters.jl
+++ b/src/Numerics/Mesh/Filters.jl
@@ -1,14 +1,101 @@
 module Filters
 
 using LinearAlgebra, GaussQuadrature, KernelAbstractions
+using KernelAbstractions.Extras: @unroll
+using StaticArrays
 using ..Grids
 using ..Grids: Direction, EveryDirection, HorizontalDirection, VerticalDirection
+
+using ...MPIStateArrays
+using ...VariableTemplates: @vars, varsize, Vars, varsindices
 
 export AbstractSpectralFilter, AbstractFilter
 export ExponentialFilter, CutoffFilter, TMARFilter
 
 abstract type AbstractFilter end
 abstract type AbstractSpectralFilter <: AbstractFilter end
+
+"""
+    AbstractFilterTarget
+
+An abstract type representing variables that the filter
+will act on
+"""
+abstract type AbstractFilterTarget end
+
+"""
+    vars_state_filtered(::AbstractFilterTarget, FT)
+
+A tuple of symbols containing variables that the filter
+will act on given a float type `FT`
+"""
+function vars_state_filtered end
+
+"""
+    compute_filter_argument!(::AbstractFilterTarget,
+                             state_filter::Vars,
+                             state::Vars,
+                             state_auxiliary::Vars)
+
+Compute filter argument `state_filter` based on `state`
+and `state_auxiliary`
+"""
+function compute_filter_argument! end
+"""
+    compute_filter_result!(::AbstractFilterTarget,
+                           state::Vars,
+                           state_filter::Vars,
+                           state_auxiliary::Vars)
+
+Compute filter result `state` based on the filtered state
+`state_filter` and `state_auxiliary`
+"""
+function compute_filter_result! end
+
+number_state_filtered(t::AbstractFilterTarget, FT) =
+    varsize(vars_state_filtered(t, FT))
+
+"""
+    FilterIndices(I)
+
+Filter variables based on their indices `I` where `I` can
+be a range or a list of indices
+
+## Examples
+```julia
+FiltersIndices(1:3)
+FiltersIndices(1, 3, 5)
+```
+"""
+struct FilterIndices{I} <: AbstractFilterTarget
+    FilterIndices(I::Integer...) = new{I}()
+    FilterIndices(I::AbstractRange) = new{I}()
+end
+vars_state_filtered(::FilterIndices{I}, FT) where {I} =
+    @vars(_::SVector{length(I), FT})
+
+function compute_filter_argument!(
+    ::FilterIndices{I},
+    filter_state::Vars,
+    state::Vars,
+    aux::Vars,
+) where {I}
+    @unroll for s in 1:length(I)
+        @inbounds parent(filter_state)[s] = parent(state)[I[s]]
+    end
+end
+
+function compute_filter_result!(
+    ::FilterIndices{I},
+    state::Vars,
+    filter_state::Vars,
+    aux::Vars,
+) where {I}
+    @unroll for s in 1:length(I)
+        @inbounds parent(state)[I[s]] = parent(filter_state)[s]
+    end
+end
+
 
 """
     spectral_filter_matrix(r, Nc, σ)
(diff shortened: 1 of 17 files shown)


mwarusz:
bors try

mwarusz:
bors try-

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
```diff
@@ -585,14 +586,22 @@ function update_auxiliary_state!(
...
-        index_u =
-            tuple(collect(varsindex(vars_state_conservative(m, FT), :u))...)
-        apply!(Q, index_u, dg.grid, vert_filter, VerticalDirection())
+        apply!(
+            Q,
+            FilterStateConservative(dg.balance_law, :u),
+            dg.grid,
+            vert_filter,
+            direction = VerticalDirection(),
+        )
```
blallen, line 595: This is much nicer! Thanks!

src/Numerics/DGMethods/balance_law.jl
```diff
@@ -13,6 +13,30 @@ function calculate_dt(dg, model, Q, Courant_number, t, direction)
     return Courant_number / CFL
 end
 
+using ..Mesh.Filters: FilterIndices
+function FilterStateConservative(
```
simonbyrne, line 17: Add a docstring.

src/Numerics/Mesh/Filters.jl
```diff
@@ -1,15 +1,66 @@
...
+using ...VariableTemplates: @vars, varsize, Vars
+
 export AbstractSpectralFilter, AbstractFilter
 export ExponentialFilter, CutoffFilter, TMARFilter
+export FilterIndices
 
 abstract type AbstractFilter end
 abstract type AbstractSpectralFilter <: AbstractFilter end
 
+abstract type AbstractFilterTarget end
```
simonbyrne, line 18: Document the required interfaces.

mwarusz:
bors try

mwarusz:
bors try

src/Numerics/DGMethods/balance_law.jl
```diff
@@ -13,6 +13,45 @@ function calculate_dt(dg, model, Q, Courant_number, t, direction)
...
+                    getproperty(model, Symbol(submodel)),
+                    Int,
+                ),
+                Symbol(var),
+            )
+        else
+            vi = varsindex(vars_state_conservative(model, Int), Symbol(var))
+        end
+        append!(I, collect(vi))
+    end
```
jkozdon, line 52: Would this be worth moving into `VariableTemplates`? Something like `varsindices` to compliment `varsindex`?

Also, maybe use `Nothing` instead of `Int`?

src/Numerics/Mesh/Filters.jl
```diff
@@ -257,27 +367,52 @@ horizontal and/or vertical reference directions.
...
     i, j, k = @index(Local, NTuple)
 
     @inbounds begin
+        ijk = i + Nq * ((j - 1) + Nq * (k - 1))
+
         s_filter[i, j] = filtermatrix[i, j]
 
+        @unroll for s in 1:nstates
+            l_Q[s] = Q[ijk, s, e]
+        end
```
jkozdon, line 396: Is there a reason we need to load the whole state and not just the elements we need?
```julia
        @unroll for (s,is) in enumerate(I)
            l_Q[s] = Q[ijk, is, e]
        end
```

Same for auxiliary below.

Should we warn somewhere the assumption is the `state` and `filtered_state` match in order? (Same with auxiliary).

That is we cannot do something like:
```julia
function vars_state_consvative(::MyModel, FT)
  @vars begin
    a::FT
    b::FT
    c::FT
  end
end
function vars_filtered_state(::MyModel, FT)
  @vars begin
    c::FT
  end
end
FilterIndices(3)
```

Since if I am not mistake, the numbers of filtered states would be 3 in this case.

lcw: I am not sure if the compiler is optimizing out these loads or not.  If not and it is straight forward to do so, it would be nice to minimize the loads.

bischtob:
Any chance that we could add just a small section about filtering at the end of the numerics design docs that matches what we have done here?

mwarusz:
bors try

mwarusz:
bors try

PR #1188, opened 2020-06-03
Add disable_custom_logger option to Driver

### LINKED ISSUE #1190: Doc build filled with `Base.IOError("stream is closed or unusable", 0)` exceptions
# Description

There are a lot of `Base.IOError("stream is closed or unusable", 0)` exceptions in the doc build, and this should be fixed.

## Additional context

# For CLIMA Developers

charleskawczynski:
# Description

The logger is not playing nice with Documenter.jl and Literate.jl (we're seeing `Base.IOError("stream is closed or unusable", 0)` exceptions). This PR fixes this issue by disabling the logger for tutorials.

 - Adds a check for `ENV["DISABLE_CUSTOM_LOGGER"]` in `init` in `Driver.jl`
 - Sets `ENV["DISABLE_CUSTOM_LOGGER"] = "OFF"` in `docs/make.jl`

Closes #1190

I have

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,7 +2,10 @@ Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
 
 using ClimateMachine, Documenter, Literate
 
-ENV["GKSwstype"] = "100" # https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
+# https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
+ENV["GKSwstype"] = "100"
+# avoid problems with Documenter/Literate when using `global_logger()`
+ENV["CLIMATEMACHINE_SETTINGS_DISABLE_CUSTOM_LOGGER"] = true
 
 generated_dir = joinpath(@__DIR__, "src", "generated") # generated files directory
 rm(generated_dir, force = true, recursive = true)
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -70,6 +70,7 @@ Base.@kwdef mutable struct ClimateMachine_Settings
     restart_from_num::Int = -1
     fix_rng_seed::Bool = false
     log_level::String = "INFO"
+    disable_custom_logger::Bool = false
     output_dir::String = "output"
     integration_testing::Bool = false
     array_type::Type = Array
@@ -219,6 +220,9 @@ function parse_commandline(
         "--fix-rng-seed"
         help = "set RNG seed to a fixed value for reproducibility"
         action = :store_true
+        "--disable-custom-logger"
+        help = "do not use a custom logger"
+        action = :store_true
         "--log-level"
         help = "set the log level to one of debug/info/warn/error"
         metavar = "<level>"
@@ -331,6 +335,8 @@ the process, otherwise the defaulting to `ClimateMachine.Settings`.
         set RNG seed to a fixed value for reproducibility
 - `log_level::String = "INFO"`:
         log level for ClimateMachine global default runtime logger
+- `disable_custom_logger::String = false`:
+        disable using a global custom logger for ClimateMachine
 - `output_dir::String = "output"`: (path)
         absolute or relative path to output data directory
 - `integration_testing::Bool = false`:
@@ -426,9 +432,17 @@ function _init_driver(settings::ClimateMachine_Settings)
     loglevel = log_level_str == "DEBUG" ? Logging.Debug :
         log_level_str == "WARN" ? Logging.Warn :
         log_level_str == "ERROR" ? Logging.Error : Logging.Info
-    # TODO: write a better MPI logging back-end and also integrate Dlog for large scale
-    logger_stream = MPI.Comm_rank(MPI.COMM_WORLD) == 0 ? stderr : devnull
-    global_logger(ConsoleLogger(logger_stream, loglevel))
+    # TODO: write a better MPI logging back-end and also integrate Dlog
+    # for large scale
+    if !settings.disable_custom_logger
+        # cannot use `NullLogger` here because MPI collectives may be
+        # used in logging calls!
+        logger_stream = MPI.Comm_rank(MPI.COMM_WORLD) == 0 ? stderr : devnull
+        prev_logger = global_logger(ConsoleLogger(logger_stream, loglevel))
+        atexit() do
+            global_logger(prev_logger)
+        end
+    end
     return
 end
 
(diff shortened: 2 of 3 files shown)


kpamnany:
I don't think this is the right solution. I think this:
```julia
    logger_stream = MPI.Comm_rank(MPI.COMM_WORLD) == 0 ? stderr : devnull
    global_logger(ConsoleLogger(logger_stream, loglevel))
```
Should become something like:
```julia
    last_logger = global_logger(
        MPI.Comm_rank(MPI.COMM_WORLD) == 0 ?
        ConsoleLogger(stderr, loglevel) :
        NullLogger()
    )
    atexit() do
        global_logger(last_logger)
    end
```
The `atexit()` might not be sufficient, resetting the global logger may need to happen before process exit? Maybe the reset can happen at the end of `ClimateMachine.invoke!()` or we need a `ClimateMachine.shutdown()`.

Because the real problem is that we set the global logger but don't reset it. When run under `Literate.notebook()` (which eventually uses `Documenter.withoutput()`), `stderr` is actually a pipe. We set a global logger using that redirected `stderr`. When `Documenter.withoutput()` returns, that pipe is closed, so all subsequent calls to logging within Literate are trying to use the global logger we set, which is using a closed pipe, and thus throwing those errors.

jakebolewski:
Yeah the semantics here might be a bit confusing, this doesn't actually disable logging (which is an option provided by the Logging module) it just doesn't replace the global logger with a custom defined one.

kpamnany:
If you do want to use this solution, you could name the keyword `custom_logger`. Worth noting that if you run with `custom_logger = false` then not only can you not set the log level, but if you run with multiple ranks then every MPI rank will output messages.

simonbyrne:
Rather than a keyword option to `init`, it might be better as a command line argument to the driver, as that is how we control other options.

charleskawczynski:
> Maybe the reset can happen at the end of ClimateMachine.invoke!() or we need a ClimateMachine.shutdown().

This seems like a reasonable solution

> so custom_logger=true or a better name might be more revealing

I'm fine with this

> Rather than a keyword option to init, it might be better as a command line argument to the driver, as that is how we control other options.

I'm not sure how we could pass that information to Documenter. Would we just define environment variables in `docs/make.jl`?

BTW, this PR does remove all of the errors! 🎉

simonbyrne:
> I'm not sure how we could pass that information to Documenter. Would we just define environment variables in `docs/make.jl`?

Yes, @jakebolewski added functionality to set all the default of any keyword args via environment variables.

simonbyrne:
It might be better to use the functionality added in #1172

jakebolewski:
I would follow the conventions in #1172 just so things are consistent.

`CLIMATEMACHINE_SETTINGS_CUSTOM_LOGGER=false`

`--disable-custom-logger` for the cli

`ClimateMachine.init(;custom_logger=false)`

internally to `init` you could set the sentinel value loglevel to disable setting the custom logger

charleskawczynski:
> I would follow the conventions in #1172 just so things are consistent. ... CLIMATEMACHINE_SETTINGS_CUSTOM_LOGGER=false

Fixed the `ENV` grab in `docs/make.jl`.

> --disable-custom-logger for the cli ... ClimateMachine.init(;custom_logger=false)

The double negative here is a bit confusing: `--disable-custom-logger = "false"`.

> internally to init you could set the sentinel value loglevel to disable setting the custom logger

I just added a separate flag for custom logger, since it's an on/off and not really a level.

jakebolewski:
we already have a `--disable-gpu` cli flag, this would use the same mechanism

docs/make.jl
```diff
@@ -3,6 +3,7 @@ Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
 using ClimateMachine, Documenter, Literate
 
 ENV["GKSwstype"] = "100" # https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
+ENV["CLIMATEMACHINE_SETTINGS_CUSTOM_LOGGER"] = "OFF" # to avoid clashing logger with Documenter/Literate
```
jakebolewski, line 6: to be consistent we should probably stick with `"true"` / `"false"` (seems like this is widely used in the Julia world as it's easy to parse).

charleskawczynski: Why not just use Boolean, then? @simonbyrne should we switch to Bool?

@jakebolewski any idea why this is failing?
```
ERROR: LoadError: LoadError: LoadError: MethodError: no method matching tryparse(::Type{String}, ::String)
Closest candidates are:
  tryparse(!Matched::Type{Float64}, ::String) at parse.jl:245
  tryparse(!Matched::Type{Float32}, ::String) at parse.jl:265
  tryparse(!Matched::Type{Float16}, ::AbstractString) at parse.jl:285
  ...
in expression starting at string:1
when executing the following code block

```

jakebolewski: there doesn't seem to be a pass-through method defined for strings :(

jakebolewski: I would just switch to bool, I don't think we use ON / OFF anywhere else

src/Driver/Driver.jl
```diff
@@ -224,6 +225,12 @@ function parse_commandline(
         metavar = "<level>"
         arg_type = String
         default = uppercase(get_setting(:log_level, defaults, global_defaults))
+        "--custom-logger"
```
jakebolewski, line 228:
```  
 "--disable-custom-logger"
        help = "Disable using ClimateMachine's custom MPI aware logger"
        action = :store_true
```

jakebolewski: ?

charleskawczynski: Why `--disable-custom-logger`? And what does `action` do?

jakebolewski: it's consistent with the other binary flag options for the cli.  

action :store_true just sets the arg variable to true if the flag is present, see `disable-gpu`

charleskawczynski: Shouldn't the help match `custom_logger = true` indicating that we're using CM.jl's custom logger?

charleskawczynski: @simonbyrne preferred this way (I do too, to avoid the double negative `--disable-custom-logger=false`). Let's work out the details in another PR (I'd like to continue with the uphill docs battle!).

jakebolewski: sure, no problem.  just pointing out that it's not a double negative when it's a binarly cli flag.  for flags there is not "=true" or "=false", if it exists it is marked true.

charleskawczynski:
bors r +

PR #1195, opened 2020-06-04
Diagnostics groups append to NetCDF

LenkaNovak:
# Description

All diagnostics groups updated to append to their NetCDF files along the `time` dimension. No longer is there one NetCDF file for each diagnostics interval.

Order of dimensions for the NetCDF variables adjusted to allow the `time` dimension to be recognized properly by other tools -- diagnostics files can now be read in Paraview (e.g. see [this](https://drive.google.com/file/d/17wkaiViLyw8QUDlAELMtEx7vAH7TiUO5/view?usp=sharing)).

I have

- [X] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`

- [X] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.


experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -224,6 +226,45 @@ function main()
         user_callbacks = (cbfilter,),
         check_euclidean_distance = true,
     )
+    # aggregate files
+
+    fnames = filter(x -> occursin(r"^Held", x), readdir("output/"))
```
charleskawczynski, line 231:
```suggestion
    fnames = filter(x -> occursin(r"^Held", x), readdir("output"))
```

experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -224,6 +226,45 @@ function main()
...
+        vd = dimnames(mfds[v_n])
+        va = mfds[v_n].attrib
+        if v_n in vd
+            defVar(ds, v_n, vv, vd, attrib = va)
+        else
+            defVar(ds, v_n, vv, vd, attrib = va) #on Cluster
+            #defVar(ds, v_n, permutedims(vv, [2, 3, 4, 1]), vd[[2, 3, 4, 1]], attrib = va) # on Mac
+        end
+    end
+    cd("../")
```
charleskawczynski, line 265:
```suggestion
    cd("..")
```

experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -224,6 +226,45 @@ function main()
...
+        # aggregate data in output into a multi-file dataset
+        NCDataset(fnames,"a"; aggdim = "time", deferopen = false) do mfds
+            varnames = keys(mfds) # all var names
+
+            dummy = mfds[varnames[5]] # one of the 4D vars to get dim info
+            dim_n = dimnames(dummy)
+            dim_s = size(dummy)
+
+            # write data to disk as combined file
+            ds = Dataset("gcm_diags_aggregated.nc","c")
```
charleskawczynski, line 242: We need to close this after opening ☝️

src/Diagnostics/Diagnostics.jl
```diff
@@ -38,34 +38,38 @@ using ..Writers
...
+
+TODO: make `param_set` required!
 """
 function init(
     mpicomm::MPI.Comm,
     dg::DGModel,
     Q::MPIStateArray,
     starttime::String,
     output_dir::String,
+    param_set::AbstractEarthParameterSet = EarthParameterSet(),
```
charleskawczynski, line 66: This should not be optional (it should _always_ be defined in the driver).

src/Diagnostics/Diagnostics.jl
```diff
@@ -38,34 +38,38 @@ using ..Writers
 using CLIMAParameters
 using CLIMAParameters.Planet: planet_radius
 struct EarthParameterSet <: AbstractEarthParameterSet end
```
charleskawczynski, line 40: Please remove this (sorry, I missed stopping this from going in)

src/Diagnostics/Diagnostics.jl
kpamnany: Yes, it has to be passed in. But I realized that's not happening and set this up as a first step.

src/Diagnostics/Diagnostics.jl
kpamnany: This was already there, I've just started the process of cleaning up.

smarras79:
Can we dump to a simple text file the mass and energy loss? It should simply append the mass loss at the current time step (or fallback call) to a file that looks like this:
```
time (s)   mass_loss   energy_loss
0          0           0
10      1e-14        1e-14
...
1000      1e-14        1e-14
...
10000     1e-14        1e-14
```

These are quantities that we should add to all of the initial benchmarking papers that are being written now but it will be good to have it regardless of the papers.

The loss is nothing else than:

image
(Disregard the bars which are from a paper with a different meaning for barred variables than we do)

Ideally, for each test we want to have a visual of something like these curves:
image (1)

kpamnany:
@smarras79 : I'll look at adding that in another PR. Can you open an issue for this please?

smarras79:
> @smarras79 : I'll look at adding that in another PR. Can you open an issue for this please?

Sure, thanks @kpamnany

charleskawczynski:
Thanks for cleaning up the param set pieces!

PR #1197, opened 2020-06-04
src/Atmos/Model/moisture.jl: added pressure as default for vtk output

smarras79:
# Description

Modified src/Atmos/Model/moisture.jl to add `pressure` among the variables that are written to VTK by default.

I have


smarras79:
It would be very good to either write the background or the pressure perturbation as well.

smarras79:
> `air_T` and `temperature` : There is a chance here to simply correct those two `names` to be consistent across the different moisture model types as well. We can include the fix here. @smarras79 ?

Sure. Do they only appear in moisture.jl though?

smarras79:
> @kpamnany @jkozdon is there any way to still get these diagnostics without adding more fields to the aux state? I remember folks at NPS were hesitant about providing auxiliary states to avoid memory bloat, and I'd like to not abuse it unless there's no other way.

If that's the case, it would be nice to replace total energy with pressure. We can calculate total energy from all the other quantities that we are already writing to file.

kpamnany:
Diagnostics computes and outputs `pres`. VTK is just dumping state and aux variables, not doing any computation.

I agree that we should be careful about adding fields if they can be readily computed. As for replacing total energy with pressure, I can't comment.

PR #1181, opened 2020-06-01
Reduce doc build warnings, fix tutorials

charleskawczynski:
# Description

UPDATE: This PR is being split off into smaller ones since the doc build was taking too long. See #1219, #1218.

 - Fixes diagnostics API docs
 - Adds `dry_rayleigh_benard.md` to tutorials pages array, and fixes .jl comments

This also does some pretty large sweeping doc fixes.

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1199, opened 2020-06-04
Added single stack horizontal statistics and test

ilopezgp:
Added horizontal statistics diagnostics for single stack driver and a single stack tutorial of the 3D Burgers equation. Modified plothelpers.jl to admit the xlabel as a keyword argument.

# Description

I have

### DIFF
--- a/docs/list_of_tutorials.jl
+++ b/docs/list_of_tutorials.jl
@@ -20,7 +20,10 @@ if generate_tutorials
     tutorials_dir = joinpath(@__DIR__, "..", "tutorials")
 
     tutorials = [
-        "Atmos" => ["Dry Idealized GCM" => "Atmos/heldsuarez.jl"],
+        "Atmos" => [
+            "Dry Idealized GCM" => "Atmos/heldsuarez.jl",
+            "Flow in Single Element Stack" => "Atmos/burgers_single_stack.jl",
+        ],
         "Ocean" => [],
         "Land" => ["Heat" => ["Heat Equation" => "Land/Heat/heat_equation.jl"]],
         "Numerics" => [
--- a/docs/plothelpers.jl
+++ b/docs/plothelpers.jl
@@ -1,5 +1,5 @@
 
-export_plot(z, all_data, ϕ_all, filename, ylabel) = nothing
+export_plot(z, all_data, ϕ_all, filename, ylabel; xlabel) = nothing
 export_plot_snapshot(z, all_data, ϕ_all, filename, ylabel) = nothing
 
 # using Requires
@@ -27,7 +27,7 @@ end
 Export plot of all variables, or all
 available time-steps in `all_data`.
 """
-function export_plot(z, all_data, ϕ_all, filename, ylabel)
+function export_plot(z, all_data, ϕ_all, filename, ylabel; xlabel = nothing)
     ϕ_all isa Tuple || (ϕ_all = (ϕ_all,))
     p = plot()
     for n in 0:(length(keys(all_data)) - 1)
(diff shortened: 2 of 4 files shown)


test/Driver/single_stack_test.jl
```diff
@@ -0,0 +1,615 @@
...
+# The `ClimateMachine`'s time-steppers provide hooks, or callbacks, which
+# allow users to inject code to be executed at specified intervals. In this
+# callback, the state and aux variables are collected, combined into a single
+# `OrderedDict` and written to a NetCDF file (for each output step `step`).
+step = [0];
+callback = GenericCallbacks.EveryXSimulationTime(
+    every_x_simulation_time,
+    solver_config.solver,
+) do (init = false)
+    state_vars_var = get_horizontal_variance(
```
charleskawczynski, line 543: Just to be clear, this is not computing averages at every timestep. For that behavior, you'd want to put `get_horizontal_variance` in `update_auxiliary_state!`. In other words, right now, your solution depends on the frequency of your callbacks.

ilopezgp: I know. I just implemented functions that provide the horizontal statistics diagnostics for outputs. I did not want to compute them for every timestep since I am not using the horizontal statistics for the evolution of the system.

test/Driver/single_stack_test.jl
```diff
@@ -0,0 +1,615 @@
...
+#  the vertical component.
+function source!(
+    m::SingleStack,
+    source::Vars,
+    state::Vars,
+    diffusive::Vars,
+    aux::Vars,
+    args...,
+)
+    ẑ = [0.0, 0.0, 1.0]
```
charleskawczynski, line 278:
```suggestion
    ẑ = @SVector FT[0, 0, 1]
```
If you want this to be able to run on GPU

test/Driver/single_stack_test.jl
```diff
@@ -0,0 +1,615 @@
...
+    ρ̄ū =
+        state.ρ *
+        [1 - 4 * (aux.z - m.zmax / 2)^2, 1 - 4 * (aux.z - m.zmax / 2)^2, 0.0] ./
```
charleskawczynski, line 281: Same, can wrap with SVector, and add type cast for type stability

ilopezgp:
I agree with what you said. I would personally move it to `tutorials/` and keep the plots. What directory do you think it should be in? `tutorials/Atoms/`?

test/Driver/single_stack_test.jl
ilopezgp: As discussed, wrapping with SVector and type casting leads to substantial slow down of the code. Waiting until we can fix this to update lines 278-281.

test/Driver/single_stack_test.jl
```diff
@@ -0,0 +1,613 @@
...
+    state⁺::Vars,
+    aux⁺::Vars,
+    n⁻,
+    state⁻::Vars,
+    aux⁻::Vars,
+    bctype,
+    t,
+    _...,
+)
+    if bctype == 1 # bottom
```
tapios, line 346: Don't we have a more abstract way of specifying boundary conditions here and below?

charleskawczynski: We'll clean this up eventually. This is what we have to do for now.

docs/list_of_tutorials.jl
```diff
@@ -65,6 +67,7 @@ if generate_tutorials
         "Atmos/dry_rayleigh_benard.jl",               # takes too long
         "Atmos/heldsuarez.jl",                        # broken
         "Atmos/risingbubble.jl",                      # broken
+        "Atmos/burgers_single_stack.jl",              # too long
```
charleskawczynski, line 70: @jakebolewski 😢

charleskawczynski: > How long does it take?
🤷‍♂️ @ilopezgp ? But there are several more that will extend the overall doc build (#1219, #1226, and more from #1181).

> can you open an issue for this?

Opened #1233

ilopezgp: > How long does it take?

Now it takes 4 minutes running on CPU on my iMac (4.2 GHz Intel Core i7, 16GB RAM). I was not sure if that was too much time so I added the case to the skip list.

ilopezgp:
> You may want to use `OrderedDict` instead of `Dict`? But anyway, the utilities functions need to be refactored for efficiency -- I'll look at that separately.

I have changed the instances of `Dict` to `OrderedDict` in the utilities in the new commit.

docs/list_of_tutorials.jl
ilopezgp: Got it. I will remove it from the skip list.

PR #1200, opened 2020-06-05
Add varsindices

mwarusz:
# Description

This PR:
- Adds `varsindices`, a refactored version of the indexing functionality needed
  in #1182 moved to `VariableTemplates` for general use
- Adds mulit-argument method of `varsindex` for getting indices of submodel variables

I have

### DIFF
--- a/src/Utilities/VariableTemplates/VariableTemplates.jl
+++ b/src/Utilities/VariableTemplates/VariableTemplates.jl
@@ -1,9 +1,26 @@
 module VariableTemplates
 
-export varsize, Vars, Grad, @vars, varsindex
+export varsize, Vars, Grad, @vars, varsindex, varsindices
 
 using StaticArrays
 
+"""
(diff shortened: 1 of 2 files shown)


mwarusz:
I know, with a different version of this I managed to get something like:
```
A[varsindex(vars, :a, :b)] .= 1
```
to compile on the GPU, but the following still wouldn't work
```
for s in (:a, :b)
  A[varsindex(vars, s)] .= 1
end
```
I think right now it may be better to precompute indices and pass them to kernels unless
somebody needs functionality like the first snippet above. @jkozdon @lcw do you see any use for this ?

lcw:
> @jkozdon @lcw do you see any use for this ?

Not in the kernels right now.  I was thinking about using something like this when modifying the `wavespeed` array but have worked around it by wrapping the array with a `Vars`.

PR #1206, opened 2020-06-05
factor out refstate contributions from AtmosAcousticGravityLinearModel

simonbyrne:
# Description

This applies the same factoring we do for the AtmosModel to the AtmosAcousticGravityLinearModel, which should significant reduce the numerical error in both this and the RemainderModel.

Should we do the same thing for the AtmosAcousticLinearModel? There the benefits are not as clear.

cc: @akshaysridhar @thomasgibson @bischtob 

I have

### DIFF
--- a/src/Atmos/Model/linear.jl
+++ b/src/Atmos/Model/linear.jl
@@ -214,7 +214,11 @@ function flux_first_order!(
         state,
         aux,
     )
-    flux.ρu += pL * I
+    if lm.atmos.ref_state isa HydrostaticState
+        flux.ρu += (pL - aux.ref_state.p) * I
+    else
+        flux.ρu += pL * I
+    end
     flux.ρe = ((ref.ρe + ref.p) / ref.ρ) * state.ρu
     nothing
 end
@@ -229,7 +233,11 @@ function source!(
 )
     if direction isa VerticalDirection || direction isa EveryDirection
         ∇Φ = ∇gravitational_potential(lm.atmos.orientation, aux)
-        source.ρu -= state.ρ * ∇Φ
+        if lm.atmos.ref_state isa HydrostaticState
+            source.ρu -= (state.ρ - aux.ref_state.ρ) * ∇Φ
+        else
+            source.ρu -= state.ρ * ∇Φ
+        end
     end
     nothing
 end


src/Atmos/Model/linear.jl
```diff
@@ -229,7 +233,11 @@ function source!(
 )
     if direction isa VerticalDirection || direction isa EveryDirection
         ∇Φ = ∇gravitational_potential(lm.atmos.orientation, aux)
-        source.ρu -= state.ρ * ∇Φ
+        if lm.atmos.ref_state isa HydrostaticState
+            source.ρu -= (state.ρ - aux.ref_state.ρ) * ∇Φ
```
mwarusz, line 237: Won't this make the operator non-linear ? Our ARK schemes could be changed to handle systems of the form `L * x + b = c` by moving `b` to the right hand side. I have an old branch that has this functionality but was never tested (mw/affine-ark).

simonbyrne: Yes, this would need some checking.

tapios: It'll remain linear (`aux.ref_state.ρ` is constant). But it does add an intercept term.

mwarusz: Right, I meant linearity in the linear algebra sense (which is what we currently assume).

tapios:
If you do this for acoustic+gravity waves, it might be good to do this for the acoustic waves alone too. It avoids the exponential pressure decrease, and roundoff issues coming from it. But it does add an artificial source from the geopotential gradient that otherwise would not occur. Again, an additional intercept term.

src/Atmos/Model/linear.jl
simonbyrne: How is discrete balance defined?

simonbyrne: Would this be easier if we changed our state variable from ρ to δρ = ρ - ρ_ref?

mwarusz: Basically that the two added terms will cancel out to machine precision using DG discretization.

tapios: Discrete balance means that \rho = -1/g dp/dz discretely. This is being implemented from what I know. 

Working with perturbation variables (for pressure and density) would help here. But not for the acoustic waves (the intercept there would arise anyway). And you'd need to go back to native variables for all thermodynamics, microphysics etc. I'd prefer to avoid that back and forth between variables.

jkozdon: We'd also need to account for this when forming the linear system.

tapios:
Do we know whether factoring out the reference state here makes a difference? If not, I suggest setting up a test where you work with perturbation variables and try it out. This would require no changes to linear solvers.

bischtob:
@simonbyrne, is this still current? if not, we should perhaps close this (it's not deleted, I think)

PR #1208, opened 2020-06-06
Remove assumption in ARK schemes about `Q`

jkozdon:
There was an assumption that the ARK solver is always called with the
same `Q` that it was initialized with. We remove that by not storing `Q`
itself in the struct.


jkozdon:
bors try

Issue #1213, opened 2020-06-08
Simplify debugging initial conditions / experiment configurations


charleskawczynski:
# Description

Debugging any sort of PDE solution almost always relies on testing several things in a specific order. For us, this procedure looks as follows:

 1. Verify `init_state_auxiliary!` (should rely on nothing).
 2. Verify `init_state_conservative!` (relies on the correctness of `init_state_auxiliary!`).
 3. Verify that the solution does not have NaNs after 1 timestep (relies on the correctness of `init_state_auxiliary!`, `init_state_auxiliary!`, `update_auxiliary_state!`, and `dostep!`).
 4. Verify that the solution does not have NaNs after multiple timesteps (relies on the correctness of `init_state_auxiliary!`, `init_state_auxiliary!`, `update_auxiliary_state!`, `dostep!`, and potentially callbacks).

After step 4, typically one has to start looking at stability, which requires a whole other set of debugging strategies/tools.

Right now, we cannot easily do several of these steps because of the current solver configuration. Adding a flag, `export_init_aux` to [SolverConfiguration](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/solver_configs.jl#L75), and exporting a VTK/JLD2 file (for visualization) [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/solver_configs.jl#L109) and [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/solver_configs.jl#L133) would significantly help decoupling the first two steps. However, we'd need to incorporate diagnostics somehow. IMO, this is currently a severe debugging bottleneck because both the initial aux state _and_ the initial conservative state must be defined correctly in order to visualize, and they're entangled via Thermodynamic functions.

## Additional context

Add any other reasons why this should be addressed.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

christophernhill:
@charleskawczynski etc.... a couple of vaguely related things, in case they are useful? None of them are as polished as Diagnostics viz, so not sure where they are going to end up.....

1. there is a little gist here 

https://gist.github.com/christophernhill/9d176cb0f2ebc9fe9361b03dc6512012 

[...]

Issue #1174, opened 2020-05-29
Add hook for `init_state_auxiliary!` similar to `update_auxiliary_state!`


charleskawczynski:
# Description

We'll need to offer a way to compute indefinite stack integrals at initialization-- and not at every timestep. This is needed to compute reference states that require numerical integration along the vertical element stack. So, we should add a hook to define `init_state_auxiliary!` which has access to entire MPIStateArray, similar to how `update_auxiliary_state!` calls `nodal_update_auxiliary_state!`.

# Additional context

Add any other reasons why this feature will help with the goals of CLIMA.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

blallen:
So you want to turn `init_state_auxiliary!` into a wrapper like `update_auxiliary_state!` where you can do operations across the full domain and then create a `nodal_init_state_auxiliary!` that acts on the individual nodal points to assign the output of those calculations?

charleskawczynski:
Exactly

charleskawczynski:
I think that this was effectively closed by #1421.

PR #1142, opened 2020-05-21
Remove unnecessary zero initializations in HBModel

blallen:
# Description

Added these at some point when debugging, turns out they aren't necessary. Code is cleaner without doing this.

I have

- [x] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [x] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [x] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [x] CLIMA developers with relevant expertise have been assigned to review this submission
- [x] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [x] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1214, opened 2020-06-08
Remove call to `update_auxiliary_state!` in driver

charleskawczynski:
# Description

At the moment, users cannot visualize the results of what is computed in `init_state_auxiliary!`, this PR fixes this by removing the call to `update_auxiliary_state!` in the solver configuration.

Step towards fixing #1213

I have

### DIFF
--- a/src/Driver/solver_configs.jl
+++ b/src/Driver/solver_configs.jl
@@ -134,7 +134,6 @@ function SolverConfiguration(
         @info @sprintf("Initializing %s", driver_config.name,)
         Q = init_ode_state(dg, FT(0), init_args...; init_on_cpu = init_on_cpu)
     end
-    update_auxiliary_state!(dg, bl, Q, FT(0), dg.grid.topology.realelems)
 
     # default Courant number
     # TODO: Think about revising this or drop it entirely.


charleskawczynski:
@blallen any idea what's wrong here?

blallen:
> @blallen any idea what's wrong here?

The ocean tests passed perfectly, but `test/Driver/driver_test.jl` failed due to a `NaN` during initialization. My best guess is that one of the auxiliary variables for the AtmosModel isn't being initialized properly but it receives a reasonable value during `update_auxiliary_state!`. The integral variables as the most likely suspects, but you have four `nodal_update_auxiliary_state!` functions to check in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/AtmosModel.jl#L517-L527.

jkozdon:
My guess is that @blallen is right. The aux initialization is pointwise, so any non-pointwise initialization is likely being done by the update function. If these values are needed for the state initialization, then this has to be done before the ode state is created.

A case could be made for a more general aux initialization function, since it is a bit weird to have to call update aux separately here so you could add that work their.

blallen:
Hey @charleskawczynski it looks like we found even more motivation for https://github.com/CliMA/ClimateMachine.jl/issues/1174

charleskawczynski:
> The ocean tests passed perfectly, but test/Driver/driver_test.jl failed due to a NaN during initialization. My best guess is that one of the auxiliary variables for the AtmosModel isn't being initialized properly but it receives a reasonable value during update_auxiliary_state!. The integral variables as the most likely suspects, but you have four nodal_update_auxiliary_state! functions to check in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/AtmosModel.jl#L517-L527.

Actually, these methods are already point-wise. The problem is that the Courant calculation depends on the [auxiliary state](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel_kernels.jl#L2623). @kpamnany any ideas what to do here?

blallen:
> > The ocean tests passed perfectly, but test/Driver/driver_test.jl failed due to a NaN during initialization. My best guess is that one of the auxiliary variables for the AtmosModel isn't being initialized properly but it receives a reasonable value during update_auxiliary_state!. The integral variables as the most likely suspects, but you have four nodal_update_auxiliary_state! functions to check in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/AtmosModel.jl#L517-L527.
> 
> Actually, these methods are already point-wise. The problem is that the Courant calculation depends on the [auxiliary state](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel_kernels.jl#L2623). @kpamnany any ideas what to do here?

Hey @charleskawczynski, take a look at #1142 and #1158. You might just need to add some extra initializations in `init_state_auxiliary!` so that the Courant functions have something reasonable to work with.

charleskawczynski:
Closing in favor of #1386

Issue #1085, opened 2020-05-11
Save diagnostics from a later timestep


LenkaNovak:
# Description

Currently all diagnostic output can only be saved into nc/jld2 files from timestep 1. It would be great to be able to start the saving from a later timestep (e.g. after spinup). 
# Additional context

This will enable faster runs and a more efficient use of storage. 

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

kpamnany:
@blallen pointed out that it's useful to have a snapshot at `t0` and also at `timeend`. So then:
- collect at `t0` (adding in #1133)
- optionally wait some interval X (TODO)
- collect at specified intervals of Y (current behavior)
- collect at `timeend` (current behavior)

Yes?

Do note that the collection at `timeend` may not be at interval Y from the last collection (so the ticks on your time axis may not be evenly spaced)! This is the current behavior. The `simtime` variable in the diagnostics file will, however, be accurate.

Cc: @bischtob

PR #1224, opened 2020-06-09
Separate out initializer from callbacks

### LINKED ISSUE #1212: Remove `try`/`catch` from `EveryXWallTimeSeconds`
# Description

This appears to be swallowing errors and causing problems

Apparently it is for optional initialization.

@jkozdon on Slack:

> Maybe change the struct to have an init function:
> ```
> struct EveryXWallTimeSeconds
>     "time of the last callback"
>     lastcbtime_ns::Array{UInt64}
>     "time between callbacks"
>     Δtime::Real
>     "MPI communicator"
>     mpicomm
>     "function to execute for callback"
>     func::Function
>     "function to call to initialize the callback"
>     init::Function
>     function EveryXWallTimeSeconds(func, Δtime, mpicomm; init=()->nothing)
>         lastcbtime_ns = [time_ns()]
>         new(lastcbtime_ns, Δtime, mpicomm, func, init)
>     end
> end
> ```
> 

# For CLIMA developers

simonbyrne:
Fixes #1212

I ended up completely changing the interface, which I think is much more elegant and flexible.

I have


src/Numerics/ODESolvers/GenericCallbacks.jl
```diff
@@ -12,123 +12,111 @@ using ..ODESolvers
...
-        try
-            cb.func(true)
-        catch
-        end
-        return 0
-    end
 
+function initialize!(cb::EveryXWallTimeSeconds, solver, Q, param, t0)
+    cb.lastcbtime_ns = time_ns()
+    cb.init()
```
simonbyrne, line 34: I'm not sure if we would want to pass `solver, Q, param, t` to `init` and the main `func` as well?

kpamnany: It would be useful as it would allow us to move at least some of the callbacks to being functions rather than closures.

src/Utilities/Callbacks/Callbacks.jl
```diff
@@ -90,18 +88,34 @@ function diagnostics(
...
                     """
 Diagnostics: %s
     %s at %s""",
                     dgngrp.name,
-                    init ? "initializing" : "collecting",
+                    "collecting",
                     string(currtime),
                 )
-                dgngrp(currtime, init = init)
+                dgngrp(currtime, init = false)
```
kpamnany, line 118:
```suggestion
                dgngrp(currtime)
```

src/Numerics/ODESolvers/GenericCallbacks.jl
simonbyrne: okay, let me look into this.

src/Utilities/Callbacks/Callbacks.jl
```diff
@@ -358,41 +372,35 @@ function CB_constructor(interval::String, solver_config, default = nothing)
...
-            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
-    elseif endswith(interval, "mins")
-        secs = 60 * parse(Int, interval[1:(end - 4)])
-        return (func) ->
-            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
-    elseif endswith(interval, "secs")
-        secs = parse(Int, interval[1:(end - 4)])
-        return (func) ->
-            GenericCallbacks.EveryXWallTimeSeconds(func, secs, mpicomm)
+    if (m = match(r"^([0-9\.]+)(smonths|sdays|shours|smins|ssecs)$",interval); m !== nothing)
```
kpamnany, line 375: Not sure this change makes the code any clearer or more efficient, tbh. But this is fine too. Intervals will need to change to support #1085 anyway.

simonbyrne: Yeah, I know you weren't a fan, but it made it a bit easier to see which callback was used.

src/Numerics/ODESolvers/GenericCallbacks.jl
simonbyrne, line 34: @kpamnany what if we changed the interface to make it so that you called `initialize!` on the callback "function"?

Of course, this means that if you wanted to use anything that required initializing, you would have to define a custom struct.

```suggestion
function initialize!(cb::EveryXWallTimeSeconds, solver, Q, param, t0)
    cb.lastcbtime_ns = time_ns()
    initialize!(cb)
```

kpamnany: > Of course, this means that if you wanted to use anything that required initializing, you would have to define a custom struct.

Which would be... not easy? We can't really inherit the logic then and would have to reimplement it.

If this is not easy to do, I think we can skip it -- it's really no big deal if `init` in particular is a closure.

simonbyrne: okay, will leave this, can re-evaluate later.

src/Utilities/Callbacks/Callbacks.jl
```diff
@@ -63,7 +61,7 @@ Update
                 )
                 isnan(energy) && error("norm(Q) is NaN")
             end
-            user_info_callback(init)
+            user_info_callback()
```
kpamnany, line 64: Requires [this](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Driver/gcm_driver_test.jl#L110) and [this](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Driver/driver_test.jl#L139) to be fixed.

src/Numerics/ODESolvers/ODESolvers.jl
```diff
@@ -122,31 +118,15 @@ function solve!(
...
-            !(thisretval in (0, 1, 2)) &&
-            error("callback #$(i) returned invalid value. It should return either:
-                  `nothing` (continue time stepping)
-                  `0`       (continue time stepping)
-                  `1`       (stop time stepping after all callbacks)
-                  `2`       (stop time stepping immediately)")
-            retval = max(thisretval, retval)
-            retval == 2 && return gettime(solver)
+        val = GenericCallbacks.call!(callbacks, solver, Q, param, t)
+        if val !== nothing && val > 0
```
charleskawczynski, line 127:
```suggestion
        if !(val === nothing) && val > 0
```

simonbyrne: any reason why?

src/Numerics/ODESolvers/GenericCallbacks.jl
simonbyrne: I came up with a really nice way to do this!

src/Numerics/ODESolvers/ODESolvers.jl
charleskawczynski: Isn't the `===` evaluation more efficient than `==` for `nothing`?

jakebolewski: that would be `!=`, the above was correct for egal

simonbyrne: yep, `!==(a,b)` is defined as `!(a === b)`.

src/Numerics/ODESolvers/GenericCallbacks.jl
```diff
@@ -1,137 +1,193 @@
...
+
+- `GenericCallbacks.init!(cb, solver, Q, param, t0)`, to be called at solver
+  initialization.
+
+- `GenericCallbacks.call!(cb, solver, Q, param, t)`, to be called after each time step:
+  the return value dictates what action should be taken:
+
+   * `0` or `nothing`: continue time stepping as usual
+   * `1`: stop time stepping after all callbacks have been executed
+   * `2`: stop time stepping immediately
```
kpamnany, line 17:
```suggestion
- `GenericCallbacks.call!(cb, solver, Q, param, t)`, to be called after each time step.
```

src/Numerics/ODESolvers/GenericCallbacks.jl
```diff
@@ -1,137 +1,193 @@
...
+end
+
+function call!(cb::AtStart, solver, Q, param, t)
+    call!(cb.callback, solver, Q, param, t)
+end
+
+
+
+
+
```
kpamnany, line 89:
```suggestion
```
Does JuliaFormatter take care of these additional blank lines?

charleskawczynski: I think there's an option for it

src/Numerics/ODESolvers/GenericCallbacks.jl
simonbyrne: We still do depend on the return value (at least for the time being).

simonbyrne:
squashed and ready to go!

PR #1232, opened 2020-06-10
Add simple hydrostatic spin-down test for Ocean Model

blallen:
# Description

This PR adds a simple test of the hydrostatic dynamics of the Ocean model. It is intended to be used as the basic test case when implementing the split-explicit time stepper. The analytic solution (and more) can be found in this overleaf: https://www.overleaf.com/project/5e9dfd2716ba2600018de266

Currently, I'm just testing that the solution is within 0.5% of the analytic solution after a day. If anyone has better ideas please suggest them. I'd also like to turn this into a Literature tutorial, so I don't think I want to go the full convergence testing route with this, altho it is possible. 

I have


blallen:
> Not sure I know of a good way to test. As we discussed on Slack, using the state check stuff would be good.

I think I'm going to try relative error wrt analytic solution for now, mainly to see how good of a metric that is. I guess I could also do a state check and check both. That's probably better future-proofing.

blallen:
> Build failed:
> 
> * [ci/slurmci](https://gist.github.com/3516b4e37759841ca992475f6301d670)

@christophernhill can you help explain why the state check is failing? Should results be different between CPU and GPU? Would Julia 1.3 versus 1.4 matter for the numbers (I haven't updated my local Julia yet)? How can min and max pass to 15 places but mean fail at 0 or 1?

christophernhill:
I think the mean fields are essentially zero. They will be pretty
random, so typically we don't bother matching that. We could add an
order of magnitude test, but not sure its worth it?

Same for u[2] thats a zero field, it gets some random noise from DG
imprecision I think.

GPU and CPU can differ for effectively zero fields. They may well
differ in last digits for
stable fields too under optimization due to ordering changes, fp
rounding optimizations etc.... - so matching to 16 digits is a bit
ambitious!

blallen:
bors try

blallen:
thanks @christophernhill . It looks like the means and $y$-related momentum variables were causing the issues, so I've stopped checking those. I also have the tolerance set to 12 for everything dynamical and for 15 for the temperature-related fields as those should stay exactly zero. 

For some reason, the state check test seems to add three digits of precision to what I have in the refvals file. Do you know why this is?

christophernhill:
OK - I'll check the digit counting, that was version 0.1....

I will also see about adding some way to exclude/include fields by
name as an option for future - I think now there are new vars_ index
funcs that might be easier.  It would probably be useful to record
which fields/vector components are not relevant to an experiment.

PR #1234, opened 2020-06-11
Switch CI to use Github Actions, run tests in parallel and fast fail when possible 

jakebolewski:
* Run subset on every push in parallel
* Kill jobs that fail quickly
* Kill stale jobs quickly upon new commits to a branch
* Run all OS unit tests only on bors try / bors merge
* Remove Azure Devops Pipelines

# Description

I have


charleskawczynski:
Looks awesome. We still need to configure bors dependence, right?

slurmci-test.toml
```diff
@@ -2,11 +2,16 @@
 #
 
 cpu = [
+  { file = "test/Numerics/DGMethods/advection_diffusion/advection_diffusion_model_1dimex_bgmres.jl", slurmargs = ["--ntasks=1"], args = [] },
```
jkozdon, line 5: Move the `cpu_gpu` section and remove gpu only test?

slurmci-test.toml
```diff
@@ -2,11 +2,16 @@
...
   { file = "test/Numerics/DGMethods/advection_diffusion/pseudo1D_advection_diffusion.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/Numerics/DGMethods/advection_diffusion/pseudo1D_advection_diffusion_1dimex.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/Numerics/DGMethods/advection_diffusion/pseudo1D_advection_diffusion_mrigark_implicit.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/Numerics/DGMethods/advection_diffusion/pseudo1D_heat_eqn.jl", slurmargs = ["--ntasks=3"], args = [] },
   { file = "test/Numerics/DGMethods/advection_diffusion/periodic_3D_hyperdiffusion.jl", slurmargs = ["--ntasks=3"], args = [] },
+  { file = "test/Numerics/DGMethods/compressible_Navier_Stokes/ref_state.jl", slurmargs = ["--ntasks=1"], args = [] },
+  { file = "test/Numerics/DGMethods/Euler/isentropicvortex.jl", slurmargs = ["--ntasks=1"], args = [] },
+  { file = "test/Numerics/DGMethods/Euler/isentropicvortex_imex.jl", slurmargs = ["--ntasks=1"], args = [] },
+  { file = "test/Numerics/DGMethods/Euler/isentropicvortex_multirate.jl", slurmargs = ["--ntasks=1"], args = [] },
```
jkozdon, line 14: Move the `cpu_gpu` section and remove gpu only test?

.github/workflows/Linux-UnitTests.yml
```diff
@@ -0,0 +1,45 @@
+name: Linux Unit Tests
+
+on:
+  push:
+    branches-ignore:
```
jkozdon, line 5:
```suggestion
    branches:
```
Is this correct?

jkozdon:
> Looks awesome. We still need to configure bors dependence, right?

Do we need each github action job to have it's own name to work with bors?

See [this discussion](https://forum.bors.tech/t/bors-with-github-workflows/426)

Alternatively we can rename the [jobs this way](https://github.com/bors-ng/bors-ng/issues/730)

.github/workflows/Linux-UnitTests.yml
jakebolewski: we want to run these on every push, but not an staging / trying branches or merge to master as the os specific unit tests are run for those.

jakebolewski:
I haven't configured bors yet as I didn't know how it worked with the status indicator.  It seem like you can reference the job name so we'll just have to make those two unique instead of both being named `test`?

jakebolewski:
The other behavior would be to decide if we only want to run the tests for PR's and not for every push.  I think this might be better default, if you need to test some behavior out you can create a draft / WIP PR.

charleskawczynski:
> I haven't configured bors yet as I didn't know how it worked with the status indicator.

From what I tested [here](https://github.com/CliMA/CalibrateEmulateSample.jl/pull/48), the syntax of the status name depends on whether we're using a matrix job or not.

> The other behavior would be to decide if we only want to run the tests for PR's and not for every push. I think this might be better default, if you need to test some behavior out you can create a draft / WIP PR.

IMO, every push seems unnecessary. Often I find myself doing what @kpamnany does (canceling docs/azure after pushing to not waste resources). It'd be really nice to have a way to trigger incremental tests. Could [mergify](https://github.com/marketplace/mergify) somehow help?

jakebolewski:
if you push now, then its configured to cancel any previous jobs on that branch.

jakebolewski:
I'm guessing that is the desired behavior, automatically kill stale jobs only build the latest commit on every branch.

I guess my question is if these jobs should only be triggered for PR's and not every branch that get's pushed.

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
bors try-

kpamnany:
> I guess my question is if these jobs should only be triggered for PR's and not every branch that get's pushed.

Currently, CI is not triggered for branch pushes and only runs when you open a PR for that branch. I think this behavior is fine.

kpamnany:
I think what Charlie was getting at was if pushes to PRs should always trigger full CI. If we can just do a Linux CI build for pushes to a PR branch and do the full CI on `bors try` or `bors r+` that would perhaps be a good thing?

jakebolewski:
bors try

jakebolewski:
bors try

jakebolewski:
I think this is ready to go, it has / had the behavior that @kpamnany that you were describing :).   I put a reasonable time limit on the jobs as well.

test/runtests.jl
```diff
@@ -14,6 +14,14 @@ end
 @testset "ClimateMachine" begin
     all_tests = isempty(ARGS) || "all" in ARGS ? true : false
 
+    function has_submodule(sm)
+        any(ARGS) do a
+            a == sm && return true
+            first(split(a, '/')) == sm && return true
```
charleskawczynski, line 20: Not a blocker, but this string parsing won't work on Windows

charleskawczynski: There's quite a bit of hard-coded `/`'s in the tests at the moment, though. We can change that later if we need

.github/workflows/OS-UnitTests.yml
```diff
@@ -0,0 +1,43 @@
+name: OS Unit Tests
+
+on:
+  push:
+    branches:
+      - staging
+      - trying
+
```
charleskawczynski, line 8: This won't run on master?

charleskawczynski:
Also, not super important, but we could add badges for the new tests to the README (and remove the one for azure)

.github/workflows/OS-UnitTests.yml
jakebolewski: No, we can add it but if it passes staging isn't it redundant to re-run the job?

jakebolewski:
@charleskawczynski I can add some documentation, where would be the best place to do so?  (wiki, docs, etc.?)

test/runtests.jl
jakebolewski: I think it's ok here as it's just specifying a label and not a full path

```Numerics/DGMethods``` -> run the subset of tests under under the numerics group

maybe there is a less confusing prefix we can use to define test groupings

File: .github/workflows/OS-UnitTests.yml (hunk 5 above)
Comment (kpamnany) [reply .github/workflows/OS-UnitTests.yml:8]: > if it passes staging isn't it redundant to re-run the job?

Yes, I agree. We will hopefully add nightly or weekly test runs on master going forward.

**Comment by charleskawczynski** [comment]
> @charleskawczynski I can add some documentation, where would be the best place to do so? (wiki, docs, etc.?)

Good question. The badge you can add to the [README](https://github.com/CliMA/ClimateMachine.jl/blob/master/README.md), [this](https://github.com/CliMA/CalibrateEmulateSample.jl/blob/master/README.md) is what I've done in the past for a GHA badge. We could also add a blurb of how this works to the [continuous integration](https://clima.github.io/ClimateMachine.jl/latest/Contributing/#Continuous-integration-1) section of the contributing docs. But wiki sounds good too. But I don't have strong feelings as to where it belongs (I do like the idea of the badges on the front page though!).

**Comment by jakebolewski** [comment]
Ok I added some docs describing the current CI checks to `Contributing` / `continuous_integration` section.

**Comment by jakebolewski** [comment]
lets see how this goes, I'm sure with some experience we'll need to make some tweaks to the config so don't hesitate to raise issues

File: .github/workflows/Linux-UnitTests.yml [hunk 6]
```diff
@@ -0,0 +1,42 @@
...
+    timeout-minutes: 60
+    strategy:
+      fail-fast: true
+      matrix:
+          test-modules: ["Arrays,Atmos,Common,Diagnostics,Driver,InputOutput,Utilities",
+                         "Numerics/DGMethods",
+                         "Numerics/Mesh",
+                         "Numerics/ODESolvers",
+                         "Numerics/SystemSolvers",
+                         "Ocean"]
```
simonbyrne, line 18: this is super neat!

jakebolewski:
hmm all checks passed but it doesn't look like bors was misconfigured to check the status of the github action matrix?   It seems stuck in a running state.

@simonbyrne / @charleskawczynski  any ideas?

bors.toml
```diff
@@ -1,5 +1,5 @@
 status = [
-  "azure-ci",
+  "test-os",
```
jkozdon, line 2:
```suggestion
  "test-os (ubuntu-latest)",
  "test-os (windows-latest)",
  "test-os (macos-latest)",
```

jkozdon: I think it reads the names list on the left side of https://github.com/CliMA/ClimateMachine.jl/runs/766357298

PR #1235, opened 2020-06-11
Fix help text for command line options

jkozdon:
Args are now parsed in the cli function, not the init function. This
moves that argument parsing description help text and corrects a few
typos.

Also clarifies environment variable usage.

### DIFF
--- a/docs/src/GettingStarted/RunningClimateMachine.md
+++ b/docs/src/GettingStarted/RunningClimateMachine.md
@@ -102,8 +102,8 @@ There may also be driver-specific command line arguments.
 ## Environment variables
 
 Every command line option can be set as an environment variable
-(`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). For example to disable gpu
-use set `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`.
+(`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). For example to disable the
+use of the gpu set `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`.
 
 ## Running with MPI
 
--- a/docs/src/GettingStarted/RunningClimateMachine.md
+++ b/docs/src/GettingStarted/RunningClimateMachine.md
@@ -101,8 +101,9 @@ There may also be driver-specific command line arguments.
 
 ## Environment variables
 
-Currently, the `ClimateMachine` only looks at one environment variable --
-`CLIMATEMACHINE_GPU`, which if set to `false`, disables GPU use.
+Every command line option can be set as an environment variable
+(`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). For example to disable gpu
+use set `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`.
 
 ## Running with MPI
 
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -266,47 +266,11 @@ default system defaults at runtime, default values will be merged with the
 parsed argument settings, with parsed cli argument values taking precedent
 over runtime defined default values.
 
-
-Returns a `Dict` containing parsed process ARGS values.
-"""
-function cli(;
-    custom_settings::Union{Nothing, ArgParseSettings} = nothing,
-    init_driver::Bool = true,
-    kwargs...,
-)
-    kw_defaults = Dict{Symbol, Any}(kwargs)
-    parsed_args = parse_commandline(kw_defaults, custom_settings)
-    # we need to munge the parsed arg dict a bit as parsed arg keys
-    # and climatemachine initialization keywords are not 1:1
-    parsed_args["checkpoint_keep_one"] = !parsed_args["checkpoint_keep_all"]
-    parsed_kwargs = Dict{Symbol, Any}((Symbol(k), v) for (k, v) in parsed_args)
-    # allow for setting cli arguments as hard defaults that override parsed process ARGS
-    init_kwargs = merge(kw_defaults, parsed_kwargs)
-    # call init with munged kw arguments
-    ClimateMachine.init(; init_driver = init_driver, init_kwargs...)
-    return parsed_args
-end
-
-
-"""
-    ClimateMachine.init(;init_driver::Bool=true, kwargs...)
-
-Perform necessary initializations for ClimateMachine:
-- Initialize MPI.
 - Parse command line arguments. To support experiment-specific arguments,
 `arg_settings` may be specified (it is an `ArgParse.ArgParseSettings`);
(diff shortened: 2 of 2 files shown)


jkozdon:
When running a test with `--fix-rng-seed` I realized that the command line arguments were not being parsed in the tutorials. Maybe this was intention, but wasn't sure.

jakebolewski:
We don't want to use `cli` for any of the tests, or things that need to be in a notebook as it doesnt compose with arbitrary `ARGS`.  Its better in these cases to override what setting you need by passing  the kwarg through `init()`.

jakebolewski:
It was my impression that the tutorials are for running in a notebook and only the `experiments/` needed command line drivers.

jkozdon:
OK. I will just start modify args rather than using the command line arguments for the tutorials.

Are my corrections to the docs right?

src/Driver/Driver.jl
```diff
@@ -342,6 +306,42 @@ the process, otherwise the defaulting to `ClimateMachine.Settings`.
...
+"""
+    ClimateMachine.init(;init_driver::Bool=true, kwargs...)
+
+Perform necessary initializations for ClimateMachine:
+- Initialize MPI.
+- Determine whether GPU(s) is available and should be used (pass
+`disable-gpu = true` if not) and set the ClimateMachine array type appropriately.
+- Set up the global logger.
+
+Setting `init_driverd = false` will set the `ClimateMachine.Settings` singleton
```
jakebolewski, line 337: typo: `init_driver = false`

src/Driver/Driver.jl
```diff
@@ -342,6 +306,42 @@ the process, otherwise the defaulting to `ClimateMachine.Settings`.
...
+end
+
+
+"""
+    ClimateMachine.init(;init_driver::Bool=true, kwargs...)
+
+Perform necessary initializations for ClimateMachine:
+- Initialize MPI.
+- Determine whether GPU(s) is available and should be used (pass
+`disable-gpu = true` if not) and set the ClimateMachine array type appropriately.
```
jakebolewski, line 334: typo: `disable_gpu = true`

jakebolewski:
the docs look ok but it's difficult to tell what the change was

jkozdon:
> the docs look ok but it's difficult to tell what the change was

I just copy and pasted the help text for the command line arguments from `init` to `cli`

src/Driver/Driver.jl
jkozdon:
```suggestion
`disable_gpu = true` if not) and set the ClimateMachine array type appropriately.
```

src/Driver/Driver.jl
jkozdon:
```suggestion
Setting `init_driver = false ` will set the `ClimateMachine.Settings` singleton
```

src/Driver/Driver.jl
```diff
@@ -342,6 +306,42 @@ the process, otherwise the defaulting to `ClimateMachine.Settings`.
...
+`disable_gpu = true` if not) and set the ClimateMachine array type appropriately.
+- Set up the global logger.
+
+Setting `init_driverd = false` will set the `ClimateMachine.Settings` singleton
+values without initializing the ClimateMachine driver runtime.
+
+`ClimateMachine.Settings` values can be overloaded at runtime upon initialization.
+If keyword argument overloads are not supplied, the `init` routine will try and
+fallback on any `CLIMATEMACHINE_SETTINGS_<VALUE>` `ENV` variables defined for
+the process, otherwise the defaulting to `ClimateMachine.Settings`.
```
kpamnany, line 343:
```suggestion
the process, otherwise defaulting to `ClimateMachine.Settings`.
```

jkozdon:
> Unfortunately, now the arguments in `slurmci-test.toml` are not parsed for those tests that use `init()` and are parsed for those tests using `cli()` and there's no obvious distinction between the two apart from the path. Not sure what's the best way to resolve this, but perhaps we should remove the arguments such as `--fix-rng-seed` from stuff in `test/` from `slurmci-test.toml`?

RIght. This is the other problem.

We added the flag `--fix-rng-seed` to fix the CI so it was reproducible. Do we need to turn this into and environment variable option like the disable GPU environment variable?

kpamnany:
Every command line option can be set as an environment variable (`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). But it's reasonable for a test (which should be reproducible) to always explicitly pass in `fix_rng_seed = true`, which removes the need for them to use the command line argument.

jkozdon:
> Every command line option can be set as an environment variable (`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). But it's reasonable for a test (which should be reproducible) to always explicitly pass in `fix_rng_seed = true`, which removes the need for them to use the command line argument.

@kpamnany: https://github.com/CliMA/ClimateMachine.jl/pull/1235/commits/3c89e4cee350335c0cb9561418ecedf27a3c2f7d

docs/src/GettingStarted/RunningClimateMachine.md
```diff
@@ -101,8 +101,9 @@ There may also be driver-specific command line arguments.
 
 ## Environment variables
 
-Currently, the `ClimateMachine` only looks at one environment variable --
-`CLIMATEMACHINE_GPU`, which if set to `false`, disables GPU use.
+Every command line option can be set as an environment variable
+(`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). For example to disable gpu
+use set `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`.
```
jkozdon, line 106:
```suggestion
(`CLIMATEMACHINE_SETTINGS_<option-in-upper-case>`). For example to disable the
use of the  gpu set `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`. 
```

jkozdon:
> ut it's reasonable for a test (which should be reproducible) to always explicitly pass in `fix_rng_seed = true`

Should we just universally set `CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED=true` for all CI testing?

kpamnany:
> Should we just universally set `CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED=true` for all CI testing?

That makes sense to me. I guess in `/test/runtests.jl`? Not sure...

jkozdon:
> > Should we just universally set `CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED=true` for all CI testing?
> 
> That makes sense to me. I guess in `/test/runtests.jl`? Not sure...

Probably `experiments` and `tutorials` for CI as well. These were the ones that were causing problems before.

jakebolewski:
you might have to export this if you are running with `runmpi` as it will launch sub-processes that won't inherit the spawning process `ENV` settings.

Issue #1216, opened 2020-06-08
Move orientation.jl, turbulence.jl, hyperdiffusion.jl out of Atmos model


tapios:
# Description

There are several pieces of Atmos code that will be useful for other models (e.g., ocean, land). It would be good to move them out of the Atmos code and make them accessible to other models. 

# Additional context

Specifically,

- `orientation.jl` (latitude, longitude, altitude etc. should be generically useable by model components (ocean, land, sea ice)
- `turbulence.jl` and `hyperdiffusion.jl` would be useful for fluid components (ocean+atmosphere)

It would be nice to make these commonly accessible, e.g., as modules under `src/Common` or under `DGMethods` for `hyperdiffusion.jl`.

# For CLIMA Developers

- [x] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

charleskawczynski:
Still need to move turbulence and hyperdiffusion out

charleskawczynski:
We still need to move `hyperdiffusion.jl`

PR #1248, opened 2020-06-15
Move orientations to Common/

charleskawczynski:
# Description

Moves `orientations.jl` into its own module so that it can be shared by Land, Ocean and Atmosphere. A step towards #1216.

I have

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -9,6 +9,17 @@ using CLIMAParameters.Atmos.SubgridScale: C_smag
 using DocStringExtensions
 using LinearAlgebra, StaticArrays
 using ..ConfigTypes
+using ..Orientations
+import ..Orientations:
+    vertical_unit_vector,
+    altitude,
+    latitude,
+    longitude,
+    projection_normal,
+    gravitational_potential,
+    ∇gravitational_potential,
+    projection_tangential
+
 using ..VariableTemplates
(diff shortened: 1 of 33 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -569,7 +595,7 @@ Store Cartesian coordinate information in `aux.coord`.
 """ init_state_auxiliary!
 function init_state_auxiliary!(m::AtmosModel, aux::Vars, geom::LocalGeometry)
     aux.coord = geom.coord
-    atmos_init_aux!(m.orientation, m, aux, geom)
+    init_aux!(m.orientation, m.param_set, aux)
```
akshaysridhar, line 598: Would this be an issue if other `Common` components export `init_aux(balancelaw.property, ...)`

charleskawczynski: No, we'd just need to `import` (and export) it if we wanted to use the same function (and define new methods). Maybe we should change the name to `init_aux_orientation!`? Idk what's the best naming strategy here. There's also the question of whether we should pass in `aux` or `aux.orientation` and change `init_aux!` appropriately. @simonbyrne, thoughts? Since we'll likely apply similar patterns to `TurbulenceClosures` and `Hyperdiffusion` modules.

charleskawczynski: Let's merge and we can change the name later.

PR #1250, opened 2020-06-16
generate tutorial output in parallel

jakebolewski:
# Description

Generating the tutorial output when building the documentation can take a long time, so enable parallel generation of literate output.   This should not impact normal single core builds (ex. `julia --project=docs/ docs/make.jl`).

I have

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,18 +1,25 @@
 Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
 
-using ClimateMachine, Documenter, Literate
-
 # https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
 ENV["GKSwstype"] = "100"
+# some of the tutorials cannot be run on the GPU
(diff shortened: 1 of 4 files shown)


docs/make.jl
```diff
@@ -1,18 +1,25 @@
 Base.HOME_PROJECT[] = abspath(Base.HOME_PROJECT[]) # JuliaLang/julia/pull/28625
 
-using ClimateMachine, Documenter, Literate
-
 # https://github.com/jheinen/GR.jl/issues/278#issuecomment-587090846
 ENV["GKSwstype"] = "100"
+# some of the tutorials cannot be run on the GPU
+ENV["CLIMATEMACHINE_SETTINGS_DISABLE_GPU"] = true
```
charleskawczynski, line 6: Is this necessary given that we’ve set this in the Documenter yaml? I think the only tutorial that isn’t GPU friendly is the StateCheck one.

jakebolewski: this is just convenience (for myself personally) because generating the docs locally doesn't work with the default behavior of deferring to use the GPU by default.

with multiple processes, the ENV distributed to workers is only the process ENV defined at invocation (shell env) not the ENV of the master process during runtime.

PR #1251, opened 2020-06-16
add separate precompile step for unit build matrix

jakebolewski:
# Description

@kpamnany pointed out that there are some sporadic `EEXIST` errors in some `runmpi` tests when decomposed as part of the build matrix.  Add explicit `Pkg.precompile` step to build stage to see if this fixes the possible race issue.

I have

### DIFF
--- a/.github/workflows/OS-UnitTests.yml
+++ b/.github/workflows/OS-UnitTests.yml
@@ -32,7 +32,7 @@ jobs:
 
     - name: Install Project Packages
       run: |
-        julia --project=@. -e 'using Pkg; Pkg.instantiate(); Pkg.build()'
+        julia --project=@. -e 'using Pkg; Pkg.instantiate()'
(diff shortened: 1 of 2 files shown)


.github/workflows/OS-UnitTests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
     - name: Install Project Packages
       run: |
-        julia --project=@. -e 'using Pkg; Pkg.instantiate(); Pkg.build()'
+        julia --project=@. -e 'using Pkg; Pkg.instantiate()'
```
charleskawczynski, line 35: Just out of curiosity, why no `precompile` here?

jakebolewski: we use the system image for these tests, which invokes pre-compilation

PR #1116, opened 2020-05-18
Add `--fix-rng-seed` for reproducibility

kpamnany:
# Description

Closes #977. Will update documentation in #1091.

I have

- [X] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [X] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1259, opened 2020-06-18
init with fix_rng_seed for all tests

jakebolewski:
# Description

For the tests, fixing the rng seed was done though a CLI argument, this forces the behavior for all test cases upon initialization.

I have

### DIFF
--- a/test/Arrays/basics.jl
+++ b/test/Arrays/basics.jl
@@ -3,7 +3,7 @@ using Test, MPI
 using ClimateMachine
 using ClimateMachine.MPIStateArrays
 
-ClimateMachine.init()
+ClimateMachine.init(fix_rng_seed=true)
 const ArrayType = ClimateMachine.array_type()
 const mpicomm = MPI.COMM_WORLD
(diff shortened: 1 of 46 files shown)


kpamnany:
See [this comment](https://github.com/CliMA/ClimateMachine.jl/pull/1116#issuecomment-631673614) and [this one](https://github.com/CliMA/ClimateMachine.jl/pull/1116#issuecomment-632285276).

jakebolewski:
Slurm has been updated, this is just a quick search replace job but I'll go through to see where we are setting these explicitly.

It seems like a better API would be to pass the seed value upon initialization of the driver runtime, rather than fixing to some (pre-defined) constant.  That would cover all cases.

kpamnany:
> It seems like a better API would be to pass the seed value upon initialization of the driver runtime, rather than fixing to some (pre-defined) constant. That would cover all cases.

The experiment doesn't decide how many processes/threads it is run with -- using the same seed for all might skew results?

PR #1266, opened 2020-06-21
Move balance law to its own module

charleskawczynski:
# Description

Right now, the `BalanceLaw` and `DGMethods`/`DGModel` are a bit entangled but I think they have a clear separation of concerns:

 - The `BalanceLaw` is concerned with specifying a spatially and temporally continuous system of PDEs.
 - `DGMethods`/`DGModel` is concerned with specifying the spatial discretization of the given PDE system.

This PR moves `BalanceLaw` to its own folder and module.

I have


charleskawczynski:
bors try

charleskawczynski:
bors try

docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
```diff
@@ -35,44 +35,44 @@ the following methods, which are computed locally at each nodal point:
...
+| [`flux_second_order!`](@ref)    |  specify `F_{second_order}` for each conservative state variable. can be functions of the conservative state, gradient flux state, and auxiliary variables. |
+| [`source!`](@ref)            |  specify `S_{non_conservative}`  for each conservative state variable. can be functions of the conservative state, gradient flux state, and auxiliary variables. |
 
 ## Methods to compute numerical fluxes
 | **Method** | Purpose |
 |:-----|:-----|
-| [`wavespeed`](@ref DGMethods.wavespeed) | specify how to compute the local wavespeed if using the `RusanovNumericalFlux`. |
-| [`boundary_state!`](@ref DGMethods.boundary_state!) | define exterior nodal values of the conservative state and gradient flux state used to compute the numerical boundary fluxes. |
+| [`wavespeed`](@ref) | specify how to compute the local wavespeed if using the `RusanovNumericalFlux`. |
+| [`boundary_state!`](@ref) | define exterior nodal values of the conservative state and gradient flux state used to compute the numerical boundary fluxes. |
```
blallen, line 69: I think this functions are pretty integrally tied to the DG discretization as they are used for computing numerical fluxes. `

boundary_state!` should probably be replaced with a `boundary_condition!` interface of which `boundary_state!` is the DG specific implementation.

the `wavespeed` function here is not a generic wavespeed but a particular weighting coefficient that is representative of a wavespeed used when calculating the rusanov flux. maybe this could have a name separate from `wavespeed` to distinguish from the physical wavespeed (altho that might still be used for the rusanov flux in some cases)

docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
```diff
@@ -35,44 +35,44 @@ the following methods, which are computed locally at each nodal point:
...
-| [`update_auxiliary_state_gradient!`](@ref DGMethods.update_auxiliary_state_gradient!) | same as above, but after computing gradients and gradient fluxes in case these variables are needed during the update. |
+| [`compute_gradient_argument!`](@ref) | specify how to compute the arguments to the gradients. can be functions of conservative state  and auxiliary variables. |
+| [`compute_gradient_flux!`](@ref) | specify how to compute gradient fluxes. can be a functions of the gradient state, the conservative state, and auxiliary variables.|
+| [`integral_load_auxiliary_state!`](@ref) | specify how to compute integrands. can be functions of the conservative state and auxiliary variables. |
+| [`integral_set_auxiliary_state!`](@ref) | specify which auxiliary variables are used to store the output of the integrals. |
+| [`reverse_integral_load_auxiliary_state!`](@ref) | specify auxiliary variables need their integrals reversed. |
+| [`reverse_integral_set_auxiliary_state!`](@ref) | specify which auxiliary variables are used to store the output of the reversed integrals. |
```
blallen, line 53: These are also pretty hard-cored tied into how the DG rhs! compute kernel works. I guess moving them here is a good first step to a proper integral interface, but these are really just interfacing with the `indefinite_stack_integral!` and `reverse_indefinite_stack_integral!` functions, which assume a DG discretization.

docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
```diff
@@ -35,44 +35,44 @@ the following methods, which are computed locally at each nodal point:
...
+| [`update_auxiliary_state!`](@ref) | perform any updates to the auxiliary variables needed at the beginning of each time-step. Can be used to solve non-linear equations, calculate integrals, and apply filters. |
+| [`update_auxiliary_state_gradient!`](@ref) | same as above, but after computing gradients and gradient fluxes in case these variables are needed during the update. |
```
blallen, line 55: Same here, these are place holders for "any operation that is not a gradient of a state variable or a divergence of a flux". They don't really make sense at the continuous level, only at the DG discretization level. The DG algorithm is being assumed here, because in a true continuous formulation everything is done at once.

docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
charleskawczynski: I agree with the suggested name changes, I didn't want to make too many changes for this PR.

I would argue that the `wavespeed` is a point-wise method and it may be needed for components in the discretization layer but one can write a spatially continuous form. Also, this method could be used for different spatial discretizations (FDM, FVM), so I don't think it's really DG specific.

charleskawczynski:
I'm fine with splitting things out in a way that makes sense but, I think that every method moved to `BalanceLaws` could have a FDM/FVM/FEM implementation, so I'd consider them discretization-independent (every moved method is point-wise or global, but no particular discretization is specified).

PR #1279, opened 2020-06-23
Add land framework

kmdeck:
# Description

This code sets up the framework for the land model. One of the subcomponents of the LandModel is the soil_model, which in turn has the subcomponents soil_heat and soil_water. All of these will be Balance Laws, but we want to be able to turn off and on different components, have everything still run, and to be able to test these different versions of the land model.

We have also defined types and functions that will be used in the soil_water model. In particular, the conductivity (that will appear in the diffusive flux term of the balance law) is the result of three factors multiplied together. We designed the conductivity such that we can turn on and off different factors (which may introduce other state variables, when coupled to heat, e.g.) and still run the soil_water model. We also have added in some functions used in the equation (e.g. pressure head, hydraulic head, matric potential). These are all functions of the state variable(s) for the water model. 

I have

- [X] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`


kmdeck:
We do have a question about how we defined the different methods of hydraulic conductivity. 

Hydraulic conductivity depends on three factors -  viscosity_factor, moisture_factor, and impedance_factor. I'll just refer to impedance_factor in the examples here. Depending on the concrete type (NoImpedance vs IceImpedance), the impedance_factor method require different arguments. It either returns 1 for NoImpedance and does not require theta_ice, or requires theta_ice for IceImpedance, and returns a function of that variable.  

We defined multiple methods of hydraulic_conductivity which take a different number of arguments, based on which method of the _factor functions we want it to use. E.g. we have a method of hydraulic conductivity which takes IceImpedance and theta_ice, and one that takes only NoImpedance. An alternative (@cfranken) would be to define a single function hydraulic_conductivity which takes all possible arguments  (e.g. theta_ice) that might be needed by the factor functions, and redefine the factor functions to always accept the arguments which might be needed by them. Depending on the type of e.g. Impedance model passed to hydraulic conductivity, and then onto the impedance_factor, the variable theta_ice either is used or not. This would be a lot cleaner, but we would be passing data in the form of arguments that would not always be used. 

Anyways, this is an area where I'm sure we can improve things. thank you for any advice/thoughts!

kmdeck:
After speaking with @charleskawczynski I will clean up the issues with hydraulic_conductivity in the above comment today, and resubmit. I converted this back to a draft in the meantime.

test/Land/Model/test_water_parameterizations.jl
```diff
@@ -0,0 +1,78 @@
...
+using StaticArrays
+using Test
+
+using CLIMAParameters
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+
+using ClimateMachine
+using ClimateMachine.Land.SoilWaterParameterizations
+
```
charleskawczynski, line 12:
```suggestion
@testset "Land water parameterizations" begin
```

test/Land/Model/test_water_parameterizations.jl
```diff
@@ -0,0 +1,78 @@
...
+#test branching in pressure head
+@test pressure_head(vg_model, 1.0, 0.001, 1.5) == 500
+
+@test pressure_head.(Ref(hk_model), Ref(1.0), Ref(0.001), test_array) ≈
+      .-((-1 .+ test_array .^ (-1 / m)) .* α^(-n)) .^ (1 / n)
+
+
+m = FT(0.5)
+ψb = FT(0.1656)
+@test pressure_head(bc_model, 1.0, 0.001, 0.5) ≈ -ψb * 0.5^(-1 / m)
```
charleskawczynski, line 78:
```suggestion
@test pressure_head(bc_model, 1.0, 0.001, 0.5) ≈ -ψb * 0.5^(-1 / m)
end
```

src/Land/Model/soil_water.jl
```diff
@@ -0,0 +1,45 @@
...
+        typeof(impedance_factor),##here we get the concrete types
+        typeof(viscosity_factor),
+        typeof(moisture_factor),
+        typeof(hydraulics),
+    }(
+        impedance_factor,
+        viscosity_factor,
+        moisture_factor,
+        hydraulics,
+    )
```
charleskawczynski, line 44:
```suggestion
    args = (
        impedance_factor,
        viscosity_factor,
        moisture_factor,
        hydraulics,
        )
    return SoilWaterModel{FT,typeof.(args)...}(args...)
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,300 @@
...
+function vars_state_conservative(land::LandModel, FT)
+    @vars begin
+        soil::vars_state_conservative(land.soil, FT)
+    end
+end
+
+"""
+    vars_state_auxiliary(land::LandModel, FT)
+
+Names of variables required for the balance law that aren't related to derivatives of the state variables (e.g. spatial coordinates or various integrals) or those needed to solve expensive auxiliary equations (e.g., temperature via a non-linear equation solve)
```
charleskawczynski, line 72:
```suggestion
Names of variables required for the balance law that aren't related to derivatives of the
 state variables (e.g. spatial coordinates or various integrals) or those needed to solve
  expensive auxiliary equations (e.g., temperature via a non-linear equation solve)
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,300 @@
...
+function vars_state_auxiliary(land::LandModel, FT)
+    @vars begin
+        soil::vars_state_auxiliary(land.soil, FT)
+    end
+end
+
+"""
+    vars_state_gradient(land::LandModel, FT)
+
+Names of the gradients of functions of the conservative state variables. Used to represent values before **and** after differentiation
```
charleskawczynski, line 83:
```suggestion
Names of the gradients of functions of the conservative state variables.
 Used to represent values before **and** after differentiation
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,300 @@
...
+        dg::DGModel,
+        land::LandModel,
+        Q::MPIStateArray,
+        t::Real,
+        elems::UnitRange,
+    )
+
+Perform any updates to the auxiliary variables needed at the beginning of each time-step.
+"""
+
```
charleskawczynski, line 227:
```suggestion
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,300 @@
...
+    function land_nodal_update_auxiliary_state!(
+        land::LandModel,
+        state::Vars,
+        aux::Vars,
+        t::Real,
+    )
+
+Update the auxiliary state array
+"""
+
```
charleskawczynski, line 255:
```suggestion
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,300 @@
...
+        aux::Vars,
+        t::Real,
+        direction,n
+    )
+Computes (and assembles) source terms `S(Y)` in:
+∂Y
+-- = - ∇ • F + S(Y)
+∂t
+"""
+
```
charleskawczynski, line 280:
```suggestion
```

src/Land/Model/soil_heat.jl
```diff
@@ -0,0 +1,32 @@
...
+The balance law for internal energy in soil.
+
+"""
+struct SoilHeatModel <: AbstractHeatModel end
+
+"""
+    vars_state_conservative(soil::AbstractHeatModel, FT)
+
+"""
+
```
charleskawczynski, line 18:
```suggestion
```

src/Land/Model/soil_heat.jl
```diff
@@ -0,0 +1,32 @@
...
+struct ConstantInternalEnergy{FT} <: AbstractHeatModel
+    T::FT
+end
+
+"""
+    get_temperature(m::ConstantInternalEnergy)
+
+"""
+
```
charleskawczynski, line 31:
```suggestion
```

src/Land/Model/soil_model.jl
```diff
@@ -0,0 +1,253 @@
...
+        heat::vars_state_auxiliary(soil.heat, FT)
+    end
+end
+
+"""
+    vars_state_gradient(soil::SoilModel, FT)
+
+Names of the gradients of functions of the conservative state variables. Used to represent values before **and** after differentiation
+"""
+
```
charleskawczynski, line 58:
```suggestion
```

src/Land/Model/LandModel.jl
charleskawczynski: There's a bunch of these (we cannot have spaces between docstrings and structs/functions), just FYI

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,512 @@
+"""
+    Soil Water Parameterizations
+
+van Genuchten, Brooks and Corey, and Haverkamp parameters for and formulation of
+  - hydraulic conductivity
+  - matric potential
+  Hydraulic conductivity can be chosen to be dependent or independent of impedance, viscosity and moisture.
+  Expressions for hydraulic head, effective saturation and pressure head are also included.
+"""
+
```
charleskawczynski, line 10:
```suggestion
```

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,512 @@
...
+AbstractMoistureFactor{FT <:AbstractFloat}
+"""
+abstract type AbstractMoistureFactor{FT <: AbstractFloat} end
+
+
+"""
+AbstractsHydraulicsModel{FT <: AbstractFloat}
+
+    Hydraulics model is used in the moisture factor in hydraulic conductivity and in the matric potential. The single hydraulics model choice sets both of these.
+"""
```
charleskawczynski, line 56:
```suggestion
"""
    AbstractsHydraulicsModel{FT <: AbstractFloat}

Hydraulics model is used in the moisture factor in hydraulic conductivity and
 in the matric potential. The single hydraulics model choice sets both of these.
"""
```

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,512 @@
...
+        A::FT = FT(124.6 / 100.0^1.77),
+        n::FT = FT(1.43),
+        α::FT = FT(2.6),
+    ) where {FT}
+        new(k, A, n, α, 1 - 1 / FT(n))
+    end
+end
+
+"""
+struct MoistureIndependent{FT} <: AbstractMoistureFactor{FT} end
```
charleskawczynski, line 128:
```suggestion
    MoistureIndependent{FT} <: AbstractMoistureFactor{FT}
```

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,512 @@
...
+"""
+struct MoistureDependent{FT} <: AbstractMoistureFactor{FT} end
+
+
+"""
+moisture_factor(
+    mm::MoistureDependent{FT},
+    hm::vanGenuchten{FT},
+    S_l::FT,
+) where {FT}
```
charleskawczynski, line 148:
```suggestion
    moisture_factor(
        mm::MoistureDependent{FT},
        hm::vanGenuchten{FT},
        S_l::FT,
    ) where {FT}
```

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,512 @@
...
+
+
+θ_l is defined to be zero or positive. If θ_l is negative, hydraulic functions that take it as an argument will return imaginary numbers, resulting in domain errors. However, it is possible that our current solver returns a negative θ_l due to numerical issues. Provide a warning in this case, and correct the value of θ_l so that the integration can proceed. We will remove this once the numerical issues are resolved.
+
+"""
+function effective_saturation(porosity::FT, θ_l::FT) where {FT}
+
+    if θ_l < 0
+        @show θ_l
+        @warn("Augmented liquid fraction is negative - domain error. Artificially setting equal to zero to proceed. ")
```
charleskawczynski, line 423: I don't think we can leave the `@show` and `@warn` here because it's not GPU compatible and it will result in very large print logs (this function will be called at every nodal point for every timestep). Perhaps this should throw an error. See [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Common/Thermodynamics/relations.jl#L1107-L1127) how print/error debugging is done in Thermodynamics.jl. `error_on_non_convergence` is defined in `Thermodynamics.jl`.

src/Land/Model/LandModel.jl
kmdeck: Thank you! I guess we only fixed the ones where we exported the struct/function, and were getting errors with them in the documentation check. will make sure this is fixed in the future.

src/Land/Model/SoilWaterParameterizations.jl
kmdeck: ok! this is just a more compact way of writing that?

kmdeck: oh wait, nevermind. this is in the documentation. sorry!

src/Land/Model/SoilWaterParameterizations.jl
kmdeck: yeah - I actually think that we dont need to make a version with an error_on_theta_negative attribute that either sets theta to zero when negative (if False) or gives an error when theta is negative (if True). If the user wants to set set it to zero in this case for debugging, they can do so, but we dont need to have clean code in master that allows for debugging. @cfranken @charleskawczynski  I guess I am proposing only this:
```
function effective_saturation(porosity::FT, θ_l::FT) where {FT}

    if θ_l < 0
        @print("-----------------------------------------\n")
        @print("Effective saturation is negative:\n")
        @print("θ_l =",
                θ_l,
               "\n"
               )
        error("Effective saturation is negative. Exiting...\n")
    end
    S_l = θ_l / porosity
    return S_l
end
```

kmdeck: (if someone wants to debug, they can make the change to debug, and then change it back when bug is fixed)

kmdeck: Edit - going with throw(), because then I can test that the correct exception is thrown in unit tests:

```
function effective_saturation(porosity::FT, θ_l::FT) where {FT}

    if θ_l < 0
        throw(DomainError(θ_l, "Effective saturation is negative."))
    end
    S_l = θ_l / porosity
    return S_l
end
```

charleskawczynski: You can use `test_throws ErrorException effective_saturation(...)`. Also, you'll need to add `using KernelAbstractions: @print` to the top of the module.

kmdeck: ok awesome! did so. this doesnt need @print then, so i think we are good (the message is printed in the exception).

going to resolve, I also updated repo to reflect these changes + the last formatting issues you mentioned (line lengths, """ on line before functions)

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,531 @@
...
+    "Reference temperature"
+    T_ref::FT = FT(288.0)
+end
+
+
+"""
+function viscosity_factor(
+    vm::ConstantViscosity{FT},
+    T::FT,
+) where {FT}
```
charleskawczynski, line 281:
```suggestion
    function viscosity_factor(
        vm::ConstantViscosity{FT},
        T::FT,
    ) where {FT}
```

src/Land/Model/soil_heat.jl
```diff
@@ -0,0 +1,30 @@
...
+"""
+struct SoilHeatModel <: AbstractHeatModel end
+
+"""
+    vars_state_conservative(soil::AbstractHeatModel, FT)
+
+"""
+function vars_state_conservative(soil::AbstractHeatModel, FT)
+    @vars begin end
```
charleskawczynski, line 19:
```suggestion
    @vars()
```

test/Land/Model/constant_energy_model.jl
```diff
@@ -0,0 +1,64 @@
...
+# solver_config =
+#     ClimateMachine.SolverConfiguration(t0, timeend, driver_config, ode_dt = dt);
+
+# ClimateMachine.invoke!(solver_config);
+
+
+# function init_state_auxiliary!(land::LandModel, aux::Vars, geom::LocalGeometry)
+#     aux.coord = geom.coord
+#     init_state_auxiliary!(land, land.soil, aux, geom)
+# end
```
charleskawczynski, line 64: Can we add a test from this? Or add tests that these objects are initialized?

test/Land/Model/constant_energy_model.jl
```diff
@@ -0,0 +1,64 @@
...
+using ClimateMachine.VariableTemplates
+using ClimateMachine.SingleStackUtils
+
+FT = Float64;
+ClimateMachine.init(; disable_gpu = true);
+
+const clima_dir = dirname(dirname(pathof(ClimateMachine)));
+
+# Load some helper functions for plotting
+# include(joinpath(clima_dir, "docs", "plothelpers.jl"));
```
charleskawczynski, line 28: Let's remove this

kmdeck: removed this file from tracking as it isnt being used

src/Land/Model/soil_heat.jl
kmdeck: fixed in  local repo & will push.

test/Land/Model/constant_energy_model.jl
kmdeck: removed this file from tracking as it was not being used; adding the equivalent test in constant_moisture_model.jl

.github/workflows/Linux-UnitTests.yml
```diff
@@ -15,7 +15,8 @@ jobs:
                          "Numerics/Mesh",
                          "Numerics/ODESolvers",
                          "Numerics/SystemSolvers",
-                         "Ocean"]
+                         "Ocean",
+			 "Land"]
```
charleskawczynski, line 19: The spacing/indenting needs to match, here.

kmdeck: it's strange, they seem lined up on my computer editor. I just added Land to the tail after "Arrays,...Utilities" instead, then.

charleskawczynski: Ah, maybe spaces vs tabs?

kmdeck: yes, something like that. Ocean is failing now! haha

kmdeck: lol, there is a reason why everyone else has tacked on their names after Arrays...

kmdeck:
It looks like everything will pass, and the documents look fine, but now there are a lot of conflicts when I try and rebase. will look at on monday. 
Screen Shot 2020-06-26 at 7 18 37 PM

docs/list_of_apis.jl
```diff
@@ -12,7 +12,11 @@ apis = [
     ],
     "Ocean" =>
         ["Hydrostatic Boussinesq" => "APIs/Ocean/HydrostaticBoussinesq.md"],
-    "Land" => [],
+    "Land" => [
+        "LandModel" => "APIs/Land/LandModel.md",
+        "SoilWaterParameterizations" =>
```
charleskawczynski, line 17:
```suggestion
        "Soil Water Parameterizations" =>
```
This is the title of the navigation page, so it should read as plain text and not code.

docs/list_of_apis.jl
```diff
@@ -12,7 +12,11 @@ apis = [
     ],
     "Ocean" =>
         ["Hydrostatic Boussinesq" => "APIs/Ocean/HydrostaticBoussinesq.md"],
-    "Land" => [],
+    "Land" => [
+        "LandModel" => "APIs/Land/LandModel.md",
```
charleskawczynski, line 16:
```suggestion
        "Land Model" => "APIs/Land/LandModel.md",
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,302 @@
...
+Users may over-ride prescribed default values for each field.
+
+# Usage
+
+    LandModel(
+        param_set,
+        soil,
+        source
+    )
+
```
charleskawczynski, line 44:
```suggestion
    LandModel{PS, S, SRC} <: BalanceLaw
```

src/Land/Model/LandModel.jl
```diff
@@ -0,0 +1,302 @@
...
+        land::LandModel,
+        flux::Grad,
+        state::Vars,
+        diffusive::Vars,
+        hyperdiffusive::Vars,
+        aux::Vars,
+        t::Real,
+    )
+
+Specify F_{second_order} for each conservative state variable
```
charleskawczynski, line 196:
```suggestion
Specify ``F_{\mathrm{second_order}}`` for each conservative state variable
```
I think this is right?

src/Land/Model/SoilWaterParameterizations.jl
```diff
@@ -0,0 +1,530 @@
...
+
+
+
+"""
+   effective_saturation(porosity::FT, θ_l::FT)
+
+Compute the effective saturation of soil.
+
+
+θ_l is defined to be zero or positive. If θ_l is negative,
```
charleskawczynski, line 433:
```suggestion
`θ_l` is defined to be zero or positive. If `θ_l` is negative, 
```
Without adding back-ticks, the underscore triggers _italics_ in markdown mode (see what this looks like [here](https://clima.github.io/ClimateMachine.jl/previews/PR1279/APIs/Land/SoilWaterParameterizations/)). Entering code mode (single back-tics), or math mode (with double back-tics), seems reasonable here.

src/Land/Model/soil_model.jl
```diff
@@ -0,0 +1,251 @@
...
+A BalanceLaw for soil modeling.
+Users may over-ride prescribed default values for each field.
+
+# Usage
+
+    SoilModel(
+        water,
+        heat,
+    )
+
```
charleskawczynski, line 17:
```suggestion
    SoilModel{W, H} <: BalanceLaw
```
The properties are already expanded by `$(DocStringExtensions.FIELDS)` (see [here](https://clima.github.io/ClimateMachine.jl/previews/PR1279/APIs/Land/LandModel/#ClimateMachine.Land.SoilModel)), so including signature/type parameters I think provides better information here.

src/Land/Model/SoilWaterParameterizations.jl
kmdeck: got it! I also noticed that some of the "where"s are bolded in the lines looking like : ) where {FT}, but in the code there is no apparent difference. It seems like __word__ and **word** would bold it, but that isn't happening...

src/Land/Model/LandModel.jl
kmdeck: I think I'll go with ``F_{\mbox{second order}}`` so that there is a space (that looks right in latex editor). but yeah, ill make sure it looks ok! That second_order makes the o a subscript, and \mathrm{second order} removes the space.

kmdeck: ah, well both seem to be giving me "ERROR: LoadError: LoadError: syntax: invalid escape sequence" in the documentation build.

kmdeck: just going to write it out as second order flux :)

PR #1280, opened 2020-06-23
Add types and unify functions in BalanceLaws

charleskawczynski:
# Description

 - Adds types to balance law (I have no strong opinions about the names here)

```julia
abstract type AbstractStateType end
struct Prognostic <: AbstractStateType end
struct Auxiliary <: AbstractStateType end
struct Gradient <: AbstractStateType end
struct GradientFlux <: AbstractStateType end
struct GradientLaplacian <: AbstractStateType end
struct Hyperdiffusive <: AbstractStateType end
struct UpwardIntegrals <: AbstractStateType end
struct DownwardIntegrals <: AbstractStateType end
```

 - Unifies `vars_state` (BalanceLaws) into a single function
 - Unifies `number_states` (BalanceLaws) into a single function
 - Unifies `create_state` (DGMethods) into a single function
 - Unifies `extract_state` (Diagnostics) into a single function

This PR also does a replace-all for `state_conservative` -> `state_prognostic`.

I have

### DIFF
--- a/docs/src/APIs/BalanceLaws/BalanceLaws.md
+++ b/docs/src/APIs/BalanceLaws/BalanceLaws.md
@@ -10,15 +10,23 @@ CurrentModule = ClimateMachine.BalanceLaws
 BalanceLaw
 ```
 
+## State variable types
+
+```@docs
+Prognostic
+Auxiliary
+Gradient
+GradientFlux
+GradientLaplacian
+Hyperdiffusive
+UpwardIntegrals
+DownwardIntegrals
+```
+
 ## Variable specification methods
 
 ```@docs
-vars_state_conservative
-vars_state_auxiliary
-vars_state_gradient
-vars_integrals
-vars_reverse_integrals
-vars_state_gradient_flux
+vars_state
 ```
 
 ## Initial condition methods
--- a/docs/src/APIs/BalanceLaws/BalanceLaws.md
+++ b/docs/src/APIs/BalanceLaws/BalanceLaws.md
@@ -32,7 +32,7 @@ vars_state
 ## Initial condition methods
 
 ```@docs
-init_state_conservative!
+init_state_prognostic!
 init_state_auxiliary!
 ```
 
--- a/experiments/OceanBoxGCM/homogeneous_box.jl
+++ b/experiments/OceanBoxGCM/homogeneous_box.jl
@@ -7,7 +7,7 @@ using ClimateMachine.ODESolvers
 using ClimateMachine.Mesh.Filters
 using ClimateMachine.VariableTemplates
 using ClimateMachine.Mesh.Grids: polynomialorder
-using ClimateMachine.BalanceLaws: vars_state_conservative
+using ClimateMachine.BalanceLaws: vars_state, Prognostic
 using ClimateMachine.Ocean.HydrostaticBoussinesq
 
 using Test
@@ -91,11 +91,11 @@ function run_homogeneous_box(; imex::Bool = false, BC = nothing)
 
     result = ClimateMachine.invoke!(solver_config)
 
-    maxQ = Vars{vars_state_conservative(driver_config.bl, FT)}(maximum(
+    maxQ = Vars{vars_state(driver_config.bl, Prognostic(), FT)}(maximum(
         solver_config.Q,
         dims = (1, 3),
     ))
-    minQ = Vars{vars_state_conservative(driver_config.bl, FT)}(minimum(
+    minQ = Vars{vars_state(driver_config.bl, Prognostic(), FT)}(minimum(
         solver_config.Q,
         dims = (1, 3),
     ))
--- a/src/BalanceLaws/interface.jl
+++ b/src/BalanceLaws/interface.jl
@@ -16,69 +16,15 @@ Subtypes `L` should define the methods below
 abstract type BalanceLaw end # PDE part
 
 """
-    vars_state_conservative(::L, FT)
+    vars_state(::L, ::AbstractStateType, FT)
 
 a tuple of symbols containing the state variables
 given a float type `FT`.
 """
-function vars_state_conservative end
+function vars_state end
 
-"""
-    vars_state_auxiliary(::L, FT)
-
-a tuple of symbols containing the auxiliary variables
-given a float type `FT`.
-"""
-function vars_state_auxiliary end
-
-"""
-    vars_state_gradient(::L, FT)
-
-a tuple of symbols containing the transformed variables
-of which gradients are computed given a float type `FT`.
-"""
-function vars_state_gradient end
-
-"""
-    vars_state_gradient_flux(::L, FT)
-
-a tuple of symbols containing the diffusive variables
-given a float type `FT`.
-"""
-function vars_state_gradient_flux end
-
-"""
-    vars_gradient_laplacian(::L, FT)
-
-a tuple of symbols containing the transformed variables
-of which gradients of laplacian are computed, they must
-be a subset of `vars_state_gradient`, given a float type `FT`.
-"""
-vars_gradient_laplacian(::BalanceLaw, FT) = @vars()
-
-"""
-    vars_hyperdiffusive(::L, FT)
-
-a tuple of symbols containing the hyperdiffusive variables
-given a float type `FT`.
-"""
-vars_hyperdiffusive(::BalanceLaw, FT) = @vars()
-
-"""
-    vars_integrals(::L, FT)
-
-a tuple of symbols containing variables to be integrated
-along a vertical stack, given a float type `FT`.
-"""
-vars_integrals(::BalanceLaw, FT) = @vars()
-
-"""
-    vars_reverse_integrals(::L, FT)
-
-a tuple of symbols containing variables to be integrated
-along a vertical stack, in reverse, given a float type `FT`.
-"""
-vars_reverse_integrals(::BalanceLaw, FT) = @vars()
+# Fallback: no variables
+vars_state(::BalanceLaw, ::AbstractStateType, FT) = @vars()
 
 """
     init_state_conservative!(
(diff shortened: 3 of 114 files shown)


blallen:
Maybe rename

```julia
struct VerticalIntegrals <: AbstractStateType end
struct ReverseVerticalIntegrals <: AbstractStateType end
```

to 

```julia
struct UpwardIntegrals <: AbstractStateType end
struct DownwardIntegrals <: AbstractStateType end
```

in order to have more descriptive names? You only know which direction an integral is going based on looking at the kernels.

OTOH, we maybe want to rewrite the integral kernels so you just specify the direction for each variable and don't require a corresponding `VerticalIntegral` variable for each `ReverseVerticalIntegral` variable.

More comments coming later.

experiments/OceanBoxGCM/homogeneous_box.jl
```diff
@@ -91,11 +91,11 @@ function run_homogeneous_box(; imex::Bool = false, BC = nothing)
 
     result = ClimateMachine.invoke!(solver_config)
 
-    maxQ = Vars{vars_state_conservative(driver_config.bl, FT)}(maximum(
+    maxQ = Vars{vars_state(driver_config.bl, Conservative(), FT)}(maximum(
```
blallen, line 94: why `Conservative()` and not `::Conservative`? What's the difference?

charleskawczynski: `Conservative()` is just a singleton. The function calls need `Conservative()` and method definitions need `::Conservative`. Thanks for catching this!

blallen:
Why do we need to pass `FT` as an argument to `vars_state()`? Isn't `@vars` just a short cut to various views/elements of an array? Why does it need to know the type of the elements in the array?

On a similar note, could we make `vars_state` a callable object? That way we could just store `FT` in the struct and not need to pass it to every function call.

experiments/OceanBoxGCM/homogeneous_box.jl
charleskawczynski: @kpamnany Is one preferred over the other? I think these should all inline anyway

charleskawczynski:
> Why do we need to pass `FT` as an argument to `vars_state()`? Isn't `@vars` just a short cut to various views/elements of an array? Why does it need to know the type of the elements in the array?

Variable templates needs to know the type (I think it may be needed for `MArray`s) and rank of each variable.

> On a similar note, could we make `vars_state` a callable object? That way we could just store `FT` in the struct and not need to pass it to every function call.

🤷‍♂️ I think then we'd have to pass `vs = vars_state()` through the call graph, which doesn't seem desirable. @simonbyrne ?

charleskawczynski:
> `struct UpwardIntegrals <: AbstractStateType end`, `struct DownwardIntegrals <: AbstractStateType end`

❤️

charleskawczynski:
> OTOH, we maybe want to rewrite the integral kernels so you just specify the direction for each variable and don't require a corresponding VerticalIntegral variable for each ReverseVerticalIntegral variable.

We could make another supertype :)

blallen:
> > OTOH, we maybe want to rewrite the integral kernels so you just specify the direction for each variable and don't require a corresponding VerticalIntegral variable for each ReverseVerticalIntegral variable.
> 
> We could make another supertype :)

Sure, but how `kernel_reverse_indefinite_stack_integral!` works is that it literally flips the sign on an already calculated integral. 

`kernel_indefinite_stack_integral!` actually does `\int_z1^z (integrand)` but `kernel_indefinite_stack_integral!` does \int_z^z2 (integrand) - \int_z1^z2 (integrand)`. After typing that, I realize links to actual code might be better:

https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel_kernels.jl#L1723-L1730
versus
https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel_kernels.jl#L1808-L1841

They really aren't equivalent operations at this point.

charleskawczynski:
> They really aren't equivalent operations at this point.

Ah, I misunderstood your original statement:

> maybe want to rewrite the integral kernels so you just specify the direction for each variable and don't require a corresponding VerticalIntegral variable for each ReverseVerticalIntegral variable.

Yes, that seems reasonable. I imagine we could optionally pack this information into `VerticalIntegral`/`ReverseVerticalIntegral`

charleskawczynski:
bors try

experiments/OceanBoxGCM/homogeneous_box.jl
kpamnany: I don't know if one is preferred but would like to! @vchuravy: thoughts please?

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
bors try

src/BalanceLaws/interface.jl
```diff
@@ -316,18 +262,11 @@ Compute reverse indefinite integral along stack.
...
-    varsize(vars_gradient_laplacian(m, FT))
-num_hyperdiffusive(m::BalanceLaw, FT) = varsize(vars_hyperdiffusive(m, FT))
-num_integrals(m::BalanceLaw, FT) = varsize(vars_integrals(m, FT))
-num_reverse_integrals(m::BalanceLaw, FT) =
-    varsize(vars_reverse_integrals(m, FT))
+number_states(m::BalanceLaw, st::AbstractStateType, FT) =
+    varsize(vars_state(m, st, FT))
+
+rank_multiplier(st::AbstractStateType) = 1
+rank_multiplier(st::GradientLaplacian) = 3
```
blallen, line 269: We might want to expand this to include `::Gradient` and maybe `::GradientFlux`

charleskawczynski: `Gradient` and `GradientFlux` are subtypes of `AbstractStateType`, so these methods provide fallbacks. Maybe I’m misunderstanding. Can you give an example?

charleskawczynski: Or maybe this is just a bad name. I think I wasn't sure what to call this function, actually. It's only used to account for this factor of 3 [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/create_states.jl#L139). The other methods (for [Hyperdiffusive](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/create_states.jl#L156), [GradientFlux](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/create_states.jl#L106), [Auxiliary](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/create_states.jl#L49), and [Conservative](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/create_states.jl#L19)) all have coefficients 1. Is this a good name? or do you have a better idea for a name here?

blallen: Is the rank multiplier the change from one component scalars to three component gradients? I just meant to include more things in the 3 category.

charleskawczynski: > Is the rank multiplier the change from one component scalars to three component gradients?

🤷‍♂️ I was just generalizing `create_conservative_state`, `create_auxiliary_state`, etc.. Maybe @jkozdon or @lcw can comment? Using

`rank_multiplier(::Gradient) = 3`
`rank_multiplier(::GradientFlux) = 3`

would be behavior changing, and I'm not sure it would work / be correct.

jkozdon: I personally don't like this. Can we just leave the 3 where its needed. This is going to obfuscate what's happening.

src/BalanceLaws/interface.jl
```diff
@@ -316,18 +262,11 @@ Compute reverse indefinite integral along stack.
...
-number_state_gradient_flux(m::BalanceLaw, FT) =
-    varsize(vars_state_gradient_flux(m, FT))
-num_gradient_laplacian(m::BalanceLaw, FT) =
-    varsize(vars_gradient_laplacian(m, FT))
-num_hyperdiffusive(m::BalanceLaw, FT) = varsize(vars_hyperdiffusive(m, FT))
-num_integrals(m::BalanceLaw, FT) = varsize(vars_integrals(m, FT))
-num_reverse_integrals(m::BalanceLaw, FT) =
-    varsize(vars_reverse_integrals(m, FT))
+number_states(m::BalanceLaw, st::AbstractStateType, FT) =
+    varsize(vars_state(m, st, FT))
```
jkozdon, line 266:
```suggestion
number_states(m::BalanceLaw, st::AbstractStateType, FT=Int) =
    varsize(vars_state(m, st, FT))
```
The type doesn't matter here. So this will allow `number_states(m, st)` to work.

docs/src/APIs/BalanceLaws/BalanceLaws.md
```diff
@@ -10,15 +10,23 @@ CurrentModule = ClimateMachine.BalanceLaws
 BalanceLaw
 ```
 
+## State variable types
+
+```@docs
+Conservative
```
jkozdon, line 16: I've mentioned this before, but what we call the "conservative state" isn't really conservative. Its really just the state or field or something like that.

Not sure what a better name would be.

- `Field`
- `State`
- ????

charleskawczynski: Isn't this for the conserved variables, though? Maybe `Prognostic`?

jkozdon: We have sources and sinks, so it's not conserved.

blallen: I think conservative means different things in applied math versus GFD?

I'm fine with `Prognostic` instead. I mainly needed something to distinguish it from the other state vectors we have so chose conservative

charleskawczynski: I see. Hmm... we're aiming for these variables to be conserved. I'd like to call it `Conserved`, but I agree that it's not strictly conserved (since `source!` is now potentially a function of gradients) so it's probably better to avoid the misnomer. I think `Prognostic` is a good name, I'll change it.

src/BalanceLaws/interface.jl
charleskawczynski: > I personally don't like this. Can we just leave the 3 where its needed. This is going to obfuscate what's happening.

So, specialize on `create_state(:: GradientLaplacian)` (have a method for `AbstractStateType` and `GradientLaplacian` methods) and hard-code the 3 in the `GradientLaplacian` method?

jkozdon: Sure.

It just seems a bit opaque to have this radom `rank_multiplier` function elsewhere.

charleskawczynski: `rank_multiplier` is only needed for `create_state`, we could just move it right above `create_state` instead

src/Numerics/DGMethods/create_states.jl
```diff
@@ -1,22 +1,24 @@
...
-    Np = dofs_per_element(grid)
     DA = arraytype(grid)
 
     weights = view(h_vgeo, :, grid.Mid, :)
     weights = reshape(weights, size(weights, 1), 1, size(weights, 2))
 
-    V = vars_state_conservative(balance_law, FT)
-    state_conservative = MPIStateArray{FT, V}(
+    # TODO: Clean up this MPIStateArray interface...
+    ns = number_states(balance_law, st, FT)
```
jkozdon, line 15:
```suggestion
    ns = number_states(balance_law, st, FT)
    st isa GradientLaplacian && (ns = 3ns)
```
```suggestion
    ns = number_states(balance_law, st, FT)
```

PR #1293, opened 2020-06-25
Add scaled-increment interface to main DG evaluation

simonbyrne:
# Description

Adds a 6-argument form `f(du, u, param, t, α, β)` to the DG evaluation that computes

    du .= α .* f(u, param, t) .+ β .* du

which will be used by the new timestepping interface.

See https://github.com/SciML/DifferentialEquations.jl/issues/615

I have

### DIFF
--- a/src/Numerics/DGMethods/DGModel.jl
+++ b/src/Numerics/DGMethods/DGModel.jl
@@ -46,13 +46,31 @@ end
 # Include the remainder model for composing DG models and balance laws
 include("remainder.jl")
 
+"""
+    (dg::DGModel)(tendency, state_conservative, nothing, t, α, β)
+
+Computes the tendency terms compatible with `IncrementODEProblem`
+
+    tendency .= α .* dQdt(state_conservative, p, t) .+ β .* tendency
+
+The 4-argument form will just compute
+
+    tendency .= dQdt(state_conservative, p, t) 
+
+"""
 function (dg::DGModel)(
     tendency,
     state_conservative,
-    ::Nothing,
+    param,
     t;
     increment = false,
 )
+    # TODO deprecate increment argument
+    dg(tendency, state_conservative, param, t, true, increment)
+end
+
+function (dg::DGModel)(tendency, state_conservative, _, t, α, β)
+
 
     balance_law = dg.balance_law
     device = array_device(state_conservative)
@@ -87,8 +105,10 @@ function (dg::DGModel)(
     ndrange_interior_surface = Nfp * length(grid.interiorelems)
     ndrange_exterior_surface = Nfp * length(grid.exteriorelems)
 
-    if !increment && num_state_conservative < num_state_tendency
-        tendency .= -zero(FT)
+    if num_state_conservative < num_state_tendency && β != 1
+        # if we don't operate on the full state, then we need to scale here instead of volume_tendency!
+        tendency .*= β
+        β = β != 0 # if β==0 then we can avoid the memory load in volume_tendency!  
     end
 
     communicate =
@@ -412,7 +432,8 @@ function (dg::DGModel)(
         grid.ω,
         grid.D,
         topology.realelems,
-        increment;
+        α,
+        β;
         ndrange = (nrealelem * Nq, Nq),
         dependencies = (comp_stream,),
     )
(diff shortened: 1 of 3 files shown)


simonbyrne:
@lcw @jkozdon I'd appreciate your thoughts on how this might affect the kernels.

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -87,10 +105,6 @@ function (dg::DGModel)(
     ndrange_interior_surface = Nfp * length(grid.interiorelems)
     ndrange_exterior_surface = Nfp * length(grid.exteriorelems)
 
-    if !increment && num_state_conservative < num_state_tendency
-        tendency .= -zero(FT)
-    end
-
```
jkozdon, line 93: We might still need something like this for when we mask out the tracers. @lcw?

src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -302,7 +302,12 @@ Computational kernel: Evaluate the volume integrals on right-hand side of a
         @unroll for k in 1:Nqk
             ijk = i + Nq * ((j - 1) + Nq * (k - 1))
             @unroll for s in 1:num_state_conservative
-                tendency[ijk, s, e] = local_tendency[k, s]
+                if β != 0
+                    T = α * local_tendency[k, s] + β * tendency[ijk, s, e]
```
jkozdon, line 306: We may want to prefetch tendency earlier, but you'd have to time it to see. Probably a minor performance thing anyway.

src/Numerics/DGMethods/DGModel.jl
simonbyrne: yup, I just realised what that was for.

src/Numerics/DGMethods/DGModel_kernels.jl
simonbyrne: That was my main worry.

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -87,8 +105,10 @@ function (dg::DGModel)(
     ndrange_interior_surface = Nfp * length(grid.interiorelems)
     ndrange_exterior_surface = Nfp * length(grid.exteriorelems)
 
-    if !increment && num_state_conservative < num_state_tendency
-        tendency .= -zero(FT)
+    if num_state_conservative < num_state_tendency && β != 1
+        # if we don't operate on the full state, then we need to scale here instead of volume_tendency!
+        tendency .*= β
+        β = true
```
jkozdon, line 111: Why true and not
```suggestion
        β = 1
```

src/Numerics/DGMethods/DGModel_kernels.jl
jkozdon: I would guess it is pretty minor. Probably in the noise of all the other kernels.

jkozdon: If you did want to prefetch, you could add to the top
```julia
local_tendency[k, s] =  β != 0 ? tendency[ijk, s, e] : -zero(FT)
```
then add later (before the first use)
```julia
local_tendency[k, s] =  β * local_tendency[k, s]
```

jkozdon: > If you did want to prefetch

Nevermind, that doesn't work...

jkozdon: > Nevermind, that doesn't work...

Because then you'd be scaling the current tendency by `α * β` so I guess really you'd want
```julia
local_tendency[k, s] =  (β / α) * local_tendency[k, s]
```

src/Numerics/DGMethods/DGModel.jl
simonbyrne: `false` is a "strong zero" such that `NaN*false == 0`: the compiler is able to elide it `x*false === 0.0`, which is why i used `Bool`s in the 4-arg form.

`true` doesn't really matter (since `x*1.0 === x`), but we might as well do the same.

jkozdon: Makes sense.

src/Numerics/DGMethods/DGModel_kernels.jl
simonbyrne: that should be fine, since I don't think anyone is going to run this with `α=0`. 

The other option would be to scale by `α` at every addition to `local_tendency[k, s]`.

simonbyrne:
> If we do the scaling this way we might be able to avoid a write in the LSRK kernels(?)

That's coming soon!

PR #1294, opened 2020-06-25
Add `reduce_[nodal|element]_stack()` in `SingleStackUtils`

### LINKED ISSUE #1175: Add support to call functions (`min`, `max`, `sum`) along stack `up_stack(fun, ...)`
# Description

We'll need to be able to compute things like `max(ϕ)` where `ϕ` is a variable along the vertical stack. So, we should implement some function (`up_stack(fun, ...)`, `down_stack(fun, ...)`) to apply along the vertical stack. It would also be nice to return the `z` location of such `min`/`max`.

# Additional context

# For CLIMA Developers

- [x] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [x] There is a timeline for when this feature can be implemented

kpamnany:
# Description

And a couple of tests. Closes #1175.

I have

### DIFF
--- a/src/Utilities/SingleStackUtils/SingleStackUtils.jl
+++ b/src/Utilities/SingleStackUtils/SingleStackUtils.jl
@@ -1,7 +1,11 @@
 module SingleStackUtils
 
-export get_vars_from_nodal_stack, get_vars_from_element_stack
-export get_horizontal_variance, get_horizontal_mean
+export get_vars_from_nodal_stack,
+    get_vars_from_element_stack,
+    get_horizontal_variance,
+    get_horizontal_mean,
+    reduce_nodal_stack,
+    reduce_element_stack
 
 using OrderedCollections
 using StaticArrays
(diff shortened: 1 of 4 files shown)


src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -204,4 +207,58 @@ function get_horizontal_variance(
...
+        var::String;
+        vrange::UnitRange = 1:size(Q, 3),
+    ) where {T, dim, N}
+
+Reduce `var` from `vars` within `Q` over all nodal points in the specified
+`vrange` of elements with `op`. Return a tuple `(result, z)` where `result` is
+the final value returned by `op` and `z` is the index within `vrange` where the
+`result` was determined.
+"""
+function reduce_element_stack(
```
charleskawczynski, line 225: I think we need this to be `reduce_nodal_stack`, we can probably write two methods and use the same format as `get_vars_from_nodal_stack`/`get_vars_from_element_stack`. Aside from that, the behavior looks good.

ilopezgp: Does `op` in the description correspond to `f::Function` in the function handle of the description?

charleskawczynski: Yeah, that doc string should be fixed. Good catch.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -204,4 +208,108 @@ function get_horizontal_variance(
...
+    op::Function,
+    grid::DiscontinuousSpectralElementGrid{T, dim, N},
+    Q::MPIStateArray,
+    vars::Type,
+    var::String;
+    vrange::UnitRange = 1:size(Q, 3),
+) where {T, dim, N}
+    Nq = N + 1
+    nodal_stacks = [
```
charleskawczynski, line 286: I think we just need
```julia
    return [
```

This way the user can decide how to parse the nodal results.

PR #1298, opened 2020-06-27
Move NumericalFluxes up a level

jkozdon:
Since these will be useful for future entropy stable discretizations it
makes sense for them to be separate from the DGMethods module.


jkozdon:
bors try

jkozdon:
bors try

jkozdon:
@thomasgibson said we are trying to freeze anything that impacts the GCM dycore functionality for the time being, so I will close this for now.

PR #1299, opened 2020-06-28
sync cb_ctk with #1224

christophernhill:
# Description

A one line change to src/Driver/Driver.jl that fixes call to cb_vtk. It looks like this fix was missed in #1224 

I have

- [ x] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ x] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ x] Updated the documentation to reflect changes from this PR.

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -647,7 +647,8 @@ Starting %s
     end
 
     # fini VTK
-    !isnothing(cb_vtk) && cb_vtk()
+    !isnothing(cb_vtk) &&
+    GenericCallbacks.call!(cb_vtk, nothing, nothing, nothing, nothing)
 
     engf = norm(Q)
     @info @sprintf(


christophernhill:
@simonbyrne I think this Driver.jl change maybe got overlooked in #1224 changes?  I came across it when uncommenting the lines
```
# vtk_interval = ceil(Int64, timeout / solver_config.dt)
# ClimateMachine.Settings.vtk = "$(vtk_interval)steps"
```
in https://github.com/CliMA/ClimateMachine.jl/blob/978bc60a5c9d03a240967e7e04733038010bec50/experiments/OceanBoxGCM/ocean_gyre.jl#L91 

I think this also means there isn't a test that exercises direct vtk calls any more? Otherwise I think this would have shown up in CI?

christophernhill:
Hi @simonbyrne any thoughts on this?

Without it ```--vtk {NN}steps``` seems to give

```
ERROR: LoadError: MethodError: objects of type EveryXSimulationSteps are not callable
```

at the end of a run. 

@jm-c may also have noticed some possibly slight non-deterministic behavior about step numbering and which fields are written to files when using MPI and --vtk.

christophernhill:
bors try

kpamnany:
Looks like a missed change to me, good catch.

> @jm-c may also have noticed some possibly slight non-deterministic behavior about step numbering and which fields are written to files when using MPI and --vtk.

@jm-c: can you please open an issue to expand on this?

jm-c:
@kpamnany  I opened an issue #1310 ; checking more carefully, vtk output things are reproduceable (just some StateCheck label
added some confusion for me).

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
@blallen and/or @kpamnany I think this should be good to bors r+ when someone feels the urge!

blallen:
> ## try
> Build failed:
> 
> * [ci/slurmci](https://gist.github.com/ede12ec63da91a3815995880d304e3bc)

@christophernhill @kpamnany, is this a real issue or just a temporary failure?

kpamnany:
😲 That doesn't look transient to me. It passes on the GPU but not on the CPU? @jkozdon any idea what's up?

bors r+

christophernhill:
@blallen hmm - that looks to be a failure in remainder model CI test. Seems odd that it would be related to this. 

@jkozdon this PR changed Driver.jl . 
Before this, code using vtk with master (since #1224 was merged) would fail in Driver.jl after all stepping but before fully exiting etc...
With this PR the vtk call error no longer happens, so things return from Driver.jl - possibly that then uncovers something in Test? Not sure that is possible, but maybe?

PR #1305, opened 2020-06-29
Use generic functions instead of MPI specific ones

jkozdon:
This allows the use of non-MPI types in ClimateMachine (such as `DoubleFloat`)

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -523,7 +523,7 @@ function LinearAlgebra.norm(
         locnorm = norm_impl(Q.realdata, Val(p), dims)
     end
 
-    mpiop = isfinite(p) ? MPI.SUM : MPI.MAX
+    mpiop = isfinite(p) ? (+) : max
     if locnorm isa AbstractArray
         locnorm = convert(Array, locnorm)
     end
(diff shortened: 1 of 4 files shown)


src/Arrays/MPIStateArrays.jl
```diff
@@ -523,7 +523,7 @@ function LinearAlgebra.norm(
         locnorm = norm_impl(Q.realdata, Val(p), dims)
     end
 
-    mpiop = isfinite(p) ? MPI.SUM : MPI.MAX
+    mpiop = isfinite(p) ? (+) : max
```
vchuravy, line 526: This might be problematic on platforms that don't support generic ops. I think Cray MPI and PowerPC

jkozdon:
On those platforms will this work for the standard types? Or do we have to use MPI.SUM and MPI.MAX here? This is really more for testing purposes, than real performance runs.

jkozdon:
bors try

src/Arrays/MPIStateArrays.jl
jkozdon: On those platforms will this work for the standard types? Or do we have to use MPI.SUM and MPI.MAX here? This is really more for testing purposes, than real performance runs.

simonbyrne: No, this is fine. It will use MPI.SUM and MPI.MAX for MPI-defined types.

Issue #1306, opened 2020-06-29
Atmospheric GCM Dycore Debugging


thomasgibson:
# Description

I am logging this issue to keep track of the progress on issues relating to the functionality of the GCM dycore. We should use this issue to centralize discussion on what the current problems are and identifying solutions.

Problems that people discover should be updated here. Related PRs that are focused on benchmarks/tests for the GCM dycore include: #1205, #1193

## Known issues

- Discrete balancing issues: Hydrostatic balance issues detailed in #1247, and a problem has been identified in the geopotential gradient (proposed fix in #1292)

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

thomasgibson:
@bischtob If you can add some of the VizClima plots you've saved and/or add more to the `Known Issues` section, please drop some stuff in. I will update this issue gradually as well.

thomasgibson:
Update: the following run was performed using the domain setup:
```
┌ Info: Establishing Atmos GCM configuration for BaroclinicWave
│     precision        = Float64
│     polynomial order = 4
│     #horiz elems     = 20
│     #vert elems      = 5
│     domain height    = 3.00e+04 m
│     MPI ranks        = 1
│     min(Δ_horz)      = 61096.87 m
└     min(Δ_vert)      = 1036.04 m
```
and with an IMEX ARK solver `ARK2GiraldoKellyConstantinescu`. Key things about this run are the inclusion of the turbulence closure `SmagorinskyLilly{Float64}(0.21)`, no hyperdiffusion, and the `CuttoffFilter`. The simulation remains stable until after 6 days of simulation time. Lat-lon and lat-alt plots are shown at 561262.432 simulation seconds (about 6.5 days). There are the last visualizations I have before the run getting killed due to `NaN`s.
[latlon-6p5d.pdf](https://github.com/CliMA/ClimateMachine.jl/files/4858865/latlon-6p5d.pdf)
[latalt-6p5d.pdf](https://github.com/CliMA/ClimateMachine.jl/files/4858867/latalt-6p5d.pdf)

There appears to be noise in the vertical component of velocity that I suspect may be causing our problems. We should probably take a closer look at the SGS turbulence models.

smarras79:
[...]

PR #1316, opened 2020-07-02
As/baroclinic box

akshaysridhar:
# Description

Adds an experiment - channel-flow : Baroclinic-wave LES test case. 

Linked to issue #1306. 

I have


experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+        driver_config.name,
+        interpol = interpol,
+    )
+    return ClimateMachine.DiagnosticsConfiguration([
+        default_dgngrp,
+    ])
+end
+
+function main()
+    FT = Float32
```
smarras79, line 243: We saw that conservation at Float32 is almost completely destroyed. 
Set this up for Float64 as default.

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+    # Prescribe domain parameters
+    xmax = FT(4e7) 
+    ymax = FT(6e6)
+    zmax = FT(30e3)
+
+    t0 = FT(0)
+
+    # For a full-run, please set the timeend to 3600*6 seconds
+    # For the test we set this to == 30 minutes
+    days = FT(86400)
```
smarras79, line 263: replace `days` with `seconds` or it will create confusion.

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
+using ClimateMachine
+ClimateMachine.init()
```
kpamnany, line 2:
```suggestion
ClimateMachine.init(parse_clargs = true)
```

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+using DocStringExtensions
+using LinearAlgebra
+using Logging
+
+"""
+    baroclinic_instability_cube(...)
+Initialisation helper for baroclinic-wave (channel flow) test case for iterative
+determination of η = p/pₛ coordinate for given z-altitude. 
+"""
+
```
kpamnany, line 38:
```suggestion
```

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+    ics = init_baroclinicwave!     # Initial conditions
+    C_smag = FT(0.30)     # Smagorinsky coefficient
+    
+    # Assemble source components
+    source = (
+        Gravity(),
+    )
+
+    # Choose default IMEX solver
+    ode_solver_type = ClimateMachine.IMEXSolverType()
```
smarras79, line 170: Your current run shows that single-rate LSRK144 is more stable than IMEX and is still pretty fast. Let's use either single- or multi-rate LSRK144 by default.

experiments/AtmosLES/baroclinicwave.jl
akshaysridhar: days should be available from the parameter set, I can remove this.

experiments/AtmosLES/baroclinicwave.jl
akshaysridhar: Right : Currently sweeping through the timestepper options to check. There seems to be some vtk output issue ...

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+
+    # Assemble model components
+    model = AtmosModel{FT}(
+        AtmosLESConfigType,
+        param_set;
+        turbulence = SmagorinskyLilly{FT}(C_smag),
```
smarras79, line 176: Although it's ok to use Smagorinsky for the sake of testing  LES, eventually this should be replaced by a 4th or 6th order (maybe higher?) hyper-diffusion for numerical stability purposes.

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+        interpol = interpol,
+    )
+    aux_dgngrp = setup_dump_aux_diagnostics(
+        AtmosLESConfigType(),
+        "1shours",
+        driver_config.name,
+        interpol = interpol,
+    )
+    return ClimateMachine.DiagnosticsConfiguration([
+        default_dgngrp,
```
kpamnany, line 238: You're setting up four diagnostics groups but only adding this one to the configuration?

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+    ])
+end
+
+function main()
+    FT = Float32
+
+    # DG polynomial order
+    N = 4
```
smarras79, line 246: order 3 should be enough and reduce the Courant constraint a bit. The accuracy won't be majorly affected but the solution time may. Worth a try although not a necessary change.

akshaysridhar: Noted

experiments/AtmosLES/baroclinicwave.jl
akshaysridhar: Noted. What does this do ?

experiments/AtmosLES/baroclinicwave.jl
```diff
@@ -0,0 +1,332 @@
...
+    # Domain resolution and size
+    Δx = FT(1e5)
```
smarras79, line 248: For a faster run in testing mode, Dx could be increased to 2e5. Based on the test shown with numa in Marras, Nazarov, Giraldo 2015, this problem should stay stable for up to Dx = 4e5. Worth testing.

experiments/AtmosLES/baroclinicwave.jl
akshaysridhar: For now yes; even this single diagnostic group isn't written out  - so I wanted to just place 1 of these in the configuration while testing.

PR #1315, opened 2020-07-02
Add VerticalIntegralModel for integrating flow and tendency

blallen:
# Description

This PR adds a VerticalIntegralModel that can be used to integrate both the flow field and flow tendency for use in the split explicit model in #1268. This model is tested for correctness using the 2D and 3D hydrostatic spindown initial states.

I have

- [x] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [x] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [x] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [x] CLIMA developers with relevant expertise have been assigned to review this submission
- [x] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [x] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1318, opened 2020-07-02
Remove some duplicate test cases from Ocean CI for faster turn around

blallen:
# Description

I noticed that the Ocean tests were taking the longest, so I removed a few cases that our new tests cover.

I have


blallen:
@jakebolewski do you know how heavy the Ocean tests are for slurm CI? We could potentially remove a few test cases there as well.

jakebolewski:
@blallen the longest ocean tests on slurm are:

```
scripts/test-gpu.sh test/Ocean/ShallowWater/GyreDriver.jl  00:20:54
scripts/test-cpu.sh test/Ocean/ShallowWater/GyreDriver.jl 00:12:47
scripts/test-gpu.sh experiments/OceanBoxGCM/homogeneous_box.jl  00:11:19
```

but these can be run in parallel and are not serialized (like on github actions), so the overall runtime is not as consequential for an individual test case.

jakebolewski:
This doesn't seem to materially reduce the ocean runtime though (compared to past runs?).

blallen:
> This doesn't seem to materially reduce the ocean runtime though (compared to past runs ~40 minutes? ).

The most recent attempt on a different PR took 54 minutes, so we are saving some time: https://github.com/CliMA/ClimateMachine.jl/pull/1315/checks?check_run_id=831686327

When I run locally, a lot of the time is spent compiling ClimateMachine for each test and the run times are rather short, no more than 5 mins at the extreme on my 2015 MBP.

jakebolewski:
I see, I referenced the time when I submitted the switch to GitHub actions PR :) but more tests might have been added since then.  Github actions also runs on shared virtualized hardware and I've noticed some longer tailed runtimes.  Another strategy would be to break the tests up into sub-modules (like Numerics) and run them as separate jobs.

However if the majority of the time is being spent in compilation overhead then that is something that should be addressed somehow, is it really the case that the majority of the time is spent in pre-compilation overhead?

jakebolewski:
Well the times do seem to decrease substantially (times of tests in order):

Running with per-compilation should / might help, although we would need a way to re-generate a per-compilation file.  Maybe just per-compiling our project dependencies and re-triggering that build on manifest changes would help for most cases.

```
/Ocean/HydrostaticBoussinesq/3D_hydrostatic_spindown.jl 16 minutes
/Ocean/HydrostaticBoussinesq/test_ocean_gyre.jl  10 minutes
/Ocean/ShallowWater/2D_hydrostatic_spindown.jl 2 minutes
/Ocean/SplitExplicit/hydrostatic_spindown.jl 3.75 minutes
```

blallen:
> Well the times do seem to decrease substantially (times of tests in order):
> 
> Running with pre-compilation should / might help, although we would need a way to re-generate a pre-compilation file. Maybe just pre-compiling our project dependencies and re-triggering that build on manifest changes would help for most cases.
> 
> ```
> /Ocean/HydrostaticBoussinesq/3D_hydrostatic_spindown.jl 16 minutes
> /Ocean/HydrostaticBoussinesq/test_ocean_gyre.jl  10 minutes
> /Ocean/ShallowWater/2D_hydrostatic_spindown.jl 2 minutes
> /Ocean/SplitExplicit/hydrostatic_spindown.jl 3.75 minutes
> ```

Huh, I did not expect 3D hydrostatic spindown to take the longest. I guess it makes sense though

```
test           : Domain Size : Timestep : Simulation Time
3D_spindown    :       5x5x8 :      120 :           86400
split_spindown :       5x5x8 :      300 :           86400
ocean_gyre     :       4x4x5 :      150 :           86400
div_free       :       4x4x4 :      600 :           36000
```

I expected it to be similar to the split spindown, but we have a 2.5x smaller timestep and we run it twice (explicit, imex). I guess I can try bumping the timestep on the 3D spindown to see if a larger timestep is still stable. I can also thinking about how to shorten the ocean gyre as well.

jakebolewski:
@blallen it might be good to merge this as we are hitting some (infrequent) timeouts.

blallen:
> @blallen it might be good to merge this as we are hitting some (infrequent) timeouts.

okay, I'll do that. @christophernhill, can you think about how else we can shorter these tests?

PR #1321, opened 2020-07-02
More pedagogical computation of continuity equation

blallen:
# Description

@jm-c pointed out that our current method for computing the continuity equation `∇u = 0` is a little counterintuitive. Here is one possible change we can make to improve the clarity while maintaining computational efficiency at the cost of some extra memory usage.

I have


blallen:
bors try

blallen:
@jm-c and @christophernhill, take a look and let me know what you think. This method is kinda half-way between what is in @jm-c branch and in master.

christophernhill:
@blallen how does this map to

https://github.com/jm-c/CLIMA/blob/jmc/split_explicit_draft/src/Ocean/SplitExplicit01/Continuity3dModel.jl

should we get in sync using ```Continuity3dModel.jl``` before changing to another new way?

blallen:
> @blallen how does this map to
> 
> https://github.com/jm-c/CLIMA/blob/jmc/split_explicit_draft/src/Ocean/SplitExplicit01/Continuity3dModel.jl
> 
> should we get in sync using `Continuity3dModel.jl` before changing to another new way?

In terms of the math, the only difference is that between calculating a divergence directly and calculating it from the trace of the gradient tensor, e.g. none. 

Algorithmically, this uses more memory than JMC's way (altho I just realized I can make a fix to make it use the same or less (will push it soon)) but saves us the step of having to do another iteration across all the elements. Seeing as we currently only do two of those per time step, adding a third is an ~50% increase in compute time. 

The main advantage of JMC's method is that we can choose the numerical flux for the continuity equation separately from that of the rest of the gradients calculated. Currently, this just means we can do central or Rusanov fluxes with JMC's method, but only central with mine. In the future, this might be a bigger difference, but I also hope/expect that we will have more flexibility in specifying the numerical fluxes for gradients by the time we have the more complicated numerical fluxes.

christophernhill:
Can we mark it WIP for now? I think we want to be able to check reference
numbers for things we merge?

blallen:
> Can we mark it WIP for now? I think we want to be able to check reference numbers for things we merge?

This is marked as a draft PR :) But this change doesn't change any of the state values in the state check. I simply turned off the gradient check for now, as I am adding new gradient variables so the order changes and the state check fails. We can still check reference numbers for PRs on top of this one.

PR #1323, opened 2020-07-03
Revamp bc interface to remove `X_bottom1` from DG kernels

jkozdon:
The bottom `X_bottom1` arrays in the `DGModel_kernels.jl` are atmos specific, this makes the interface a bit more general purpose for future extensions.

### DIFF
--- a/src/Atmos/Model/bc_momentum.jl
+++ b/src/Atmos/Model/bc_momentum.jl
@@ -126,7 +126,7 @@ function atmos_momentum_boundary_state!(
     aux⁻,
     bctype,
     t,
-    args...,
+    bc_extra_data,
 ) where {DL <: DragLaw}
     atmos_momentum_boundary_state!(
         nf,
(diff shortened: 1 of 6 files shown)


jkozdon:
bors try

jkozdon:
bors try

jkozdon:
> I like it overall. Might be interesting to make a function for `bc_extra_data` and `face_movement_interior` that can dispatch on the balance law type in case anyone else wants to use this and to pass `nothing` for all models besides the `AtmosModel`.

I made it as minimal as possible and extendable as I could, figuring it can be easily added to as needed (hence the reason I made it a named tuple rather than a function callback or a type).

> Is the `state_conservative` in `bc_extra_data` the entire MPIStateArray? How does passing this as an argument affect the performance of models which don't need it?

There shouldn't be any performance impact if the function doesn't load any data from the array. This is certainly a power-user feature / hack. There are some possible performance improvements though since now data isn't being loaded all the time that's not needed. (Now whether Julia is removing the unused loaded automatically already I don't know.)

PR #1330, opened 2020-07-07
more stable ice microphysics test

trontrytel:
# Description

@charleskawczynski and @tapios I have found a couple of places where to guard ourselves from negative values in water tracers it makes sense for me to add something like `max(FT(0), tracer)`. Could you please take a look and tell me if thats what you had in mind on the standup meeting last week? 

I have


src/Common/Thermodynamics/relations.jl
```diff
@@ -845,7 +845,7 @@ function supersaturation(
 ) where {FT <: Real}
 
     q_sat::FT = q_vap_saturation_generic(param_set, T, ρ; phase = Liquid())
-    q_vap::FT = q.tot - q.liq - q.ice
+    q_vap::FT = max(FT(0), q.tot - q.liq - q.ice)
```
tapios, line 848: This comes up a number of times. We should probably make it a `Thermodynamics` function, which takes care of it being returned non-negative.

src/Common/Thermodynamics/relations.jl
```diff
@@ -908,7 +908,7 @@ saturation_excess(ts::ThermodynamicState) = saturation_excess(
 
 Condensate of the phase partition.
 """
-condensate(q::PhasePartition) = q.liq + q.ice
+condensate(q::PhasePartition) = max(0, q.liq) + max(0, q.ice)
```
tapios, line 911: Maybe better `max(0, q.liq + q.ice)`? (And perhaps also make it a `Thermodynamics` function.)

test/Atmos/Parameterizations/Microphysics/KM_ice.jl
```diff
@@ -178,13 +179,13 @@ function kinematic_model_nodal_update_auxiliary_state!(
         aux.e_pot = _grav * aux.z
         aux.e_int = aux.e_tot - aux.e_kin - aux.e_pot
         # supersaturation
-        q = PhasePartition(aux.q_tot, aux.q_liq, aux.q_ice)
+        q = PhasePartition(max(FT(0), aux.q_tot), max(FT(0), aux.q_liq), max(FT(0), aux.q_ice))
         aux.T = air_temperature(param_set, aux.e_int, q)
         ts_neq = TemperatureSHumNonEquil(param_set, aux.T, state.ρ, q)
-        # TODO: add super_saturation method in moist thermo
-        aux.S = max(0, aux.q_vap / q_vap_saturation(ts_neq) - FT(1)) * FT(100)
         aux.RH = relative_humidity(ts_neq) * FT(100)
```
tapios, line 185: Wouldn't it be safer to define RH as a fraction, rather than percentage?

test/Atmos/Parameterizations/Microphysics/KM_warm_rain.jl
```diff
@@ -116,10 +117,9 @@ function kinematic_model_nodal_update_auxiliary_state!(
         q = PhasePartition(aux.q_tot, aux.q_liq, aux.q_ice)
         aux.T = air_temperature(param_set, aux.e_int, q)
         ts_neq = TemperatureSHumNonEquil(param_set, aux.T, state.ρ, q)
-        # TODO: add super_saturation method in moist thermo
-        aux.S = max(0, aux.q_vap / q_vap_saturation(ts_neq) - FT(1)) * FT(100)
         aux.RH = relative_humidity(ts_neq) * FT(100)
```
tapios, line 120: Same as above: RH as fraction rather than percentage throughout?

jkozdon:
Just an FYI, that I don't think that the `FT` is necessary here for type stability since `max` for [promotion](https://github.com/JuliaLang/julia/blob/df8f8ae88793e265c14fd3eee140774fe523901d/base/promotion.jl#L358)
```julia
julia> max(0, -1.0)
0.0
``

**Comment by trontrytel** [comment]
Following on the Issue https://github.com/CliMA/ClimateMachine.jl/issues/1357 I think that `KM_ice.jl` is a good test to look at problems that come from time integration of various source terms. For now I limited the number of source terms to 2 (transition from water vapour to liquid water and ice). 

The test is a little like a raising thermal bubble, but the velocity is prescribed and designed to mimic a squall-line. I will work some more on visualising the problems but for now here is a quick look of velocity, humidity and my water tracers:

KM_ice_plots 001

I'm getting negative values because around the freezing level the tendency from the source terms is big. It is small everywhere else. One solution would be to reduce the time step even further. Another would be to switch to implicit timestepping, but as mentioned in Issue https://github.com/CliMA/ClimateMachine.jl/issues/1357 I don't know how to evaluate equilibrium supersaturation at the t+1 time level that is needed in my source terms. Another solution would be to artificially limit or correct the tendencies at the freezing level.

Once that problem is handled the next step would be to uncomment the rest of the source terms from that test and figure out how to ensure that the sum of all the sources will not result in negative values. 

The test to look at is in `test/Atmos/Parameterizations/Microphysics/KM_ice.jl`. I'm doing output both to vtk and netcdf. For netcdf output there is a corresponding plotting script in `VizCLIMA.jl` branch `aj/negative_plots` in `src/scripts/negative_values_test.jl`. The scalars to look at are `q_liq` and `q_ice`.

I definitely lack the expertise to solve it myself. I also know that there are already other people working on similar problems. Let me know how I could contribute or if you could help me: @jkozdon , @fxgiraldo , @mwarusz, @blallen

**Comment by trontrytel** [comment]
> Just an FYI, that I don't think that the `FT` is necessary here for type stability since `max` for [promotion](https://github.com/JuliaLang/julia/blob/df8f8ae88793e265c14fd3eee140774fe523901d/base/promotion.jl#L358)
> 
> ```julia
> julia> max(0, -1.0)
> 0.0
> ``
> ```

Yes, sorry. I'm a little superstitious and adding FT in too many places. Will fix it.

**Comment by jkozdon** [comment]
> Another would be to switch to implicit timestepping

Why would switching to implicit time stepping help? Are you just thinking that it would be easier to limit the source size in this case?

**Comment by trontrytel** [comment]
I'm trying to see if there is anything useful in this stale PR.

File: src/Common/Thermodynamics/relations.jl [hunk 5]
```diff
@@ -956,7 +956,7 @@ function liquid_fraction(
 ) where {FT <: Real}
     q_c = condensate(q)     # condensate specific humidity
     if has_condensate(q_c)
-        return q.liq / q_c
+        return max(0, q.liq) / q_c
```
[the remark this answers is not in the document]
trontrytel: I think it's ok - it has the `if has_condensate()` condition.

src/Common/Thermodynamics/relations.jl
```diff
@@ -904,7 +904,7 @@ function supersaturation(
 ) where {FT <: Real}
 
     q_sat::FT = q_vap_saturation_generic(param_set, T, ρ, Liquid())
-    q_vap::FT = q.tot - q.liq - q.ice
+    q_vap::FT = max(FT(0), q.tot - q.liq - q.ice)
```
charleskawczynski, line 907:
```suggestion
    q_vap = vapor_specific_humidity(q)
```
Then we can use `max(0, q.tot - q.liq - q.ice)` in `vapor_specific_humidity`.

src/Common/Thermodynamics/relations.jl
```diff
@@ -917,7 +917,7 @@ function supersaturation(
 ) where {FT <: Real}
 
     q_sat::FT = q_vap_saturation_generic(param_set, T, ρ, Ice())
-    q_vap::FT = q.tot - q.liq - q.ice
+    q_vap::FT = max(FT(0), q.tot - q.liq - q.ice)
```
charleskawczynski, line 920:
```suggestion
    q_vap = vapor_specific_humidity(q)
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -335,22 +337,24 @@ a constant timescale.
...
     q_sat::PhasePartition{FT},
-    q::PhasePartition{FT},
+    q::PhasePartition{FT};
+    slowdown_liq::FT = FT(1),
 ) where {FT <: Real}
 
     _τ_cond_evap::FT = τ_relax(liquid_param_set)
 
-    return (q_sat.liq - q.liq) / _τ_cond_evap
+    return (q_sat.liq - max(FT(0), q.liq)) / _τ_cond_evap / slowdown_liq
```
charleskawczynski, line 346:
```suggestion
    return (q_sat.liq - q.liq) / _τ_cond_evap / slowdown_liq
```
We can filter `q.liq` on the construction of `PhasePartition`

src/Atmos/Parameterizations/CloudPhysics/Microphysics.jl
```diff
@@ -335,22 +337,24 @@ a constant timescale.
...
     q_sat::PhasePartition{FT},
-    q::PhasePartition{FT},
+    q::PhasePartition{FT};
+    slowdown_ice::FT = FT(1),
 ) where {FT <: Real}
 
     _τ_sub_dep::FT = τ_relax(ice_param_set)
 
-    return (q_sat.ice - q.ice) / _τ_sub_dep
+    return (q_sat.ice - max(FT(0), q.ice)) / _τ_sub_dep / slowdown_ice
```
charleskawczynski, line 357:
```suggestion
    return (q_sat.ice - q.ice) / _τ_sub_dep / slowdown_ice
```

src/Common/Thermodynamics/relations.jl
trontrytel: I was doing exactly that. I'll have a separate PR with just those `relations.jl` fixes.

PR #1334, opened 2020-07-08
Validate `ClimateMachine.init()` keyword arguments

### LINKED ISSUE #1333: ClimateMachine.init() doesn't error if you misspell an optional argument
# Description

If you type `ClimateMachine.init(dsisable_gpu = true)`, the file will execute perfectly fine on the GPU without raising an error.  I imagine similar things happen for the other named arguments. It would be nice if it raised a warning or error instead of silently executing with the default options.

## Steps to reproduce

Make a typo in any named argument when using ClimateMachine.init().

## Expected behavior

Tell me that I passed a non-existent named argument.

## Additional context

# For CLIMA developers

kpamnany:
# Description

Closes #1333.

I have

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -133,7 +133,8 @@ end
 
 """
     parse_commandline(
-        defaults::Union{Nothing, Dict{Symbol,Any}) = nothing,
+        defaults::Dict{Symbol, Any},
+        global_defaults::Dict{Symbol, Any},
         custom_clargs::Union{Nothing, ArgParseSettings} = nothing,
     )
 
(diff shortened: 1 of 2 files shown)


src/Driver/Driver.jl
```diff
@@ -340,10 +335,24 @@ function init(;
...
+    # `Settings` contains global defaults
+    global_defaults = Dict{Symbol, Any}(
+        (n, getproperty(Settings, n)) for n in propertynames(Settings)
+    )
+
+    # keyword arguments must be applicable to `Settings`
     all_args = Dict{Symbol, Any}(keyword_args)
+    for kwarg in keys(all_args)
+        if get(global_defaults, kwarg, nothing) === nothing
+            @warn("Unrecognized keyword argument '$(string(kwarg))'; ignoring")
```
blallen, line 347: Are there any disadvantages to making this return an error instead of a warning?

leios, line 347: Maybe it's a good idea to error instead of warn? I don't really have a preference either way, but if users configure their simulation and get results when they actually wanted a different sim, that could be a problem.

blallen: Are there any disadvantages to making this return an error instead of a warning?

jakebolewski: I think an error would be better as well.  For custom CI arguments you can pass in the defaults with the argparse settings so I don't see a need to not strictly validate the arguments

kpamnany: I'm now making it throw an `ArgumentError` but am happy to do something else if anyone has a suggestion.

PR #1337, opened 2020-07-09
Allow equispaced output in VTK

jkozdon:
Previously only VTK output at the LGL nodes was supported, this allows interpolation to an equidistant grid of points which allows for better visualization.

Co-authored-by:  Thomas Gibson <thomas.gibson@nps.edu>


jkozdon:
bors try

jkozdon:
Just to be concrete, this is a sample of the output from the acoustic wave with N=4 comparing the new feature (with sampling rate of 10) to the existing plotting of LGL points.

mesh_compare

blallen:
cc: @christophernhill @jm-c this should be useful for visualizing our output

PR #1339, opened 2020-07-09
Add Roe flux

mwarusz:
# Description

Adds Roe numerical flux.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -70,9 +70,13 @@ import ..DGMethods.NumericalFluxes:
     NumericalFluxGradient,
     NumericalFluxSecondOrder,
     CentralNumericalFluxHigherOrder,
-    CentralNumericalFluxDivergence
+    CentralNumericalFluxDivergence,
+    CentralNumericalFluxFirstOrder,
+    numerical_flux_first_order!
+using ..DGMethods.NumericalFluxes: RoeNumericalFlux
 
 import ..Courant: advective_courant, nondiffusive_courant, diffusive_courant
+
 
 """
     AtmosModel <: BalanceLaw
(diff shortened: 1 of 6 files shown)


mwarusz:
Lots still left to be done. Not sure how to test that this is really a Roe flux. Results for the rising thermal bubble without any diffusion except for the numerical flux look better with Roe though.
comp2

smarras79:
WOW, this is amazing! Thanks for putting that in, @mwarusz I look forward to trying it with Dycoms!

bischtob:
@mwarusz, does this mean that the simulation is stable just with Roe fluxes all the way to equilibrium?

mwarusz:
> @mwarusz, does this mean that the simulation is stable just with Roe fluxes all the way to equilibrium?

It is stable for 1000 secs with both fluxes. Haven't tried running longer.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -716,4 +718,107 @@ function init_state_conservative!(
...
+        balance_law,
+        balance_law.moisture,
+        state_conservative⁺,
+        state_auxiliary⁺,
+    )
+    h⁺ = (ρe⁺ + p⁺) / ρ⁺
+
+    ũ = (sqrt(ρ⁻) * u⁻ + sqrt(ρ⁺) * u⁺) / (sqrt(ρ⁻) + sqrt(ρ⁺))
+    h̃ = (sqrt(ρ⁻) * h⁻ + sqrt(ρ⁺) * h⁺) / (sqrt(ρ⁻) + sqrt(ρ⁺))
+    c̃ = sqrt((γ - 1) * (h̃ - ũ' * ũ / 2 - Φ + _cv_d * _T_0))
```
tapios, line 785: Use ` c̃ = soundspeed_air(...)` from `Thermodynamics` here for this to be correct in the moist case.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -716,4 +718,107 @@ function init_state_conservative!(
...
+        state_auxiliary⁻,
+        state_conservative⁺,
+        state_auxiliary⁺,
+        t,
+        direction,
+    )
+
+    FT = eltype(fluxᵀn)
+    param_set = balance_law.param_set
+    _cv_d::FT = cv_d(param_set)
```
tapios, line 751: You don't need ` _cv_d`, `_T_0` and `γ` here if you use the Thermodynamics functions below (which properly generalize to the moist case).

src/Atmos/Model/AtmosModel.jl
tapios: (Note that the error here otherwise is O(1) because the enthalpy has O(1) contributions from latent enthalpies.)

tapios:
Thanks @mwarusz. I'm happy to see this coming along. Please talk with @charleskawczynski how to get a correct sound speed out in the moist case after the weighting averaging of the total enthalpies.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -716,4 +718,107 @@ function init_state_conservative!(
...
+        abs(ũᵀn),
+        abs(ũᵀn),
+        abs(ũᵀn + c̃),
+    )
+
+    M = hcat(
+        SVector(1, ũc̃⁻[1], ũc̃⁻[2], ũc̃⁻[3], h̃ - c̃ * ũᵀn),
+        SVector(0, τ1[1], τ1[2], τ1[3], τ1' * ũ),
+        SVector(0, τ2[1], τ2[2], τ2[3], τ2' * ũ),
+        SVector(1, ũ[1], ũ[2], ũ[3], ũ' * ũ / 2 + Φ - _T_0 * _cv_d),
```
tapios, line 813: This looks odd. It seems it depends on the arbitrary absolute temperature scale `_T_0`, which it should not depend on (this is a reference scale that should drop out of all dynamics).

fxgiraldo:
> Lots still left to be done. Not sure how to test that this is really a Roe flux. Results for the rising thermal bubble without any diffusion except for the numerical flux look better with Roe though.
> 
> comp2
> 
> 

@mwarusz looks good. Can you confirm whether the Roe flux satisfies the contact property?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -716,4 +718,107 @@ function init_state_conservative!(
...
+    ρe⁻ = state_conservative⁻.ρe
+
+    u⁻ = ρu⁻ / ρ⁻
+    p⁻ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_conservative⁻,
+        state_auxiliary⁻,
+    )
+    h⁻ = (ρe⁻ + p⁻) / ρ⁻
```
tapios, line 768: Use the `total_specific_enthalpy` here (which @charleskawczynski will provide in `Thermodynamics`) so that this generalizes appropriately.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -716,4 +718,107 @@ function init_state_conservative!(
...
+    ρe⁺ = state_conservative⁺.ρe
+
+    u⁺ = ρu⁺ / ρ⁺
+    p⁺ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_conservative⁺,
+        state_auxiliary⁺,
+    )
+    h⁺ = (ρe⁺ + p⁺) / ρ⁺
```
tapios, line 781: Same as above: use `total_specific_enthalpy`.

src/Atmos/Model/AtmosModel.jl
tapios: To improve the low-Mach number behavior of the fluxes (which we need), you may want to first compute 
- `c = soundspeed_air(...)`
and then take 
- `c̃ =c * max(min(M, 1), M_cut)` where `M = abs(u)/c` and `M_cut` is a cutoff Mach number parameter (default value `M_cut = 0`.

See https://doi.org/10.1016/j.jcp.2008.01.037 (their Eq. (18) uses a cutoff function that goes smoothly to 1, and you could use that too.) This gives better results at low Mach numbers, which we have in LES. See https://arxiv.org/pdf/2004.01627.pdf for numerical examples.

src/Atmos/Model/AtmosModel.jl
tapios: Shouldn't the last term here be only `ũ' * ũ / 2`? The quantity that is there, adding potential energy and subtracting an arbitrary reference internal energy, is not a meaningful thermodynamic quantity. For dry dynamics, everything should be completely independent of `T_0`, which is arbitrary (the dynamics are unchanged for `T_0 = 0` or `T_0 = 1e12`). So it's difficult to see how `T_0` arises here. (While the linearized pressure depends on `T_0`, only pressure gradients matter dynamically, and those do not depend on `T_0`.)

@mwarusz: Could you please provide a reference from which you are working here?

src/Atmos/Model/AtmosModel.jl
szy21: > Use ` c̃ = soundspeed_air(...)` from `Thermodynamics` here for this to be correct in the moist case. (@charleskawczynski can help if there are questions about how to access this.)

@charleskawczynski I wonder if you have already looked at this? (I am trying to implement moist Roe flux, and it seems there is no straightforward way to get soundspeed_air from enthalpy and velocity?)

tapios: Why don't you do something like this:

- Define function `Roe_average(A⁻, A⁺, ρ⁻, ρ⁺) = ...`
- Compute `ũ = Roe_average(...), `\tilde e = Roe_average(e⁻, e⁺, ...)`, `\tilde T  = Roe_average(T⁻, T⁺, ...)`, `\tilde q = Roe_average...`
- Define thermodynamic state on basis of Roe averages
- Compute `c  = soundspeed_air(...)` [as function of thermodynamic state]
- Compute `h̃ = total_specific_enthalpy(\tilde e, Rm(\tilde q), tilde T)`

szy21: Yes, that's the plan actually. Thanks!

szy21: Although the `\tilde h` you get from this would not be the same as the Roe average of h?

tapios: Yes, not exactly because of the nonlinearity in the moist gas constant. But this is a thermodynamically consistent way of defining the Roe averaged enthalpy through a Legendre transformation of thermodynamic states from the energy. My understanding is that it holds to the same degree of accuracy as the Roe averages in general.

szy21: Good to know, thank you!

szy21: What should be the density used for defining the thermodynamic state on basis of Roe averages? (sorry if it's a too basic question...)

tapios: Yes, maybe better to avoid the thermodynamic state. 

If you want, you can also avoid the issue of the different `p/\rho` and `R_m T` Roe averages by simply Roe averaging `R_m T = p/\rho`, and then using that to compute the speed of sound (e.g., add another overloaded method that takes `R_m T` as input to compute `c`). However, practically the small difference between the two averages probably makes no difference.

src/Atmos/Model/AtmosModel.jl
mwarusz: Done.

src/Atmos/Model/AtmosModel.jl
mwarusz: Done.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -711,4 +713,126 @@ function init_state_prognostic!(
...
+        state_auxiliary⁻,
+    )
+
+    u⁻ = ρu⁻ / ρ⁻
+    p⁻ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_conservative⁻,
+        state_auxiliary⁻,
+    )
```
szy21, line 771: Do we still need this?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -711,4 +713,126 @@ function init_state_prognostic!(
...
+        state_auxiliary⁺,
+    )
+
+    u⁺ = ρu⁺ / ρ⁺
+    p⁺ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_conservative⁺,
+        state_auxiliary⁺,
+    )
```
szy21, line 792: same as above - do we still need this?

src/Atmos/Model/AtmosModel.jl
mwarusz: No, we don't. Good catch!

src/Atmos/Model/AtmosModel.jl
```diff
@@ -711,4 +713,114 @@ function init_state_prognostic!(
...
+    ω = FT(π) / 3
+    δ = FT(π) / 5
+    random_unit_vector = SVector(sin(ω) * cos(δ), cos(ω) * cos(δ), sin(δ))
+
+    # tangent space basis
+    τ1 = random_unit_vector × normal_vector
+    τ2 = τ1 × normal_vector
+
+    ũᵀn = ũ' * normal_vector
+    ũc̃⁻ = ũ + c̃ * normal_vector
```
szy21, line 800: Should it be u - c * normal_vector here? Maybe I miss something in the derivation.

mwarusz: Yes it should. Thanks ! Wonder why this didn't cause obvious issues.

szy21:
Is this ready for merging?

mwarusz:
Not yet. Still need to handle tracers and rewrite it in a matrix-free form for performance.

szy21:
> Not yet. Still need to handle tracers and rewrite it in a matrix-free form for performance.

Thanks! @LenkaNovak @jiahe23 just to let you know it will probably take a few more days, so if you would like to try it maybe just pull this branch.

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try

PR #1205, opened 2020-06-05
Add moist-baroclinic-wave benchmark case to GCM experiments

bischtob:
# Description

This adds the Moist Baroclinic Wave DCMIP 2016 benchmark to the GCM experiments.

co-authored by: @LenkaNovak & @bischtob 

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [x] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1193, opened 2020-06-03
DCMIP Test Case: Non-orographic gravity waves on a small planet

thomasgibson:
# Description

This PR is a result of a productive programming session between myself, @bischtob, @akshaysridhar, and @charleskawczynski. This adds a non-hydrostatic gravity wave test featured in the 2012 Dynamical Core Model Intercomparison Project (DCMIP): https://www.earthsystemcog.org/projects/dcmip-2012/test_cases

This test is a fairly simple GCM test, which is targeting the dynamical core specifically (no moisture, physics, radiation). It's a well-known test and a good one to have!

Currently in a "draft" phase, but will go ahead and post this here for comments!

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://CliMA.github.io/CLIMA/latest/CodingConventions.html) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1340, opened 2020-07-10
Add GCM Benchmarks

thomasgibson:
This PR is co-authored by: @bischtob, @mwarusz, and @akshaysridhar, and supersedes both #1193 and #1205. We have added a suite of experiments for the GCM model. This includes the following:
1. A non-hydrostatic gravity wave,
2. A balanced isothermal flow,
3. A baroclinic wave;
4. An updated Held-Suarez experiment (with tracer-support).

This PR addresses issue #1306, but we still need to sort out how we will test these experiments (similarly for the LES) in order to ensure we don't regress.

# Description

I have

### DIFF
--- a/src/Common/Orientations/Orientations.jl
+++ b/src/Common/Orientations/Orientations.jl
@@ -39,7 +39,9 @@ export init_aux!,
     projection_normal,
     gravitational_potential,
     ∇gravitational_potential,
-    projection_tangential
+    projection_tangential,
+    sphr_to_cart_vec,
+    cart_to_sphr_vec
 
 
 abstract type Orientation <: BalanceLaw end
@@ -141,6 +143,57 @@ latitude(orientation::SphericalOrientation, aux::Vars) =
 longitude(orientation::SphericalOrientation, aux::Vars) =
     @inbounds atan(aux.coord[2], aux.coord[1])
 
+"""
+    sphr_to_cart_vec(orientation::SphericalOrientation, state::Vars, aux::Vars)
+
+Projects a vector defined based on unit vectors in spherical coordinates to cartesian unit vectors.
+"""
+function sphr_to_cart_vec(
+    orientation::SphericalOrientation,
+    vec::AbstractVector,
+    aux::Vars,
+)
+    FT = eltype(aux)
+    lat = latitude(orientation, aux)
+    lon = longitude(orientation, aux)
+
+    slat, clat = sin(lat), cos(lat)
+    slon, clon = sin(lon), cos(lon)
+
+    u = MVector{3, FT}(
+        -slon * vec[1] - slat * clon * vec[2] + clat * clon * vec[3],
+        clon * vec[1] - slat * slon * vec[2] + clat * slon * vec[3],
+        clat * vec[2] + slat * vec[3],
+    )
+
+    return u
+end
+
+"""
+    cart_to_sphr_vec(orientation::SphericalOrientation, state::Vars, aux::Vars)
+
+Projects a vector defined based on unit vectors in cartesian coordinates to a spherical unit vectors.
+"""
+function cart_to_sphr_vec(
+    orientation::SphericalOrientation,
+    vec::AbstractVector,
+    aux::Vars,
+)
+    FT = eltype(aux)
+    lat = latitude(orientation, aux)
+    lon = longitude(orientation, aux)
+
+    slat, clat = sin(lat), cos(lat)
+    slon, clon = sin(lon), cos(lon)
+
+    u = MVector{3, FT}(
+        -slon * vec[1] + clon * vec[2],
+        -slat * clon * vec[1] - slat * slon * vec[2] + clat * vec[3],
+        clat * clon * vec[1] + clat * slon * vec[2] + slat * vec[3],
+    )
+
+    return u
+end
 
 #####
 ##### FlatOrientation
--- a/docs/src/APIs/Common/Orientations.md
+++ b/docs/src/APIs/Common/Orientations.md
@@ -14,4 +14,6 @@ Orientations
 NoOrientation
 FlatOrientation
 SphericalOrientation
+sphr_to_cart_vec
+cart_to_sphr_vec
 ```
--- a/experiments/AtmosGCM/baroclinic-wave.jl
+++ b/experiments/AtmosGCM/baroclinic-wave.jl
@@ -1,18 +1,6 @@
 #!/usr/bin/env julia --project
 using ClimateMachine
-using ArgParse
-
-s = ArgParseSettings()
-@add_arg_table! s begin
-    "--number-of-tracers"
-    help = "Number of dummy tracers"
-    metavar = "<number>"
-    arg_type = Int
-    default = 0
-end
-
-parsed_args = ClimateMachine.init(parse_clargs = true, custom_clargs = s)
-const number_of_tracers = parsed_args["number-of-tracers"]
+ClimateMachine.init(parse_clargs = true)
 
 using ClimateMachine.Atmos
 using ClimateMachine.Orientations
@@ -26,7 +14,8 @@ using ClimateMachine.Mesh.Filters
 using ClimateMachine.Mesh.Grids
 using ClimateMachine.TemperatureProfiles
 using ClimateMachine.Thermodynamics:
-    air_pressure, air_density, air_temperature, total_energy, internal_energy
+    air_density, air_temperature, total_energy, internal_energy, PhasePartition
+using ClimateMachine.TurbulenceClosures
 using ClimateMachine.VariableTemplates
 
 using LinearAlgebra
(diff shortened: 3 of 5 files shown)


jkozdon:
> This PR addresses issue #1306, but we still need to sort out how we will test these experiments (similarly for the LES) in order to ensure we don't regress.

I would take a look at what @blallen and @christophernhill have been doing for the ocean model with @christophernhill's `StateCheck` functionality. See [here](https://github.com/CliMA/ClimateMachine.jl/blob/1171a998ca49b9e7a3efc92309ecfd4797110428/test/Ocean/HydrostaticBoussinesq/3D_hydrostatic_spindown.jl#L131-L135) for example.

thomasgibson:
> > This PR addresses issue #1306, but we still need to sort out how we will test these experiments (similarly for the LES) in order to ensure we don't regress.
> 
> I would take a look at what @blallen and @christophernhill have been doing for the ocean model with @christophernhill's `StateCheck` functionality. See [here](https://github.com/CliMA/ClimateMachine.jl/blob/1171a998ca49b9e7a3efc92309ecfd4797110428/test/Ocean/HydrostaticBoussinesq/3D_hydrostatic_spindown.jl#L131-L135) for example.

Thanks for the information! I'll take a look.

experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -1,121 +1,137 @@
...
 const param_set = EarthParameterSet()
 
-struct HeldSuarezDataConfig{FT}
-    T_ref::FT
-end
+import CLIMAParameters
+#CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 6.371e6 / 125.0
 
-function init_heldsuarez!(bl, state, aux, coords, t)
+function init_baroclinic_wave!(bl, state, aux, coords, t)
```
tapios, line 33: Why initialize HS with this detailed a baroclinic wave?

experiments/AtmosGCM/heldsuarez-tracers.jl
```diff
@@ -0,0 +1,273 @@
...
+        domain_height,
+        param_set,
+        init_heldsuarez!;
+        model = model,
+    )
+
+    return config
+end
+
+function held_suarez_forcing!(
```
tapios, line 120: Can we abstract this out somewhere, so we do not need to duplicate code here and in heldsuarez.jl? I'd prefer for this forcing to be defined and coded in only one place.

experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -173,73 +189,85 @@ function held_suarez_forcing!(
     k_v = k_f * height_factor
 
     # Apply Held-Suarez forcing
-    source.ρu -= k_v * projection_tangential(bl, aux, ρu)
+    source.ρu -= k_v * ρu
```
tapios, line 192: Why are we back to damping vertical velocities too, rather than just horizontal?

bischtob: It looks like they damp all velocity components in Held & Suarez 1994.

experiments/AtmosGCM/heldsuarez-tracers.jl
bischtob: I agree. I don't know where we would put this very specific forcing function. Would need to create a separate space.

experiments/AtmosGCM/heldsuarez.jl
bischtob: That's a mistake. Needs to be reverted.

tapios:
I'm glad to see this coming together. 

The discrete hydrostatic balance issue in #1306 still remains to be addressed, right? 

More generally (not for this PR), we need a cleaner separation of initial conditions/boundary conditions/forcing ("experiment setup") and "model configuration" concerns (LES vs GCM vs single stack, at different resolutions etc.). It would be good to be able to run these different experiments with less duplication of boiler plate code. We'd want a composable structure that allows us to combine different experiment setups with different model configurations. This PR is not the place to do this, but it would be good to get it on people's radar.

experiments/AtmosGCM/heldsuarez.jl
tapios: No, they don't. The model is hydrostatic, hence `v` is horizontal velocity only.

bischtob: Makes sense. I'll revert to original state.

experiments/AtmosGCM/heldsuarez-tracers.jl
```diff
@@ -0,0 +1,273 @@
...
+using ClimateMachine
+using ArgParse
+
+s = ArgParseSettings()
+@add_arg_table! s begin
+    "--number-of-tracers"
+    help = "Number of dummy tracers"
+    metavar = "<number>"
+    arg_type = Int
+    default = 0
```
thomasgibson, line 11: @bischtob Why do we have two versions of held suarez? by default, no tracers are set in this script. If the only functional difference between this held suarez and the other one is the tracer argument, why not just add this tracer stuff to `heldsuarez.jl`?

bischtob: Works for me!

experiments/AtmosGCM/heldsuarez-tracers.jl
bischtob: Perhaps we just combine the two drivers into one.

bischtob:
> I'm glad to see this coming together.
> 
Thanks!

> The discrete hydrostatic balance issue in #1306 still remains to be addressed, right?
> 
Yes, @mwarusz is putting this together as he is putting together the Roe fluxes in #1339.

experiments/AtmosGCM/heldsuarez-tracers.jl
thomasgibson: Done!

bischtob:
@simonbyrne, docs breaking?

jakebolewski:
fixed by #1362

thomasgibson:
bors try

bischtob:
bors try

blallen:
> ## try
> Build failed:
> 
> * [docs-build](https://github.com/CliMA/ClimateMachine.jl/runs/894368212)

I think you need to rebase on top of master for CI to work properly

bischtob:
bors try

simonbyrne:
LGTM. 

The only question is how we want to test this: these aren't currently hooked up to the Slurm CI pipeline.

@thomasgibson @bischtob what did you have in mind?

simonbyrne:
Can you coordinate the @jakebolewski to figure out what you want to do with this?

jakebolewski:
Is there a reduced problem size that we could run just to check that it doesn't trivially break (with some code changes) and run that with Slurm-CI? 

For real runs that we want to interpret I think I would just need the driver files / slurm batch scripts you are using to generate the output.

bischtob:
> Is there a reduced problem size that we could run just to check that it doesn't trivially break (with some code / api changes) and run that with Slurm-CI?
> 
> For real runs that we want to interpret I think I would just need the driver files / slurm batch scripts you are using to generate the output.

@jiahe23 & @LenkaNovak are the best to decide :-). They are running the GCM.

tapios:
> > Is there a reduced problem size that we could run just to check that it doesn't trivially break (with some code / api changes) and run that with Slurm-CI?
> > For real runs that we want to interpret I think I would just need the driver files / slurm batch scripts you are using to generate the output.
> 
> @jiahe23 & @LenkaNovak are the best to decide :-). They are running the GCM.

Just run ~10 time steps?

jiahe23:
> > > Is there a reduced problem size that we could run just to check that it doesn't trivially break (with some code / api changes) and run that with Slurm-CI?
> > > For real runs that we want to interpret I think I would just need the driver files / slurm batch scripts you are using to generate the output.
> > 
> > 
> > @jiahe23 & @LenkaNovak are the best to decide :-). They are running the GCM.
> 
> Just run ~10 time steps?

@jakebolewski 

10 steps sounds reasonable for a Slurm-CI test

An example of bash script we (mainly Lenka and myself) use a lot in visualize GCM output is at the [wiki page](https://github.com/CliMA/ClimateMachine.jl/wiki/Bash-Run-Scripts) which only output zonal mean zonal wind for now. 

Define the diagnostics/visualization for CI is on-going here #1363

src/Common/Orientations/Orientations.jl
```diff
@@ -141,6 +143,57 @@ latitude(orientation::SphericalOrientation, aux::Vars) =
...
+        clat * vec[2] + slat * vec[3],
+    )
+
+    return u
+end
+
+"""
+    cart_to_sphr_vec(orientation::SphericalOrientation, state::Vars, aux::Vars)
+
+Projects a vector defined based on unit vectors in cartesian coordinates to a spherical unit vectors.
```
charleskawczynski, line 175:
```suggestion
Projects a vector defined based on unit vectors in cartesian coordinates to a spherical unit vector.
```

src/Common/Orientations/Orientations.jl
```diff
@@ -141,6 +143,57 @@ latitude(orientation::SphericalOrientation, aux::Vars) =
 longitude(orientation::SphericalOrientation, aux::Vars) =
     @inbounds atan(aux.coord[2], aux.coord[1])
 
+"""
+    sphr_to_cart_vec(orientation::SphericalOrientation, state::Vars, aux::Vars)
+
+Projects a vector defined based on unit vectors in spherical coordinates to cartesian unit vectors.
```
charleskawczynski, line 149:
```suggestion
Projects a vector defined based on unit vectors in spherical coordinates to cartesian unit vector.
```

PR #1351, opened 2020-07-15
Add ViscousSponge to TurbulenceClosures

bischtob:
# Description

Adds a viscous sponge layer to turbulence closures. This sponge later can be used (just like the Rayleigh sponge layer that we already have) to damp waves near the top of the domain. 

`turbconv` and `turbulence` models can be used simultaneously with viscous sponges in this configuration.

I have

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -93,6 +93,7 @@ default values for each field.
         ref_state,
         turbulence,
         hyperdiffusion,
+        spongelayer,
         moisture,
         radiation,
         source,
@@ -103,7 +104,7 @@ default values for each field.
 # Fields
 $(DocStringExtensions.FIELDS)
 """
-struct AtmosModel{FT, PS, PR, O, RS, T, TC, HD, M, P, R, S, TR, DC} <:
+struct AtmosModel{FT, PS, PR, O, RS, T, TC, HD, VS, M, P, R, S, TR, DC} <:
        BalanceLaw
     "Parameter Set (type to dispatch on, e.g., planet parameters. See CLIMAParameters.jl package)"
     param_set::PS
@@ -119,6 +120,8 @@ struct AtmosModel{FT, PS, PR, O, RS, T, TC, HD, M, P, R, S, TR, DC} <:
     turbconv::TC
     "Hyperdiffusion Model (Equations for dynamics of high-order spatial wave attenuation)"
     hyperdiffusion::HD
+    "Viscous sponge layers"
+    viscoussponge::VS
     "Moisture Model (Equations for dynamics of moist variables)"
     moisture::M
     "Precipitation Model (Equations for dynamics of precipitating species)"
@@ -149,6 +152,7 @@ function AtmosModel{FT}(
     turbulence::T = SmagorinskyLilly{FT}(0.21),
     turbconv::TC = NoTurbConv(),
     hyperdiffusion::HD = NoHyperDiffusion(),
+    viscoussponge::VS = NoViscousSponge(),
     moisture::M = EquilMoist{FT}(),
     precipitation::P = NoPrecipitation(),
     radiation::R = NoRadiation(),
@@ -160,7 +164,7 @@ function AtmosModel{FT}(
     ),
     tracers::TR = NoTracers(),
     data_config::DC = nothing,
-) where {FT <: AbstractFloat, ISP, PR, O, RS, T, TC, HD, M, P, R, S, TR, DC}
+) where {FT <: AbstractFloat, ISP, PR, O, RS, T, TC, HD, VS, M, P, R, S, TR, DC}
 
     atmos = (
         param_set,
@@ -170,6 +174,7 @@ function AtmosModel{FT}(
         turbulence,
         turbconv,
         hyperdiffusion,
+        viscoussponge,
         moisture,
         precipitation,
         radiation,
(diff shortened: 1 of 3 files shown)


bischtob:
I have not tested this yet. Best would be to run it with a mountain and document that waves are properly absorbed vs case without sponge.

tapios:
Can we still use other turbulence closures (e.g., Smagorinsky) together with this viscous sponge? Or are the options mutually exclusive as it is? It would be important that we can use the sponge together with other turbulence closures. If it cannot be done in this framework, perhaps the sponge needs to become a wrapper that uses turbulence code?

bischtob:
> Can we still use other turbulence closures (e.g., Smagorinsky) together with this viscous sponge? Or are the options mutually exclusive as it is? It would be important that we can use the sponge together with other turbulence closures. If it cannot be done in this framework, perhaps the sponge needs to become a wrapper that uses turbulence code?

Good point. I think what we need to similar to what is done for the sources. @simonbyrne, can we do that?

smarras79:
> > Can we still use other turbulence closures (e.g., Smagorinsky) together with this viscous sponge? Or are the options mutually exclusive as it is? It would be important that we can use the sponge together with other turbulence closures. If it cannot be done in this framework, perhaps the sponge needs to become a wrapper that uses turbulence code?
> 
> Good point. I think what we need to similar to what is done for the sources. @simonbyrne, can we do that?

By "viscous sponge layer" are you considering a region where viscosity is increased drastically to damp some modes? If this is the case, no matter what turbulence does in that region, the viscous sponge will kill everything else, especially turbulence. But this is OK. It is a typical way of handling outflow b.c. in classical CFD problems (I do that all the time with my tsunami simulations based on  3D incompressible LES with two fluids).

I may have completely misunderstood the issue though.

bischtob:
> > > Can we still use other turbulence closures (e.g., Smagorinsky) together with this viscous sponge? Or are the options mutually exclusive as it is? It would be important that we can use the sponge together with other turbulence closures. If it cannot be done in this framework, perhaps the sponge needs to become a wrapper that uses turbulence code?
> > 
> > 
> > Good point. I think what we need to similar to what is done for the sources. @simonbyrne, can we do that?
> 
> By "viscous sponge layer" are you considering a region where viscosity is increased drastically to damp some modes? If this is the case, no matter what turbulence does in that region, the viscous sponge will kill everything else, especially turbulence. But this is OK. It is a typical way of handling outflow b.c. in classical CFD problems (I do that all the time with my tsunami simulations based on 3D incompressible LES with two fluids).
> 
> I may have completely misunderstood the issue though.

I think the question was more a software and interface question. Since this sponge is treated like a turbulence closure (it is based on those) the question is if we can apply multiple turbulence closures in the same experiment withing the current framework.

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -402,6 +405,81 @@ function turbulence_tensors(
     return ν, D_t, τ
 end
 
+# ### [Viscous Sponge](@id viscous-sponge)
+# `ViscousSponge` requires a user to specify a constant viscosity (kinematic),
```
akshaysridhar, line 409: w.r.t line 427, dynamic this should be "dynamic visc..." (?)

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -402,6 +405,81 @@ function turbulence_tensors(
...
+    ViscousSponge <: TurbulenceClosureModel
+
+Sponge layer with constant dynamic viscosity (`ρν`).
+
+# Fields
+
+$(DocStringExtensions.FIELDS)
+"""
+struct ViscousSponge{FT} <: TurbulenceClosureModel
+    "Dynamic Viscosity [kg/m/s]"
```
akshaysridhar, line 427: Do you intend to use this in addition to one of these turb models ? If this is the case then setting this up as a source term might be more convenient (\nabla u is available through the state_gradient_flux variables in the source function)

bischtob: Ok, so we move this over to source? @akshaysridhar, do we need anything else?

src/Common/TurbulenceClosures/TurbulenceClosures.jl
bischtob: correct

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -97,6 +103,20 @@ vars_state(::TurbulenceClosureModel, ::AbstractStateType, FT) = @vars()
...
+"""
+abstract type ViscousSponge end
+
+vars_state_conservative(::TurbulenceClosureModel, FT) = @vars()
+vars_state_auxiliary(::TurbulenceClosureModel, FT) = @vars()
+vars_state_gradient(::TurbulenceClosureModel, FT) = @vars()
+vars_state_gradient_flux(::TurbulenceClosureModel, FT) = @vars()
+vars_gradient_laplacian(::TurbulenceClosureModel, FT) = @vars()
+vars_integrals(::TurbulenceClosureModel, FT) = @vars()
+vars_reverse_integrals(::TurbulenceClosureModel, FT) = @vars()
```
charleskawczynski, line 118:
```suggestion
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -138,13 +141,13 @@ and single stack configurations.
 function AtmosModel{FT}(
     ::Union{Type{AtmosLESConfigType}, Type{SingleStackConfigType}},
     param_set::AbstractParameterSet;
-    init_state_prognostic::ISP = nothing,
```
kpamnany, line 141: You probably don't want to do this. :)

PR #1304, opened 2020-06-29
Upgrade WriteVTK

kpamnany:
# Description

Closes #1282 with suggested fix.

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [X] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [X] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [X] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1352, opened 2020-07-16
Switch VTK to use VTK_LAGRANGE_XXX

jkozdon:
By switching from `VTK_LINE`, `VTK_PIXEL`, and `VTK_VOXEL` to `VTK_LAGRANGE_CURVE`, `VTK_LAGRANGE_QUADRILATERAL`, and `VTK_LAGRANGE_HEXAHEDRON` the degree of freedom connecting lines are removed from the mesh.

Additionally, there is hope of eventually doing [high-order viz in paraview], but this requires an additional interpolation to an equally spaced mesh. Ideally we would want this done in paraview itself, since then the aliasing errors occur at the interpolation nodes (as in the code)

[high-order viz in paraview]: https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/

### DIFF
--- a/src/InputOutput/VTK/writemesh.jl
+++ b/src/InputOutput/VTK/writemesh.jl
@@ -1,60 +1,110 @@
 using WriteVTK
 
-#=
-This is the 1D WriteMesh routine
-=#
-function writemesh(base_name, x1; fields = (), realelems = 1:size(x1)[end])
-    (Nqr, _) = size(x1)
-    Nsubcells = (Nqr - 1)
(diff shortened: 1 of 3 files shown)


jkozdon:
An example of what I mean:
mesh

kpamnany:
Sweet. Might be good to rebase this on #1304?

jkozdon:
> Sweet. Might be good to rebase this on #1304?

Will do once it lands.

jkozdon:
bors try

jkozdon:
bors try

PR #1359, opened 2020-07-17
Yt/boyd vandeven

slifer50:
# Description

Addition of the Boyd-Vandeven filter.

Me and @thomasgibson have worked on adding the Boyd-Vandeven filter. This currently appears to be working and is only pending the addition of a unit test and some cleanup.

### DIFF
--- a/src/Numerics/Mesh/Filters.jl
+++ b/src/Numerics/Mesh/Filters.jl
@@ -1,5 +1,6 @@
 module Filters
 
+using SpecialFunctions
 using LinearAlgebra, GaussQuadrature, KernelAbstractions
 using KernelAbstractions.Extras: @unroll
 using StaticArrays
@@ -10,7 +11,7 @@ using ...MPIStateArrays
 using ...VariableTemplates: @vars, varsize, Vars, varsindices
 
 export AbstractSpectralFilter, AbstractFilter
-export ExponentialFilter, CutoffFilter, TMARFilter
+export ExponentialFilter, CutoffFilter, TMARFilter, BoydVandevenFilter
 
 abstract type AbstractFilter end
 abstract type AbstractSpectralFilter <: AbstractFilter end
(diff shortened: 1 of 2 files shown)


tapios:
Thanks for putting this together, @slifer50!

src/Numerics/Mesh/Filters.jl
```diff
@@ -160,6 +161,66 @@ struct ExponentialFilter <: AbstractSpectralFilter
...
+
+Returns the spectral filter using the logorithmic error function of
+the form:
+```math
+σ(η) = 1/2 erfc(2*sqrt(s)*χ(η)*(abs(η)-0.5))
+```
+whenever s ≤ i ≤ N, and 1 otherwise. The function `χ(η)` is defined
+as
+```math
+χ(x) = sqrt(-log(1-4*(abs(x)-0.5)^2)/4*(abs(x)-0.5)^2)
```
tapios, line 175: Maybe easier to read if you replace `x -> η` here?

src/Numerics/Mesh/Filters.jl
```diff
@@ -160,6 +161,66 @@ struct ExponentialFilter <: AbstractSpectralFilter
...
+        α = -log(eps(eltype(grid))),
+    )
+        AT = arraytype(grid)
+        N = polynomialorder(grid)
+        ξ = referencepoints(grid)
+
+        @assert iseven(s)
+        @assert 0 <= Nc <= N
+
+        σ(η) = 0.5*erfc(2*sqrt(s)*χ(η)*(abs(η)-0.5))
```
tapios, line 209: The expressions here and in `χ` simplify if you substitute `\bar η = abs(η)-0.5` and make `χ` a function of this transformed `η`.

tapios:
Looks good to me!

slifer50:
I've done a few tests with the filter to see how it affects the solution of an inviscid rising bubble. The first image shows solution when using the Boyd-Vandeven filter. The second and third images show the solution of the unfiltered flow. The second image showcases the large extrema while the third uses an adjusted scale for a clearer image.
Both of these tests show the presence of element imprinting, with and without the filter.
 
Screen Shot 2020-07-17 at 10 39 50 PM
Screen Shot 2020-07-17 at 10 57 35 PM
Screen Shot 2020-07-17 at 11 29 08 PM

slifer50:
To investigate the source of the element imprinting I ran the unfiltered rising bubble with Constant diffusions of 75,10, 0.5 and 0. The images are presented in that order. 
Screen Shot 2020-07-18 at 12 14 43 AM
Screen Shot 2020-07-18 at 12 29 12 AM
Screen Shot 2020-07-18 at 12 41 44 AM
Screen Shot 2020-07-17 at 11 29 08 PM
We clearly see that the element imprinting seems to be tied to a lack of diffusion in the flow. 
This is what I think is happening:
Without enough diffusion in the flow, the effects of the numerical fluxes between elements becomes much more apparent. This is especially since the Rusanov flux is diffusive which might be creating surfaces of diffusion specifically around element boundaries which results in the imprinting we see (note that this does not break the continuity of the solution). This imprinting as shown by the bubble case is also only visible where we have gradients and thus active numerical fluxes. The areas with constant delta_theta do not show any imprinting

tapios:
> To investigate the source of the element imprinting I ran the unfiltered rising bubble with Constant diffusions of 75,10, 0.5 and 0. The images are presented in that order.
> Screen Shot 2020-07-18 at 12 14 43 AM
> Screen Shot 2020-07-18 at 12 29 12 AM
> Screen Shot 2020-07-18 at 12 41 44 AM
> Screen Shot 2020-07-17 at 11 29 08 PM
> We clearly see that the element imprinting seems to be tied to a lack of diffusion in the flow.
> This is what I think is happening:
> Without enough diffusion in the flow, the effects of the numerical fluxes between elements becomes much more apparent. This is especially since the Rusanov flux is diffusive which might be creating surfaces of diffusion specifically around element boundaries which results in the imprinting we see (note that this does not break the continuity of the solution). This imprinting as shown by the bubble case is also only visible where we have gradients and thus active numerical fluxes. The areas with constant delta_theta do not show any imprinting

I expect this to improve with numerical fluxes that have proper limits for low-Mach number flow, e.g., Roe fluxes with a Mach number corrections that reduces them to standard upwinding in the low-Mach number limit. Rusanov with the low-Mach number flows such as you have here is extremely over-upwinding.

smarras79:
Good job @slifer50, thanks for adding this so quickly and having it to work. The filter seems to be working correctly when it comes to stabilizing the solution. I agree with @tapios that the imprinting is related to the flux and the filter could not handle it unless somewhat tuned for that (not worth a try). 
After this PR has been approved and is ready to be merged, please, in this order, do the following:
- run the julia formatter on the files that you modified 
- rebase (I can help you with the process if you don't know the steps)
- push -f

thanks!

bischtob:
@slifer50, can we squash and get this in :-). Looks like it can be part of the codebase now.

PR #1361, opened 2020-07-17
Enforce discrete hydrostatic balance

### LINKED ISSUE #1247: Maintaining discrete hydrostatic balance
# Description

Numerics should maintain hydrostatically balanced states to a very high precision.
For that to happen the pressure gradient term must be in discrete balance with
the gravity source term and the numerical flux shouldn't break this balance.
Calculating `rho -= 1 / g dp/dz` using DG results in discrete balance but Rusanov
flux breaks it due to the penalty term `wavespeed * (rho+ - rho-)` and discontinuous `rho`.   

Ideas:
- calculate `p` from `rho` using discrete numerical integration that uses projection to
  a space that is one order lower  to make the state continuous
- use a different numerical flux (current strategy is 1) to test central fluxes and split in IMEX between Rusanov & central and 2) implement Roe fluxes later)
- approach based on https://arxiv.org/pdf/1511.08739

# Additional context

# For CLIMA Developers

mwarusz:
This PR enforces discrete hydrostatic balance of the reference state by computing its density from the balance equation.
For the discrete balance to be maintained using a numerical flux with the contact property is necessary. Currently in CLIMA only the central flux has this property, and it is likely to lead to unstable code. Another flux that has this property is the Roe flux which is getting implemented in #1339.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -63,7 +63,8 @@ import ClimateMachine.DGMethods:
     lengthscale,
     resolutionmetric,
     DGModel,
-    nodal_update_auxiliary_state!
+    nodal_update_auxiliary_state!,
+    nodal_init_state_auxiliary!
 import ..DGMethods.NumericalFluxes:
     boundary_state!,
     boundary_flux_second_order!,
@@ -640,24 +641,53 @@ function reverse_integral_set_auxiliary_state!(
     reverse_integral_set_auxiliary_state!(m.radiation, aux, integ)
 end
 
+function atmos_nodal_init_state_auxiliary!(
+    m::AtmosModel,
+    aux::Vars,
+    tmp::Vars,
+    geom::LocalGeometry,
+)
+    aux.coord = geom.coord
+    init_aux_turbulence!(m.turbulence, m, aux, geom)
+    atmos_init_aux!(m.ref_state, m, aux, tmp, geom)
+    init_aux_hyperdiffusion!(m.hyperdiffusion, m, aux, geom)
+    atmos_init_aux!(m.tracers, m, aux, geom)
+    init_aux_turbconv!(m.turbconv, m, aux, geom)
+end
 
 @doc """
     init_state_auxiliary!(
         m::AtmosModel,
-        aux::Vars,
-        geom::LocalGeometry
+        aux::MPIStateArray,
+        grid,
         )
 Initialise auxiliary variables for each AtmosModel subcomponent.
 Store Cartesian coordinate information in `aux.coord`.
 """ init_state_auxiliary!
-function init_state_auxiliary!(m::AtmosModel, aux::Vars, geom::LocalGeometry)
-    aux.coord = geom.coord
-    init_aux!(m.orientation, m.param_set, aux)
-    init_aux_turbulence!(m.turbulence, m, aux, geom)
-    atmos_init_aux!(m.ref_state, m, aux, geom)
-    init_aux_hyperdiffusion!(m.hyperdiffusion, m, aux, geom)
-    atmos_init_aux!(m.tracers, m, aux, geom)
-    init_aux_turbconv!(m.turbconv, m, aux, geom)
+function init_state_auxiliary!(
+    m::AtmosModel,
+    state_auxiliary::MPIStateArray,
+    grid,
+)
+    init_aux!(m, m.orientation, state_auxiliary, grid)
+
+    nodal_init_state_auxiliary!(
+        m,
+        (m, aux, tmp, geom) ->
+            atmos_init_ref_state_pressure!(m.ref_state, m, aux, geom),
+        state_auxiliary,
+        grid,
+    )
+
+    ∇p = ∇reference_pressure(m.ref_state, state_auxiliary, grid)
+
+    nodal_init_state_auxiliary!(
+        m,
+        atmos_nodal_init_state_auxiliary!,
+        state_auxiliary,
+        grid,
+        state_temporary = ∇p,
+    )
 end
 
 @doc """
--- a/src/Atmos/Model/ref_state.jl
+++ b/src/Atmos/Model/ref_state.jl
@@ -19,6 +19,7 @@ atmos_init_aux!(
     ::ReferenceState,
     ::AtmosModel,
     aux::Vars,
+    tmp::Vars,
     geom::LocalGeometry,
 ) = nothing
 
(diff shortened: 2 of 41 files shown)


mwarusz:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -401,11 +402,7 @@ equations.
 
     # pressure terms
     p = pressure(m, m.moisture, state, aux)
-    if m.ref_state isa HydrostaticState
-        flux.ρu += (p - aux.ref_state.p) * I
-    else
-        flux.ρu += p * I
-    end
+    flux.ρu += p * I
```
tapios, line 405: Why is the factoring out of the hydrostatic balance in the momentum equation removed? We had subtracted this for numerical stability. Maybe I'm missing it, but I do not see where this subtraction occurs now.

src/Atmos/Model/source.jl
```diff
@@ -65,11 +65,7 @@ function atmos_source!(
     t::Real,
     direction,
 )
-    if atmos.ref_state isa HydrostaticState
-        source.ρu -= (state.ρ - aux.ref_state.ρ) * aux.orientation.∇Φ
-    else
-        source.ρu -= state.ρ * aux.orientation.∇Φ
-    end
+    source.ρu -= state.ρ * aux.orientation.∇Φ
```
tapios, line 68: Same as above: Why is the subtraction of hydrostatic balance removed from the momentum equation? Or does it occur somewhere else now where I missed it?

tapios:
Thanks, @mwarusz. I'm glad to see this coming along. I do not understand why the subtraction of hydrostatic balance is removed from the momentum equation (we needed that for numerical stability), but it's possible I'm missing something. It would be good for @simonbyrne to have a look at this.

src/Atmos/Model/AtmosModel.jl
mwarusz: Unless I am missing something if the reference state is discretely balanced then the terms that we used to factor out should cancel out to machine precision.

src/Atmos/Model/source.jl
tapios: No, the only cancel out in your initial state, where ρ and p are hydrostatically balanced. But this is a special initial state, and not one that is generally maintained. In general, there is a nonhydrostatic residual, and this is essential (e.g., it's what's responsible for convection).

src/Atmos/Model/AtmosModel.jl
tapios: No, they only cancel out in your initial state, where ρ and p are hydrostatically balanced. But this is a special initial state, and not one that is generally maintained. In general, there is a nonhydrostatic residual, and this is essential (e.g., it's what's responsible for convection).

mwarusz:
bors try

src/Atmos/Model/ref_state.jl
```diff
@@ -95,3 +108,63 @@ function atmos_init_aux!(
...
 end
+
+using ..MPIStateArrays: vars
+using ..DGMethods: init_ode_state
+using ..DGMethods.NumericalFluxes:
+    CentralNumericalFluxFirstOrder,
+    CentralNumericalFluxSecondOrder,
+    CentralNumericalFluxGradient
+
+struct PressureGradientModel <: BalanceLaw end
```
thomasgibson, line 119: Even though it's clear to us who have been following this issue, I think a short docstring describing what this micro balance law is representing is needed.

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -905,3 +949,45 @@ function MPIStateArrays.MPIStateArray(dg::DGModel)
 
     return state_conservative
 end
+
+function contiguous_field_gradient!(
```
thomasgibson, line 953: A short docstring describing what this function is doing will be very nice.

src/Atmos/Model/AtmosModel.jl
jkozdon: So Tapio, evidence suggests that even if the reference field is discretely hydrostatically balanced using the perturbation here (and below) is more stable? Because I agree with Maciek that that with exact math, if the reference state is discretely balance this is a no-op.

mwarusz:
@jkozdon  
> Don't we need to use a numerical flux with the contact property for this? Since we use Rusanov I thought this wouldn't work. (It's likely I am just missing something!)

No, you are not missing anything, we still need a flux with the contact property. I have updated the PR description.

tapios:
Several separate issues are mixed up in the discussion here:

1. What this PR needs to do is *only* to ensure that reference state that is being subtracted in the momentum equation is discretely in hydrostatic balance (so we subtract an exact 0). 

2. The first issue is *separate* from preserving a hydrostatically balanced state at rest. The letter requires that contact discontinuities are preserved by numerical fluxes when velocities vanish. The Roe flux, with the modification for the sound speed by the Mach number I suggested there, will preserve this. But we do not need to deal with this here. This is only about hydrostatic balance of the reference state.

3. As a consequence of 1. but 2. being a separate issue, it's important that the numerical flux with which the reference state pressure gradient is computed is the same as that which is used in the momentum flux computation, whichever it is (and whether it preserves contact discontinuities or not).

@bischtob can help if the concepts need more clarification.

jkozdon:
> No, you are not missing anything, we still need a flux with the contact property. I have updated the PR description.

@mwarusz. Thanks for clarifying and working through all the issues with this.

> Several separate issues are mixed up in the discussion here:

@tapios. Thanks for the info. 

@tapios and @mwarusz: you all have more experience here than I do, so I was just trying to make sure I understand the issues. Seems to me that we are all roughly on the same page.

mwarusz:
bors try

mwarusz:
bors try

mwarusz:
bors try

src/Atmos/Model/ref_state.jl
```diff
@@ -95,3 +108,63 @@ function atmos_init_aux!(
...
+∇reference_pressure(::NoReferenceState, state_auxiliary, grid) = nothing
+function ∇reference_pressure(::ReferenceState, state_auxiliary, grid)
+    FT = eltype(state_auxiliary)
+    ∇p = similar(state_auxiliary; vars = @vars(∇p::SVector{3, FT}))
+
+    grad_model = PressureGradientModel()
+    grad_dg = DGModel(
+        grad_model,
+        grid,
+        CentralNumericalFluxFirstOrder(),
```
bischtob, line 159: Let's make sure that the numerical flux with which the reference state pressure gradient is computed is the same as that which is used in the momentum flux computation. @tapios made the point below (point 3 in the list).

src/Atmos/Model/AtmosModel.jl
mwarusz: I reverted this change.

src/Atmos/Model/source.jl
mwarusz: Same as above: I reverted this change

src/Atmos/Model/ref_state.jl
mwarusz: Done.

src/Numerics/DGMethods/DGModel.jl
mwarusz: Done.

src/Atmos/Model/ref_state.jl
mwarusz: The choice of numerical fluxes doesn't matter for taking the reference state pressure gradient since it is a continuous field. I added a comment that says so. Do we need to handle reference states with discontinuous pressure ?

mwarusz:
bors try

src/Atmos/Model/ref_state.jl
```diff
@@ -54,23 +55,35 @@ end
...
+    k = vertical_unit_vector(atmos, aux)
+    ∇Φ = ∇gravitational_potential(atmos, aux)
 
-    # Replace density by computation from pressure
-    # ρ = -1/g*dpdz
-    ρ = p / (_R_d * T_virt)
+    # density computation from pressure ρ = -1/g*dpdz
+    ρ = -k' * tmp.∇p / (k' * ∇Φ)
     aux.ref_state.ρ = ρ
-    aux.ref_state.p = p
```
charleskawczynski, line 73: Don't we still need pressure here?

mwarusz: I don't think so. To take the gradient of pressure the reference state is initialized in two phases. The reference state pressure is initialized in `atmos_init_ref_state_pressure!`, before this function is called. Maybe it would be clearer to get `p` directly from `aux.ref_state` here ?

charleskawczynski: Ah, I see, thanks!

simonbyrne:
bors try

mwarusz:
bors try

mwarusz:
bors try

simonbyrne:
Looks good, should it be squashed?

mwarusz:
Of course, doing this right now.

mwarusz:
bors try

mwarusz:
bors try

PR #1365, opened 2020-07-19
add a memory usage monitor for the GPU

vchuravy:
Adds a simple memory usage monitor for the GPU.


valeriabarra:
Thanks for the addition!

Would this close #1628 ?

vchuravy:
Partially yes. I just need to re-add the test to the buildkite system and then this should be good to go.

kpamnany:
🤔 What will this do if you're running on the CPU?

PR #1366, opened 2020-07-21
add les diagnostic variable attributes

### LINKED ISSUE #1230: Add diagnostic variable attributes
# Description

Every diagnostic variable listed [here](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/DiagnosticVariables/) needs to have an entry added in [this function](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Diagnostics/variables.jl#L25).

There will be common variables between GCM and LES. If there are any conflicts in long names or anything, we should discuss here.

We may want to split `setup_variables()` into a number of separate functions -- `setup_common_variables()`, `setup_les_default_variables()`, etc.?

# For CLIMA Developers

- [X] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)

szy21:
# Description

Closes #1230.

I have

### DIFF
--- a/src/Diagnostics/atmos_les_core.jl
+++ b/src/Diagnostics/atmos_les_core.jl
@@ -129,7 +129,8 @@ function atmos_les_core_init(dgngrp::DiagnosticsGroup, currtime)
         )
         append!(varnames, ho_varnames)
         for varname in varnames
-            vars[varname] = (("z",), FT, Dict())
+            var = Variables[varname]
+            vars[varname] = (("z",), FT, var.attrib)
         end
 
         # create the output file
--- a/src/Diagnostics/atmos_les_default.jl
+++ b/src/Diagnostics/atmos_les_default.jl
@@ -265,12 +265,13 @@ function atmos_les_default_init(dgngrp::DiagnosticsGroup, currtime)
         )
         append!(varnames, ho_varnames)
         for varname in varnames
-            vars[varname] = (("z",), FT, Dict())
+            var = Variables[varname]
+            vars[varname] = (("z",), FT, var.attrib)
         end
-        vars["cld_frac"] = (("z",), FT, Dict())
-        vars["cld_top"] = ((), FT, Dict())
-        vars["cld_base"] = ((), FT, Dict())
-        vars["cld_cover"] = ((), FT, Dict())
+        vars["cld_frac"] = (("z",), FT, Variables["cld_frac"].attrib)
+        vars["cld_top"] = ((), FT, Variables["cld_top"].attrib)
+        vars["cld_base"] = ((), FT, Variables["cld_base"].attrib)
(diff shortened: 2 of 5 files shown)


szy21:
@kpamnany I have added all the les diagnostic variables.

kpamnany:
The only conflicting descriptions appear to be `u`, `v` and `w`. A couple of questions:

1. Do you think there will be other variables that have identical names, but different descriptions in the LES and the GCM?
2. Can we change these names for either LES or GCM? That is, can we name these variables `xvel`, `yvel`, `zvel` for LES, or can we call them `east`, `north`, `up` for GCM (could be any names)?

Re: `rho`, it is a 3 dimensional field in the GCM 'default' group, so it is not averaged.

Cc: @LenkaNovak

szy21:
> The only conflicting descriptions appear to be `u`, `v` and `w`. A couple of questions:
> 
> 1. Do you think there will be other variables that have identical names, but different descriptions in the LES and the GCM?

I can't think of any right now.
> 2. Can we change these names for either LES or GCM? That is, can we name these variables `xvel`, `yvel`, `zvel` for LES, or can we call them `east`, `north`, `up` for GCM (could be any names)?

I'm ok with changing u,v,w to xvel, yvel, zvel in LES, although we will need to change other variables (e.g., var_u, cov_w_qt, etc.) as well. I'm also ok with calling u,v,w zonal wind, meridional wind, vertical wind (same as GCM). Right now these long names make sense for LES, although they won't make sense if we change the orientation of LES. @akshaysridhar ?
> Re: `rho`, it is a 3 dimensional field in the GCM 'default' group, so it is not averaged.

In the GCM the diagnostics are time-averaged? Is it density-weighted for time average?
> 
> Cc: @LenkaNovak

LenkaNovak:
I think changing the `(u,v,w)` long names should be sufficient. There will be things like the `meridional temperature flux` or `zonal wind stress`, but we could get away with just saying `v-wind temperature flux` or `u-stress`. It may sound a bit clunky, but it's probably worth keeping this as general as possible, at least for now.   

I would also strongly vote for keeping the symbols for `(u,v,w)`. I think changing that now would cause a lot of confusion for current users and would go against the conventional notation. 

Would it be possible to have a switch so that the relevant long names of `u` and `v` get selected for the appropriate model? 

As for time averaging for GCM diagnostics, this is not yet implemented, but yes, it will follow the same format as the current LES averaging. :)

kpamnany:
> Would it be possible to have a switch so that the relevant long names of `u` and `v` get selected for the appropriate model?

Sure it's possible, just clunky to have a special case for 3 variables which is why I asked if there might be others.

Since @szy21 is fine with using the GCM long names for now, I'll just push a commit that will add the use of these attributes into the various diagnostic groups and merge this.

PR #1378, opened 2020-07-23
Add UpDownIntegrals, unify Upward and reverse_ integral methods

charleskawczynski:
# Description

 - Adds `UpdDownIntegrals`
 - Adds a `UpDownIntegrals` argument to both `integral_` and `reverse_integral` methods
 - Fixes the radiation argument list in `ClimateMachine.jl/src/Atmos/Model/radiation.jl`

I have

### DIFF
--- a/src/BalanceLaws/state_types.jl
+++ b/src/BalanceLaws/state_types.jl
@@ -7,6 +7,7 @@ export AbstractStateType,
     GradientFlux,
     GradientLaplacian,
     Hyperdiffusive,
+    IntegralType,
     UpwardIntegrals,
     DownwardIntegrals
 
@@ -62,15 +63,22 @@ Hyper-diffusive variables
 struct Hyperdiffusive <: AbstractStateType end
 
 """
-    UpwardIntegrals <: AbstractStateType
+    IntegralType <: AbstractStateType
+
+Super-type for UpwardIntegrals and DownwardIntegrals
+"""
+abstract type IntegralType <: AbstractStateType end
+
+"""
+    UpwardIntegrals <: IntegralType
 
 Variables computed in upward integrals
 """
-struct UpwardIntegrals <: AbstractStateType end
+struct UpwardIntegrals <: IntegralType end
 
 """
-    DownwardIntegrals <: AbstractStateType
+    DownwardIntegrals <: IntegralType
 
 Variables computed in downward integrals
 """
-struct DownwardIntegrals <: AbstractStateType end
+struct DownwardIntegrals <: IntegralType end
--- a/docs/src/APIs/BalanceLaws/BalanceLaws.md
+++ b/docs/src/APIs/BalanceLaws/BalanceLaws.md
@@ -48,11 +48,8 @@ source!
 
 ```@docs
 indefinite_stack_integral!
-reverse_indefinite_stack_integral!
 integral_load_auxiliary_state!
 integral_set_auxiliary_state!
-reverse_integral_load_auxiliary_state!
-reverse_integral_set_auxiliary_state!
 ```
 
 ## Gradient/Laplacian kernels
--- a/docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
+++ b/docs/src/HowToGuides/Numerics/DGMethods/how_to_make_a_balance_law.md
@@ -52,8 +52,6 @@ The `vars_state` method is used to specify the names of the variables for the fo
 | [`compute_gradient_flux!`](@ref) | specify how to compute gradient fluxes. can be a functions of the gradient state, the prognostic state, and auxiliary variables.|
 | [`integral_load_auxiliary_state!`](@ref) | specify how to compute integrands. can be functions of the prognostic state and auxiliary variables. |
 | [`integral_set_auxiliary_state!`](@ref) | specify which auxiliary variables are used to store the output of the integrals. |
-| [`reverse_integral_load_auxiliary_state!`](@ref) | specify auxiliary variables need their integrals reversed. |
-| [`reverse_integral_set_auxiliary_state!`](@ref) | specify which auxiliary variables are used to store the output of the reversed integrals. |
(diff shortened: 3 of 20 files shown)


blallen:
This doesn't change how the kernels function, just how the functions are called? E.g., you still always have to call an upwards integral before calling a downwards integral?

charleskawczynski:
> This doesn't change how the kernels function, just how the functions are called?

Correct, this PR does not change the behavior of the kernels, it just uses `UpwardIntegrals` and `DownwardIntegrals`, to reduce

 - `indefinite_stack_integral!`
 - `integral_load_auxiliary_state!`
 - `integral_set_auxiliary_state!`
 - `reverse_indefinite_stack_integral!`
 - `reverse_integral_load_auxiliary_state!`
 - `reverse_integral_set_auxiliary_state!`

to

 - `indefinite_stack_integral!(..., [UpwardIntegrals / DownwardIntegrals])`
 - `integral_load_auxiliary_state!(..., [UpwardIntegrals / DownwardIntegrals])`
 - `integral_set_auxiliary_state!(..., [UpwardIntegrals / DownwardIntegrals])`

> you still always have to call an upwards integral before calling a downwards integral?

This PR doesn't change anything about the order the kernels are called.

charleskawczynski:
> In that case, this looks good to me. Maybe make the direction the first argument for indefinite_stack_integral!?

That's fine with me. Let's wait for Jeremy to review. Frankly, I was a bit hesitant with these changes anyway, but I thought I'd open a PR and see what people think.

To allow for a generic up/down integral (either or both), I think we'd need to make these kernels work with tuples, and I think that may make tracing a bit complicated-- It's already a bit painful to trace what's in `source!` and `boundary_state!`.

charleskawczynski:
Going to peel off the radiation integral kernel fixes, since it's a bug fix and we need that sooner than later.

src/BalanceLaws/state_types.jl
```diff
@@ -62,15 +63,22 @@ Hyper-diffusive variables
 struct Hyperdiffusive <: AbstractStateType end
 
 """
-    UpwardIntegrals <: AbstractStateType
+    UpDownIntegrals <: AbstractStateType
+
+Super-type for UpwardIntegrals and DownwardIntegrals
+"""
+abstract type UpDownIntegrals <: AbstractStateType end
```
thomasgibson, line 70: Why not just `IntegralType`?

charleskawczynski: Yeah, I hadn't thought long enough. `IntegralType` sounds much better 👍

jkozdon:
> When @jkozdon gets back, we probably want to build on this to have `indefinite_stack_integral!(::DownwardIntegrals)` not require having called `indefinite_stack_integral!(::UpwardIntegrals)` first. Then this code will act like how it looks.

Let me know if you need help with this.

charleskawczynski:
> we probably want to build on this to have indefinite_stack_integral!(::DownwardIntegrals) not require having called indefinite_stack_integral!(::UpwardIntegrals) first. Then this code will act like how it looks.

Actually, unless I'm missing something, I don't think the `DownwardIntegrals` method requires having the `UpwardIntegrals` method first. Where is that assumption @blallen ?

jkozdon:
> Actually, unless I'm missing something, I don't think the `DownwardIntegrals` method requires having the `UpwardIntegrals` method first. Where is that assumption @blallen ?

The current [`DownwardsIntegrals`](https://github.com/CliMA/ClimateMachine.jl/blob/24c0efab30442254c2568999663a5d2db0e39189/src/Numerics/DGMethods/DGModel.jl#L726-L766) in this pr uses `kernel_reverse_indefinite_stack_integral!`, which assumes that the integral has already been initialized by `kernel_indefinite_stack_integral!` (which is called by [`UpwardsIntegrals`](https://github.com/CliMA/ClimateMachine.jl/blob/24c0efab30442254c2568999663a5d2db0e39189/src/Numerics/DGMethods/DGModel.jl#L682-L724)).

What needs to happen is that `kernel_indefinite_stack_integral!` needs to be made direction aware, i.e., someone needs to write a version of `kernel_indefinite_stack_integral!` that does downwards (and not upwards) integrals.

charleskawczynski:
> > Actually, unless I'm missing something, I don't think the `DownwardIntegrals` method requires having the `UpwardIntegrals` method first. Where is that assumption @blallen ?
> 
> The current [`DownwardsIntegrals`](https://github.com/CliMA/ClimateMachine.jl/blob/24c0efab30442254c2568999663a5d2db0e39189/src/Numerics/DGMethods/DGModel.jl#L726-L766) in this pr uses `kernel_reverse_indefinite_stack_integral!`, which assumes that the integral has already been initialized by `kernel_indefinite_stack_integral!` (which is called by [`UpwardsIntegrals`](https://github.com/CliMA/ClimateMachine.jl/blob/24c0efab30442254c2568999663a5d2db0e39189/src/Numerics/DGMethods/DGModel.jl#L682-L724)).
> 
> What needs to happen is that `kernel_indefinite_stack_integral!` needs to be made direction aware, i.e., someone needs to write a version of `kernel_indefinite_stack_integral!` that does downwards (and not upwards) integrals.

Maybe we can leave that for a subsequent PR?

jkozdon:
> Maybe we can leave that for a subsequent PR?

Why the rush? I see this as having real potential for bugs to be introduced.

From what I can tell this is a purely cosmetic code change, and I can see no reason to not get it right from the start.

The way its written right now could lead to someone doing a `DownwardIntegrals` without calling `UpwardsIntegrals` first which will lead to bugs (because you will be reversing garbage).

blallen:
> > Maybe we can leave that for a subsequent PR?
> 
> Why the rush? I see this as having real potential for bugs to be introduced.
> 
> From what I can tell this is a purely cosmetic code change, and I can see no reason to not get it right from the start.
> 
> The way its written right now could lead to someone doing a `DownwardIntegrals` without calling `UpwardsIntegrals` first which will lead to bugs (because you will be reversing garbage).

I agree with this. I've already had that issue before and it's not fun to debug.

charleskawczynski:

PR #1384, opened 2020-07-24
Add `init_auxiliary_state!()` hook

### LINKED ISSUE #1213: Simplify debugging initial conditions / experiment configurations

kpamnany:
# Description

Closes #1213.

I have

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -95,13 +95,14 @@ Users may over-ride prescribed default values for each field.
         source,
         tracers,
         boundarycondition,
-        init_state_prognostic
+        init_state_prognostic,
+        init_state_auxiliary,
     )
 
 # Fields
 $(DocStringExtensions.FIELDS)
 """
-struct AtmosModel{FT, PS, O, RS, T, TC, HD, M, P, R, S, TR, BC, IS, DC} <:
+struct AtmosModel{FT, PS, O, RS, T, TC, HD, M, P, R, S, TR, BC, ISP, ISA, DC} <:
        BalanceLaw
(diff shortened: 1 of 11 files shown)


kpamnany:
We don't like this solution.

PR #1385, opened 2020-07-24
Add low-storage 3N Runge-Kutta time-stepping scheme

jm-c:
Using the formulation of Fyfe (1966) that can accommodate any 4-stage, RK scheme,
add a new RK scheme implementation that requires only 3-N storage.

Two standard RK schemes are provided:
1) the Classic 4-stage, fourth order Runge-Kutta ;
2) the 3-stage, third order Heun's method.

# Description

Fyfe (1966) formulation for Runge-Kutta scheme, that uses 3-N storage

I have

- [ x] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`


jm-c:
Few comments:
1) I think it could be useful to have a cheap (only 3-N storage) version of the classic RK-4 scheme (e.g., when comparing with other
 models since this scheme is widely used). Also, I am planning to use the Heun's RK-3 (a simple, cheap RK scheme that is
stable for pure oscillatory problems, i.e., pure imaginary axis) for ocean split-explicit time-stepping.
2) naming issue: not sure I made the right choice; may be "Fyfe3NStorageRungeKutta3NMethod.jl", F3NRK44Classic, F3NRK33Heuns would be better ? 
3) I did not bother going through formatter at this point; I tried to add some math comments but not sure they are any good.
4) regarding implementation, can be view as an extension of LowStorageRungeKutta2N since here RKA(:,1) & RKB(:,1) are
used the same way as RKA(:), RKB(:) in LowStorageRungeKutta2N.

simonbyrne:
How do these relate to the ones in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/ODESolvers/StrongStabilityPreservingRungeKuttaMethod.jl?

jm-c:
> How do these relate to the ones in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/ODESolvers/StrongStabilityPreservingRungeKuttaMethod.jl?

Although both use 3-N storage, the way RK is applied is different: the SSPRK use Shu & Osher formulation while the new one
LowStorageRK-3N use Fyfe formulation. As a consequence, I cannot fit neither the Classic RK-4 nor the Heun's RK-3 into the SSPRK (nor does they fit into the LowStorageRK2N ), so if someone want to try one of these standard RK scheme, it needs to be implemented outside the SSPRK current formulation, in a new RK time-stepping. This is what I propose here.

simonbyrne:
Thanks, that makes sense.

Just need to apply the formatter, and add the docstrings to the docs. I can walk you through that if you like.

src/Numerics/ODESolvers/LowStorageRungeKutta3NMethod.jl
```diff
@@ -0,0 +1,273 @@
...
+   @article {Blum1962,
+       title = {A Modification of the Runge-Kutta Fourth-Order Method}
+       author = {Blum, E. K.},
+       journal = {Mathematics of Computation},
+       volume = {16},
+       pages = {176-187},
+       year = {1962}
+    }
+"""
+function LS3NRK44Classic(
```
jkozdon, line 197: Other places we have used the convention of using the author's names in the type.

I'm fine either way, just thought I'd point this out.

src/Numerics/ODESolvers/LowStorageRungeKutta3NMethod.jl
```diff
@@ -0,0 +1,273 @@
...
+    rhs!, dQ, dR = lsrk3n.rhs!, lsrk3n.dQ, lsrk3n.dR
+
+    rv_Q = realview(Q)
+    rv_dQ = realview(dQ)
+    rv_dR = realview(dR)
+    groupsize = 256
+
+    rv_dR .= -0
+    for s in 1:length(RKC)
+        rhs!(dQ, Q, p, time + RKC[s] * dt, increment = true)
```
jkozdon, line 127: `increment` is going to be depreciated, should this be switch? (@simonbyrne and @thomasgibson)

src/Numerics/ODESolvers/LowStorageRungeKutta3NMethod.jl
jm-c: I could change this, but thought the original scheme was from Runge + Kutta in the early 1900s, and was known as this
"classic" RK-4. The Blum 1962 paper is more relevant for the implementaion (3-N storage) rather than the property of the scheme.

jm-c:
@jkozdon and others, in term of names, would you advise to stay with "LowStorageRungeKutta3N" or switch to a name that 
contains "Fyfe" like "Fyfe3NStorageRungeKutta" ?

src/Numerics/ODESolvers/LowStorageRungeKutta3NMethod.jl
simonbyrne: it's still supported for now, we will probably get rid of it once we integrate TimeMachine.jl though

src/Numerics/ODESolvers/LowStorageRungeKutta3NMethod.jl
jkozdon: Sounds good.

jkozdon:
> @jkozdon and others, in term of names, would you advise to stay with "LowStorageRungeKutta3N" or switch to a name that
> contains "Fyfe" like "Fyfe3NStorageRungeKutta" ?

Names is fine by me

jm-c:
@simonbyrne I run the formatter, this should be good now. I need to check the docs but I am not sure how to do this (+ have problems to see how the updated docs look like). Do you have some tips ?

jm-c:
I made little changes so that the documentation test is now passing and in my local web-browser the new piece of doc looks good.
Does anyone has some comments on this doc part ? Otherwise I am going to move to next step (rebase).

jm-c:
bors try

PR #1360, opened 2020-07-17
Spectrum calculator and diagnostics group

slifer50:
# Description

A spectrum calculator built around the Taylor Green problem

I have implemented a spectrum calculator that works for problems with the same size and points in all directions. This has been implemented to work through a diagnostic group. This will form a baseline which can then be extended to work on anisotropic grids and spectrum calculations on the sphere. I've tested this with the Taylor-Green problem and correctly reproduces the results I've gotten by post-processing the vtks from this case. It's been tested to work on a single rank run and with multiple-ranks without issues. @tapios @smarras79. @kpamnany It would be good if you could have a look at the diagnostics implementation to double check everything.

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1388, opened 2020-07-26
Add turbulence statistics diagnostics

slifer50:
# Description

Defines a callback for calculating volumetrically averaged dimensionless kinetic energy and dissipation

I have built a module that allows for a callback to calculate volumetrically averaged dimensionless kinetic energy and kinetic energy dissipation. This will be useful as it replaces the tedious procedure of calculating these during post-processing. These represent useful insight into certain turbulence dominated cases like the Taylor-Green vortex.

### DIFF
--- a/docs/src/APIs/Diagnostics/Diagnostics.md
+++ b/docs/src/APIs/Diagnostics/Diagnostics.md
@@ -25,6 +25,7 @@ Diagnostics.setup_atmos_default_diagnostics
 Diagnostics.setup_atmos_core_diagnostics
 Diagnostics.setup_atmos_default_perturbations
 Diagnostics.setup_atmos_refstate_perturbations
+Diagnostics.setup_atmos_turbulence_stats
 Diagnostics.setup_dump_state_diagnostics
 Diagnostics.setup_dump_aux_diagnostics
 ```
--- a/experiments/AtmosLES/taylor_green.jl
+++ b/experiments/AtmosLES/taylor_green.jl
@@ -21,7 +21,8 @@ using DocStringExtensions
 using LinearAlgebra
 
 using CLIMAParameters
-using CLIMAParameters.Planet: cp_d, MSLP, grav, LH_v0
+using CLIMAParameters.Planet: R_d, cv_d, cp_d, MSLP, grav, LH_v0
+using CLIMAParameters.Atmos.SubgridScale: C_smag
 struct EarthParameterSet <: AbstractEarthParameterSet end
 const param_set = EarthParameterSet()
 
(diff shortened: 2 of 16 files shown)


slifer50:
bors try

src/Common/TurbulenceStatistics/kinetic_energy.jl
```diff
@@ -0,0 +1,62 @@
...
+    driver_config,
+    E_0,
+    iter,
+    outfile,
+    nor,
+)
+    mpicomm = MPI.COMM_WORLD
+    mpirank = MPI.Comm_rank(mpicomm)
+    if mpirank == 0
+        E_k0 = average_kinetic_energy(solver_config, driver_config, nor)
```
kpamnany, line 12: This won't work on more than one rank since you're calling `MPI_Reduce()` in `average_kinetic_energy()` below, but only from rank 0.

experiments/AtmosLES/taylor-green.jl
```diff
@@ -202,11 +204,22 @@ function main()
         init_on_cpu = true,
         Courant_number = CFL,
     )
-    dgn_config = config_diagnostics(driver_config)
-
+    #dgn_config = config_diagnostics(driver_config)
+    energy_file = "/home/yassine/Kinetic_energy.txt"
```
kpamnany, line 208: We cannot merge such an absolute path.

kpamnany:
This doesn't seem sufficiently general a solution to me.

experiments/AtmosLES/taylor-green.jl
slifer50: @kpamnany Good point, how would we best get around this ?

slifer50:
> This doesn't seem sufficiently general a solution to me.

@kpamnany I would have liked to have this included in a diagnostics group but I need access to data from previous time steps and I'm not sure how to approach from within the diagnostics framework.

src/Common/TurbulenceStatistics/kinetic_energy.jl
slifer50: Good point, I'll fix that

experiments/AtmosLES/taylor-green.jl
akshaysridhar: @slifer50 We could permit write operations into the `--output-dir` similar to the VTK and NetCDF diagnostics (?)

slifer50: Yes, unfortunately I'm not sure how to do that.

smarras79:
> This doesn't seem sufficiently general a solution to me.

@kpamnany how can this be made more general? turbulence diagnostics need to be part of the code; we should be able to turn them on and off as needed. 
Can you please help @slifer50 make it general? thanks

kpamnany:
@slifer50: I changed this into a diagnostics group. Please check out the output from the Taylor Green experiment (run with `--diagnostics=default`).

As another note, you probably need to change #1360 -- instead of adding `nor` into `DiagnosticsGroup`, you can follow the method used in this PR, [here](https://github.com/CliMA/ClimateMachine.jl/blob/yt/turbulence_stats/src/Diagnostics/atmos_turbulence_stats.jl#L52).

slifer50:
@kpamnany Everything looks good to me. Awesome work!

Issue #1391, opened 2020-07-27
Boyd-Vandeven filter


jiahe23:
# Description

A bug is identified in the `χ(x)` formula at the filter [description](https://github.com/CliMA/ClimateMachine.jl/blob/765e41d598bc7647043b9424524c20ed78ddfbdd/src/Numerics/Mesh/Filters.jl#L175) and the [function](https://github.com/CliMA/ClimateMachine.jl/blob/765e41d598bc7647043b9424524c20ed78ddfbdd/src/Numerics/Mesh/Filters.jl#L215).

## Expected behavior
A simple fix is 
`sqrt(-log(1 - 4 * (abs(x) - 0.5)^2) / 4 / (abs(x) - 0.5)^2)`

# For CLIMA developers

- [ ] The bug can be reproduced
- [ ] Labels have been added and developers assigned
- [ ] The bug can be fixed (if not, please explain why not in a comment below)
- [ ] There is a timeline to fix the bug
- [ ] The bug has been fixed (Please link the PR)

slifer50:
@jiahe23 I've added your suggestion in PR #1392

PR #1392, opened 2020-07-27
Fixed Boyd_Vandeven filter bug

slifer50:
# Description

Fix suggested by @jiahe23 in issue #1391.

I have implemented the correction suggested by @jiahe23 in issue #1391 

### DIFF
--- a/src/Numerics/Mesh/Filters.jl
+++ b/src/Numerics/Mesh/Filters.jl
@@ -172,7 +172,7 @@ the form:
 whenever s ≤ i ≤ N, and 1 otherwise. The function `χ(η)` is defined
 as
 ```math
-χ(x) = sqrt(-log(1-4*(abs(x)-0.5)^2)/4*(abs(x)-0.5)^2)
+χ(η) = sqrt(-log(1-4*(abs(η)-0.5)^2)/(4*(abs(η)-0.5)^2))
 ```
 if `x != 0.5` and `1` otherwise. Here, `s` is the filter order,
 the filter starts with polynomial order `Nc`, and `alpha` is a parameter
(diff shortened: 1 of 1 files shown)


src/Numerics/Mesh/Filters.jl
```diff
@@ -212,7 +212,7 @@ function χ(x)
     if (x == 0.5)
         return 1
     else
-        return sqrt(-log(1 - 4 * (abs(x) - 0.5)^2) / 4 * (abs(x) - 0.5)^2)
+        return sqrt(-log(1 - 4 * (abs(x) - 0.5)^2) / (4 * (abs(x) - 0.5)^2))
```
jkozdon, line 215: Probably should not be `0.5` by `FT(1/2)` or 
```suggestion
        return sqrt(-log(1 - (2abs(x) - 1)^2) / ((2abs(x) - 1)^2))
```

jkozdon: (otherwise you won't be respecting the type of `x`)

src/Numerics/Mesh/Filters.jl
```diff
@@ -212,7 +212,7 @@ function χ(x)
     if (x == 0.5)
```
jkozdon, line 212: Should this really be exact equality? Also, some floating point instability issue (since `0.5` is a `Float64` and `x` isn't necessarily)

slifer50: The equality should be exact since 0.5 generates a division by zero in the else statement. I'll adjust it to deal with the floating point instability issue, thanks for pointing this out.

src/Numerics/Mesh/Filters.jl
slifer50: @jkozdon you're right I'll adjust this accordingly

src/Numerics/Mesh/Filters.jl
jkozdon: Still not sure this is right then.

Maybe:

```suggestion
    if (2x -1 == 0)
```

@simonbyrne?

Really there should probably be an asymptotic expansion when `|2x-1|` is small since even with small values you can get issues (I would think)

slifer50: This seems plausible, I'll differ to your expertise on this.

jkozdon: 🤷

simonbyrne: Yeah, this does look like it will suffer from cancellation error: using `log1p` might help a bit, but the division might cause problems.

simonbyrne:
We will also need to add `erfc` to KernelAbstractions

jkozdon:
> We will also need to add `erfc` to KernelAbstractions

This is all done host side (it's a precompute) so don't think that's needed.

src/Numerics/Mesh/Filters.jl
simonbyrne: The division is fine, since `2abs(η) - 1` has to be at least `eps()/2`: it was the cancellation in the `log(1+...)` which was the problem.

slifer50:
> It looks like using `log1p` was sufficient. I also simplified the code slightly.
> 
> @slifer50 can you double check I didn't break anything

@simonbyrne Looks good to me.

Issue #1368, opened 2020-07-21
Unit tests may be leaking scope


charleskawczynski:
# Description

At one time, I found that this `AtmosModel` [call](https://github.com/CliMA/ClimateMachine.jl/pull/1265/files#diff-5e87599165833809de99132c510d213cR52) was picking up this `AtmosModel` [reference](https://github.com/CliMA/ClimateMachine.jl/pull/1050/files#diff-02a333c064d8b529d7692e47086e731fL6). Maybe we should implement something like [SafeTestsets.jl](https://github.com/YingboMa/SafeTestsets.jl) to avoid leaking scope in our unit tests.

## Additional context

Add any other reasons why this should be addressed.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

charleskawczynski:
Just got stung by this [here](https://github.com/CliMA/ClimateMachine.jl/pull/1280#issuecomment-662003220).

charleskawczynski:
I noticed in ForwardDiff.jl, they just put each unit test inside a module. Doing that seems like a simple solution/pattern.

PR #1394, opened 2020-07-27
Adjust ocean tests for shorter Github Actions CI

blallen:
# Description

Reorganize ocean tests to run fewer integration tests on Github Actions CI. Also, restructure ocean experiments/tests so that all tests are in `test/Ocean` directory but reference a single `run` function defined in `experiments/OceanBoxGCM`. Also add new files to `experiments/OceanBoxGCM` that allow users to run experiments and change parameters without affecting the tests as well as use command line arguments for output and such.

I have


blallen:
@kpamnany can you confirm that the files in `experiments/OceanBoxGCM` have everything they need to use command-line arguments properly? Also that this setup is in line with the philosophy for `Driver.jl`.

@christophernhill double-check and make sure that I didn't remove any crucial tests by accident. I tried to remove all multi-BC tests and just make sure we have everything tested at least once across the different tests.

@jakebolewski please confirm that this actually shortens things and if not tell me which tests are the worst offenders.

Thanks guys!

blallen:
bors try

blallen:
bors try

blallen:
> These changes are a bit confusing -- e.g. are you running `test_2D_spindown.jl` on Actions _and_ on SlurmCI? Seems unnecessary?
> 

The run on slurm is to check that things work on the GPU. The Actions CI is because that is generally faster than a `bors try` attempt, in order to get faster feedback in case things break.

> Generally, I think you would want to use `parse_clargs = true` for code running on SlurmCI, but _not_ for code running on Actions although it doesn't really matter if you aren't passing in any command line arguments.
> 
> Edit: sorry, started looking without seeing your comment. `homogeneous_box.jl` and `ocean_gyre.jl` have the right parameter, but `simple_box.jl` is presumably included from other places and isn't calling `ClimateMachine.init()` itself, so depends on who is including it.

Yep, just wanted to make sure those two were using it properly. `simple_box.jl` is meant to be a "header" file that can be called by a lot of other different scripts, so it doesn't do any initialization itself.

test/Ocean/HydrostaticBoussinesq/test_ocean_gyre_long.jl
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env julia --project
+using Test
+
+include("../../../experiments/OceanBoxGCM/simple_box.jl")
+ClimateMachine.init()
+
+#################
+# RUN THE TESTS #
+#################
+const FT = Float64
```
jakebolewski, line 10: will these toplevel constants be an issue if all the tests are evaluated in the same scope?  it might be better to include them in the testset?

blallen: ah that sounds like a good idea. it hasn't caused an issue previously and we have a few existing tests that use this setup

blallen:
Okay timings from the Actions run: https://github.com/CliMA/ClimateMachine.jl/pull/1394/checks?check_run_id=915586403

`test_vertical_integral_model.jl` - 130 seconds
`test_2D_spindown.jl` - 155 seconds
`test_3D_spindown.jl` - 478 seconds
`test_windstress_short.jl` - 792 seconds
`test_ocean_gyre_short.jl` - 201 seconds
`test_spindown_short.jl` - 284 seconds

`test_windstress_short.jl` actually runs two integrations, missed that earlier. but in general, I might be able to get away with dropping everything besides `test_spindown_short.jl` and `test_ocean_gyre_short.jl` since if those work everything else works. And then I can run the rest on slurm with `bors try` to see what is broken if those fail.

blallen:
>  but in general, I might be able to get away with dropping everything besides `test_spindown_short.jl` and `test_ocean_gyre_short.jl` since if those work everything else works. And then I can run the rest on slurm with `bors try` to see what is broken if those fail.

Awesome, with this change, the Ocean tests take 15 minutes and are the fastest now :) https://github.com/CliMA/ClimateMachine.jl/pull/1394/checks?check_run_id=915803091

test/Ocean/HydrostaticBoussinesq/test_ocean_gyre_long.jl
blallen: I think I was trying to cheat with some of these and actually use them as globals

jakebolewski: This shows up other places: https://github.com/CliMA/ClimateMachine.jl/issues/1368

blallen: okay, I managed to work around everything but the `const FT`. Would be nice if we could set `FT` as part of something like `ClimateMachine.init()` so we don't have to pass it through every single function.

blallen:
bors try

jakebolewski:
bors ping

PR #1398, opened 2020-07-29
Minor fixes for moist GCM diagnostics

LenkaNovak:
# Description

A typo was fixed to enable saving GCM diagnostics and added the `qi` variable (output tested). 

`experiments/AtmosGCM/moist-baroclinic-wave.jl` is a copy of `baroclinic-wave.jl` with moisture switched on. Perhaps this can be merged into one file once the config interface is cleaned up.

I have

### DIFF
--- a/src/Diagnostics/variables.jl
+++ b/src/Diagnostics/variables.jl
@@ -122,13 +122,17 @@ function setup_variables()
     )
     Variables["qt"] = DiagnosticVariable(
         "qt",
-        var_attrib("kg kg^-1", "total specific humidity", ""),
+        var_attrib(
+            "kg kg^-1",
+            "mass fraction of total water in air (qv+ql+qi)",
+            "mass_fraction_of_water_in_air",
+        ),
     )
     Variables["ql"] = DiagnosticVariable(
         "ql",
         var_attrib(
             "kg kg^-1",
-            "liquid water specific humidity",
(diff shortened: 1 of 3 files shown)


src/Diagnostics/variables.jl
```diff
@@ -137,9 +137,19 @@ function setup_variables()
         var_attrib(
             "kg kg^-1",
             "water vapor specific humidity",
-            "specific humidity",
+            "water vapor specific humidity",
```
szy21, line 140: I think it's specific_humidity in cf standard names (sorry I forgot _ in the original code). I do think water_vapor_specific_humidity is more accurate - up to you.

src/Diagnostics/variables.jl
```diff
@@ -137,9 +137,19 @@ function setup_variables()
...
         ),
     )
+
+    Variables["qi"] = DiagnosticVariable(
+        "qi",
+        var_attrib(
+            "kg kg^-1",
+            "ice specific humidity",
+            "ice specific humidity",
```
szy21, line 149: The cf standard name is mass_fraction_of_cloud_ice_in_air.

src/Diagnostics/variables.jl
LenkaNovak: Sounds good. We should stick to the CF convention :)

PR #1400, opened 2020-07-29
Shrink dg kernels

charleskawczynski:
# Description

This PR adds a few convenience functions and shortens some names of some things in the DG kernels (mostly to battle the formatter). I'm not 100% sure about these changes, but I wanted to open this PR to hear people's thoughts.

I have

### DIFF
--- a/src/Numerics/DGMethods/DGModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGModel_kernels.jl
@@ -24,6 +24,36 @@ const _n1, _n2, _n3 = Grids._n1, Grids._n2, Grids._n3
 const _sM, _vMI = Grids._sM, Grids._vMI
 # }}}
 
+dofs_per_elem_per_direction(N::Int) = (N + 1) # better name for this method?
+
+dofs_per_elem(N::Int, dim::Int) = (N + 1)^dim
+
+get_Nqk(N::Int, dim::Int) = dim == 2 ? 1 : N + 1 # better name for this method?
+
+n_faces(dim::Int) = 2 * dim
+
+get_faces(dim::Int, direction) = 1:n_faces(dim)
+
+get_faces(dim::Int, ::VerticalDirection) = (n_faces(dim) - 1):n_faces(dim)
+
+get_faces(dim::Int, ::HorizontalDirection) = 1:(n_faces(dim) - 2)
+
+"""
+    get_face_direction(face, dim)
+
+The remainder model needs to know which direction of face the model is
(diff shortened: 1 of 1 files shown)


jkozdon:
In my opinion a better approach would be to get `Var` / `Grad` out of the kernels completely. Have the kernels deal with straight arrays, and then have the wrapping by `Var` / `Grad` happen in the balance law and / or Numerical flux file.

This would also clean up things like the remainder model where we have to pull out the parent arrays.

So mainly only have `Vars` come into the play when it's needed.

I played around with this idea earlier on one of my branches which you can see here: https://github.com/CliMA/ClimateMachine.jl/commit/a82f049872e16446c8095e2ab43ee46f68bca86e

blallen:
I like this a lot! Seems like a better idea than #1062. 

Maybe to go with @jkozdon's comment, you could make wrapper functions that interface from pure array function calls to the Vars versions?

something like

```julia
function flux_first_order!(
    balance_law, 
    local_flux,
    local_state_prognostic,
    local_state_auxiliary,
    t,
    (direction,),
)
    vs = all_vars_states(balance_law, FT)
    return flux_first_order!(
        balance_law,
        Grad{vs.prognostic}(local_flux),
        Vars{vs.prognostic}(local_state_prognostic),
        Vars{vs.auxiliary}(local_state_auxiliary),
        t, 
        (direction,),
    )
end
```

I think we've mentioned this before but never gotten around to it

jkozdon:
That said, I think we should hold off on this for now.

There is no functionality added with this, but it's going to cause pain for anyone who has PRs working on the kernels.

cc @mwarusz, @thomasgibson

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

jkozdon:
Personally I liked stuff the old way. This seems more obfuscated to me.

But if others think this is good that’s fine.

charleskawczynski:
> Personally I liked stuff the old way. This seems more obfuscated to me. But if others think this is good that’s fine.

I've updated the PR to remove the `all_number_states`-related changes, but I left the added methods in `dg_element_helper_funcs.jl`, since this will help us document and generalize towards supporting different poly orders in different directions.

charleskawczynski:

PR #1404, opened 2020-07-31
Add bulk surface flux to boundaryconditions

szy21:
# Description

Add bulk surface flux of moisture and energy to boundaryconditions.

I have

### DIFF
--- a/src/Atmos/Model/bc_energy.jl
+++ b/src/Atmos/Model/bc_energy.jl
@@ -114,3 +114,65 @@ function atmos_energy_normal_boundary_flux_second_order!(
     # we want to prescribe the inward flux
     fluxᵀn.ρe -= bc_energy.fn(state⁻, aux⁻, t)
 end
+
+"""
+    BulkFormulaEnergy(fn) :: EnergyBC
+
+Calculate the net inward energy flux across the boundary.
+The drag coefficient is `C_h = fn_C_h(state, aux, t, normu_int_tan)`.
+The surface temp is `T_sfc= fn_T_sfc(state, aux, t)`.
+The surface q_tot is `q_tot_sfc = fn_q_tot_sfc(state, aux, t)`.
+Return the flux (in W m^-2).
+"""
+struct BulkFormulaEnergy{FNX, FNT, FNM} <: EnergyBC
+    fn_C_h::FNX
+    fn_T_sfc::FNT
+    fn_q_tot_sfc::FNM
+end
+function atmos_energy_boundary_state!(
+    nf,
+    bc_energy::BulkFormulaEnergy,
+    atmos,
+    args...,
+) end
+function atmos_energy_normal_boundary_flux_second_order!(
+    nf,
+    bc_energy::BulkFormulaEnergy,
+    atmos,
+    fluxᵀn,
+    n⁻,
+    state_sfc⁻,
+    diff_sfc⁻,
+    hyperdiff_sfc⁻,
+    aux_sfc⁻,
+    state_sfc⁺,
+    diff_sfc⁺,
+    hyperdiff_sfc⁺,
+    aux_sfc⁺,
+    bctype,
+    t,
+    state_int⁻,
+    diff_int⁻,
+    aux_int⁻,
+)
+
+    u_int⁻ = state_int⁻.ρu / state_int⁻.ρ
+    u_int⁻_tan = projection_tangential(atmos, aux_int⁻, u_int⁻)
+    normu_int⁻_tan = norm(u_int⁻_tan)
+    C_h = bc_energy.fn_C_h(state_sfc⁻, aux_sfc⁻, t, normu_int⁻_tan)
+    T_sfc = bc_energy.fn_T_sfc(state_sfc⁻, aux_sfc⁻, t)
+    q_tot_sfc = bc_energy.fn_q_tot_sfc(state_sfc⁻, aux_sfc⁻, t)
+
+    # calculate MSE from the states at the surface and at the interior point
+    ts_sfc =
+        TemperatureSHumEquil(atmos.param_set, T_sfc, state_sfc⁻.ρ, q_tot_sfc)
+    ts_int = thermo_state(atmos, atmos.moisture, state_int⁻, aux_int⁻)
+    e_pot_sfc = gravitational_potential(atmos.orientation, aux_sfc⁻)
+    e_pot_int = gravitational_potential(atmos.orientation, aux_int⁻)
+    MSE_sfc = moist_static_energy(ts_sfc, e_pot_sfc)
+    MSE_int = moist_static_energy(ts_int, e_pot_int)
+
+    ρ_avg = average_density_sfc_int(state_sfc⁻.ρ, state_int⁻.ρ)
+    # NOTE: difference from design docs since normal points outwards
+    fluxᵀn.ρe -= C_h * ρ_avg * normu_int⁻_tan * (MSE_sfc - MSE_int)
+end
--- a/src/Atmos/Model/bc_moisture.jl
+++ b/src/Atmos/Model/bc_moisture.jl
@@ -60,3 +60,57 @@ function atmos_moisture_normal_boundary_flux_second_order!(
     # assumes EquilMoist
     fluxᵀn.moisture.ρq_tot += nρd_q_tot
 end
+
+"""
+    BulkFormulaMoisture(fn) :: MoistureBC
+
+Calculate the net inward moisture flux across the boundary using
+the bulk formula.
+The drag coefficient is `C_q = fn_C_q(state, aux, t, normu_int_tan)`.
+The surface q_tot is `q_tot_sfc = fn_q_tot_sfc(state, aux, t)`.
+Return the flux (in kg m^-2 s^-1).
+"""
+struct BulkFormulaMoisture{FNX, FNM} <: MoistureBC
+    fn_C_q::FNX
+    fn_q_tot_sfc::FNM
+end
+function atmos_moisture_boundary_state!(
+    nf,
+    bc_moisture::BulkFormulaMoisture,
+    atmos,
+    args...,
+) end
+function atmos_moisture_normal_boundary_flux_second_order!(
+    nf,
+    bc_moisture::BulkFormulaMoisture,
+    atmos,
+    fluxᵀn,
+    n⁻,
+    state_sfc⁻,
+    diff_sfc⁻,
+    hyperdiff_sfc⁻,
+    aux_sfc⁻,
+    state_sfc⁺,
+    diff_sfc⁺,
+    hyperdiff_sfc⁺,
+    aux_sfc⁺,
+    bctype,
+    t,
+    state_int⁻,
+    diff_int⁻,
+    aux_int⁻,
+)
+
+    u_int⁻ = state_int⁻.ρu / state_int⁻.ρ
+    u_int⁻_tan = projection_tangential(atmos, aux_int⁻, u_int⁻)
+    normu_int⁻_tan = norm(u_int⁻_tan)
+    C_q = bc_moisture.fn_C_q(state_sfc⁻, aux_sfc⁻, t, normu_int⁻_tan)
+    q_tot_sfc = bc_moisture.fn_q_tot_sfc(state_sfc⁻, aux_sfc⁻, t)
+    q_tot_int = state_int⁻.moisture.ρq_tot / state_int⁻.ρ
+    ρ_avg = average_density_sfc_int(state_sfc⁻.ρ, state_int⁻.ρ)
+    # NOTE: difference from design docs since normal points outwards
+    fluxᵀn.moisture.ρq_tot -=
+        C_q * ρ_avg * normu_int⁻_tan * (q_tot_sfc - q_tot_int)
+    fluxᵀn.ρ -= C_q * ρ_avg * normu_int⁻_tan * (q_tot_sfc - q_tot_int)
+    fluxᵀn.ρu -= C_q * normu_int⁻_tan * (q_tot_sfc - q_tot_int) .* state_int⁻.ρu
+end
--- a/src/Atmos/Model/boundaryconditions.jl
+++ b/src/Atmos/Model/boundaryconditions.jl
@@ -10,10 +10,14 @@ export AtmosBC,
     Insulating,
     PrescribedTemperature,
     PrescribedEnergyFlux,
+    BulkFormulaEnergy,
     Impermeable,
     ImpermeableTracer,
     PrescribedMoistureFlux,
+    BulkFormulaMoisture,
     PrescribedTracerFlux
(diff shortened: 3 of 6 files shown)


src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,67 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    aux⁺,
+    bctype,
+    t,
+    state1⁻,
+    diff1⁻,
+    aux1⁻,
+)
+
+    u1⁻ = state1⁻.ρu / state1⁻.ρ
+    Pu1⁻ = u1⁻ - dot(u1⁻, n⁻) .* SVector(n⁻)
```
tapios, line 163: Didn't we have a function (`tangential_velocity`?) for this somewhere? (I know it came up in the Held-Suarez setup.) If we don't have it, we should probably add it, since we'll need in several places.

tapios: And could we have some more descriptive variable names instead of `state1⁻` and `state⁻` etc.? (e.g., `...int` and `...sfc`?)

szy21: > And could we have some more descriptive variable names instead of `state1⁻` and `state⁻` etc.? (e.g., `...int` and `...sfc`?)

I agree with this! I am inheriting the code from bc_momentum.jl. As someone who looks at bc for the first time it's really hard to understand what these variables mean.

szy21:
@akshaysridhar @tapios This is ready for review. Thanks!

src/Atmos/Model/AtmosModel.jl
```diff
@@ -27,7 +27,12 @@ using ..TurbulenceClosures
 import ..TurbulenceClosures: turbulence_tensors
 using ..TurbulenceConvection
 
-import ..Thermodynamics: internal_energy, total_specific_enthalpy
+import ..Thermodynamics:
```
charleskawczynski, line 30:
```suggestion
using ..Thermodynamics: 
```

`import` should be used when we want to define new methods for the same function

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    q_pt_sfc = PhasePartition(q_tot_sfc)
+    R_m_sfc = gas_constant_air(atmos.param_set, q_pt_sfc)
+    e_int_sfc = internal_energy(atmos.param_set, T_sfc, q_pt_sfc)
+    h_int_sfc = specific_enthalpy(e_int_sfc, R_m_sfc, T_sfc)
+    e_pot_sfc = gravitational_potential(atmos.orientation, aux_sfc⁻)
+    MSE_sfc = h_int_sfc + e_pot_sfc
+    # MSE at the interior point
+    TS_int = thermo_state(atmos, atmos.moisture, state_int⁻, aux_int⁻)
+    e_pot_int = gravitational_potential(atmos.orientation, aux_int⁻)
+    MSE_int = moist_static_energy(TS_int, e_pot_int)
```
charleskawczynski, line 178: Can we just establish a thermodynamic state for surface and int? For example:
```julia
    ts_sfc = thermo_state(atmos, atmos.moisture, state_sfc⁻, aux_sfc⁻)
    ts_int = thermo_state(atmos, atmos.moisture, state_int⁻, aux_int⁻)

    e_pot_sfc = gravitational_potential(atmos.orientation, aux_sfc⁻)
    e_pot_int = gravitational_potential(atmos.orientation, aux_int⁻)

    # MSE at the interior point
    MSE_sfc = moist_static_energy(ts_sfc, e_pot_sfc)
    MSE_int = moist_static_energy(ts_int, e_pot_int)
```

Also, isn't `z_sfc==z_int`? I think we may just be able to compute `e_pot` from either `aux_sfc⁻` or `aux_int⁻`.

src/Atmos/Model/bc_moisture.jl
```diff
@@ -60,3 +60,59 @@ function atmos_moisture_normal_boundary_flux_second_order!(
...
+    state_int⁻,
+    diff_int⁻,
+    aux_int⁻,
+)
+
+    u_int⁻ = state_int⁻.ρu / state_int⁻.ρ
+    Pu_int⁻ = projection_tangential(atmos, aux_int⁻, u_int⁻)
+    normPu_int⁻ = norm(Pu_int⁻)
+    # NOTE: difference from design docs since normal points outwards
+    C_q = bc_moisture.fn_C_q(state_sfc⁻, aux_sfc⁻, t, normPu_int⁻)
```
charleskawczynski, line 108: What's the implementation of `fn_C_q` look like?

src/Atmos/Model/bc_energy.jl
szy21: That would be more convenient. The problem is right now this is applied to prescribed surface temperature and moisture, and the surface state is not consistent. (All the simulations we have is with prescribed surface fluxes, so the state for the surface is not used or specified correctly.)

I think z_sfc is zero when there is no topography, and z_int is the height of the first interior point, which is larger than zero? @akshaysridhar

src/Atmos/Model/bc_moisture.jl
szy21: Right now it's just a constant. Akshay and I decided to keep it as a function for flexibility. It could be a function of the surface state (e.g. u) or time.

szy21:
> This looks like some nice enhancements. I left some comments on some things that I think can be improved on.

Thanks for the quick review! I will change the code accordingly. I hope this serves as a basic working example that we can build the surface module and/or coupler on. For example, M-O similarity theory will provide C_q and C_h, which are prescribed here. And a coupler will provide the surface state, which is also prescribed here.

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
     # we want to prescribe the inward flux
     fluxᵀn.ρe -= bc_energy.fn(state⁻, aux⁻, t)
 end
+
+"""
+    BulkFormulaEnergy(fn) :: EnergyBC
+
+Calculate the net inward energy flux across the boundary.
+The drag coefficient is `C_q = fn_C_q(state, aux, t, normPu_int)`.
+The surface temp is `T_sfc= fn_T_sfc(state, aus, t)`.
```
tapios, line 123:
```suggestion
The surface temp is `T_sfc= fn_T_sfc(state, aux, t)`.
```

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+The surface q_tot is `q_tot_sfc = fn_q_tot_sfc(state, aux, t)`.
+with signature `fn(state, aux, t)`, returning the flux (in W m^-2).
```
tapios, line 125:
```suggestion
The function with signature `fn(state, aux, t)` returns the energy flux (in W m^-2).
```

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    aux_sfc⁺,
+    bctype,
+    t,
+    state_int⁻,
+    diff_int⁻,
+    aux_int⁻,
+)
+
+    u_int⁻ = state_int⁻.ρu / state_int⁻.ρ
+    Pu_int⁻ = projection_tangential(atmos, aux_int⁻, u_int⁻)
```
tapios, line 161:
```suggestion
    u_int⁻_tan = projection_tangential(atmos, aux_int⁻, u_int⁻)
```
(or something like this, here and elsewhere, to be more descriptive)?

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    normPu_int⁻ = norm(Pu_int⁻)
+    # NOTE: difference from design docs since normal points outwards
+    C_h = bc_energy.fn_C_h(state_sfc⁻, aux_sfc⁻, t, normPu_int⁻)
+    T_sfc = bc_energy.fn_T_sfc(state_sfc⁻, aux_sfc⁻, t)
+    q_tot_sfc = bc_energy.fn_q_tot_sfc(state_sfc⁻, aux_sfc⁻, t)
+    #TODO: add kinetic energy? also figure out a better way...
```
tapios, line 167: Don't need KE.

src/Atmos/Model/bc_energy.jl
```diff
@@ -114,3 +114,68 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    # MSE at the surface
+    q_pt_sfc = PhasePartition(q_tot_sfc)
+    R_m_sfc = gas_constant_air(atmos.param_set, q_pt_sfc)
+    e_int_sfc = internal_energy(atmos.param_set, T_sfc, q_pt_sfc)
+    h_int_sfc = specific_enthalpy(e_int_sfc, R_m_sfc, T_sfc)
+    e_pot_sfc = gravitational_potential(atmos.orientation, aux_sfc⁻)
+    MSE_sfc = h_int_sfc + e_pot_sfc
```
tapios, line 174: Use the `moist_static_energy` function in thermodynamics. Right now it requires a thermodynamic state. If you cannot define that for the surface, you can overload an explicit function in `Thermodynamics` that takes the variables you have here as input. @charleskawczynski can help.

src/Atmos/Model/bc_energy.jl
tapios: It might be worthwhile to specify the surface state consistently (in drivers?)

src/Atmos/Model/bc_moisture.jl
```diff
@@ -60,3 +60,59 @@ function atmos_moisture_normal_boundary_flux_second_order!(
...
+    q_tot_sfc = bc_moisture.fn_q_tot_sfc(state_sfc⁻, aux_sfc⁻, t)
+    q_tot_int = state_int⁻.moisture.ρq_tot / state_int⁻.ρ
+    # both sides involve projections of normals, so signs are consistent
+    # TODO: change q_tot to q_vap
```
tapios, line 112: It's ok as it is with `q_tot`.

src/Atmos/Model/bc_energy.jl
charleskawczynski: Ok, what we may need then is
```julia
    T_sfc_prescribed = ...
    ts_sfc = TemperatureSHumEquil(atmos.param_set, T_sfc_prescribed, state.ρ, state.moisture.ρq_tot / state.ρ)
```

charleskawczynski: Let's go with

```julia
ts_sfc = TemperatureSHumEquil(atmos.param_set, T_sfc, state.ρ, q_tot_sfc)
```

for now (until the BC layer is better abstracted), so that we don't need to think about numerical fluxes.

szy21:
@akshaysridhar This is ready I think. Could you take a look and let me know if you have any further comments? I plan to run a full bomex to check the results. It would be good to change all the variable names for consistency (state, state1, Pu, etc.), but I prefer to do that in a separate PR. What do you think?

src/Atmos/Model/bc_moisture.jl
```diff
@@ -60,3 +60,56 @@ function atmos_moisture_normal_boundary_flux_second_order!(
...
+    u_int⁻_tan = projection_tangential(atmos, aux_int⁻, u_int⁻)
+    normu_int⁻_tan = norm(u_int⁻_tan)
+    C_q = bc_moisture.fn_C_q(state_sfc⁻, aux_sfc⁻, t, normu_int⁻_tan)
+    q_tot_sfc = bc_moisture.fn_q_tot_sfc(state_sfc⁻, aux_sfc⁻, t)
+    q_tot_int = state_int⁻.moisture.ρq_tot / state_int⁻.ρ
+    # NOTE: difference from design docs since normal points outwards
+    fluxᵀn.moisture.ρq_tot -=
+        C_q * state_sfc⁻.ρ * normu_int⁻_tan * (q_tot_sfc - q_tot_int)
+    fluxᵀn.ρ -= C_q * state_sfc⁻.ρ * normu_int⁻_tan * (q_tot_sfc - q_tot_int)
+    fluxᵀn.ρu -= C_q * normu_int⁻_tan * (q_tot_sfc - q_tot_int) .* state_sfc⁻.ρu
```
szy21, line 114: Should we use the density at the surface or at the interior point here? Does it matter? @tapios @akshaysridhar

szy21:
I ran a small domain bomex (3x3x3km) and the results show some strong element imprinting in variance and covariance. There is imprinting in bomex with prescribed flux too, but it's weaker. I think what I have changed should only affect the fluxes at the surface though?

image

src/Atmos/Model/bc_moisture.jl
tapios: Good question. What would make most sense is using an average density between surface and interior point (for a "central difference approximation"). The problem is that the surface density is probably not accurate because we are not resolving the surface layer. I suggest you write a function for the density, which for now can return either just the interior density, or the mean of interior and surface densities. It does not matter much what you use. If we want something more precise one day, we can then modify that function.

szy21: @charleskawczynski where would be a good place to add a function like this? boundaryconditions.jl? or relations.jl?

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -206,6 +208,10 @@ function atmos_normal_boundary_flux_second_order!(
     turbconv_normal_boundary_flux_second_order!(nf, bc.turbconv, atmos, args...)
 end
 
+function average_density_sfc_int(ρ_sfc::FT, ρ_int::FT) where {FT <: Real}
+    return 0.5 * (ρ_sfc + ρ_int)
```
charleskawczynski, line 212:
```suggestion
    return FT(0.5) * (ρ_sfc + ρ_int)
```

szy21: Ah right, thanks!

src/Atmos/Model/bc_moisture.jl
charleskawczynski: I think boundaryconditions.jl is the right place.

szy21:
bors try

szy21:
bors try

szy21:
bors try

szy21:
@akshaysridhar It seems I don't have the permission. Could you do `bors r+`? Thanks!

szy21:
@charleskawczynski @akshaysridhar I have added a driver (bomex_bulk_sfc_flux.jl). Please let me know what you think. Thanks!

PR #1411, opened 2020-08-06
add Performance Benchmarks to docs

jiahe23:
# Description

This PR adds a `Performance benchmarks` section to the documentation. The GPU scaling for GCM simulations is added in this section.

Ideally, this should be able to automatically updated once there are changes in the code. For now, a temporary solution is to add existing test results in the documentation.

I have

### DIFF
--- a/docs/list_of_performance_benchmarks.jl
+++ b/docs/list_of_performance_benchmarks.jl
@@ -1,0 +1,8 @@
+####
+#### Defines list of performance benchmarks
+####
+
+performance_benchmarks =
+    Any["AtmosGCM" => "PerformanceBenchmarks/AtmosGCM-GPUscaling.md",
+    # add LES performance md here
+    ]
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -25,6 +25,8 @@ include("list_of_apis.jl")                      # defines `apis`
 include("list_of_theory_docs.jl")               # defines `theory_docs`
 include("list_of_dev_docs.jl")                  # defines `dev_docs`
 
+include("list_of_performance_benchmarks.jl")    # defines `performance_benchmarks`
+
 pages = Any[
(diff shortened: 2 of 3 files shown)


jiahe23:
> I don't know if this is the right place for reporting performance benchmarks. I think we should probably just add a link somewhere in the docs that points to a performance monitoring / benchmarks page that @kpamnany started to work on.

Is it going to be part of the tutorial page? (@tapios first suggested to put it in the tutorial page under each experiment but I found the literate makes it a bit complicated, plus a separate page collects performances of major experiments looks clear and straightforward, e.g., the [oceananigan's docs](https://clima.github.io/OceananigansDocumentation/stable/) here. but I'm open to other options)

charleskawczynski:
> > I don't know if this is the right place for reporting performance benchmarks. I think we should probably just add a link somewhere in the docs that points to a performance monitoring / benchmarks page that @kpamnany started to work on.
> 
> Is it going to be part of the tutorial page? (@tapios first suggested to put it in the tutorial page under each experiment but I found the literate makes it a bit complicated, plus a separate page collects performances of major experiments looks clear and straightforward, e.g., the [oceananigan's docs](https://clima.github.io/OceananigansDocumentation/stable/) here. but I'm open to other options)

No, I think this will be separate from our docs (but we can include a link to point to the perf monitoring page). I think there's flexibility in hosting and where we put content, but all of the performance monitoring should probably be grouped together. I think @kpamnany and @jakebolewski have ideas for this.

kpamnany:
The thing is that these numbers are with a particular snapshot of the ClimateMachine code. They would need to be verified/updated as code is added. Ideally, we'd do so in an automated fashion, but that infrastructure is TBD.

Until we get that up and working, we can add this as you have, but should probably mention the commit SHA with which these numbers were gathered.

Issue #1356, opened 2020-07-16
Separate the configuration from driver (e.g. BOMEX - LES v single stack)


LenkaNovak:
- Ocean folks already have something similar...

tapios:
What we want here is a separation of concerns, specifying separately, and in a composable way, 

- Configuration (e.g., LES, GCM, single-stack)
- Experiment details (initial conditions and top and bottom boundary conditions
- Parameters (both for planet and free parameters)

The driver should compose configuration, experiment details, and parameters into a model.

The purpose is for us to be able to run, e.g., LES and a single-stack configuration with the exact same forcing and boundary conditions (e.g., derived from GCM, such as the GCP runs). Prototyping this with BOMEX for a LES and single-stack (with EDMF parameterization) would be good.

bischtob:
Does this only apply to planar geometries? Example, the Held-Suarez forcing is specific to the sphere (at least the way it is coded up now).

tapios:
> Does this only apply to planar geometries? Example, the Held-Suarez forcing is specific to the sphere (at least the way it is coded up now).

Yes, it's not arbitrarily composable. HS is spherical as coded (but you could make it into a channel, for example, once we have a configuration for that).

PR #1415, opened 2020-08-06
Introduce `AbstractProblem` and `AtmosProblem`

kpamnany:
# Description

To separate the specification of initial and boundary conditions from the specification of the model. This is already done in the Ocean code.

Towards addressing #1356. The BOMEX prototype that uses this capability remains TBD.

I have


kpamnany:
bors try

kpamnany:
bors try

kpamnany:
bors try

simonbyrne:
This seems like a good solution.

blallen:
Right now `AtmosProblem` only has `boundarycondition`, `init_state_prognostic!`, and `init_state_auxiliary!`. One of the benefits of the problem concept is that you can store parameters related to those three functions in the problem itself, so that the model doesn't store parameters that aren't always used. Are you guys considering doing that?

kpamnany:
@blallen: please expand? What sort of parameters?

kpamnany:
@szy21: I've merged the two BOMEX drivers into one (in [this commit](https://github.com/CliMA/ClimateMachine.jl/pull/1415/commits/124a3fcc9051bee4be2e784dc95da7c08f9d8565)). Can you please take a look -- pull this branch -- and make sure I haven't messed it up?

szy21:
> this commit

Oh great! Let me take a look.

szy21:
@kpamnany BOMEX works, thanks! As I mentioned it might be better to have surface_flux = 'prescribed', 'bulk', 'monin-obukhov', etc.. This is not specific to BOMEX, the only reason we add this is right now there are no unit tests or drivers that use the new boundary condition functions, and we add it to BOMEX so the functions don't go stale.

I wonder in general what we should put in command line arguments? We will have many other things that people would like to change in the driver in the future. Previously it is done through a namelist, but I'm not sure what CliMA's plan is.

blallen:
> @blallen: please expand? What sort of parameters?

Here's what we have for the Ocean https://github.com/CliMA/ClimateMachine.jl/blob/124a3fcc9051bee4be2e784dc95da7c08f9d8565/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L249-L255

`Lˣ, Lʸ, H` are the dimensions of the physical domain and are used in `ocean_init_state!`
https://github.com/CliMA/ClimateMachine.jl/blob/124a3fcc9051bee4be2e784dc95da7c08f9d8565/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L286-L296

`τₒ` is used in the velocity boundary condition (this function is called in the BC code)
https://github.com/CliMA/ClimateMachine.jl/blob/124a3fcc9051bee4be2e784dc95da7c08f9d8565/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L318-L319

`λʳ, θᴱ` are used in the temperature boundary condition (this function is called in the BC code)
https://github.com/CliMA/ClimateMachine.jl/blob/124a3fcc9051bee4be2e784dc95da7c08f9d8565/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L336-L343

Maybe it's a better idea to have the parameters for the boundary conditions be attached to the BoundaryCondition structs themselves instead of in the problem. I'm not sure where AtmosModel stores the domain parameters right now.

kpamnany:
> @kpamnany BOMEX works, thanks! As I mentioned it might be better to have surface_flux = 'prescribed', 'bulk', 'monin-obukhov', etc.. This is not specific to BOMEX, the only reason we add this is right now there are no unit tests or drivers that use the new boundary condition functions, and we add it to BOMEX so the functions don't go stale.

Ah, I understand now. Will discuss this with you to properly generalize.

> I wonder in general what we should put in command line arguments? We will have many other things that people would like to change in the driver in the future. Previously it has been done through a namelist, but I'm not sure what CliMA's plan is.

We will add namelist-like functionality, but the exact form is not yet defined. I have a couple of ideas, but again, we can discuss.

charleskawczynski:
> Right now `AtmosProblem` only has `boundarycondition`, `init_state_prognostic!`, and `init_state_auxiliary!`. One of the benefits of the problem concept is that you can store parameters related to those three functions in the problem itself, so that the model doesn't store parameters that aren't always used. Are you guys considering doing that?

Idk that we need to put parameters inside the `Problem`. I think a convenient and flexible alternative, for now, would be to define methods for pieces we need and dispatching on the `Problem`, rather than making them properties of the `Problem`. If we see a pattern that many problems need certain parameters, we can add properties, but I think problems often have quite different configurations.

kpamnany:
@LenkaNovak: I've merged moist baroclinic wave into baroclinic wave. Can you please take a look?

kpamnany:
I'm going to defer demonstrating use of `AbstractProblem` to separate PRs:
- When the single stack BOMEX version is ready, @charleskawczynski and I will merge that into the current BOMEX.
- @LenkaNovak and I will soon open a PR that will create a generalized GCM driver that can be used by Held Suarez and baroclinic wave (as well as by other experiments she is developing) which will also make use of the problem abstraction.

PR #1420, opened 2020-08-07
Add kernel vars wrappers

charleskawczynski:
# Description

This PR adds kernel wrappers that wraps arrays with `Vars{}`. This, together with #1400, should help clean up some of the DG kernels (especially the numerical flux kernels), and thin the layer in which VariableTemplates is used.

Remaining pieces:

 - [ ] Add more type restrictions?

I have

### DIFF
--- a/src/Numerics/DGMethods/DGMethods.jl
+++ b/src/Numerics/DGMethods/DGMethods.jl
@@ -57,6 +57,7 @@ export DGModel,
 include("NumericalFluxes.jl")
 include("DGModel.jl")
 include("DGModel_kernels.jl")
+include("kernel_vars_wrappers.jl")
 include("create_states.jl")
 
 """
--- a/src/Numerics/DGMethods/DGModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGModel_kernels.jl
@@ -1,14 +1,14 @@
 using .NumericalFluxes:
-    numerical_flux_gradient!,
-    numerical_flux_first_order!,
-    numerical_flux_second_order!,
-    numerical_flux_divergence!,
-    numerical_flux_higher_order!,
-    numerical_boundary_flux_gradient!,
-    numerical_boundary_flux_first_order!,
-    numerical_boundary_flux_second_order!,
-    numerical_boundary_flux_divergence!,
-    numerical_boundary_flux_higher_order!
+    wrapper_numerical_flux_gradient!,
+    wrapper_numerical_flux_first_order!,
+    wrapper_numerical_flux_second_order!,
+    wrapper_numerical_flux_divergence!,
+    wrapper_numerical_flux_higher_order!,
+    wrapper_numerical_boundary_flux_gradient!,
+    wrapper_numerical_boundary_flux_first_order!,
+    wrapper_numerical_boundary_flux_second_order!,
+    wrapper_numerical_boundary_flux_divergence!,
+    wrapper_numerical_boundary_flux_higher_order!
 
 using ..Mesh.Geometry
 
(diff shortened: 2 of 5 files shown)


christophernhill:
@charleskawczynski not sure if this comment fits here, but it might be nice for there to be an interface to the kernel functions that includes passing in ```ijk``` and ```e``` from the calling level.

There are some intermediate terms that can be useful for diags and analysis (e.g. components and products involving mixing tensors, advective products etc... ) that are only held locally in balance law engine. Long term we probably will want some generic way to add optional hooks that can get these values both for debugging and diagnostics. Being able to call back to an arbitrary function with the value(s) and ijk, e would allow this. 

At moment, because ijk and e are not visible you can get arbitrary intermediate values by inserting a user function call in a balance law, but you can't really tell where they are in the mesh and if you want to store them they don't have a convenient index available - which makes debugging hard! 

```coords``` doesn't help too much for this (although it is important too) because it is hard to go from ```coords``` to storing something at a data structure index location.

charleskawczynski:
> it might be nice for there to be an interface to the kernel functions that includes passing in `ijk` and `e` from the calling level. ... but you can't really tell where they are in the mesh... `coords` doesn't help too much for this ...

I agree that this may be a good idea for debugging, particularly for the numerics. Generally speaking, I don't know that we want to expose numerics internals to the physics kernels. We do have an [Orientations module](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Common/Orientations/Orientations.jl), which has `altitude`, `latitude` and `longitude` methods for providing spatial information but it doesn't really help with knowing which element/node we're on and, since a lot of debugging probably happens at the physics kernel level, I agree that perhaps we should think about offering some way to peek into numerics info. That said, this debuggability enhancement would probably fit better in a separate issue/PR.

kpamnany:
These functions look to me like they could be generated?

jkozdon:
> @charleskawczynski not sure if this comment fits here, but it might be nice for there to be an interface to the kernel functions that includes passing in `ijk` and `e` from the calling level.

We could possibly add some named tuple that passes all the info necessary for the user to shoot themselves in the foot. For instance, we could even pass the underlying data arrays in case extra data was needed.

I started to get at this with #1323, but then abandon it for no particularly good reason.

charleskawczynski:
> These functions look to me like they could be generated?

I'm not really sure how to do this.

> Also, the kernels need to be modified to remove the vars and grads wrapping for these functions to ever be used. (right now I think they never get called)

Yeah, I did that in a separate branch (I wanted to see how much code ends up being touched) but going to incorporate the changes very soon (running CI locally now).

I did run into two snags:

 - I tried overloading these methods (e.g., `flux_first_order!`) but I ran into method ambiguities in several driver files which provide types only for a concrete balance law (e.g., `function flux_first_order!(::MoistureModel, _...) end`). To avoid touching too much code, I prepended these wrappers with `wrapper_` (maybe `backend_` is a better name) to avoid these method ambiguities

 - I was not able to remove `Vars{}` from `update_auxiliary_state!` and `courant` because function pointers of their nodal counterparts (of which there are many) are passed in. It may be possible, but I didn't see a simple way at first glance

blallen:
> > @charleskawczynski not sure if this comment fits here, but it might be nice for there to be an interface to the kernel functions that includes passing in `ijk` and `e` from the calling level.
> 
> We could possibly add some named tuple that passes all the info necessary for the user to shoot themselves in the foot. For instance, we could even pass the underlying data arrays in case extra data was needed.
> 
> I started to get at this with #1323, but then abandon it for no particularly good reason.

Bring back #1323! I liked that a lot :)

charleskawczynski:

PR #1421, opened 2020-08-10
Rename `nodal_update_auxiliary_state!` for `MPIStateArray` methods

charleskawczynski:
# Description

It's counter intuitive that `nodal_update_auxiliary_state!` is defined for both `Vars` _and_ `MPIStateArrays`, while `update_auxiliary_state!` is only defined for `MPIStateArrays`. I think a more consistent naming convention convention would be
 - `nodal_update_auxiliary_state!` defined for `Vars` only
 - `update_auxiliary_state!` defined for `MPIStateArrays` only

This PR renames the appropriate `nodal_update_auxiliary_state!` methods to conform to this naming convention.

I have


blallen:
So is this just renaming `nodal_update_auxiliary_state!` to `update_auxiliary_state!`?

The function you are calling in `update_auxiliary_state!` is named by the user afaict. In the HBModel, we are using the very informative name `f!` for this function:  https://github.com/CliMA/ClimateMachine.jl/blob/100a6394d40e4293e592f05840fc08ebae7d6731/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl#L657-L666

charleskawczynski:
> So is this just renaming nodal_update_auxiliary_state! to update_auxiliary_state!?

Yes.

simonbyrne:
Maybe add a formal deprecation (using `@deprecate`)?

blallen:
> > So is this just renaming nodal_update_auxiliary_state! to update_auxiliary_state!?
> 
> Yes.

Sounds good to me then. As discussed in the meeting, we might want to promote the inner `nodal_update_auxiliary_state!` to be defined for all balance laws by default.

charleskawczynski:
> Maybe add a formal deprecation (using `@deprecate`)?

Done.

> Sounds good to me then. As discussed in the meeting, we might want to promote the inner nodal_update_auxiliary_state! to be defined for all balance laws by default.

I ended up adding more changes:

 - `init_state_auxiliary!(balance_law::BalanceLaw, state_auxiliary, grid)` now calls `init_state_auxiliary!(balance_law, init_f!, state_auxiliary, grid)` with `init_f! = nodal_init_state_auxiliary!` by default

and same for the update method:

 - `update_auxiliary_state!(dg::DGModel,balance_law::BalanceLaw,state_prognostic,t,elems)` now calls `update_auxiliary_state!(f!,dg::DGModel,m::BalanceLaw,state_prognostic::MPIStateArray,t::Real,elems::UnitRange = dg.grid.topology.realelems)` with `f! = nodal_update_auxiliary_state!` by default.

This way, users have the option to only overload the nodal init/update methods if their init/update is purely pointwise (which is the case for many tests, hence why the PR deletes ~200 lines).

TLDR: MPIStateArray-based init/update methods now call nodal vars-based methods be default, so that users have the option to overload either one.

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -818,8 +831,31 @@ function reverse_indefinite_stack_integral!(
...
+        balance_law,
+        state_prognostic,
+        t,
+        elems,
+    )
+end
+
+# Should we provide a fallback implementation here?
+# Maybe better to throw a method error?
+function nodal_update_auxiliary_state!(balance_law, state, aux, t) end
```
charleskawczynski, line 856: @simonbyrne we could remove this definition / throw a method error so that users must define `update_auxiliary_state!` or `nodal_update_auxiliary_state!` (and same for the init methods) Thoughts?

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
```diff
@@ -671,7 +671,7 @@ function update_auxiliary_state_gradient!(
 
         return nothing
     end
-    nodal_update_auxiliary_state!(f!, dg, m, Q, t, elems; diffusive = true)
+    update_auxiliary_state!(f!, dg, m, Q, t, elems; diffusive = true)
```
blallen, line 674: If I renamed `f!` to `nodal_update_auxiliary_state!`, could I use the new function signature you added? Also, could I move the definition of `f!` outside of `update_auxiliary_state_gradient!` and still have it work?

charleskawczynski: > If I renamed f! to nodal_update_auxiliary_state!, could I use the new function signature you added

Good point. Not at the moment, but we should be able to do that if we add/propagate that kwarg. I'll update.

> Also, could I move the definition of `f!` outside of update_auxiliary_state_gradient! and still have it work?
Yes, I believe so if we change `f!` to `nodal_update_auxiliary_state!`.

charleskawczynski: Alright, done.

src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl
```diff
@@ -662,16 +671,15 @@ function update_auxiliary_state_gradient!(
...
-    nodal_update_auxiliary_state!(f!, dg, m, Q, t, elems; diffusive = true)
+    update_auxiliary_state!(
+        nodal_update_auxiliary_state!,
+        dg,
+        m,
+        Q,
+        t,
+        elems;
+        diffusive = true,
+    )
```
blallen, line 682: looking good! What needs to be done to drop the first argument and get this on one line? 😃

charleskawczynski: Looks like we can do
```julia
    f! = nodal_update_auxiliary_state!
    update_auxiliary_state!(f!, dg, m, Q, t, elems; diffusive = true)
```

blallen: can't you just update the `update_auxiliary_state!(dg::DGModel,balance_law::BalanceLaw,state_prognostic,t,elems)` method with the `diffusive` kwarg? we have pretty similar interfaces everywhere.

charleskawczynski: I think everything is in shape.

charleskawczynski:
Fixed a couple of missed cases. bors r+

PR #1422, opened 2020-08-10
Add `ODESolvers.getsteps()`

### LINKED ISSUE #1419: Add i-th timestep to ODESolvers
# Description

For diagnostic / debugging purposes, we should add the i-th timestep to the solver somewhere, similar to time and `dt`.

# Additional context

# For CLIMA Developers

kpamnany:
# Description

An alternative way to do this that isn't invasive of the `ODESolver` types is to make [this `step`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/ODESolvers/ODESolvers.jl#L113) a global `const Ref`. Thoughts?

Closes #1419. 

I have

### DIFF
--- a/docs/src/APIs/Numerics/ODESolvers/ODESolvers.md
+++ b/docs/src/APIs/Numerics/ODESolvers/ODESolvers.md
@@ -90,8 +90,9 @@ ODESolvers.DiffEqJLSolver
 
 ```@docs
 ODESolvers.solve!
-ODESolvers.gettime
 ODESolvers.updatedt!
+ODESolvers.gettime
+ODESolvers.getsteps
 ```
 
 ## Generic Callbacks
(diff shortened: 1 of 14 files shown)


jkozdon:
> An alternative way to do this that isn't invasive of the `ODESolver` types is to make [this `step`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/ODESolvers/ODESolvers.jl#L113) a global `const Ref`. Thoughts?

Would this work when we have multiple steppers running side by side though?

kpamnany:
> Do we want this to be called somewhere in the tests to make sure the interface keeps working?

I'll look for a place to add this.

> Would this work when we have multiple steppers running side by side though?

Great point.

Issue #1417, opened 2020-08-07
Clean up boundary conditions


szy21:
# Description

A few things need to be cleaned up in boundary conditions:

- [ ] rename state to state_sfc, state1 to state_int, aux to aux_sfc, aux1 to aux1_int, etc.
- [ ] use projection_tangential to calculate u when needed, rename Pu to u_int
- [ ] use an average rho between surface and interior when needed

@akshaysridhar anything else?

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

PR #1424, opened 2020-08-10
clean up boundary conditions

szy21:
# Description

Address #1417 

I have

### DIFF
--- a/src/Atmos/Model/bc_momentum.jl
+++ b/src/Atmos/Model/bc_momentum.jl
@@ -109,8 +109,9 @@ function atmos_momentum_normal_boundary_flux_second_order!(
     DragLaw(fn) :: MomentumDragBC
 
 Drag law for momentum parallel to the boundary. The drag coefficient is
-`C = fn(state, aux, t, normPu_int)`, where `normPu_int⁻` is the internal speed
+`C = fn(state, aux, t, normu_int_tan)`, where `normu_int_tan` is the internal speed
 parallel to the boundary.
+`_int` refers to the first interior node.
 """
 struct DragLaw{FN} <: MomentumDragBC
     fn::FN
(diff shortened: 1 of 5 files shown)


szy21:
> I might suggest reverting the `_sfc` suffix. I think these states represent the surface for some `bctype`s, but they represent other boundaries for different `bctype`s.

I don't have much preference. @akshaysridhar any suggestion? What are the `bctypes` where the states don't represent the surface?

src/Atmos/Model/bc_momentum.jl
```diff
@@ -148,28 +148,28 @@ function atmos_momentum_normal_boundary_flux_second_order!(
...
+    state_sfc⁺,
+    diffusive_sfc⁺,
+    hyperdiffusive_sfc⁺,
+    aux_sfc⁺,
     bctype,
     t,
-    state1⁻,
-    diff1⁻,
-    aux1⁻,
+    state_int⁻,
```
akshaysridhar, line 161: Suggest adding a doc-string / comments to clarify these arguments; e.g. `_int` refers to first interior node; other than that this seems reasonable.

szy21:
Sounds good. I will add the comments. What do we do for state_sfc? just state? state_bdy? (I'm making up abbreviations) If you don't have strong preference I could also keep it as state_sfc?

charleskawczynski:
I think `state` is fine (these are in the `boundary_state` method after all), but I don't have a strong preference. I think `state_sfc` is fine if it makes sense to users at TOA-- I just wanted to make sure it was clear that the boundary state is for all boundaries and not just at altitude = 0.

szy21:
> I think `state` is fine (these are in the `boundary_state` method after all), but I don't have a strong preference. I think `state_sfc` is fine if it makes sense to users at TOA-- I just wanted to make sure it was clear that the boundary state is for all boundaries and not just at altitude = 0.

Sounds good, I will revert it to state then.

szy21:
bors try

PR #1433, opened 2020-08-11
Add fixed_number_of_steps option to driver

jkozdon:
For performance analysis runs and testing it is often useful to force a fixed
number of time steps. To achieve this is the addition of the keyword
argument `fixed_number_of_steps` to `SolverConfiguration`; this can also
be set with the command line argument `--fixed-number-of-steps`.

If `fixed_number_of_steps < 0` then the original number of steps
(inferred from `tend`) is used, otherwise the number of specified by
`fixed_number_of_steps` is used.

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -70,6 +70,7 @@ Base.@kwdef mutable struct ClimateMachine_Settings
     debug_init::Bool = false
     integration_testing::Bool = false
     array_type::Type = Array
+    fixed_number_of_steps::Int = -1
 end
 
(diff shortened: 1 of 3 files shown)


jkozdon:
address comment of @charleskawczynski: https://github.com/CliMA/ClimateMachine.jl/issues/1419#issuecomment-670647193

jkozdon:
bors try

jkozdon:
> I guess `fixed_number_of_steps == 0` is useful?

Wasn't sure. But didn't seem like a terrible idea...

PR #1434, opened 2020-08-12
Add atmos ref state plots to docs

charleskawczynski:
# Description

This PR adds some plots of the default atmospheric reference state for multiple polynomial orders and numbers of elements. I'm hoping that this will help us monitor the status of the reference state progress.

I had to add a driver kwarg to skip calling `update_auxiliary_state!` because it was trying to use uninitialized prognostic fields. cc @kpamnany

I have


src/Driver/solver_configs.jl
```diff
@@ -86,6 +86,7 @@ function SolverConfiguration(
     diffdir = EveryDirection(),
     timeend_dt_adjust = true,
     CFL_direction = EveryDirection(),
+    skip_update_aux = false,
```
kpamnany, line 89: This is getting pretty ugly. :(

We were looking into removing the update call altogether -- is that really not possible?

charleskawczynski: Agreed. This was the simplest thing I could think of without duplicating the first half of `SolverConfiguration`

jkozdon:
Would it be worth making this a test in some way so that when the reference state changes it requires approval? (For instancing, saving the data and comparing it with each new PR.)

christophernhill:
> Would it be worth making this a test in some way so that when the reference state changes it requires approval? (For instancing, saving the data and comparing it with each new PR.)

@jkozdon the stuff in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Diagnostics/Debug/StateCheck.jl might be useful for this. I am not sure if it can get traction outside ocean group though!

jkozdon:
> I am not sure if it can get traction outside ocean group though!

I've advocated for it before! https://github.com/CliMA/ClimateMachine.jl/pull/1340#issuecomment-656427101

charleskawczynski:
> Would it be worth making this a test in some way so that when the reference state changes it requires approval? (For instance, saving the data and comparing it with each new PR.)

I think both would be useful. We do have some tests [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Atmos/Model/ref_state.jl), but it changed a bit in #1361 and I'd like to make sure that it's fully hooked into the default atmos behavior. Perhaps I'll update that here.

> StateCheck.jl might be useful for this

Agreed. I connect with @blallen to make sure I use it correctly.

src/Driver/solver_configs.jl
blallen: Wait, why is the `update_aux` call needed here? Did you guys add something that requires it? I'm pretty sure I removed any need for it from the Ocean side

charleskawczynski: The only reason it's needed inside this function is because the atmos courant computations depend on aux variables that depend on the prognostic variables. The init aux method only initializes aux variables that are independent of the prognostic variables-- the update aux method basically initializes the remaining aux variables. The problem is that we cannot use this method (`SolverConfiguration`) without defining prognostic fields, and this unit test should not be concerned with the prognostic field because it's only concerned with aux initialization.

blallen: I ran into the same issue where I wanted an aux state that depends on the state variables. @jkozdon and I were able to find a workaround that didn't involve changing the DGModel structure too much, but, if there is more than one use case for aux depending on the state, it might make more sense move the initialization from the `DGModel` constructor [1] to the `init_ode_state` function [2]. I'm not sure how invasive this would be; the main issue I can see is if we have initial states that depend on auxiliary variables.

[1] https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel.jl#L40-L41
[2] https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel.jl#L585-L653

kpamnany: This sounds to me like a case for having a separate group of auxiliary variables -- splitting them between those that are independent of the prognostic variables and those that aren't.

blallen: @jkozdon mentioned they had considered that in the past but never went forward with it

I think that would be a good idea because we definitely have two cases of auxiliary variables: those that are constant in time and those that aren't. It would be nice to split them into two separate concepts.

jkozdon: We were initially thinking about it in terms of those that do and do not need to be communicated, but this seems like a reasonable use case as well (the comm issue doesn't matter anymore since @lcw made it so that aux is never communicated after initialization).

charleskawczynski:
I tried to use StateCheck, but it seemed a bit of a manual process, and I only want to check certain fields, and I'd rather not need to update things when new auxiliary fields are added to AtmosModel. So, I've updated this PR to simply use JLD2 and Artifacts. The only remaining issue I see is `skip_update_aux`.

kpamnany:
> The only remaining issue I see is `skip_update_aux`.

Perhaps a separate PR that splits `aux` into two per [this comment](https://github.com/CliMA/ClimateMachine.jl/pull/1434#discussion_r469976585)? I have no clue what would be appropriate names, but anyway, that PR can clean up all this update aux stuff.

charleskawczynski:
> Perhaps a separate PR that splits `aux` into two per [this comment](https://github.com/CliMA/ClimateMachine.jl/pull/1434#discussion_r469976585)? I have no clue what would be appropriate names, but anyway, that PR can clean up all this update aux stuff.

I agree that's a good idea, but maybe we can merge this in the mean time? I was really hoping to remove the "[mega hack](https://github.com/CliMA/ClimateMachine.jl/pull/1434/files#diff-5e87599165833809de99132c510d213cL56-L62)", since this effectively decouples the behavior in `AtmosModel` from the test.

kpamnany (approved):
LGTM.

src/Driver/solver_configs.jl
charleskawczynski: Going to leave this for now and tackle removing it as a part of closing #1551.

PR #1308, opened 2020-06-29
Add EDMF model

charleskawczynski:
# Description

Adds the EDMF model for a BOMEX single stack driver.

We recently had things working on the GPU ~~but it's currently CPU-only. We'll fix the gpu capability in a subsequent PR~~, and we still do 🚀.

Co-authored with: @yairchn, @trontrytel, @ilopezgp, @jiahe23, @simonbyrne, @kpamnany, @jkozdon (and likely more, since I needed a lot of tangential advice on various things)

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [x] There are no open pull requests for this already
- [x] CLIMA developers with relevant expertise have been assigned to review this submission
- [x] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [x] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1437, opened 2020-08-12
Add GPU tests for VariableTemplates

### LINKED ISSUE #1413: Cannot unroll updraft loops on GPU
# Description

Just to track this, we're not yet able to unroll updraft loops on the gpu (getting `getindex` error). This is happening in [this branch](https://github.com/CliMA/ClimateMachine.jl/tree/ck/EDMFInSrc).

## Additional context

# For CLIMA Developers

charleskawczynski:
# Description

This PR adds a GPU test for VariableTemplates which performs a memcopy for different models.

Co-authored by @simonbyrne

Closes #1413

I have


test/Utilities/VariableTemplates/test_complex_models_gpu.jl
```diff
@@ -0,0 +1,110 @@
...
+
+function mem_copy!(m::NTupleContainingModel{N}, dst::Vars, src::Vars) where {N}
+    dst.vector_model.x = src.vector_model.x
+    dst.scalar_model.x = src.scalar_model.x
+
+    up = ntuple(i -> src.ntuple_model[i].scalar_model.x, N)
+
+    unroll_map(N) do i
+        dst.ntuple_model[i].scalar_model.x = up[i]
+    end
```
jkozdon, line 55: I ran into this with the remainder model, and @vchuravy gave me the magic sauce. `ntuple' with `Val` and `Base.@_inline_meta`
```suggestion
    @inbounds ntuple(Val(N)) do i
        Base.@_inline_meta
        dst.ntuple_model[i].scalar_model.x = up[i]
    end
```
which removes the need for the generated function.

test/Utilities/VariableTemplates/test_complex_models_gpu.jl
```diff
@@ -0,0 +1,110 @@
...
+    dst.x = src.x
+end
+
+unroll_map(f, N::Int, args...) = unroll_map(f, Val(N), args...)
+@generated function unroll_map(f::F, ::Val{N}, args...) where {F, N}
+    quote
+        Base.@_inline_meta
+        Base.Cartesian.@nexprs $N i -> f(i, args...)
+    end
+end
```
jkozdon, line 45: Should this be imported from `VariableTemplates`?

test/Utilities/VariableTemplates/test_complex_models_gpu.jl
```diff
@@ -0,0 +1,110 @@
+using Test
+using StaticArrays
+using ClimateMachine.VariableTemplates
+using CUDA
+using KernelAbstractions
+using KernelAbstractions.Extras: @unroll
+CUDA.allowscalar(false)
+# CUDA.allowscalar(true) # for CPU tests
+
```
jkozdon, line 9: So the code can be run with just this file and not the runtests

```suggestion
include("complex_models.jl")
```

test/Utilities/VariableTemplates/runtests_gpu.jl
```diff
@@ -0,0 +1,8 @@
+using Test
+using StaticArrays
+using ClimateMachine.VariableTemplates
+
+@testset "VariableTemplates - GPU" begin
+    include("complex_models.jl")
+    include("test_complex_models_gpu.jl")
```
jkozdon, line 7:
```suggestion
    include("test_complex_models_gpu.jl")
```

src/Utilities/VariableTemplates/VariableTemplates.jl
```diff
@@ -374,4 +374,27 @@ Base.@propagate_inbounds function Base.getindex(
...
+end
+```
+"""
+unroll_map(f, N::Int, args...) = unroll_map(f, Val(N), args...)
+@generated function unroll_map(f::F, ::Val{N}, args...) where {F, N}
+    quote
+        Base.@_inline_meta
+        Base.Cartesian.@nexprs $N i -> f(i, args...)
+    end
+end
```
jkozdon, line 398: Same comment as below: https://github.com/CliMA/ClimateMachine.jl/pull/1437/files#r469604066 about not needing a generated function

test/Utilities/VariableTemplates/test_complex_models_gpu.jl
charleskawczynski: Oh wow! I thought we may have been missing magic sauce 😄. This works. In fact, after playing around with this, I found that simply
```julia
    ntuple(N) do i
        dst.ntuple_model[i].scalar_model.x = up[i]
    end
```
also works! Thanks @jkozdon! I'll remove the generated function 🚀

charleskawczynski: Yeah, although I don't know why it wouldn't unroll automatically since `N` is passed as a type parameter. Unfortunately, while this is working fine for the toy example here, _none_ of these solutions are working for the actual use-case that this was all designed for. Branch [here](https://github.com/CliMA/ClimateMachine.jl/blob/ck/EDMFInSrcGPUUnroll), driver [here](https://github.com/CliMA/ClimateMachine.jl/blob/ck/EDMFInSrcGPUUnroll/test/Atmos/EDMF/bomex_edmf.jl), and observed failure [here](https://github.com/CliMA/ClimateMachine.jl/blob/ck/EDMFInSrcGPUUnroll/test/Atmos/EDMF/edmf_kernels.jl#L305). 😞 @vchuravy I'm probably going to need your help with this one!

charleskawczynski: Update! Fortunately, there was an easy fix: the `unroll_map` ended up working, but needed a tweak:

```julia
@generated function unroll_map(f::F, ::Val{N}, args...) where {F, N}
    quote
        Base.@_inline_meta
        Base.Cartesian.@nexprs $N i -> f(Val(i), args...)              < ---------- need to wrap i with Val
    end
end
```
And we needed to modify the `getindex` accordingly to accept `Val{i}... where {i}`. So relieved!

jkozdon: Can you use the `Val(i)` syntax with the `ntuple` syntax?

charleskawczynski: Trying...

charleskawczynski: No,
```
    @inbounds ntuple(Val(N_up)) do i
        Base.@_inline_meta
        up_a[Val(i)].buoyancy = eps(FT)
    end
```
fails. I guess that by the time that `Val` is called, it's too late and has already left the type domain.

src/Utilities/VariableTemplates/VariableTemplates.jl
```diff
@@ -315,11 +315,22 @@ end
...
+unroll_map(f::F, N::Int, args...) where {F} = unroll_map(f, Val(N), args...)
+@generated function unroll_map(f::F, ::Val{N}, args...) where {F, N}
+    quote
+        Base.@_inline_meta
+        Base.Cartesian.@nexprs $N i -> f(Val(i), args...)
+    end
+end
+
+Base.@propagate_inbounds Base.getindex(v::AbstractVars, i::Int) =
+    Base.getindex(v, Val(i))
```
charleskawczynski, line 328: @jkozdon I'm not sure exactly why, but this was even needed for some reason (otherwise there was a method error).

jkozdon: Yeah, I dunno. I can take a crack at it next week if you want, or we can just run with the generated function. Fine by me either way.

test/Utilities/VariableTemplates/test_complex_models_gpu.jl
jkozdon: Bummer. Was just a thought.

I don't understand the whole GPU compile stack, so I would just be hacking at it like you are.

charleskawczynski: Something else is going on. Now I'm getting a failure somewhere else (when passing `i` through functions), and I think that the compiler is somehow, sometimes, able to eliminate `Base.getindex(v::AbstractVars, i::Int)`, but sometimes not. Still debugging to figure out what an end-goal solution is. Updates to come.

src/Utilities/VariableTemplates/VariableTemplates.jl
vchuravy: Yeah in the method below `i` is used for an operation on types so this makes sense.

test/Utilities/VariableTemplates/test_complex_models_gpu.jl
charleskawczynski: Okay, I figured out _what_ was triggering a dynamic runtime call, but I don't know exactly _why_. It's a bit difficult to explain, and there seem to be a lot of edge cases where certain things work in some circumstances and others do not. Ultimately, for the use-case, the `unroll_map` with `Val` seems needed. Plus a few more pieces. Going to expand on this.

charleskawczynski:
Okay, I managed to fix enough things to get through the update aux kernel, which touches quite a bit of code (yay!). I removed the fallback ``. But, idk how I feel about overriding
```julia
Base.getindex(t::Tuple, ::Val{i}) where {i} = Base.getindex(t, i)
Base.getindex(a::SArray, ::Val{i}) where {i} = Base.getindex(a, i)
```
since we need to be able to index into tuples and `SArray`'s inside `unroll_map`. I ended up removing
```julia
Base.@propagate_inbounds Base.getindex(v::AbstractVars, i::Int) =
    Base.getindex(v, Val(i))
```
to ensure that we never call `getindex` with an `Int`, and always call using `Val`.

charleskawczynski:
After an attempt with @vchuravy, we were unable to improve on the existing solution, which seems to work in the EDMF model.

charleskawczynski:
This PR shouldn’t affect anything except the EDMF model, and I’d like to rebase #1308 tomorrow and start gpu-ifying things, so I’m going to merge this for now.

bors r+

PR #1440, opened 2020-08-14
Precipitation removal scheme for GCM

trontrytel:
# Description

This is a draft of precipitation removal scheme for the GCM (aka the 0-moment microphysics scheme).

I have

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -49,7 +49,7 @@ WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 [compat]
 Adapt = "2.0.2"
 ArgParse = "1.1"
-CLIMAParameters = "0.1.6"
+CLIMAParameters = "0.1.8"
 CUDA = "1.2"
 Combinatorics = "1.0"
 Coverage = "1.0"
--- a/docs/src/Theory/Atmos/Microphysics_0M.md
+++ b/docs/src/Theory/Atmos/Microphysics_0M.md
@@ -1,0 +1,81 @@
+# Microphysics_0M
+
+The `Microphysics_0M.jl` module defines a 0-moment bulk parameterization of
+  the moisture sink due to precipitation.
+It offers a simplified way of removing the excess water
+  without assuming anything about the size distributions of cloud
+  or precipitation particles.
+
+The ``q_{tot}`` (total water specific humidity) sink due to precipitation
+  is obtained by relaxation with a constant timescale
+  to a state with condensate exceeding a threshold value removed.
+The threshold for removing excess ``q_{tot}`` is defined either by the
+  condensate specific humidity or supersaturation.
+The thresholds and the relaxation timescale are defined in
+  `CLIMAParameters.jl`.
+
+!!! note
+
+    To remove precipitation instantly, the relaxation timescale should be
+    equal to the timestep length.
+
+## Moisture sink due to precipitation
+
+If based on maximum condensate specific humidity, the sink is defined as:
+``` math
+\begin{equation}
+  \left. \mathcal{S}_{q_{tot}} \right|_{precip} =-
+    \frac{max(0, q_{liq} + q_{ice} - q_{c0})}{\tau_{precip}}
+\end{equation}
+```
+where:
+  - ``q_{liq}``, ``q_{ice}`` are cloud liquid water and cloud ice specific humidities,
+  - ``q_{c0}`` is the condensate specific humidity threshold above which water is removed,
+  - ``\tau_{precip}`` is the relaxation timescale.
+
+If based on saturation excess, the sink is defined as:
+```math
+\begin{equation}
+  \left. \mathcal{S}_{q_{tot}} \right|_{precip} =-
+    \frac{max(0, q_{liq} + q_{ice} - S_{0} \, q_{vap}^{sat})}{\tau_{precip}}
+\end{equation}
+```
+where:
+  - ``q_{liq}``, ``q_{ice}`` are cloud liquid water and cloud ice specific humidities,
+  - ``S_{0}`` is the supersaturation threshold above which water is removed,
+  - ``q_{vap}^{sat}`` is the saturation specific humidity,
+  - ``\tau_{precip}`` is the relaxation timescale.
+
+## Coupling to the state variables
+
+Following the conservation equations for
+[moisture](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Moisture)
+and [mass](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Mass),
+the ``\mathcal{S}_{q_{tot}}`` sink has to be multiplied by ``\rho`` before
+  adding it as one of the sink terms to both moisture and mass state variables.
+For the conservation equation for
+[total energy](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Energy),
+  no additional source/sink terms $M$ are considered, and the
+  the sink due to removing ``q_{tot}`` is computed as:
+```math
+\begin{equation}
+\left. \sum_{j\in\{v,l,i\}}(I_j + \Phi)  \rho C(q_j \rightarrow q_p) \right|_{precip} =
+  \left[\lambda I_l + (1 - \lambda) I_i + \Phi \right]
+  \rho \, \left.\mathcal{S}_{q_{tot}} \right|_{precip}
+\end{equation}
+```
+where:
+ - ``\lambda`` is the liquid fraction
+ - ``I_l = c_{vl} (T - T_0)`` is the internal energy of liquid water
+ - ``I_i = c_{vi} (T - T_0) - I_{i0}`` is the internal energy of ice
+ - ``T`` is the temperature,
+ - ``T_0`` is the thermodynamic reference temperature (which is unrelated to the reference temperature used in hydrostatic reference states used in the momentum equations),
+ - ``I_{i0}`` is the specific internal energy of ice at ``T_0``
+ - ``c_{vl}`` and ``c_{vi}`` are the isochoric specific heats
+     of liquid water, and ice.
+ - ``\Phi`` is the effective gravitational potential.
+
+This assumes that the ``\mathcal{S}_{q_{tot}}`` sink is partitioned between the
+  cloud liquid water and cloud ice sinks
+  ``\mathcal{S}_{q_{liq}}`` and ``\mathcal{S}_{q_{ice}}`` based on the
+  cloud liquid water and cloud ice fractions.
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -1,5 +1,13 @@
-using CLIMAParameters.Planet: Omega
-export Source, Gravity, RayleighSponge, Subsidence, GeostrophicForcing, Coriolis
+using ..Microphysics_0M
+using CLIMAParameters.Planet: Omega, e_int_i0, cv_d, cv_l, cv_i, T_0
+
+export Source,
+    Gravity,
+    RayleighSponge,
+    Subsidence,
+    GeostrophicForcing,
+    Coriolis,
+    RemovePrecipitation
 
 # kept for compatibility
 # can be removed if no functions are using this
@@ -171,5 +179,64 @@ function atmos_source!(
         r = (z - s.z_sponge) / (s.z_max - s.z_sponge)
         β_sponge = s.α_max * sinpi(r / 2)^s.γ
         source.ρu -= β_sponge * (state.ρu .- state.ρ * s.u_relaxation)
+    end
+end
+
+"""
+    RemovePrecipitation{FT} <: Source
+
+A sink to `q_tot` when cloud condensate is exceeding a threshold.
+The threshold is defined either in terms of condensate or supersaturation.
+The removal rate is implemented as a relaxation term
+in the Microphysics_0M module.
+The default thresholds and timescale are defined in CLIMAParameters.jl.
+"""
+struct RemovePrecipitation <: Source
+    " Set to true if using qc based threshold"
+    use_qc_thr::Bool
+end
+function atmos_source!(
+    s::RemovePrecipitation,
+    atmos::AtmosModel,
+    source::Vars,
+    state::Vars,
+    diffusive::Vars,
+    aux::Vars,
+    t::Real,
+    direction,
+)
+    FT = eltype(state)
+    ts = thermo_state(atmos, state, aux)
+    if has_condensate(ts)
+
+        q = PhasePartition(ts)
+        T::FT = air_temperature(ts)
+        λ::FT = liquid_fraction(ts)
+
+        _e_int_i0::FT = e_int_i0(atmos.param_set)
+        _cv_l::FT = cv_l(atmos.param_set)
+        _cv_i::FT = cv_i(atmos.param_set)
+        _T_0::FT = T_0(atmos.param_set)
+
+        S_qt::FT = 0
+        if s.use_qc_thr
+            S_qt = remove_precipitation(atmos.param_set, q)
+        else
+            q_vap_sat = q_vap_saturation(ts)
+            S_qt = remove_precipitation(atmos.param_set, q, q_vap_sat)
+        end
+
+        source.moisture.ρq_tot += state.ρ * S_qt
+
+        source.ρ += state.ρ * S_qt
+
+        source.ρe +=
+            (
+                λ * _cv_l * (T - _T_0) +
+                (1 - λ) * (_cv_i * (T - _T_0) - _e_int_i0) +
+                gravitational_potential(atmos.orientation, aux)
+            ) *
+            state.ρ *
+            S_qt
     end
 end
--- a/src/Atmos/Parameterizations/CloudPhysics/Microphysics_0M.jl
+++ b/src/Atmos/Parameterizations/CloudPhysics/Microphysics_0M.jl
@@ -1,0 +1,59 @@
+"""
+    Microphysics_0M
+
+Zero-moment bulk microphysics scheme that instantly removes
+moisture above certain threshold.
+This is equivalent to instanteneous conversion of cloud condensate
+into precipitation and precipitation fallout with infinite
+terminal velocity.
+
+"""
+module Microphysics_0M
+
+using ..Thermodynamics
+
+using CLIMAParameters
+using CLIMAParameters.Atmos.Microphysics_0M
+
+const APS = AbstractParameterSet
+
+export remove_precipitation
+
+"""
+    remove_precipitation(param_set::APS, q; q_vap_sat)
+
+ - `param_set` - abstract parameter set
+ - `q` - current PhasePartition
+ - `q_vap_sat` - water vapor specific humidity at saturation
+
+Returns the `q_tot` tendency due to the removal of precipitation.
+The tendency is obtained assuming a relaxation with a constant timescale
+to a state with precipitable water removed.
+The threshold for when to remove `q_tot` is defined either by the
+condensate specific humidity or supersaturation.
+The thresholds and the relaxation timescale are defined in
+CLIMAParameters.
+"""
+function remove_precipitation(
+    param_set::APS,
+    q::PhasePartition{FT},
+) where {FT <: Real}
+
+    _τ_precip::FT = τ_precip(param_set)
+    _qc_0::FT = qc_0(param_set)
+
+    return -max(0, (q.liq + q.ice - _qc_0)) / _τ_precip
+end
+function remove_precipitation(
+    param_set::APS,
+    q::PhasePartition{FT},
+    q_vap_sat::FT,
+) where {FT <: Real}
+
+    _τ_precip::FT = τ_precip(param_set)
+    _S_0::FT = S_0(param_set)
+
+    return -max(0, (q.liq + q.ice - _S_0 * q_vap_sat)) / _τ_precip
+end
+
+end #module Microphysics_0M.jl
--- a/docs/list_of_apis.jl
+++ b/docs/list_of_apis.jl
@@ -7,6 +7,7 @@ apis = [
     "Driver" => "APIs/Driver/index.md",
     "Atmos" => [
         "AtmosModel" => "APIs/Atmos/AtmosModel.md",
+        "Microphysics_0M" => "APIs/Atmos/Microphysics_0M.md",
         "Microphysics" => "APIs/Atmos/Microphysics.md",
         "Temperature Profiles" => "APIs/Atmos/TemperatureProfiles.md",
     ],
--- a/docs/list_of_theory_docs.jl
+++ b/docs/list_of_theory_docs.jl
@@ -9,6 +9,7 @@ theory_docs = Any[
     ],
     "Atmos" => Any[
         "AtmosModel" => "Theory/Atmos/AtmosModel.md",
+        "Microphysics_0M" => "Theory/Atmos/Microphysics_0M.md",
         "Microphysics" => "Theory/Atmos/Microphysics.md",
         "EDMF equations" => "Theory/Atmos/EDMFEquations.md",
         "Tracers" => "Theory/Atmos/Model/tracers.md",
--- a/docs/src/APIs/Atmos/AtmosModel.md
+++ b/docs/src/APIs/Atmos/AtmosModel.md
@@ -58,3 +58,9 @@ ClimateMachine.Atmos.Insulating
 ClimateMachine.Atmos.NoSlip
 ClimateMachine.Atmos.average_density
 ```
+
+## Sources
+
+```@docs
+ClimateMachine.Atmos.RemovePrecipitation
+```
--- a/docs/src/APIs/Atmos/Microphysics_0M.md
+++ b/docs/src/APIs/Atmos/Microphysics_0M.md
@@ -1,0 +1,15 @@
+# Microphysics_0M
+
+```@meta
+CurrentModule = ClimateMachine
+```
+
+```@docs
+Microphysics_0M
+```
+
+## Methods
+
+```@docs
+Microphysics_0M.remove_precipitation
+```
--- a/experiments/AtmosGCM/baroclinic-wave.jl
+++ b/experiments/AtmosGCM/baroclinic-wave.jl
@@ -168,9 +168,14 @@ function config_baroclinic_wave(FT, poly_order, resolution, with_moisture)
     if with_moisture
         hyperdiffusion = EquilMoistBiharmonic(FT(8 * 3600))
         moisture = EquilMoist{FT}()
+        # TODO - switch to line below if you want to start removing q_tot
+        #        due to precipitation
+        source = (Gravity(), Coriolis())
+        #source = (Gravity(), Coriolis(), RemovePrecipitation(true))
     else
         hyperdiffusion = DryBiharmonic(FT(8 * 3600))
         moisture = DryModel()
+        source = (Gravity(), Coriolis())
     end
     model = AtmosModel{FT}(
         AtmosGCMConfigType,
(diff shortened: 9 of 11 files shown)


src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,73 @@ function atmos_source!(
...
+
+        _e_int_i0::FT = e_int_i0(atmos.param_set)
+        _cv_d::FT = cv_d(atmos.param_set)
+        _cv_l::FT = cv_l(atmos.param_set)
+        _cv_i::FT = cv_i(atmos.param_set)
+        _T_0::FT = T_0(atmos.param_set)
+
+        source.moisture.ρq_tot += state.ρ / q_dry * dqt_dt
+
+        source.ρ += state.ρ / q_dry * dqt_dt
```
tapios, line 240:
```suggestion
        source.ρ += source.moisture.ρq_tot
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,92 @@
...
+  - `S_{0}` is the supersaturation threshold above which water is removed,
+  - `q_{vap}^{sat}` is the saturation specific humidity,
+  - `\tau_{precip}` is the relaxation timescale.
+
+## Coupling to the state variables
+
+Removing the `q_tot` changes the mass of the working fluid.
+Therefore the state variables `\rho` and `\rho q_{tot}` are by:
+
+```math
```
tapios, line 61: Better to express this as the source in the conservation law, which is not the same as the material derivative (but is what we need here). See Eq. (3.7) in the design doc.

src/Atmos/Model/source.jl
tapios: My preference would be to collect all mass sources owing moisture sources and add them as a density source where we specify the equations of motion, not piece-by-piece where we define parameterizations.

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,73 @@ function atmos_source!(
...
+            dqt_dt = remove_precipitation(atmos.param_set, q, q_vap_sat)
+        end
+
+        _e_int_i0::FT = e_int_i0(atmos.param_set)
+        _cv_d::FT = cv_d(atmos.param_set)
+        _cv_l::FT = cv_l(atmos.param_set)
+        _cv_i::FT = cv_i(atmos.param_set)
+        _T_0::FT = T_0(atmos.param_set)
+
+        source.moisture.ρq_tot += state.ρ / q_dry * dqt_dt
```
szy21, line 238: I wonder if this should be `source.moisture.ρq_tot += state.ρ * dqt_dt`? It seems `dqt_dt` here corresponds to `S_qt` in the design doc.

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,73 @@ function atmos_source!(
...
+
+        source.ρe +=
+            (
+                q.liq / (q.liq + q.ice) * (_cv_l - _cv_d) * (T - _T_0) +
+                q.ice / (q.liq + q.ice) *
+                ((_cv_i - _cv_d) * (T - _T_0) - _e_int_i0) +
+                state.ρe / state.ρ / q_dry
+            ) *
+            state.ρ *
+            dqt_dt
```
szy21, line 250: Would it be simpler to just write it as (according to Eq.3.13, also I don't think they are exactly the same):
```
source.ρe +=
            (
                q.liq / (q.liq + q.ice) * _cv_l * (T - _T_0) +
                q.ice / (q.liq + q.ice) *
                (_cv_i * (T - _T_0) - _e_int_i0) +
                gravitational_potential(atmos.orientation, aux)
            ) *
            state.ρ *
            dqt_dt
```

src/Atmos/Model/source.jl
trontrytel: updated, thanks!

src/Atmos/Model/source.jl
trontrytel: I have added the forgotten geopotential. I still think there should also be terms proportional to `-cv_d` there. I have updated the code and the docs. Let me know what you think now.

docs/src/Theory/Atmos/Microphysics_0M.md
trontrytel: I have updated the docs and the code to follow the notation from the DDs. I'm still not sure about the total energy source term. I will send the link to the documentation once it is generated.

src/Atmos/Model/source.jl
trontrytel: I have opened an issue. I'm no sure how to do it with the current piece-by-piece source term summation. 

issue https://github.com/CliMA/ClimateMachine.jl/issues/1466

src/Atmos/Model/source.jl
szy21: Did you remove the `state.ρe / state.ρ / q_dry` term as well? Now I'm confused by why it was there before.

trontrytel: Yes, I did remove it. I think it was there because of my confusion with the notation and derivatives. I think what you suggested is right, modulo the missing `cv_d` terms (?)

docs/src/Theory/Atmos/Microphysics_0M.md
trontrytel: https://clima.github.io/ClimateMachine.jl/previews/PR1440/Theory/Atmos/Microphysics_0M/

szy21: I looked at the docs. I'm still confused by the c_v,d term. For example when you remove q_liq (assuming no_ice) you lose the total energy of liquid. It looks strange to me that the source term is zero when c_v,l = c_v,d?

trontrytel: ok - I think this is wrong and you are right. Will update again!

trontrytel:
@LenkaNovak - This is the final (I hope) version of the precipitation removing scheme. Could you take it and see how it behaves in your tests?

@jakebolewski - Could you take a look if I didn't mess up something in the toml files? (I think the changes come from adding CLIMAParameters 1.8 only).

@Tapio, @szy21 - once the documentation preview is rendered I'll send you the link. Thank you for all the help with the equations!

trontrytel:
And the corresponding docs can be previewed here: https://clima.github.io/ClimateMachine.jl/previews/PR1440/Theory/Atmos/Microphysics_0M/

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+    source::Vars,
+    state::Vars,
+    diffusive::Vars,
+    aux::Vars,
+    t::Real,
+    direction,
+)
+    # TODO - how to assert that it is used with EquilMoist module?
+    #        (if it were to be used with non-EquilMoist model
+    #         q_liq and q_ice would be part of state)
```
charleskawczynski, line 209:
```suggestion
atmos_source!(
    s::RemovePrecipitation,
    atmos::AtmosModel,
    source::Vars,
    state::Vars,
    diffusive::Vars,
    aux::Vars,
    t::Real,
    direction,
) = atmos_source!(s, atmos, atmos.moisture, source, state, diffusive, aux, t, direction)

function atmos_source!(
    s::RemovePrecipitation,
    atmos::AtmosModel,
    moist::EquilMoist,
    source::Vars,
    state::Vars,
    diffusive::Vars,
    aux::Vars,
    t::Real,
    direction,
)
```

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+    FT = eltype(state)
+    if aux.moisture.q_liq + aux.moisture.q_ice > eps(FT)
```
charleskawczynski, line 211: We should probably make a thermo method for this

src/Atmos/Parameterizations/CloudPhysics/Microphysics_0M.jl
```diff
@@ -0,0 +1,57 @@
...
+    Zero-moment bulk microphysics scheme that instantly removes
+    moisture above certain threshold.
+    This is equivalent to instanteneous conversion of cloud condensate
+    into precipitation and precipitation fallout with infinite
+    terminal velocity.
+
+"""
+module Microphysics_0M
+
+using ClimateMachine.Thermodynamics
```
charleskawczynski, line 11:
```suggestion
using ..Thermodynamics
```

src/Atmos/Model/source.jl
```diff
@@ -1,5 +1,13 @@
-using CLIMAParameters.Planet: Omega
-export Source, Gravity, RayleighSponge, Subsidence, GeostrophicForcing, Coriolis
+using ClimateMachine.Microphysics_0M
```
charleskawczynski, line 1:
```suggestion
using ..Microphysics_0M
```

src/Atmos/Parameterizations/CloudPhysics/Microphysics_0M.jl
```diff
@@ -0,0 +1,57 @@
+"""
+    Zero-moment bulk microphysics scheme that instantly removes
+    moisture above certain threshold.
+    This is equivalent to instanteneous conversion of cloud condensate
+    into precipitation and precipitation fallout with infinite
+    terminal velocity.
+
```
charleskawczynski, line 7:
```suggestion
    Microphysics_0M

Zero-moment bulk microphysics scheme that instantly removes
moisture above certain threshold.
This is equivalent to instanteneous conversion of cloud condensate
into precipitation and precipitation fallout with infinite
terminal velocity.
```

src/Atmos/Model/source.jl
charleskawczynski: Probably need to format this, but this is how we can restrict to the `EquilMoist` model

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
+# Microphysics_0M
+
+The `Microphysics_0M.jl` module defines a 0-moment bulk parameterization of
+  moisture sink due to precipitation.
```
tapios, line 4:
```suggestion
  the moisture sink due to precipitation.
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+It offers a simplified way of removing the excess water
+  without assuming anything about the size distributions of cloud
+  or precipitation particles.
+
+The ``q_{tot}`` (total water specific humidity) sink due to precipitation
+  is obtained by relaxation with a constant timescale
+  to a state with precipitable water removed.
```
tapios, line 11:
```suggestion
  to a state with condensate exceeding a threshold value removed.
```
(note that precipitable water usually, and misleadingly, is understood to include water vapor)

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+The threshold for when to remove ``q_{tot}`` is defined either by the
```
tapios, line 12:
```suggestion
The threshold for removing excess ``q_{tot}`` is defined either by the
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+
+!!! note
+
+    The implementation assumes that
+    ``q_{liq}`` (liquid water specific humidity) and
+    ``q_{ice}`` (ice water specific humidity) are in the auxiliary variables.
+
+!!! note
+
+    To remove precipitation instantly the relaxation timescale should be
```
tapios, line 25:
```suggestion
    To remove precipitation instantly, the relaxation timescale should be
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+    equal to the timestep length.
+
+## Moisture sink due to precipitation
+
+If based on maximum condensate specific humidity the sink is defined as:
```
tapios, line 30:
```suggestion
If based on maximum condensate specific humidity, the sink is defined as:
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  \left. \mathcal{S}_{q_{tot}} \right|_{precip} =-
+    \frac{max(0, q_{liq} + q_{ice} - q_{c0})}{\tau_{precip}}
+\end{equation}
+```
+where:
+  - ``q_{liq}``, ``q_{ice}`` are cloud liquid water and cloud ice specific humidities,
+  - ``q_{c0}`` is the condensate specific humidity threshold above which water is removed,
+  - ``\tau_{precip}`` is the relaxation timescale.
+
+If based on saturation excess the sink is defined as:
```
tapios, line 42:
```suggestion
If based on saturation excess, the sink is defined as:
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  - ``q_{vap}^{sat}`` is the saturation specific humidity,
+  - ``\tau_{precip}`` is the relaxation timescale.
+
+## Coupling to the state variables
+
+Following the conservation equations for
+[moisture](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Moisture)
+and [mass](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Mass)
+the ``\mathcal{S}_{q_{tot}}`` sink has to be multiplied by ``\rho`` before
+  adding as one of the sink terms to both moisture and mass state variables.
```
tapios, line 61:
```suggestion
  adding it as one of the sink terms to both moisture and mass state variables.
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  - ``q_{liq}``, ``q_{ice}`` are cloud liquid water and cloud ice specific humidities,
+  - ``S_{0}`` is the supersaturation threshold above which water is removed,
+  - ``q_{vap}^{sat}`` is the saturation specific humidity,
+  - ``\tau_{precip}`` is the relaxation timescale.
+
+## Coupling to the state variables
+
+Following the conservation equations for
+[moisture](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Moisture)
+and [mass](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Mass)
```
tapios, line 59:
```suggestion
and [mass](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Mass),
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+the ``\mathcal{S}_{q_{tot}}`` sink has to be multiplied by ``\rho`` before
+  adding as one of the sink terms to both moisture and mass state variables.
+For the conservation equation for
+[total energy](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Energy)
```
tapios, line 63:
```suggestion
[total energy](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Energy),
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  no additional source/sink terms $M$ are considered and the
```
tapios, line 64:
```suggestion
  no additional source/sink terms $M$ are considered, and the
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  the sink due to removing ``q_{tot}`` is computed as:
+```math
+\begin{equation}
+\left. \sum_{j\in\{v,l,i\}}(I_j + \Phi)  \rho C(q_j \rightarrow q_p)) \right|_{precip} =
```
tapios, line 68:
```suggestion
\left. \sum_{j\in\{v,l,i\}}(I_j + \Phi)  \rho C(q_j \rightarrow q_p) \right|_{precip} =
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  \left(lf I_l + (1 - lf) I_i + \Phi \right)
```
tapios, line 69:
```suggestion
  \left[\lambda I_l + (1 - \lambda) I_i + \Phi \right)
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  \rho \, \left.\mathcal{S}_{q_{tot}} \right|_{precip}
+\end{equation}
+```
+where:
+ - ``lf`` is the liquid fraction
```
tapios, line 74:
```suggestion
 - ``\lambda`` is the liquid fraction
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+ - ``I_l = c_{vl} (T - T_0)`` is the internal energy of liquid water
+ - ``I_i = c_{vi} (T - T_0) - I_{i0}`` is the internal energy of ice
+ - ``T`` is the temperature,
+ - ``T_0`` is the reference temperature,
```
tapios, line 78:
```suggestion
 - ``T_0`` is the thermodynamic reference temperature (which is unrelated to the reference temperature used in hydrostatic reference states used in the momentum equations),
```

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+            S_qt = remove_precipitation(atmos.param_set, q, q_vap_sat)
+        end
+
+        source.moisture.ρq_tot += state.ρ * S_qt
+
+        source.ρ += state.ρ * S_qt
+
+        source.ρe +=
+            (
+                lf * _cv_l * (T - _T_0) +
```
tapios, line 242: `\lambda` instead of lf?

tapios:
Looks good to me. Thank you for being so diligent with the documentation!

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+        _cv_d::FT = cv_d(atmos.param_set)
+        _cv_l::FT = cv_l(atmos.param_set)
+        _cv_i::FT = cv_i(atmos.param_set)
+        _T_0::FT = T_0(atmos.param_set)
+
+        S_qt::FT = 0
+        if s.use_qc_thr
+            S_qt = remove_precipitation(atmos.param_set, q)
+        else
+            ts_neq = TemperatureSHumNonEquil(atmos.param_set, T, state.ρ, q)
```
szy21, line 231: Why do we need `TemperatureSHumNonEquil` instead of `TemperatureSHumEquil` here?

charleskawczynski: I assume a non-equilibrium thermo state is needed. I think I'll add a convenience method for computing non-equilibrium thermo states since microphysics seems to need them even when using an equilibrium moist model.

szy21: I'm a bit confused because we have `lf::FT = liquid_fraction(atmos.param_set, T, PhaseEquil, q)` here. I could be wrong though, it seems I need to familiarize myself again with Thermodynamics.jl!

trontrytel: What I want to do is to compute saturation vapour pressure. In order to do that, I have to create the thermo state and pass it into `q_vap_saturation`. I'm not a big fan of that, but that is the general strategy when using functions from `Thermodynamics.jl` module. However, I think that if I pass into  `TemperatureSHumNonEquil` T, rho and q it will not do anything to them, just populate the thermo_state fields. This is what I want, because I just want to pass them as arguments to `q_vap_saturation`.

szy21: For this, I think you could simply do:

```
ts = thermo_state(atmos, atmos.moisture, state, aux)
q_vap = q_vap_saturation(ts)
```
@charleskawczynski is this correct?

trontrytel: What does it do? Will it create phase partition based on q_liq and q_ice saved in aux? Or will it do another saturation adjustment?

szy21: I don't think it does SA, but it may return a `PhaseEquil` state. I guess I'm just confused why the liquid fraction is `PhaseEquil` but the thermodynamic state is `PhaseNonEquil`.

trontrytel: I just treat them as containers for temperature and phase partition. Will try to understand what functions are being called when you do what you are suggesting. I want to make sure I don't call another saturation adjustment here.

charleskawczynski: What @trontrytel has done is fine for now, but we should probably discuss if there is a better/more proper alternative to creating non-equilibrium states when we’re using an equilibrium model. `thermo_state` recovers (but does not itself call SA) the saved thermo state for whichever moisture model is being used.

charleskawczynski: After re-reading this, I think we should actually call `thermo_state`. Let’s discuss this more @trontrytel

szy21: If I understand correctly, `lf::FT = liquid_fraction(atmos.param_set, T, PhaseEquil, q)` will only return 1 or 0? But we would like to have phase partition based on q_liq and q_ice?

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+    #        (if it were to be used with non-EquilMoist model
+    #         q_liq and q_ice would be part of state)
+    FT = eltype(state)
+    if aux.moisture.q_liq + aux.moisture.q_ice > eps(FT)
+
+        q = PhasePartition(
+            state.moisture.ρq_tot / state.ρ,
+            aux.moisture.q_liq,
+            aux.moisture.q_ice,
+        )
```
charleskawczynski, line 217:
```suggestion
        ts = thermo_state(atmos, state, aux)
        q = PhasePartition(ts)
```

charleskawczynski: Let's not build ontop of `aux.moisture.q_liq`/`aux.moisture.q_ice`, since these were just added for debugging. We should actually remove these soon, since we want to keep the aux state lean to avoid memory bloat. At the moment, those should be

src/Atmos/Model/source.jl
charleskawczynski: > If I understand correctly, `lf::FT = liquid_fraction(atmos.param_set, T, PhaseEquil, q)` will only return 1 or 0? But we would like to have phase partition based on q_liq and q_ice?

It returns `q.liq / q_c` for `PhaseNonEquil` _if_ there is condensate. Otherwise, it returns `FT(T > T_freeze(param_set))`. So, in this case it returns `FT(T > T_freeze(param_set))`.

src/Atmos/Model/source.jl
charleskawczynski, line 211:
```suggestion
    if has_condensate(ts)
```

charleskawczynski, line 211:
```suggestion
    FT = eltype(state)
    ts = thermo_state(atmos, state, aux)
    if has_condensate(ts)
```

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,69 @@ function atmos_source!(
...
+        T::FT = aux.moisture.temperature
+        lf::FT = liquid_fraction(atmos.param_set, T, PhaseEquil, q)
```
charleskawczynski, line 219:
```suggestion
        lf::FT = liquid_fraction(ts)
```

src/Atmos/Model/source.jl
charleskawczynski: IIUC, this thermo state will be exactly the same as an equilibrium thermo state.

src/Atmos/Model/source.jl
trontrytel: Thanks! Now that I'm not using `aux.q_liq` and `aux.q_ice`. I think this is not needed. `thermo_state` should do the job of switching between the Equil and NonEquil moisture models.

trontrytel:
@charleskawczynski - could you take a look again?

Project.toml
```diff
@@ -49,7 +49,7 @@ WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 [compat]
 Adapt = "2.0.2"
 ArgParse = "1.1"
-CLIMAParameters = "0.1.6"
+CLIMAParameters = "= 0.1.8"
```
charleskawczynski, line 52: Is this what we want? @jakebolewski

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,63 @@ function atmos_source!(
...
+    FT = eltype(state)
+    ts = thermo_state(atmos, state, aux)
+    if has_condensate(ts)
+
+        q = PhasePartition(ts)
+        T::FT = aux.moisture.temperature
+        λ::FT = liquid_fraction(ts)
+
+        _e_int_i0::FT = e_int_i0(atmos.param_set)
+        _cv_d::FT = cv_d(atmos.param_set)
```
charleskawczynski, line 217:
```suggestion
```

src/Atmos/Model/source.jl
```diff
@@ -173,3 +181,63 @@ function atmos_source!(
...
+    aux::Vars,
+    t::Real,
+    direction,
+)
+    FT = eltype(state)
+    ts = thermo_state(atmos, state, aux)
+    if has_condensate(ts)
+
+        q = PhasePartition(ts)
+        T::FT = aux.moisture.temperature
```
charleskawczynski, line 213:
```suggestion
        T::FT = air_temperature(ts)
```

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+the ``\mathcal{S}_{q_{tot}}`` sink has to be multiplied by ``\rho`` before
+  adding it as one of the sink terms to both moisture and mass state variables.
+For the conservation equation for
+[total energy](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/AtmosModel/#Energy),
+  no additional source/sink terms $M$ are considered, and the
+  the sink due to removing ``q_{tot}`` is computed as:
+```math
+\begin{equation}
+\left. \sum_{j\in\{v,l,i\}}(I_j + \Phi)  \rho C(q_j \rightarrow q_p) \right|_{precip} =
+  \left[\lambda I_l + (1 - \lambda) I_i + \Phi \right)
```
charleskawczynski, line 69:
```suggestion
  \left[\lambda I_l + (1 - \lambda) I_i + \Phi \right]
```
I think?

docs/src/Theory/Atmos/Microphysics_0M.md
```diff
@@ -0,0 +1,87 @@
...
+  condensate specific humidity or supersaturation.
+The thresholds and the relaxation timescale are defined in
+  `CLIMAParameters.jl`.
+
+!!! note
+
+    The implementation assumes that
+    ``q_{liq}`` (liquid water specific humidity) and
+    ``q_{ice}`` (ice water specific humidity) are in the auxiliary variables.
+
```
charleskawczynski, line 22:
```suggestion
```

Project.toml
trontrytel: In package manager I tried simply `up CLIMAParameters` but was constantly getting package errors, even after I fixed my Julia version. The only thing that worked for me was to `rm CLIMAParameters` and then `add CLIMAParameters`. I don't know if thats the result of it. 

I can manually edit this file and remove the exact version. I don't think I can achieve the same inside the package manager.

charleskawczynski: Ah, that should be fine then. Thanks!

jakebolewski: I think the min version is probably better, even if for now defacto there is an assumed pinned version for ClimateMachine (another project that depends on ClimaParameters could add some parameters that are not needed by ClimateMachine).

PR #1448, opened 2020-08-18
Small bugfix: CUDAevice -> CUDADevice

charleskawczynski:
# Description

Small bugfix: `CUDAevice()` -> `CUDADevice()`

I have

### DIFF
--- a/test/Numerics/SystemSolvers/bgmres.jl
+++ b/test/Numerics/SystemSolvers/bgmres.jl
@@ -48,7 +48,7 @@ let
         if isa(x, Array)
             kernel! = multiply_A_kernel!(CPU(), cpu_threads)
         else
-            kernel! = multiply_A_kernel!(CUDAevice(), gpu_threads)
+            kernel! = multiply_A_kernel!(CUDADevice(), gpu_threads)
         end
         return kernel!(x, A, y, n1, n2, ndrange = ndrange)
     end


jkozdon:
Is this not being tested / run of the GPU? I don’t understand how this wasn’t hit before?

jakebolewski:
I don't think any of the unit tests are currently tested on the cluster (CPU or GPU).  Should we add that back and is the assumption that all unit tests will pass with the GPU backend?

kpamnany:
> I don't think any of the unit tests are currently tested on the cluster (CPU or GPU). Should we add that back and is the assumption that all unit tests will pass with the GPU backend?

I think so. SlurmCI used to run the unit tests on the GPU (but not on the GPU -- we'd disabled that as being redundant). See [here](https://github.com/CliMA/slurmci/blob/master/scripts/test-gpu-tests.sh).

jakebolewski:
GPU unittests added in #1450

charleskawczynski:
> Is this not being tested / run of the GPU? I don’t understand how this wasn’t hit before?

To add on what's been said, I found this when running `test/runtests.jl` with GPU resources available and, fortunately, I think this was the only failure I found.

jkozdon:
> To add on what's been said, I found this when running test/runtests.jl with GPU resources available and, fortunately, I think this was the only failure I found.

Thanks for checking on this!

jakebolewski:
@charleskawczynski maybe add this as an additional commit to #1450 if that is the case?

kpamnany:
Hmm. Not sure why (`runtests_gpu.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/runtests_gpu.jl) was running so little then... is this file even being used by anyone else?

charleskawczynski:
> Hmm. Not sure why [`runtests_gpu.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/runtests_gpu.jl) was running so little then... is this file even being used by anyone else?

Good question, I haven't tried testing that.

Closing this now, since this commit was added to #1450.

PR #1453, opened 2020-08-19
Fix `fill!` return on an MPIStateArrays

jkozdon:
Previously the underlying data array was returned for `fill!` with an
`MPIStateArray`, so that calls like
```julia
   B = fill!(A, 0)
```
would return result in `B === A.data` not `B === A` as is standard for
other array types in Julia.


jkozdon:
bors try

PR #1462, opened 2020-08-24
Make `init_state_auxiliary!` direction aware

charleskawczynski:
# Description

One problem I ran into (and @jkozdon helped in tracking down), when working on the BOMEX single stack, was that `direction` currently does not propagate to `contiguous_field_gradient!`. As a result, the horizontal components of `orientation.∇Φ`, which should be zero for single stack configurations (and maybe LES configurations? @akshaysridhar), had precision errors which significantly impacted results at `Float32` precision. This PR makes `init_state_auxiliary!` and `contiguous_field_gradient!` direction-aware so that the horizontal components of `orientation.∇Φ` are exactly zero for single stack configurations.

In addition, this PR renames `contiguous_field_gradient!` to `continuous_field_gradient!` and fixes the implementation for when `direction = VerticalDirection()`.

## Before
prog_ρu 1

## After*
prog_ρu 1

 - *Note that differences near the surface are the result from a separate issue, which we'll resolve in a subsequent PR.

Co-authored with: @mwarusz 

I have

### DIFF
--- a/docs/src/APIs/Numerics/DGMethods/DGMethods.md
+++ b/docs/src/APIs/Numerics/DGMethods/DGMethods.md
@@ -7,7 +7,7 @@ CurrentModule = ClimateMachine.DGMethods
 ```@docs
 DGModel
 remainder_DGModel
-contiguous_field_gradient!
+continuous_field_gradient!
 courant
 ```
 
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -671,8 +671,9 @@ function init_state_auxiliary!(
     m::AtmosModel,
     state_auxiliary::MPIStateArray,
     grid,
+    direction,
 )
-    init_aux!(m, m.orientation, state_auxiliary, grid)
+    init_aux!(m, m.orientation, state_auxiliary, grid, direction)
 
     init_state_auxiliary!(
         m,
@@ -680,6 +681,7 @@ function init_state_auxiliary!(
             atmos_init_ref_state_pressure!(m.ref_state, m, aux, geom),
         state_auxiliary,
         grid,
+        direction,
     )
 
     ∇p = ∇reference_pressure(m.ref_state, state_auxiliary, grid)
@@ -689,6 +691,7 @@ function init_state_auxiliary!(
         atmos_nodal_init_state_auxiliary!,
         state_auxiliary,
         grid,
+        direction;
         state_temporary = ∇p,
     )
 end
--- a/src/Atmos/Model/linear.jl
+++ b/src/Atmos/Model/linear.jl
@@ -152,7 +152,12 @@ function boundary_state!(
 )
     nothing
 end
-init_state_auxiliary!(lm::AtmosLinearModel, aux::MPIStateArray, grid) = nothing
+init_state_auxiliary!(
+    lm::AtmosLinearModel,
+    aux::MPIStateArray,
+    grid,
+    direction,
+) = nothing
 init_state_prognostic!(
     lm::AtmosLinearModel,
     state::Vars,
--- a/src/Atmos/Model/ref_state.jl
+++ b/src/Atmos/Model/ref_state.jl
@@ -132,6 +132,7 @@ function init_state_auxiliary!(
     m::PressureGradientModel,
     state_auxiliary::MPIStateArray,
     grid,
+    direction,
 ) end
 function init_state_prognostic!(
     ::PressureGradientModel,
--- a/src/Common/Orientations/Orientations.jl
+++ b/src/Common/Orientations/Orientations.jl
@@ -30,7 +30,7 @@ using ..BalanceLaws
 using ..MPIStateArrays: MPIStateArray
 import ..BalanceLaws: vars_state
 using ..DGMethods:
-    init_state_auxiliary!, contiguous_field_gradient!, LocalGeometry
+    init_state_auxiliary!, continuous_field_gradient!, LocalGeometry
 
 export Orientation, NoOrientation, FlatOrientation, SphericalOrientation
 
@@ -98,22 +98,30 @@ function projection_tangential(
     return u⃗ .- projection_normal(orientation, param_set, aux, u⃗)
 end
 
-function init_aux!(m, ::Orientation, state_auxiliary::MPIStateArray, grid)
+function init_aux!(
+    m,
+    ::Orientation,
+    state_auxiliary::MPIStateArray,
+    grid,
+    direction,
+)
     init_state_auxiliary!(
         m,
         (m, aux, tmp, geom) ->
             orientation_nodal_init_aux!(m.orientation, m.param_set, aux, geom),
         state_auxiliary,
         grid,
+        direction,
     )
 
-    contiguous_field_gradient!(
+    continuous_field_gradient!(
         m,
         state_auxiliary,
         ("orientation.∇Φ",),
         state_auxiliary,
         ("orientation.Φ",),
         grid,
+        direction,
     )
 end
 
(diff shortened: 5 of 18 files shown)


akshaysridhar:
Screen Shot 2020-08-24 at 8 48 07 AM
Current LES-BOMEX results I see (min,max)[∇Φ]=0 for the horizontal derivatives. The direction keywords are not used in the LES configurations

jkozdon:
> As a result, the horizontal components of `orientation.∇Φ`, which should be zero for single stack configurations (and maybe LES configurations? @akshaysridhar), had precision errors which significantly impacted results at `Float32` precision.

I'm not sure that artificially zeroing these components is the right way to go. We used analytic expressions before, then switched for numerical for a reason (@mwarusz or @simonbyrne might be able to comment).

Is `Float64` OK? If so this seems to be symptom of a larger problem.

Also, why is it only the single stack that has a problem and not the full BOMEX simulation?

charleskawczynski:
> Current LES-BOMEX (`Float32`) results I see (min,max)[∇Φ]_h=0 for the horizontal derivatives. The direction keywords are not used in the LES configurations

Can you confirm this by printing values of the horizontal direction (the values I'm seeing are ~0.001)? This plot is showing `∇Φ[3]`, which is not the component with the problem.

charleskawczynski:
> I'm not sure that artificially zeroing these components is the right way to go. We used analytic expressions before, then switched for numerical for a reason (@mwarusz or @simonbyrne might be able to comment).

It seems fine to me since the simulation should be effectively 1D.

> Is `Float64` OK? If so this seems to be symptom of a larger problem.

I need to double check this.

> Also, why is it only the single stack that has a problem and not the full BOMEX simulation?

One potential reason is that BOMEX / LES resolve the horizontal domain, which is much larger.

I guess one simple solution would be to simply use a very large horizontal domain for single stack configurations, but I'm not sure if this is the most elegant either.

jkozdon:
> > I'm not sure that artificially zeroing these components is the right way to go. We used analytic expressions before, then switched for numerical for a reason (@mwarusz or @simonbyrne might be able to comment).
> 
> It seems fine to me since the simulation should be effectively 1D.

In this case yes, but my concern is that this is just revealing a deeper `Float32` issue in the code (or some other bug)

I think in general, we need to run some simulations in both `Float64` and `Float32` to make sure that the results are actually correct for `Float32` (we are relying on `Float32` in a number of places, but we have never verified that for real problems the code is OK with the lower precision).

> > Also, why is it only the single stack that has a problem and not the full BOMEX simulation?
> 
> One potential reason is that BOMEX / LES resolve the horizontal domain, which is much larger.

I believe that the important thing is the element aspect ratio, not the overall domain size.

> I guess one simple solution would be to simply use a very large horizontal domain for single stack configurations, but I'm not sure if this is the most elegant either.

Like I said above, I think we need to understand why `Float32` is causing a problem and/or whether there is another bug in the code we need to track down.

charleskawczynski:
> In this case yes, but my concern is that this is just revealing a deeper `Float32` issue in the code (or some other bug)...I think in general, we need to run some simulations in both `Float64` and `Float32` to make sure that the results are actually correct for `Float32` (we are relying on `Float32` in a number of places, but we have never verified that for real problems the code is OK with the lower precision).... Like I said above, I think we need to understand why `Float32` is causing a problem and/or whether there is another bug in the code we need to track down.

I can confirm that the horizontal components, which should be zero, are really only problematic for `Float32`. I spoke with @mwarusz, and it seems that the errors may just be related to the magnitude of the geopotential. I completely agree that we should look into `Float32`, but perhaps in a separate PR.

Regardless, the current implementation of `contiguous_field_gradient!` is broken for `direction=VerticalDirection()` (precision-independent). When passing `direction=VerticalDirection()`, ∇Φ[3] ends up being zero when it should be ~9.8. @mwarusz is working on fixing this issue in this PR/ hook in a test.

> I believe that the important thing is the element aspect ratio, not the overall domain size.

Perhaps, but I think it makes sense to allow users to write 1D simulations that do not depend on the horizontal domain size.

mwarusz:
@charleskawczynski 
I pushed out fixes and tests for the `VerticalDirection` case. Also I renamed the gradient function since it has been pointed out to me that I used slightly incorrect terminology. Hope that's ok.

charleskawczynski:
> @charleskawczynski
> I pushed out fixes and tests for the `VerticalDirection` case. Also I renamed the gradient function since it has been pointed out to me that I used slightly incorrect terminology. Hope that's ok.

Looks like it's working correctly, thank you for the speedy fix!

charleskawczynski:
@kpamnany this effectively fixes the single stack simulation PR that I'll be opening up. There's one more non-breaking discrepancy between the LES and single stack, related to the Smagorinsky model.

akshaysridhar:
> > Current LES-BOMEX (`Float32`) results I see (min,max)[∇Φ]_h=0 for the horizontal derivatives. The direction keywords are not used in the LES configurations
> 
> Can you confirm this by printing values of the horizontal direction (the values I'm seeing are ~0.001)? This plot is showing `∇Φ[3]`, which is not the component with the problem.

Printing min-max for the 1,2 components results in zeros: This is from the VTK dataset at the nodal DoF
```
orientation.∇Φ_lb_1_rb_ -- Min = 0 (node 0 at coord <0, 0, 0>)
orientation.∇Φ_lb_1_rb_ -- Max = 0 (node 0 at coord <0, 0, 0>)
```

charleskawczynski:
> Printing min-max for the 1,2 components results in zeros: This is from the VTK dataset at the nodal DoF

Okay, good to know. This all still makes sense since the LES horizontal domain from the figure is ~6km.

jkozdon:
> Okay, good to know. This all still makes sense since the LES horizontal domain from the figure is ~6km.

But what is the element aspect ratio?

charleskawczynski:
> > Okay, good to know. This all still makes sense since the LES horizontal domain from the figure is ~6km.
> 
> But what is the element aspect ratio?

For LES BOMEX:
```julia
    xmax = FT(6400)
    ymax = FT(6400)
    zmax = FT(3000)
```
For single stack BOMEX
```julia
    xmax = FT(1)
    ymax = FT(1)
    zmax = FT(3000)
```

charleskawczynski:
bors try

jkozdon:
> ```julia
> 3000
> ```

That's the domain aspect ratio, not the element aspect ratios.

Looks like BOMEX uses 10:4 (horizontal:vertical) elements
https://github.com/CliMA/ClimateMachine.jl/blob/7c7a0f0a951332cd8153d7ccc3427c949e2ecc2e/experiments/AtmosLES/bomex.jl#L529-L531

If I understand your input file correctly you are using 1:60 (horizontal:vertical) elements
https://github.com/CliMA/ClimateMachine.jl/blob/e306feb498825e9e528b44f018d374eb9c27923b/experiments/AtmosLES/bomex_single_stack.jl#L31-L33

Another difference is that BOMEX uses polynomial order 4 and the single stack uses polynomial order 1.

charleskawczynski:
> That's the domain aspect ratio, not the element aspect ratios.
> 
> Looks like BOMEX uses 10:4 (horizontal:vertical) elements
> https://github.com/CliMA/ClimateMachine.jl/blob/7c7a0f0a951332cd8153d7ccc3427c949e2ecc2e/experiments/AtmosLES/bomex.jl#L529-L531
> 
> If I understand your input file correctly you are using 1:60 (horizontal:vertical) elements
> https://github.com/CliMA/ClimateMachine.jl/blob/e306feb498825e9e528b44f018d374eb9c27923b/experiments/AtmosLES/bomex_single_stack.jl#L31-L33
> 
> Another difference is that BOMEX uses polynomial order 4 and the single stack uses polynomial order 1.

Right, my mistake. I think the numbers you're showing are correct.

PR #1467, opened 2020-08-25
Add saturated function to thermo

charleskawczynski:
# Description

Adds a function `saturated` to Thermodynamics, which returns a bool indicating if a given thermodynamic state is saturated or supersaturated.

I have

### DIFF
--- a/src/Common/Thermodynamics/relations.jl
+++ b/src/Common/Thermodynamics/relations.jl
@@ -38,6 +38,7 @@ export air_temperature_from_ideal_gas_law
 export condensate, has_condensate
 export specific_enthalpy, total_specific_enthalpy
 export moist_static_energy
+export saturated
 
 """
     gas_constant_air(param_set, [q::PhasePartition])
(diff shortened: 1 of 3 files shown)


src/Common/Thermodynamics/relations.jl
```diff
@@ -2074,3 +2075,11 @@ function moist_static_energy(
...
     return specific_enthalpy(ts) + e_pot
 end
+
+"""
+    saturated(ts::ThermodynamicState)
+
+Boolean indicating if thermodynamic
+state is saturated.
+"""
+saturated(ts::ThermodynamicState) = (relative_humidity(ts) ≈ 1)
```
trontrytel, line 2085: @charleskawczynski - I think this assumes that you are running in equilibrium moisture setting. Wouldn't it be better to have `relative_humidity(ts) >= 1). This would include supersaturated conditions as well

charleskawczynski: The [situation](https://github.com/CliMA/ClimateMachine.jl/pull/1308/files#diff-22c202328e500e0b200455bb04b99648R53) that @yairchn tested for was `relative_humidity(ts) ≈ 1`. @yairchn should we change this?

yairchn: I think Anna's suggestion is correct - it should be larger or equal 1

charleskawczynski: Updated.

trontrytel:
LGTM. Looks like `>=` is more complicated to `approx`! ;)

charleskawczynski:
> LGTM. Looks like `>=` is more complicated to `approx`! ;)

Yeah, I actually really which Julia had `a \approx > b` lower to `a \approx b || a > b`. I'm just trying to protect against edge cases like
```julia
julia> 1<1+eps(Float64)/2
false

julia> 1≈1+eps(Float64)/2
true
```

PR #1485, opened 2020-08-27
GCM ref_state refinement

LenkaNovak:
Small fix that refines the GCM ref_state to be more physical, allowing for more stable runs.

# Description

I have

### DIFF
--- a/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
+++ b/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
@@ -140,7 +140,7 @@ function config_gcm_experiment(
 ) where {FT}
     # Set up a reference state for linearization of equations
     temp_profile_ref =
-        DecayingTemperatureProfile{FT}(param_set, FT(275), FT(75), FT(45e3))
+        DecayingTemperatureProfile{FT}(param_set, FT(290), FT(220), FT(8e3))
     ref_state = HydrostaticState(temp_profile_ref)
 
     # Distance between surface and top of atmosphere (m)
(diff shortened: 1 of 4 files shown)


tapios:
The same reference state is in experiments/AtmosGCM/heldsuarez.jl, baroclinic-wave.jl, and moist-baroclinic-wave-bulksfcflux.jl. You might want to remove the reference state specifications in those experiments and have all of them just use the driver. Otherwise, these simulations will run into trouble with deeper domains.

szy21:
Do we still need drivers like heldsuarez.jl, baroclinic-wave.jl, etc.? Or should we remove them?

kpamnany:
I had briefly discussed with Lenka the possibility of simply removing the old experiments; we'd decided to defer that decision until this new GCM driver is exercised properly.

But there's another aspect also: right now, we're running a number of these experiments as integration tests for CI. This makes sense if they are well known, with well understood initial values and well understood (and verified) behavior.

If this new GCM driver will be used as a platform for more experimental work, should we keep the old experiments around, for validation purposes?

szy21:
> If this new GCM driver will be used as a platform for more experimental work, should we keep the old experiments around, for validation purposes?

Sounds good. If the GCM driver will be mostly used for experimental work, I would actually prefer keeping the drivers for benchmark experiments, such as heldsuarez and baroclinic-wave.

LenkaNovak:
The ref states are now changed for all drivers. However, these individual drivers, such as `heldsuarez.jl`, should not be needed once we're happy with the GCM driver, because the default options for the GCM driver are set to be the benchmark configurations.

PR #1490, opened 2020-08-28
Bugfix for sponge modifier

akshaysridhar:
# Description
`\tau` is an output of `turbulence_tensors` ; for optional  
upper atmosphere sponges this fixes the viscous stress modifier. 

No effect on existing master branch experiments/ tutorials. 
Independent of the rayleigh linear sponge in `source.jl`

I have


szy21:
So fast, thanks! In the comment there is "It works like `ConstantViscosityWithDivergence` but without divergence", but the `τ` here is still with the divergence term I think?

akshaysridhar:
> So fast, thanks! In the comment there is "It works like `ConstantViscosityWithDivergence` but without divergence", but the `τ` here is still with the divergence term I think?
> 
Updated comments to clarify this : The viscosity / diffusivity should be taken from some arbitrary model.

PR #1491, opened 2020-08-28
Kd/agl/soil_tutorials_pt_1

kmdeck:
# Description

This code adds three tutorials for the land model. We also clarified some terminology - we had the user supplying Neumann BC on minus the gradient flux, but were calling it the flux. we added a sign correction and now the user supplies the BC on the gradient flux itself. 

These tutorials are using an interpolation helper file to smooth output. The idea was that the DG output for a state or aux variable would have multiple values on the boundaries, and this led to graphs with step like discontinuities. We implement this using a couple shared functions in `tutorials/Land/Soil/interpolation_helper.jl`. For now, it seems like it is fine to have our own functions for this, as they are quite simple. If we start to require more flexibility we might want to switch to using the Diagnostics framework. 

I have


tutorials/Land/Soil/Water/hydraulic_functions.jl
```diff
@@ -0,0 +1,352 @@
...
+plot(
+    S_l,
+    log10.(K),
+    xlabel = "effective saturation",
+    ylabel = "Log10(K)",
+    label = "Base case",
+)
+plot!(S_l, log10.(K_T), label = "Temperature Dependent Viscosity")
+plot!(S_l_accounting_for_ice, log10.(K_i), label = "Ice Impedance")
+savefig("./T_ice_K.png")
```
charleskawczynski, line 270:
```suggestion
savefig("T_ice_K.png")
```
`/` is an OS-specific path separator. Better to always avoid this, and use `joinpath` if we're saving somewhere other than `pwd`.

tutorials/Land/Soil/Water/hydraulic_functions.jl
```diff
@@ -0,0 +1,352 @@
...
+    label = "van Genuchten",
+)
+plot!(
+    S_l,
+    log10.(K_bc),
+    xlabel = "effective saturation",
+    ylabel = "Log10(K)",
+    label = "Brooks and Corey",
+)
+savefig("./bc_vg_k.png")
```
charleskawczynski, line 310:
```suggestion
savefig("bc_vg_k.png")
```

charleskawczynski: There are a few more of these

tutorials/Land/Soil/Water/hydraulic_functions.jl
kmdeck: ah, ok! I thought it universally just meant in the current directory. will change.

kmdeck:
@charleskawczynski I think I will do a last edit/read-through of the tutorials, then would it be OK to merge (after squashing/rebasing)?

src/Land/Model/SoilHeatParameterizations.jl
```diff
@@ -184,23 +187,102 @@ end
...
+        κ_solid::FT,
+        porosity::FT,
+        κ_air::FT,
+        ρp::FT
+    ) where {FT}
+
+Computes the thermal conductivity of dry soil.
+"""
+function k_dry(κ_solid::FT, porosity::FT, κ_air::FT, ρp::FT) where {FT}
+    a = FT(0.053)
```
charleskawczynski, line 281: Can we pull this constant out? It probably belongs as a model or free parameter.

charleskawczynski:
> @charleskawczynski I think I will do a last edit/read-through of the tutorials, then would it be OK to merge (after squashing/rebasing)?

Reviewing now..

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -0,0 +1,556 @@
...
+
+
+# As we are not including the equations for phase changes in this tutorial,
+# we chose temperatures that are above the freezing point of water.
+
+# The boundary conditions for the heat equation:
+surface_heat_flux = (aux, t) -> eltype(aux)(0.0)
+bottom_heat_flux = (aux, t) -> eltype(aux)(0.0)
+surface_heat_state = nothing
+bottom_heat_state = nothing;
```
charleskawczynski, line 249: What does a `nothing` surface state imply?

kmdeck: It means that no BC is applied to that surface. We have four possibilities per eqn (dirichlet top, bottom, neumann top, bottom) but should only specify two. In the BC functions, we check, and if the BC is not nothing, we apply the BC supplied

src/Land/Model/SoilHeatParameterizations.jl
kmdeck: It's a good question, I'll think about it! Specifying these functions helped with model validation, but Im not sure down the road how much we will use them (vs. e.g. reading k_dry from a soil data set)

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -0,0 +1,556 @@
...
+plot!(all_data[3]["ϑ_l"], iz, label = "t = 2.25 days")
+plot!(all_data[4]["ϑ_l"], iz, label = "t = 3 days")
+
+
+K∇h_z_plot = plot(
+    all_data[1]["K∇h_vert"],
+    iz,
+    label = "0.75 days",
+    ytickfontcolor = "white",
+    xlabel = "K∇h_z (m/s)",
```
charleskawczynski, line 485: It seems like we should be using `dict_of_nodal_states` and `export_state_plots`. If not, I think it would be worth generalizing this and moving into `src/Utilities/SingleStackUtils.jl` and or `docs/plothelpers.jl` (same with pieces in the callback).

tutorials/Land/Soil/interpolation_helper.jl
```diff
@@ -0,0 +1,22 @@
+function create_interpolation_grid(xbnd, xres, thegrid)
+    x1g = collect(range(xbnd[1, 1], xbnd[2, 1], step = xres[1]))
+    nx1 = length(x1g)
+    x2g = collect(range(xbnd[1, 2], xbnd[2, 2], step = xres[2]))
+    nx2 = length(x2g)
+    x3g = collect(range(xbnd[1, 3], xbnd[2, 3], step = xres[3]))
+    nx3 = length(x3g)
```
charleskawczynski, line 7:
```suggestion
    x1g = collect(range(xbnd[1, 1], xbnd[2, 1], step = xres[1]))
    x2g = collect(range(xbnd[1, 2], xbnd[2, 2], step = xres[2]))
    x3g = collect(range(xbnd[1, 3], xbnd[2, 3], step = xres[3]))
```

tutorials/Land/Soil/Coupled/equilibrium_test.jl
kmdeck: ok! Maybe we can chat briefly about this?

tutorials/Land/Soil/interpolation_helper.jl
```diff
@@ -0,0 +1,22 @@
+function create_interpolation_grid(xbnd, xres, thegrid)
```
charleskawczynski, line 1: Can we add docs? And should this live in `SingleStackUtils`?

tutorials/Land/Soil/Coupled/equilibrium_test.jl
charleskawczynski: Does no BC mean no flux? or 0 Dirichlet? I think avoiding `nothing` for BC application is more clear. I would just suggest not including the BC, and letting the constructor handle default BCs (whether they're Dirichlet / Neumann)

charleskawczynski:
Can we update the title?

tutorials/Land/Soil/Coupled/equilibrium_test.jl
kmdeck: no BC means no BC (no flux and 0 dirichlet are BCs). The idea is that since the Eqns are 2nd order in spatial derivatives, we need to supply two pieces of information (which could be zero flux of zero state) to uniquely determine the solution. But, if we restrict to Neumann and Dirichlet BC types, there are four possible pieces of information to supply. so, the two that arent picked are "no BC" or `nothing`.

I think this defn: https://github.com/CliMA/ClimateMachine.jl/blob/4607e850cbe7d33247cfd5141d7907c9afdafaac/src/Land/Model/land_bc.jl#L40

allows us to do what you suggested though - For example, we can instead call the soil model with:
```soil_water_model = SoilWaterModel(
           FT;
           viscosity_factor = TemperatureDependentViscosity{FT}(),
           moisture_factor = MoistureDependent{FT}(),
           hydraulics = vanGenuchten{FT}(α = vg_α, n = vg_n),
           initialϑ_l = ϑ_l0,
           dirichlet_bc = Dirichlet(),
           neumann_bc = Neumann(
               surface_flux = surface_water_flux,
               bottom_flux = bottom_water_flux,
           ),)```

and this works the same as specifying "nothing" for both the Dirichlet options.

What do you think?

Comment (charleskawczynski) [reply tutorials/Land/Soil/Coupled/equilibrium_test.jl:249]: > no BC means no BC

I think we need to see what the default behavior is. Is this equivalent to leaving the `boundary_state!` kernel empty? @jkozdon would this be a zero Neumann BC? I'm not sure I understand what "no BCs" mean.

Comment (kmdeck) [reply tutorials/Land/Soil/Coupled/equilibrium_test.jl:249]: Here is an example of how it is used in the `boundary_state!` function:
https://github.com/CliMA/ClimateMachine.jl/blob/25bc89432e183fd53b96afe0cfd09b5165407eb4/src/Land/Model/soil_bc.jl#L501

Comment (kmdeck) [reply tutorials/Land/Soil/Coupled/equilibrium_test.jl:249]: > no BC means no BC (no flux and 0 dirichlet are BCs). The idea is that since the Eqns are 2nd order in spatial derivatives, we need to supply two pieces of information (which could be zero flux of zero state) to uniquely determine the solution. But, if we restrict to Neumann and Dirichlet BC types, there are four possible pieces of information to supply. so, the two that arent picked are "no BC" or `nothing`.
> 
> I think the lines following this:
> 
> https://github.com/CliMA/ClimateMachine.jl/blob/4607e850cbe7d33247cfd5141d7907c9afdafaac/src/Land/Model/land_bc.jl#L40
> 
> allow us to do what you suggested though - For example, we can instead call the soil model with:
> 
> ```
>            FT;
>            viscosity_factor = TemperatureDependentViscosity{FT}(),
>            moisture_factor = MoistureDependent{FT}(),
>            hydraulics = vanGenuchten{FT}(α = vg_α, n = vg_n),
>            initialϑ_l = ϑ_l0,
>            dirichlet_bc = Dirichlet(),
>            neumann_bc = Neumann(
>                surface_flux = surface_water_flux,
>                bottom_flux = bottom_water_flux,
>            ),)
> ```
> 
> and this works the same as specifying "nothing" for both the Dirichlet options.
> 
> What do you think? Just a driver level change?

to clarify, the four possible options arise because this is 1d - so we have (top, bottom) x (dirichlet, neumann) as our options.

File: src/Land/Model/SoilHeatParameterizations.jl (hunk 3 above)
Comment (kmdeck) [reply src/Land/Model/SoilHeatParameterizations.jl:281]: OK, so we will be using a formula like this to compute k_dry from soil properties. I don't think we will want to tune them, though, as these functions have already been tuned by fitting soil data (and this soil data constrains them far more than our climate model would). 

Right now there are two other constants in the same category (used in the thermal conductivity model, but already tuned to soil data). These are stored in the soil model SoilParamFunctions - called a and b. I can add this one there as well.

File: tutorials/Land/Soil/interpolation_helper.jl (hunk 7 above)
Comment (kmdeck) [reply tutorials/Land/Soil/interpolation_helper.jl:7]: added docs. let's discuss the where it should live when we meet.

File: tutorials/Land/Soil/Coupled/equilibrium_test.jl (hunk 5 above)
Comment (kmdeck) [reply tutorials/Land/Soil/Coupled/equilibrium_test.jl:485]: ok! as per our discussion we will add into `dict_of_nodal_states` in a future branch. 
https://github.com/CliMA/ClimateMachine.jl/issues/1556

File: tutorials/Land/Soil/Coupled/equilibrium_test.jl (hunk 4 above)
Comment (kmdeck) [reply tutorials/Land/Soil/Coupled/equilibrium_test.jl:258]: OK - I added this text to the tutorials. https://github.com/CliMA/ClimateMachine.jl/blob/bdf33a47165f258a7eb78535941f4ca12e06306d/tutorials/Land/Soil/Coupled/equilibrium_test.jl#L230

I think the mechanics of the code is OK - as a user, it is natural that I have to give two BC. The code has the flexibility to accept any combination (Dirichlet, Neumann, top, bottom), so setting `nothing` for the fields where I dont want to give a BC feels reasonable.  What "do nothing" in the `boundary_state!` function is another question that the user shouldnt have to care about

Whether or not the two BC supplied must be at the top and bottom is another question (I put in the tutorial text that they need to be). So unless we want to add in a check that requires them to be at the top and at the bottom, not both at the top, e.g., this seems fine to me now.

File: src/Land/Model/SoilHeatParameterizations.jl (hunk 3 above)
Comment (kmdeck) [reply src/Land/Model/SoilHeatParameterizations.jl]: Waiting to make sure it all runs, but I think what I did is fine. I moved this constant out of this function:
https://github.com/CliMA/ClimateMachine.jl/blob/bdf33a47165f258a7eb78535941f4ca12e06306d/src/Land/Model/SoilHeatParameterizations.jl#L279

However, since k_dry was an attribute of soil_param_functions, and it needed that constant to evaluate, I couldnt pass the constant in via soil_param_functions. However, it seems clean enough for now to compute k_dry inside the source code (it only does it once). this just required making k_solid and rho_p (both things calculated from soil data sets; static profiles in space) attributes of soil_param_functions.

File: src/Land/Model/SoilHeatParameterizations.jl [hunk 8]
```diff
@@ -184,23 +188,108 @@ end
...
     ρcp_l = FT(cp_l(param_set) * _ρ_l)
     ρe_int_l = ρcp_l * (T - _T_ref)
     return ρe_int_l
 end
 
+"""
+    function k_solid(
+        ν_ss_om::FT,
+        ν_ss_quartz::FT,
+        ν_ss_minerals::FT,
```
kmdeck, line 212: removed this line, will push after the current CI passes (want to make sure it passes/docs look good)

charleskawczynski (approved):
I think these tutorials will be nice to get in for our doc sprint 🎉

PR #1504, opened 2020-08-31
Generalized mass+energy conservation checks

### LINKED ISSUE #1240: Dumping Mass and Energy loss to file
# Description

Can we dump the mass and energy losses to either a txt or NC file that is independent of the rest of the diagnostics? 
The dumping should simply append the mass and energy loss at the current time step (or fallback call) to a file that looks like this:

time (s)   mass_loss   energy_loss
0          0           0
10      1e-14        1e-14
...
1000      1e-14        1e-14
...
10000     1e-14        1e-14

These are quantities that we should add to all of the initial benchmarking papers that are being written now but it will be good to have it regardless of the papers.

The loss is nothing else than:
image
(Disregard the bars which are from a paper with a different meaning for barred variables than we do)

Ideally, for each test we want to have a visual of something like these curves:

image (1)

# Additional context

# For CLIMA Developers

kpamnany:
# Description

- Adds keyword parameter `check_cons` to `ClimateMachine.invoke!` which may contain a tuple of `ClimateMachine.ConservationCheck`s. Each of these will be checked at a specificable interval, the delta will optionally be displayed, and an error threshold may be specified so that an error will be thrown if the threshold is exceeded.
- Adds `AtmosMassEnergyLoss` diagnostics group to output deltas for mass and energy. Closes #1240.
- Updated the Taylor Green experiment to use both of these and the BOMEX experiments to use the `invoke!` keyword, replacing the explicit callback there.

Edit: @akshaysridhar has also pointed out that for full generality, the thresholds should be computed from sources, boundary fluxes, etc. but that remains to be done.

I have


src/Utilities/Conservation/Conservation.jl
```diff
@@ -0,0 +1,29 @@
...
+
+using MPI
+
+function compute(mpicomm::MPI.Comm, dg::DGModel, Q::MPIStateArray)
+    vgeo = dg.grid.vgeo
+
+    M = view(dg.grid.vgeo, :, [Grids._M], :)
+
+    Σρ = MPI.Reduce(sum(Q.ρ .* M), +, 0, mpicomm)
+    Σρe = MPI.Reduce(sum(Q.ρe .* M), +, 0, mpicomm)
```
blallen, line 15: these lines only work for an `AtmosModel`, they will fail on any other balance law. I think it would be better practice to pass in the symbol of which state variable you want to compute as an argument to this function and then configure the Atmos specific part in the Atmos specific diagnostics

jkozdon: Would be better to use `weightedsum` https://github.com/CliMA/ClimateMachine.jl/blob/64fbdf3556bd2ef6b8958670308dde7f96416cf3/src/Arrays/MPIStateArrays.jl#L594-L622

With this and `varsindices` https://github.com/CliMA/ClimateMachine.jl/blob/64fbdf3556bd2ef6b8958670308dde7f96416cf3/src/Utilities/VariableTemplates/VariableTemplates.jl#L58-L85

should generalize this fairly easily.

kpamnany: Ah, yeah. I was thinking of putting that off, but I'll fix it in this PR.

kpamnany: Fixed this, can you guys take a look please?

kpamnany:
@smarras79: please take a look at this and see if it addresses your issue #1240? The diagnostics group is added to the Taylor Green experiment.

kpamnany:
bors try

src/Diagnostics/atmos_mass_energy_loss.jl
```diff
@@ -0,0 +1,120 @@
...
+function atmos_mass_energy_loss_init(dgngrp, currtime)
+    mpicomm = Settings.mpicomm
+    mpirank = MPI.Comm_rank(mpicomm)
+    dg = Settings.dg
+    bl = dg.balance_law
+    Q = Settings.Q
+    FT = eltype(Q)
+
+    ρ_idx = varsindices(vars_state(bl, Prognostic(), FT), "ρ")
+    ρe_idx = varsindices(vars_state(bl, Prognostic(), FT), "ρe")
```
charleskawczynski, line 64:
```suggestion
    ρ_idx = varsindices(vars_state(bl, Prognostic(), FT), :ρ)
    ρe_idx = varsindices(vars_state(bl, Prognostic(), FT), :ρe)
```
I think this works with strings or symbols, but should we lean towards symbols since they can be used on the GPU?

kpamnany: Wasn't sure if you could do something like `:moisture.ρq_tot`, but it turns out you can (with `:(moisture.ρq_tot)`). I can change it, but does it matter @jkozdon?

mwarusz: `varsindices` doesn't work on the GPU so I think this a question of preference. I added the string version specifically because I didn't like `:(moisture.ρq_tot)`.

charleskawczynski: Ah, I was thinking of `varsindex`. I think the syntax that works is `varsindex(..., :moisture, :ρq_tot))` for `varsindex`. I think the `varsindices` are for when we want multiple variable indexes.

src/Utilities/Conservation/Conservation.jl
jkozdon: Looks fine to me.

src/Diagnostics/atmos_mass_energy_loss.jl
jkozdon: > Wasn't sure if you could do something like `:moisture.ρq_tot`, but it turns out you can (with `:(moisture.ρq_tot)`). I can change it, but does it matter @jkozdon?

Fine with me either way. I'd say symbols or strings is all personal preference.

mwarusz: Pretty much. Another difference is that`varsindex` returns a range whereas `varsindices` returns a tuple of integers. I don't think that `varsindex` with multiple `Symbol` arguments works on the GPU either but I could be wrong. Haven't tried it lately.

blallen: Here's an example we've used on the ocean side that works

https://github.com/CliMA/ClimateMachine.jl/blob/d5df1495500ae78a7fca79ddf88ca5b5561f12f3/src/Ocean/HydrostaticBoussinesq/HydrostaticBoussinesqModel.jl#L690-L699

If your symbol returns a range, you can just typecast it to a `tuple()` and then it will work on the GPU as well. Haven't tried to pass multiple symbols as arguments to `varsindex` tho.

jkozdon: @mwarusz meant inside a device kernel (not in the host-side code that generates device code).

kpamnany: I don't think specifying `(:moisture, :ρq_tot)` is an improvement over `"moisture.ρq_tot"`. Also, this code is not expected to run on the GPU anyway. The `weightedsum` might, but that's internal to `MPIStateArrays` and independent of how we find which index to use.

src/Diagnostics/atmos_mass_energy_loss.jl
```diff
@@ -0,0 +1,120 @@
...
+        interval,
+        out_prefix,
+        writer,
+        interpol,
+    )
+end
+
+# store initial mass and energy
+Base.@kwdef mutable struct AtmosMassEnergyLoss
+    Σρ₀::Float64 = 0.0
```
smarras79, line 50: Should this be explicitly Float64 or should we be consistent with the type used across the code?

Issue #1502, opened 2020-08-31
Clean up ConstantViscosityWithDivergence


szy21:
# Description

In ConstantViscosityWithDivergence, change constant dynamic viscosity to constant kinematic viscosity, also add an option to turn off the divergence term.

# For CLIMA Developers

- [x] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

PR #1505, opened 2020-08-31
clean up ConstantViscosityWithDivergence

szy21:
# Description

In ConstantViscosityWithDivergence, change constant dynamic viscosity to constant kinematic viscosity, and add an option for the divergence term. Address #1502 

Still need to modify the drivers once the structure is determined.

I have


szy21:
@simonbyrne I was told that it is not recommended to use `Bool` in `struct`. If this is the case, could you please take a look and let me know what would be a better way to do this? Thanks!

szy21:
> Doesn't the type `ConstantViscosityWithDivergence` already suggest "with divergence"? If not, then we should probably have separate types:
> 
> * `ConstantViscosityWithDivergence`
> * `ConstantViscosityNoDivergence`

Yes, I was thinking about changing it to `ConstantViscosity`. If they only differ by one divergence term, is it better to have separate types or just an option in one type?

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -385,9 +390,13 @@ function turbulence_tensors(
...
     S = diffusive.turbulence.S
-    ν = m.ρν / state.ρ
+    ν = m.ν
     D_t = ν * _inv_Pr_turb
-    τ = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I
+    if m.with_divergence
+        τ = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I
+    else
+        τ = (-2 * ν) * S
+    end
```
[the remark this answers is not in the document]
szy21: Yes, I just talked with Charlie and this is exactly the reason. I will change it to a type.

szy21:
@akshaysridhar I am leaning towards renaming `ConstantViscosityWithDivergence` to `ConstantViscosity` because I think it makes more sense. What do you think? Currently the default is WithDivergence(), and I would like to change it to WithoutDivergence() but I'm afraid it will break many tests. I may already break them by setting constant kinematic viscosity, though...

szy21:
This breaks many tests too. `ConstantViscosityWithDivergence` is used in `test/Driver/mms3.jl` and `Numerics/DGMethods/compressible_Navier_Stokes/mms_bc_atmos.jl`. Any idea on what the expected results should be in these tests?

charleskawczynski:
From what I can see, any combination of two things could be breaking CI:
 - Change from dynamic to kinematic viscosity. If we change the constructor (`ConstantViscosityWithDivergence`), we must change every driver/experiment that uses it and pass in the appropriate input.
 - Changes in the courant calculation

szy21:
> From what I can see, any combination of two things could be breaking CI:
> 
> * Change from dynamic to kinematic viscosity. If we change the constructor (`ConstantViscosityWithDivergence`), we must change every driver/experiment that uses it and pass in the appropriate input.
> * Changes in the courant calculation

Thanks! I would still like to change it because I think it is better to use kinematic viscosity. I change the courant.jl accordingly. But I wonder how to get the expected results for the tests after changing the constructor?

akshaysridhar:
> From what I can see, any combination of two things could be breaking CI:
> 
>     * Change from dynamic to kinematic viscosity. If we change the constructor (`ConstantViscosityWithDivergence`), we must change every driver/experiment that uses it and pass in the appropriate input.
> 
>     * Changes in the courant calculation

It seems the `mms3` generated solution relies on a dynamic viscosity; with density given as `3 +/- f(x,y,z,t)`

szy21:
@charleskawczynski @akshaysridhar I think this is ready. Could you please take a look? I changed `ConstantVisocsityWithDivergence` to `ConstantDynamicViscosity`, and all the related tests passed. I added `ConstantKinematicViscosity` in the gcm drivers it gives me reasonable results. Since we will probably not using it in the end, I think it's ok if we don't include a new mms test?

szy21:
bors try

szy21:
bors try

szy21:
bors try

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -373,7 +427,7 @@ function compute_gradient_flux!(
 end
```
charleskawczynski, line 428:
```suggestion

compute_stress(div_type::WithoutDivergence, ν, S) = (-2 * ν) * S
compute_stress(div_type::WithDivergence, ν, S) = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I

```
This might be a nicer / compact way to do this

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -387,7 +441,38 @@ function turbulence_tensors(
...
     ν = m.ρν / state.ρ
     D_t = ν * _inv_Pr_turb
-    τ = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I
+    if m.divergence_type isa WithDivergence
+        τ = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I
+    elseif m.divergence_type isa WithoutDivergence
+        τ = (-2 * ν) * S
+    else
+        error("divergence_type can only be WithDivergence or WithoutDivergence")
+    end
```
charleskawczynski, line 450:
```suggestion
    τ = compute_stress(m.divergence_type, ν, S)
```

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -387,7 +441,38 @@ function turbulence_tensors(
...
+    S = diffusive.turbulence.S
+    ν = m.ν
+    D_t = ν * _inv_Pr_turb
+    if m.divergence_type isa WithDivergence
+        τ = (-2 * ν) * S + (2 * ν / 3) * tr(S) * I
+    elseif m.divergence_type isa WithoutDivergence
+        τ = (-2 * ν) * S
+    else
+        error("divergence_type can only be WithDivergence or WithoutDivergence")
+    end
```
charleskawczynski, line 475:
```suggestion
    τ = compute_stress(m.divergence_type, ν, S)
```

src/Common/TurbulenceClosures/TurbulenceClosures.jl
charleskawczynski: Also, it allows us to avoid writing the `else` case-- it will just result in a method error.

szy21: Thanks! I will change it.

PR #1511, opened 2020-09-02
Use many threads for stack in set banded matrix

jkozdon:
@Zhengyu-Huang I was originally thinking that we should use multiple threads per stack element (as opposed to one thread per stack element)

### DIFF
--- a/src/Numerics/SystemSolvers/columnwise_lu_solver.jl
+++ b/src/Numerics/SystemSolvers/columnwise_lu_solver.jl
@@ -408,16 +408,19 @@ function update_banded_matrix!(
                 f!(dQ, Q, args...)
 
                 # Store the banded matrix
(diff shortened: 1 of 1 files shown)


jkozdon:
bors try

jkozdon:
Hmm... maybe we cannot bors a non-master PR

simonbyrne (approved):

jkozdon:
This wasn't meant to be merged.

thomasgibson:
@jkozdon We should talk about these changes. I was under the impression there were still issues?

jkozdon:
> @jkozdon We should talk about these changes. I was under the impression there were still issues?

The real PR is #1512. I need to fix some of the testing and look into the Float32 issue some more.

PR #1483, opened 2020-08-27
Add a Jacobian-Free Newton-Krylov solver and preconditioning hooks

thomasgibson:
# Description
Co-authored by: @Zhengyu-Huang 

This PR adds a Jacobian-Free Newton-Krylov solver and adds hooks for using preconditioners. This is still a work in progress, but comments will be really appreciated.

Some things left to do:

- Clean up documentation and expand
- Clean up PC hooks (PC abstraction)

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1512, opened 2020-09-02
banded matrix refactoring

jkozdon:
Mainly adding this PR to check CI since #1483 doesn't pass CI yet


jkozdon:
bors try

jkozdon:
see also: #1511

PR #1482, opened 2020-08-27
Add time-averaging for fast model to SplitExplicitMethod

blallen:
# Description

See title. Also adds tests with Coriolis force and temperature advection to test the time averaging.

I have

- [ ] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [ ] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [ ] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1517, opened 2020-09-02
Add traits to turn on and off advection for tests

blallen:
# Description

Testing #1482 requires running tests with and without temperature advection to test the stability of the averaging scheme (and determine that is specifically temperature advection that is failing and not something else related to temperature). I also went ahead and implemented the same thing for momentum advection because why not. 

There's probably a better version of this where we have a struct for the advection-diffusion equation that is used to turn manipulate both parts (and contains the viscosities/diffusivities). Not sure if it's worth it to go ahead and add that complexity now though.

Also, there is no test for the `::Nothing` case of tracer advection and for the `::NonLinear` case of momentum advection. The former will come in #1482 and the latter will come once we experiment with non-linear momentum advection.

I have

### DIFF
--- a/src/Ocean/Ocean.jl
+++ b/src/Ocean/Ocean.jl
@@ -1,10 +1,14 @@
 module Ocean
 
-export AbstractOceanCoupling, Uncoupled, Coupled
+export AbstractOceanCoupling,
+    Uncoupled, Coupled, AdvectionTerm, NonLinearAdvectionTerm
 
 abstract type AbstractOceanCoupling end
 struct Uncoupled <: AbstractOceanCoupling end
 struct Coupled <: AbstractOceanCoupling end
+
+abstract type AdvectionTerm end
+struct NonLinearAdvectionTerm <: AdvectionTerm end
 
 function ocean_init_state! end
 function ocean_init_aux! end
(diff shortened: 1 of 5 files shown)


src/Ocean/Ocean.jl
```diff
@@ -1,11 +1,14 @@
...
 
-export AbstractOceanCoupling, Uncoupled, Coupled
+export AbstractOceanCoupling, Uncoupled, Coupled, AdvectionTerm, NonLinear
 
 abstract type AbstractOceanCoupling end
 struct Uncoupled <: AbstractOceanCoupling end
 struct Coupled <: AbstractOceanCoupling end
 
+abstract type AdvectionTerm end
+struct NonLinear <: AdvectionTerm end
```
jkozdon, line 10:
```suggestion
struct NonLinearAdvectionTerm <: AdvectionTerm end
```

`NonLinear` seem a bit vague by CLIMA standard :)

jkozdon (approved):
Seem reasonable to me.

src/Ocean/Ocean.jl
blallen: ha, I originally had that but didn't want to type it all out. I can add it back

blallen:
bors try

PR #1521, opened 2020-09-04
add solid body rotation test

szy21:
# Description

Co-authored by: @bischtob 
Add a solid body rotation test. This driver needs to become part of our nightly evaluation set. It tests whether hydrostatic balance is respected in the GCM.

I have

### DIFF
--- a/experiments/AtmosGCM/solid-body-rotation.jl
+++ b/experiments/AtmosGCM/solid-body-rotation.jl
@@ -1,0 +1,173 @@
+#!/usr/bin/env julia --project
+using ClimateMachine
+ClimateMachine.init(parse_clargs = true)
+
+using ClimateMachine.Atmos
+using ClimateMachine.Orientations
+using ClimateMachine.ConfigTypes
+using ClimateMachine.Diagnostics
+using ClimateMachine.GenericCallbacks
+using ClimateMachine.ODESolvers
+using ClimateMachine.TurbulenceClosures
+using ClimateMachine.SystemSolvers: ManyColumnLU
+using ClimateMachine.Mesh.Filters
+using ClimateMachine.Mesh.Grids
+using ClimateMachine.TemperatureProfiles
+using ClimateMachine.VariableTemplates
+using ClimateMachine.Thermodynamics: air_density, total_energy
+
+using LinearAlgebra
+using StaticArrays
+using Test
+
+using CLIMAParameters
+using CLIMAParameters.Planet: day, planet_radius
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+
+function init_solid_body_rotation!(problem, bl, state, aux, coords, t)
+    FT = eltype(state)
+
+    # The initial state is in hydrostatic balance
+    # It is chosen to be the same as reference state
+    # But we should expect the flow to stay at rest
+    # with any hydrostatic initial state
+    temp_profile_init =
+        DecayingTemperatureProfile{FT}(param_set, FT(290), FT(220), FT(8e3))
+    z = altitude(bl.orientation, bl.param_set, aux)
+    T₀, p = temp_profile_init(bl.param_set, z)
+    ρ = air_density(bl.param_set, T₀, p)
+    e_pot = gravitational_potential(bl.orientation, aux)
+    e_kin = FT(0)
+
+    # Assign state variables
+    state.ρ = ρ
+    state.ρu = SVector{3, FT}(0, 0, 0)
+    state.ρe = ρ * total_energy(bl.param_set, e_kin, e_pot, T₀)
+
+    nothing
+end
+
+function config_solid_body_rotation(FT, poly_order, resolution)
+    # Set up a reference state for linearization of equations
+    temp_profile_ref =
+        DecayingTemperatureProfile{FT}(param_set, FT(290), FT(220), FT(8e3))
+    ref_state = HydrostaticState(temp_profile_ref)
+
+    # Set up the atmosphere model
+    exp_name = "SolidBodyRotation"
+    domain_height::FT = 30e3 # distance between surface and top of atmosphere (m)
+
+    model = AtmosModel{FT}(
+        AtmosGCMConfigType,
+        param_set;
+        init_state_prognostic = init_solid_body_rotation!,
+        ref_state = ref_state,
+        turbulence = ConstantKinematicViscosity(FT(0)),
+        #hyperdiffusion = DryBiharmonic(FT(8 * 3600)),
+        moisture = DryModel(),
+        source = (Gravity(), Coriolis()),
+    )
+
+    config = ClimateMachine.AtmosGCMConfiguration(
+        exp_name,
+        poly_order,
+        resolution,
+        domain_height,
+        param_set,
+        init_solid_body_rotation!;
+        model = model,
+    )
+
+    return config
+end
+
+function main()
+    # Driver configuration parameters
+    FT = Float64                             # floating type precision
+    poly_order = 3                           # discontinuous Galerkin polynomial order
+    n_horz = 12                              # horizontal element number
+    n_vert = 6                               # vertical element number
+    n_days::FT = 1
+    timestart::FT = 0                        # start time (s)
+    timeend::FT = n_days * day(param_set)    # end time (s)
+
+    # Set up driver configuration
+    driver_config = config_solid_body_rotation(FT, poly_order, (n_horz, n_vert))
+
+    # Set up experiment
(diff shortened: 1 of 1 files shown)


experiments/AtmosGCM/solid-body-rotation.jl
```diff
@@ -0,0 +1,161 @@
...
+using CLIMAParameters.Planet:
+    day, planet_radius
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+
+function init_solid_body_rotation!(problem, bl, state, aux, coords, t)
+    FT = eltype(state)
+
+    ## Assign state variables
+    state.ρ = aux.ref_state.ρ
```
tapios, line 32: It would be nice here to set up an initial state that is hydrostatic but differs from the reference state. For example, you could choose a state from the family of `DecayingTemperatureProfile` but with different parameters than in the ref state. (The ref state should be trivially maintained the way the model is set up, unless it is perturbed by filters. But this is not necessarily the case for other hydrostatic states.)

bischtob: Done.

experiments/AtmosGCM/solid-body-rotation.jl
```diff
@@ -0,0 +1,167 @@
...
+const param_set = EarthParameterSet()
+
+function init_solid_body_rotation!(problem, bl, state, aux, coords, t)
+    FT = eltype(state)
+
+    # The initial state is chosen to be in hydrostatic balance, but differs from
+    # the reference state.
+    temp_profile_ref =
+        DecayingTemperatureProfile{FT}(param_set, FT(300), FT(210), FT(9e3))
+    ref_state = HydrostaticState(temp_profile_ref)
```
tapios, line 35: Just use `init` in place of `ref` here to avoid confusion

szy21:
@tapios @bischtob This is ready I think. Could you take a look again? We change the initial state back to the reference state because the state doesn't remain at rest even for this case, and we want to solve this problem first. Thanks!

tapios:
> @tapios @bischtob This is ready I think. Could you take a look again? We change the initial state back to the reference state because the state doesn't remain at rest even for this case, and we want to solve this problem first. Thanks!

Could you please revert this so that you can have any initial state from the decaying temperature profiles? You can choose the parameters of that state to be the same as those for the reference state for now. But it's important that we can maintain any hydrostatic initial state eventually, once we have fluxes that preserve contact discontinuities. In the past this has caused confusion (people thought we only need to maintain the reference state). If you make it clear in the code that this should be possible for any hydrostatic state irrespective of reference state, it'll help avoid that confusion.

szy21:
> Could you please revert this so that you can have any initial state from the decaying temperature profiles? You can choose the parameters of that state to be the same as those for the reference state for now. But it's important that we can maintain any hydrostatic initial state eventually, once we have fluxes that preserve contact discontinuities. In the past this has caused confusion (people thought we only need to maintain the reference state). If you make it clear in the code that this should be possible for any hydrostatic state irrespective of reference state, it'll help avoid that confusion.

I agree. Just changed it back.

tapios:
> > Could you please revert this so that you can have any initial state from the decaying temperature profiles? You can choose the parameters of that state to be the same as those for the reference state for now. But it's important that we can maintain any hydrostatic initial state eventually, once we have fluxes that preserve contact discontinuities. In the past this has caused confusion (people thought we only need to maintain the reference state). If you make it clear in the code that this should be possible for any hydrostatic state irrespective of reference state, it'll help avoid that confusion.
> 
> I agree. Just changed it back.

Thanks so much!

experiments/AtmosGCM/solid-body-rotation.jl
```diff
@@ -0,0 +1,173 @@
...
+    # But we should expect the flow to stay at rest
+    # with any hydrostatic initial state
+    temp_profile_init =
+        DecayingTemperatureProfile{FT}(param_set, FT(290), FT(220), FT(8e3))
+    z = altitude(bl.orientation, bl.param_set, aux)
+    T₀, p = temp_profile_init(bl.param_set, z)
+    ρ = air_density(bl.param_set, T₀, p)
+    e_pot = gravitational_potential(bl.orientation, aux)
+    e_kin = FT(0)
+
```
tapios, line 42: This is not exactly the same as the reference state because in the reference state, the density is calculated discretely from hydrostatic balance and pressure, and the temperature is determined to be consistent with that density. See https://github.com/CliMA/ClimateMachine.jl/blob/8a97053e7acaa58d9e67e85abf2d07e1883d6255/src/Atmos/Model/ref_state.jl#L85-L107.

We should make the result of the reference state calculations available to initialize the model. Perhaps @charleskawczynski can help making that possible. An option might be to move the calculations of density, temperature etc. from `ref_state` to `TemperatureProfiles`, and then use those results both for initialization and the reference state.

szy21: Thanks, Tapio. @charleskawczynski What I would like to have here is to have `state.ρ = aux.ref_state.ρ` and `state.ρe = aux.ref_state.ρe` but use `TemperatureProfiles` instead of `ref_state`. What would be the best way to achieve this?

bischtob: This is actually why we reverted this back. We quickly discovered that it was a bit more involved since the discrete part is living inside this function and requires a discrete pressure gradient that I am note sure is available to the init function.

bischtob: @charleskawczynski, my intuition as a user is that when I create a hyrostatic state, I will have easy access to a set of state variables (dry or moist). Then it would be easy to use that kind of object in the init function.

charleskawczynski: We have access to the ref state vars in `init_state_prognostic!`: `aux.ref_state.x`. IIUC, I think that'd be fine

charleskawczynski: What's the benefit of computing / recomputing this from `TemperatureProfiles`?

tapios: @charleskawczynski What we need is a little different: We want to be able to set up a hydrostatic reference and *also* a hydrostatic initial state, which may differ from the reference state but is computed in the same way (i.e., is in discrete hydrostatic balance).

What would make sense to me is to have a function that computes a hydrostatic state given input parameters in the same way the reference state is currently computed. This could be an augmented version of the current `TemperatureProfiles` (`HydrostaticProfiles`?). We could then use that function both to set up a reference state (with one set of parameters) and an initial state (with another set of parameters).

charleskawczynski: I see.

> which may differ from the reference state

For example, using different input parameters to `DecayingTemperatureProfile`?

tapios: Yes, exactly.

mwarusz:
As already mentioned the current setup of this test is not in a discrete hydrostatic balance. After fixing this I see that we preserve the reference state with a numerical flux that has the contact property (Central or Roe) but not with Rusanov. This is also true when the initial state is a different discretely balanced state. The only caveat in the latter case is that filtering has to be applied on perturbations from the initial state not the reference state. My experiments can be found here:
https://github.com/CliMA/ClimateMachine.jl/tree/mw/solid_body_rotation.

There is a similar existing test https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Atmos/Model/discrete-hydrostatic-balance.jl that also tests the LES case. Seems like it would be a good idea to merge these tests.

szy21:
> https://github.com/CliMA/ClimateMachine.jl/tree/mw/solid_body_rotation

Thanks, Maciej! I think it's a good idea to merge the tests. Another thing I have noticed is it seems I cannot run the simulation (with Rusanov or Central) for more than one simulated hour without the filter. Is this as expected?

mwarusz:
That's expected. The simulation is unstable without the filter regardless of discrete balance.

szy21:
> That's expected. The simulation is unstable without the filter regardless of discrete balance.

Thanks. Could you explain a bit more on why the filtering has to be applied on perturbations from the initial state?

mwarusz:
If you start with an initial state different from the reference state but filter perturbations from the reference state then the filter will act on the initial state, disrupting the balance.

tapios:
This is one of the issues we need to fix. We need a numerical discretization that preserves any hydrostatic state, without adjustment for the state to be preserved. We'll talk about this this pm.

szy21:
This will be added in #1532 . I'm going to close this soon.

PR #1523, opened 2020-09-05
Add `horizontally_average!`, `hmax` for single stack, and prep for edmf

charleskawczynski:
# Description

This PR
 - Adds `horizontally_average!` for single stack utils. This is needed to avoid issues with horizontally-non uniform initial profiles, resulting from `rand()`.
 - Adds `hmax` to `SingleStackConfiguration` becuase the Smagorinsky model is scale-aware and causes problems at small (`hmax ~ 1m`) and large (`hmax ~ 50km`) horizontal domains. To avoid holding up the EDMF development, we'll make `hmax` adjustable per driver for now.
 - Adds `turbconv` as a kwarg to `bomex_model`.

I have

### DIFF
--- a/src/Utilities/SingleStackUtils/SingleStackUtils.jl
+++ b/src/Utilities/SingleStackUtils/SingleStackUtils.jl
@@ -6,6 +6,7 @@ export get_vars_from_nodal_stack,
     get_horizontal_mean,
     reduce_nodal_stack,
     reduce_element_stack,
+    horizontally_average!,
     dict_of_nodal_states
 
 using OrderedCollections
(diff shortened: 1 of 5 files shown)


ilopezgp (approved):

kpamnany (commented):
As a reminder, multiple nested `for` loops can be written as:
```julia
for i = 1:Nq, j = 1:Nq
    ....
end
```
Also suggest extracting the `sum` anonymous function from the call, maybe into a `let`; will make the code more readable.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -298,6 +299,35 @@ function reduce_element_stack(
...
+        Q_sum = sum((1:Nq, 1:Nq)) do ij
+            Q[ij[1] + Nq * ((ij[2] - 1) + Nq * (k - 1)), i_v, ev]
+        end
+        Q_ave = Q_sum / (Nq * Nq)
+        for i in 1:Nq, j in 1:Nq
+            ijk = i + Nq * ((j - 1) + Nq * (k - 1))
+            Q[ijk, i_v, ev] = Q_ave
+        end
+    end
+end
```
jkozdon, line 329: This should probably use the metric terms to compute the average, mainly you should compute the integral and divide by the length. Otherwise, you are giving more weight to the solution near the boundary due to the use of LGL points.

jkozdon: Also, as written this is not a GPU friendly function.

kpamnany: None of the single stack util functions are GPU friendly. They started as just a couple of helper functions and are now growing rapidly -- probably need refactoring to abstract out MPIStateArray internal knowledge and for GPU friendliness soonish now.

charleskawczynski: > This should probably use the metric terms to compute the average, mainly you should compute the integral and divide by the length. Otherwise, you are giving more weight to the solution near the boundary due to the use of LGL points.

Thanks for letting me know. The main purpose of this function is to eliminate horizontal variation introduced in `rand()`, used in `init_state_prognostic!` for some of the single stack drivers. Ultimately, I'm trying to eliminate any resulting horizontal fluxes-- do I need to compute this average using the metric terms in order to eliminate horizontal fluxes?

> Also, as written this is not a GPU friendly function.

This is not performance-critical code (it's just for initialization), but we should modify it to work when the backing arrays are CuArrays.

jkozdon: > Ultimately, I'm trying to eliminate any resulting horizontal fluxes-- do I need to compute this average using the metric terms in order to eliminate horizontal fluxes?

OK. I misunderstood what you were doing. In this case, you are right that it doesn't matter too much.

Also, I pointed to the wrong code snippet, you don't want the vertical integral but the horizontal integral stuff that was put into the diagnostics: https://github.com/CliMA/ClimateMachine.jl/blob/fa76456047bba885a17f9f73756ed3e4b2cd6eba/src/Diagnostics/atmos_les_core.jl#L65

charleskawczynski: Ok, it should now work on the GPU.

PR #1528, opened 2020-09-08
Add HLLC numerical flux

mwarusz:
Co-authored-by: Thomas Gibson <thomas.gibson@nps.edu>

# Description

Adds HLLC numerical flux.

I have

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -72,8 +72,10 @@ import ..DGMethods.NumericalFluxes:
     CentralNumericalFluxHigherOrder,
     CentralNumericalFluxDivergence,
     CentralNumericalFluxFirstOrder,
-    numerical_flux_first_order!
-using ..DGMethods.NumericalFluxes: RoeNumericalFlux
+    numerical_flux_first_order!,
+    NumericalFluxFirstOrder
+using ..DGMethods.NumericalFluxes:
+    RoeNumericalFlux, HLLCNumericalFlux, RusanovNumericalFlux
 
 import ..Courant: advective_courant, nondiffusive_courant, diffusive_courant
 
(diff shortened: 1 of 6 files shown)


mwarusz:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,138 @@ function numerical_flux_first_order!(
...
+        state_auxiliary⁺,
+        t,
+        direction,
+    )
+    fluxᵀn⁺ = flux⁺' * normal_vector
+
+    ρ⁻ = state_prognostic⁻.ρ
+    ρu⁻ = state_prognostic⁻.ρu
+    ρe⁻ = state_prognostic⁻.ρe
+    ts⁻ = thermo_state(
```
charleskawczynski, line 945: Is this expected to work at the boundaries? At the boundary, `state_auxiliary⁻` (specifically variables that are a function of the state) is not necessarily in sync with the state assigned in `boundary_state!`, in which case, you'll get different results by creating a new thermodynamic state:
```julia
ts⁻ = PhaseEquil(
        atmos.param_set,
        internal_energy(atmos, state, aux),
        state⁻.ρ,
        state⁻.moisture.ρq_tot / state.ρ,
        moist.maxiter,
        atmos.moisture.tolerance,
    )
```
The problem with this is that it will call saturation adjustment.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,138 @@ function numerical_flux_first_order!(
...
+        balance_law,
+        balance_law.moisture,
+        state_prognostic⁻,
+        state_auxiliary⁻,
+    )
+
+    ρ⁺ = state_prognostic⁺.ρ
+    ρu⁺ = state_prognostic⁺.ρu
+    ρe⁺ = state_prognostic⁺.ρe
+    ts⁺ = thermo_state(
```
charleskawczynski, line 967: Is this expected to work at the boundaries? At the boundary, `state_auxiliary⁺` (specifically variables that are a function of the state) is not necessarily in sync with the state assigned in `boundary_state!`, in which case, you'll get different results by creating a new thermodynamic state:
```julia
ts⁺ = PhaseEquil(
        atmos.param_set,
        internal_energy(atmos, state, aux),
        state⁺.ρ,
        state⁺.moisture.ρq_tot / state.ρ,
        moist.maxiter,
        atmos.moisture.tolerance,
    )
```
The problem with this is that it will call saturation adjustment.

jkozdon: Shouldn't this always be the case though? If you modify the state in a way that requires a saturation adjustment.

charleskawczynski: There are two ways to compute thermo states. One that calls SA (ideally once in `nodal_update_auxiliary_state!`)
```julia
ts = PhaseEquil(
        atmos.param_set,
        internal_energy(atmos, state, aux),
        state.ρ,
        state.moisture.ρq_tot / state.ρ,
        moist.maxiter,
        atmos.moisture.tolerance,
    )
```
And one that _recovers_ the thermodynamic state:
```julia
ts = thermo_state(atmos, state, aux)
```
This second method grabs the temperature from the `aux` state, and calls the `PhaseEquil{...}` (the default constructor) _with_ temperature (computed in `nodal_update_auxiliary_state!`), so SA is never performed with this second method.

The idea was that users should only need to call the second method. This assumes that the `aux` and `prognostic` states are in sync, but this may be invalid if we're updating/changing the prognostic state (as is the case on the boundaries).

mwarusz: Sounds like there is no way to avoid SA if we need pressure that is consistent with the boundary state ? Wondering where it should be performed. Other numerical fluxes don't need it so doing it  in `boundary_state!` would be wasteful. Maybe we should define a special method of `numerical_boundary_flux_first_order!` for `HLLC` and `Roe`. Or is it better to do it outside of surface kernels and communicate through aux like for volume SA ?

jkozdon:
> Or is it better to do it outside of surface kernels and communicate through aux like for volume SA ?

Not sure this would be possible since we do not have aux storage for the plus side of the boundary (the plus side is set to the minus side on the boundary).

This could change, but we'd have to modify the mesh generation and array creation.

src/Atmos/Model/AtmosModel.jl
jkozdon: > Or is it better to do it outside of surface kernels and communicate through aux like for volume SA ?

Not sure this would be possible since we do not have aux storage for the plus side of the boundary (the plus side is set to the minus side on the boundary).

This could change, but we'd have to modify the mesh generation and array creation.

charleskawczynski: I'm not sure what the best way forward is, but it seems like, for now, do the correct thing and call SA. We should probably have a discussion regarding this (I imagine that this may be problematic in several places).

charleskawczynski: Maybe we can meet to discuss this more?

thomasgibson: Agreed. I think we should definitely talk about this

jkozdon: It's possible that we can get away with an SA in the boundary state. We won't know until the code is profiled to see if we had register spilling or other problems.

We moved it to aux initially due to performance issues, but this situation might not be as bad.

Also, is it clear that we shouldn't be doing it for all numerical fluxes? (All the fluxes evaluate the flux with the boundary state which in turn involves pressure and wavespeed)

mwarusz: You're right. All fluxes have the plus state pressure and sound speed inconsistent with BCs when using `EquilMoist`. I will check if performing SA in `boundary_state!` is a performance problem.

charleskawczynski: > It's possible that we can get away with an SA in the boundary state. 

Agreed. Let's try it.

> We won't know until the code is profiled to see if we had register spilling or other problems.

✔️ 

> Agreed. I think we should definitely talk about this

While we have a quick/easy first pass at a solution to this particular problem, I agree that we should still talk about the general concern of accidentally using `aux⁺` (in the case that it's not in sync with `state⁺`).

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,171 @@ function numerical_flux_first_order!(
...
+        direction,
+    )
+    fluxᵀn⁺ = flux⁺' * normal_vector
+
+    # Extract relevant fields and thermodynamic variables defined on
+    # the positive + and negative - sides of the interior facets
+    ρ⁻ = state_prognostic⁻.ρ
+    ρu⁻ = state_prognostic⁻.ρu
+    ρe⁻ = state_prognostic⁻.ρe
+    ts⁻ = thermo_state(
```
charleskawczynski, line 976:
```suggestion
    ts⁻ = recover_thermo_state(
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,171 @@ function numerical_flux_first_order!(
...
+        balance_law,
+        balance_law.moisture,
+        state_prognostic⁻,
+        state_auxiliary⁻,
+    )
+
+    ρ⁺ = state_prognostic⁺.ρ
+    ρu⁺ = state_prognostic⁺.ρu
+    ρe⁺ = state_prognostic⁺.ρe
+    ts⁺ = thermo_state(
```
charleskawczynski, line 998:
```suggestion
    ts⁺ = new_thermo_state(
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,171 @@ function numerical_flux_first_order!(
...
+
+    u⁺ = ρu⁺ / ρ⁺
+    uᵀn⁺ = u⁺' * normal_vector
+    e⁺ = ρe⁺ / ρ⁺
+    p⁺ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_prognostic⁺,
+        state_auxiliary⁺,
+    )
```
charleskawczynski, line 1013:
```suggestion
    p⁺ = air_pressure(ts⁺)
```

charleskawczynski: I'm now thinking that we should remove `pressure`, since this relies on `recover_thermo_state` and will have the same issue as above. Using `ts⁺ = new_thermo_state(...)` will fix these issues.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -896,4 +898,171 @@ function numerical_flux_first_order!(
...
+    c⁻ = soundspeed_air(ts⁻)
+
+    uᵀn⁻ = u⁻' * normal_vector
+    e⁻ = ρe⁻ / ρ⁻
+    p⁻ = pressure(
+        balance_law,
+        balance_law.moisture,
+        state_prognostic⁻,
+        state_auxiliary⁻,
+    )
```
charleskawczynski, line 993:
```suggestion
    p⁻ = air_pressure(ts⁻)
```
This one doesn't matter as much, but we might as call `air_pressure` directly.

src/Atmos/Model/AtmosModel.jl
mwarusz: Should this become `recover_thermo_state` after #1542 gets merged ?

charleskawczynski: Ah, yes. So long `numerical_flux_first_order!` is called right after `boundary_state!`.

mwarusz: It is https://github.com/CliMA/ClimateMachine.jl/blob/721b67f2fc72692fe4daf63d9c091d3398b1b889/src/Numerics/DGMethods/NumericalFluxes.jl#L147-L171

bischtob:
@thomasgibson, @mwarusz is this good to go pending @charleskawczynski's changes?

src/Atmos/Model/AtmosModel.jl
mwarusz: Done

src/Atmos/Model/AtmosModel.jl
mwarusz: Done

src/Atmos/Model/AtmosModel.jl
mwarusz: Done

src/Atmos/Model/AtmosModel.jl
mwarusz: Done

mwarusz:
bors try

simonbyrne (approved):

PR #1530, opened 2020-09-09
Project density to N-1 degree polynomial in the vertical

mwarusz:
Co-authored-by: Thomas Gibson <thomas.gibson@nps.edu>

# Description

Applies the projection technique from the paper https://doi.org/10.1002/fld.4197 that may help with stability.
The highest order mode of density in the vertical is filtered out in the gravity source term.

@bischtob 

I have

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -35,6 +35,7 @@ using ..Mesh.Grids:
     min_node_distance,
     EveryDirection,
     Direction
+using ..Mesh.Filters
 
 using ClimateMachine.BalanceLaws
 using ClimateMachine.Problems
@@ -316,6 +317,7 @@ debug variables.
 """
 function vars_state(m::AtmosModel, st::Auxiliary, FT)
     @vars begin
+        projected_ρ::FT
         ∫dz::vars_state(m, UpwardIntegrals(), FT)
         ∫dnz::vars_state(m, DownwardIntegrals(), FT)
         coord::SVector{3, FT}
@@ -597,6 +599,22 @@ function update_auxiliary_state!(
 
     update_auxiliary_state!(nodal_update_auxiliary_state!, dg, m, Q, t, elems)
 
+    # Filter the highest order vertical mode from density
+    grid = dg.grid
+    if m.ref_state isa HydrostaticState
+        target = AtmosFilterDensityPerturbation(m)
+    else
+        target = (:projected_ρ,)
+    end
+    Filters.apply!(
+        state_auxiliary,
+        target,
+        grid,
+        Filters.CutoffFilter(grid),
+        state_auxiliary = state_auxiliary,
+        direction = VerticalDirection(),
+    )
+
     # TODO: Remove this hook. This hook was added for implementing
     # the first draft of EDMF, and should be removed so that we can
     # rely on a single vertical element traversal. This hook allows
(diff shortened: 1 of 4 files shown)


jkozdon (approved):
Looks to me like it does what you say it does.

On the sphere should just density be filtered or the whole source term? (Maybe also for curved meshes?)

Does it help?

jkozdon:
Also, should this be an option or the behavior that's always used? (I don't know, just asking)

mwarusz:
> Does it help?

We don't know yet. It was coded up so that others can try it out.

> Also, should this be an option or the behavior that's always used? (I don't know, just asking)

Depends on the answer to the previous question I guess.

> On the sphere should just density be filtered or the whole source term? (Maybe also for curved meshes?)

Good question. I am not sure.

szy21:
I run solid-body-rotation.jl (https://github.com/CliMA/ClimateMachine.jl/blob/zs/dycore_test_case/experiments/TestCase/solid-body-rotation.jl) with this branch and there is still some oscillation (difference between t=12h and initial state):

image

mwarusz:
>I run solid-body-rotation.jl (https://github.com/CliMA/ClimateMachine.jl/blob/zs/dycore_test_case/experiments/TestCase/solid->body-rotation.jl) with this branch and there is still some oscillation (difference between t=12h and initial state):

Dis use a flux with the contact property ? This property is still needed even with this change. This change is supposed to make simulations more stable without explicit filtering.

szy21:
> > I run solid-body-rotation.jl ([https://github.com/CliMA/ClimateMachine.jl/blob/zs/dycore_test_case/experiments/TestCase/solid->body-rotation.jl](https://github.com/CliMA/ClimateMachine.jl/blob/zs/dycore_test_case/experiments/TestCase/solid-%3Ebody-rotation.jl)) with this branch and there is still some oscillation (difference between t=12h and initial state):
> 
> Dis use a flux with the contact property ? This property is still needed even with this change. This change is supposed to make simulations more stable without explicit filtering.

No I didn't. I could run it again with Roe flux. But there is almost no oscillation with Roe flux without this change. Maybe I could try to run it without the exponential filter?

mwarusz:
> Maybe I could try to run it without the exponential filter?

That would be good to try to see if it`s more stable.

szy21:
> > Maybe I could try to run it without the exponential filter?
> 
> That would be good to try to see if it`s more stable.

It is more stable I think. Now I can run solid body rotation without any filter for 12 simulated hours (without the change it crashed at hour 1). This is IMEX with Roe flux. Here is the difference at hour 12:

image

src/Atmos/Model/AtmosModel.jl
```diff
@@ -594,6 +596,22 @@ function update_auxiliary_state!(
...
+    if m.ref_state isa HydrostaticState
+        target = AtmosFilterDensityPerturbation(m)
+    else
+        target = (:projected_ρ,)
+    end
+    Filters.apply!(
+        state_auxiliary,
+        target,
+        grid,
+        Filters.CutoffFilter(grid),
```
blallen, line 610: Question for my understanding: where is the cutoff to N-1 specified? I'm used to having to do `CutoffFilter(grid, polynomialorder(grid) - 1)`

jkozdon: The docs https://github.com/CliMA/ClimateMachine.jl/blob/2be057f82576c4ec3f95243854acc8662872c4a3/src/Numerics/Mesh/Filters.jl#L214-L219
say that without it will cut off the highest polynomial mode.

Guessing the line you have was my fault, maybe we are being too aggressive in the ocean code(?).

blallen: @sandreza does it make sense that we're preserving discrete incompressibility if we are actually zeroing out the top two modes instead of just the top mode? because we seem to still be preserving it based on our tests

PR #1531, opened 2020-09-09
add NonEquilMoist model to Atmos

trontrytel:
# Description

This adds a non-equilibrium model of moisture in Atmos by:
  - adding two more variables to state (rho_q_liq, rho_q_ice)
  - replacing saturation adjustment with relaxation to equilibrium source terms.

@akshaysridhar , @charleskawczynski - could you take a look and check if I'm missing something?

@kpamnany - I included here a draft of possible changes in the diagnostics. It's not the most elegant. Could you also take a look there?

I have

### DIFF
--- a/experiments/AtmosLES/bomex_model.jl
+++ b/experiments/AtmosLES/bomex_model.jl
@@ -77,8 +77,23 @@ using ClimateMachine.BalanceLaws:
 
 using CLIMAParameters
 using CLIMAParameters.Planet: e_int_v0, grav, day
-struct EarthParameterSet <: AbstractEarthParameterSet end
-const param_set = EarthParameterSet()
+using CLIMAParameters.Atmos.Microphysics
+
+struct LiquidParameterSet <: AbstractLiquidParameterSet end
+struct IceParameterSet <: AbstractIceParameterSet end
+
+struct MicropysicsParameterSet{L, I} <: AbstractMicrophysicsParameterSet
+    liq::L
+    ice::I
+end
+
+struct EarthParameterSet{M} <: AbstractEarthParameterSet
+    microphys::M
+end
+
+microphys = MicropysicsParameterSet(LiquidParameterSet(), IceParameterSet())
+
+const param_set = EarthParameterSet(microphys)
 
 import ClimateMachine.Atmos: atmos_source!
 using ClimateMachine.Atmos: altitude, recover_thermo_state
@@ -358,6 +373,10 @@ function init_bomex!(problem, bl, state, aux, (x, y, z), t)
     state.ρu = SVector(ρu, ρv, ρw)
     state.ρe = ρe_tot
     state.moisture.ρq_tot = ρ * q_tot
+    if bl.moisture isa NonEquilMoist
+        state.moisture.ρq_liq = FT(0)
+        state.moisture.ρq_ice = FT(0)
+    end
 
     if z <= FT(400) # Add random perturbations to bottom 400m of model
         state.ρe += rand() * ρe_tot / 100
@@ -372,6 +391,7 @@ function bomex_model(
     zmax,
     surface_flux;
     turbconv = NoTurbConv(),
+    moisture_model = "equilibrium",
 ) where {FT}
 
     ics = init_bomex!     # Initial conditions
@@ -407,7 +427,7 @@ function bomex_model(
     f_coriolis = FT(0.376e-4) # Coriolis parameter
 
     # Assemble source components
-    source = (
+    source_default = (
         Gravity(),
         BomexTendencies{FT}(
             ∂qt∂t_peak,
@@ -431,17 +451,31 @@ function bomex_model(
         BomexGeostrophic{FT}(f_coriolis, u_geostrophic, u_slope, v_geostrophic),
         turbconv_sources(turbconv)...,
     )
+    if moisture_model == "equilibrium"
+        source = source_default
+        moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1))
+    elseif moisture_model == "nonequilibrium"
+        source = (source_default..., CreateClouds())
+        moisture = NonEquilMoist()
+    else
+        @warn @sprintf(
+            """
+%s: unrecognized moisture_model in source terms, using the defaults""",
+            moisture_model,
+        )
+        source = source_default
+    end
 
     # Set up problem initial and boundary conditions
     if surface_flux == "prescribed"
-        energy = PrescribedEnergyFlux((state, aux, t) -> LHF + SHF)
-        moisture = PrescribedMoistureFlux((state, aux, t) -> moisture_flux)
+        energy_bc = PrescribedEnergyFlux((state, aux, t) -> LHF + SHF)
+        moisture_bc = PrescribedMoistureFlux((state, aux, t) -> moisture_flux)
     elseif surface_flux == "bulk"
-        energy = BulkFormulaEnergy(
+        energy_bc = BulkFormulaEnergy(
             (state, aux, t, normPu_int) -> C_drag,
             (state, aux, t) -> (T_sfc, q_sfc),
         )
-        moisture = BulkFormulaMoisture(
+        moisture_bc = BulkFormulaMoisture(
             (state, aux, t, normPu_int) -> C_drag,
             (state, aux, t) -> q_sfc,
         )
@@ -452,6 +486,7 @@ function bomex_model(
             surface_flux,
         )
     end
+
     problem = AtmosProblem(
         boundarycondition = (
             AtmosBC(
@@ -460,8 +495,8 @@ function bomex_model(
                     # P represents the projection onto the horizontal
                     (state, aux, t, normPu_int) -> (u_star / normPu_int)^2,
                 )),
-                energy = energy,
-                moisture = moisture,
+                energy = energy_bc,
+                moisture = moisture_bc,
                 turbconv = turbconv_bcs(turbconv),
             ),
             AtmosBC(),
@@ -475,10 +510,11 @@ function bomex_model(
         param_set;
         problem = problem,
         turbulence = SmagorinskyLilly{FT}(C_smag),
-        moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1)),
+        moisture = moisture,
         source = source,
         turbconv = turbconv,
     )
+
     return model
 end
 
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -1,5 +1,5 @@
 using ..Microphysics_0M
-using CLIMAParameters.Planet: Omega, e_int_i0, cv_d, cv_l, cv_i, T_0
+using CLIMAParameters.Planet: Omega, e_int_i0, cv_l, cv_i, T_0
 
 export Source,
     Gravity,
@@ -7,7 +7,8 @@ export Source,
     Subsidence,
     GeostrophicForcing,
     Coriolis,
-    RemovePrecipitation
+    RemovePrecipitation,
+    CreateClouds
 
 # kept for compatibility
 # can be removed if no functions are using this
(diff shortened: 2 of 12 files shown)


trontrytel:
@charleskawczynski - The structure from microphysics related CLIMAParameters adds a bit of chatter:

https://github.com/CliMA/ClimateMachine.jl/blob/00a61a5a267eee0bc7bc584ab23b78b225edd80c/experiments/AtmosLES/bomex_model.jl#L82-L96

Not sure how to avoid this.

trontrytel:
Also, I noticed that the linearized pressure for `EquilMoist` model assumes that `q_liq` and `q_ice` are zero:

https://github.com/CliMA/ClimateMachine.jl/blob/00a61a5a267eee0bc7bc584ab23b78b225edd80c/src/Atmos/Model/linear.jl#L48-L63

@tapios, was this intentional? Should I do the same for the `NonEquilMoist` model?

tapios:
> Also, I noticed that the linearized pressure for `EquilMoist` model assumes that `q_liq` and `q_ice` are zero:
> 
> 
> 
> https://github.com/CliMA/ClimateMachine.jl/blob/00a61a5a267eee0bc7bc584ab23b78b225edd80c/src/Atmos/Model/linear.jl#L48-L63
> 
> 
> 
> @tapios, was this intentional? Should I do the same for the `NonEquilMoist` model?

The linearized pressure is used for linearizations of the EOMs in terms of the state variables (for implicit timestepping). For the equilibrium moisture model, only the total water specific humidity is a state variable. There, liquid and ice do not appear.

In the non-equilibrium case, `q_liq` and `q_ice` are state variables and should appear.

kpamnany:
> @charleskawczynski - The structure from microphysics related CLIMAParameters adds a bit of chatter:

@charleskawczynski: I suggest a Driver-like default setup made available in CLIMAParameters, such that one can do the following in an experiment:
```julia
param_set = DefaultEarthParameterSet()
```
Which will do everything done in this "chatter". CLIMAParameters should ensure that only a single such default parameter set exists. Seems like it would help?

kpamnany:
> @kpamnany - I included here a draft of possible changes in the diagnostics. It's not the most elegant. Could you also take a look there?

This looks very neatly done to me within the constraints of the existing structure; what don't you like about it?

experiments/AtmosLES/bomex_model.jl
```diff
@@ -475,7 +493,8 @@ function bomex_model(
         param_set;
         problem = problem,
         turbulence = SmagorinskyLilly{FT}(C_smag),
-        moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1)),
+        #moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1)),
+        moisture = NonEquilMoist(),
```
charleskawczynski, line 497: The default bomex model will now use non-equilibrium thermodynamics? 😅 @kpamnany, @yairchn should we expose this as a kwarg to `bomex_model`? The EDMF model has only been tested on the `EquilMoist` so far.

src/Atmos/Model/source.jl
```diff
@@ -182,6 +183,49 @@ function atmos_source!(
...
+)
+    # current phase partition
+    FT = eltype(state)
+    ts = thermo_state(atmos, state, aux)
+    q = PhasePartition(ts)
+    T = air_temperature(ts)
+
+    # saturation specific humidity over liquid (packed into PhasePartition)
+    q_liq_sat =
+        q_vap_saturation_generic(atmos.param_set, T, state.ρ, phase = Liquid())
```
charleskawczynski, line 212: Remind me to write a wrapper for this so that you can call this with
```suggestion
    q_liq_sat = q_vap_saturation(ts, Liquid())
```

src/Diagnostics/thermo.jl
```diff
@@ -61,3 +69,14 @@ function compute_thermo!(moist::EquilMoist, state, aux, ts, thermo)
 
     return nothing
 end
+function compute_thermo!(moist::NonEquilMoist, state, aux, ts, thermo)
+    Phpart = PhasePartition(ts)
+
+    thermo.moisture.q_liq = Phpart.liq
+    thermo.moisture.q_ice = Phpart.ice
```
charleskawczynski, line 76:
```suggestion
    thermo.moisture.q_liq = liquid_specific_humidity(ts)
    thermo.moisture.q_ice = ice_specific_humidity(ts)```

File: src/Diagnostics/atmos_les_default.jl [hunk 4]
```diff
@@ -404,10 +489,21 @@ function atmos_les_default_collect(dgngrp::DiagnosticsGroup, currtime)
 
         # FIXME properly
         if isa(bl.moisture, EquilMoist)
-            if !iszero(thermo.moisture.q_liq)
+            if !iszero(thermo.moisture.q_liq + thermo.moisture.q_ice)
```
charleskawczynski, line 492: Can we call `has_condensate(::ThermodynamicState)` here?

trontrytel:
> > @kpamnany - I included here a draft of possible changes in the diagnostics. It's not the most elegant. Could you also take a look there?
> 
> This looks very neatly done to me within the constraints of the existing structure; what don't you like about it?

Thanks! It seemed that I'm adding repetitive lines and maybe it could be avoided. But it does what I want it to do.

experiments/AtmosLES/bomex_model.jl
trontrytel: Haha - no, sorry. This is me still testing. But it would be good to be able to switch between the two setups. I'll add that.

kpamnany:
> Thanks! It seemed that I'm adding repetitive lines and maybe it could be avoided. But it does what I want it to do.

Depending on when you merge this, you might have to change it a bit as I'm opening a PR shortly that touches this code. I am planning to move all the `if isa(bl.moisture, EquilMoist)` stuff into functions specialized on `MoistureModel` -- what you've done here will make it easier.

experiments/AtmosLES/bomex_model.jl
trontrytel: @charleskawczynski - I added the optional `--moisture-model` keyword argument to bomex.

trontrytel:
Also, a qiestion to @akshaysridhar . For moisture boundary conditions. Do I understand the code correctly that right now if I specify nothing for `rho_q_liq` and `rho_q_ice` it will default to zero flux?

src/Diagnostics/atmos_les_default.jl
trontrytel: done

trontrytel:
I think that the Non Equilbrium Mosture configuration is ready. I updated with the most recent changes to `thermo_state`. 

Here are the results for Bomex for liquid water when comparing between the equilibrium and non-equilibrium configurations. I'll run the simulation for longer.

eq_non_eq jpeg 001

From the couple of runs that I did it seems to be tiny little bit faster than equilibrium option. Plus, it has the potential to be less fragile, because it does not rely on saturation adjustment. 

@charleskawczynski, @kpamnany - the only thing left that I see is the cleanup from the overhead from the structures in CLIMAParameters. Let me know if you want to do it separately or in this PR.

@LenkaNovak , @szy21 - I could add it to some GCM experiment setup as well? Maybe it would help with testing. If you think it's useful, let me know which one.

src/Diagnostics/atmos_les_default.jl
```diff
@@ -404,10 +489,21 @@ function atmos_les_default_collect(dgngrp::DiagnosticsGroup, currtime)
...
+                qc_gt_0_z[evk][idx] = one(FT)
+                qc_gt_0_full[idx] = one(FT)
+
+                z = zvals[evk]
+                cld_top = max(cld_top, z)
+                cld_base = min(cld_base, z)
+            end
+        end
+        if isa(bl.moisture, NonEquilMoist)
+            if thermo.moisture.has_condensate
```
charleskawczynski, line 503: My original thought was to call `has_condensate` given a thermo state (on the fly) here. But I now see that it's a bit unclear how to do that. I'm fine with this so long this isn't a mem/perf issue in terms of storing a `Bool` in the state along with all of the other `FT`s. cc @kpamnany

tapios:
> I think that the Non Equilbrium Mosture configuration is ready. I updated with the most recent changes to `thermo_state`.
> 
> Here are the results for Bomex for liquid water when comparing between the equilibrium and non-equilibrium configurations. I'll run the simulation for longer.
> 
> eq_non_eq jpeg 001
> 
> From the couple of runs that I did it seems to be tiny little bit faster than equilibrium option. Plus, it has the potential to be less fragile, because it does not rely on saturation adjustment.
> 
> @charleskawczynski, @kpamnany - the only thing left that I see is the cleanup from the overhead from the structures in CLIMAParameters. Let me know if you want to do it separately or in this PR.
> 
> @LenkaNovak , @szy21 - I could add it to some GCM experiment setup as well? Maybe it would help with testing. If you think it's useful, let me know which one.



> I think that the Non Equilbrium Mosture configuration is ready. I updated with the most recent changes to `thermo_state`.
> 
> Here are the results for Bomex for liquid water when comparing between the equilibrium and non-equilibrium configurations. I'll run the simulation for longer.
> 
> eq_non_eq jpeg 001
> 
> From the couple of runs that I did it seems to be tiny little bit faster than equilibrium option. Plus, it has the potential to be less fragile, because it does not rely on saturation adjustment.
> 
> @charleskawczynski, @kpamnany - the only thing left that I see is the cleanup from the overhead from the structures in CLIMAParameters. Let me know if you want to do it separately or in this PR.
> 
> @LenkaNovak , @szy21 - I could add it to some GCM experiment setup as well? Maybe it would help with testing. If you think it's useful, let me know which one.

It's great this is working! Am I reading this right that q_l is larger for non-equilibrium moisture, especially at later times? Is this robust? If so, can you recover the equilibrium result when you make the relaxation time short?

src/Atmos/Model/source.jl
```diff
@@ -182,6 +183,43 @@ function atmos_source!(
...
+    # get current temperature and phase partition
+    FT = eltype(state)
+    ts = recover_thermo_state(atmos, state, aux)
+    q = PhasePartition(ts)
+    T = air_temperature(ts)
+
+    # phase partition corresponding to the current T and q.tot
+    # (this is not the same as phase partition from saturation adjustment)
+    ts_eq = TemperatureSHumEquil(atmos.param_set, T, state.ρ, q.tot)
+    q_eq = PhasePartition(ts_eq)
```
charleskawczynski, line 213: Now I understand 😄. Looks good!

charleskawczynski (approved):
I'd almost rather `moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1))` be used as the optional kwarg as the input argument, but it's nothing to hold this up. Looks good on my side!

charleskawczynski:
> If so, can you recover the equilibrium result when you make the relaxation time short?

Sounds like a _great_ test.

kpamnany:
> @charleskawczynski, @kpamnany - the only thing left that I see is the cleanup from the overhead from the structures in CLIMAParameters. Let me know if you want to do it separately or in this PR.

It's up to Charlie, but I think that we might as well use a `DefaultEarthParameterSet` in all the test/experiment code, which could be a separate PR.

charleskawczynski:
> It's up to Charlie, but I think that we might as well use a `DefaultEarthParameterSet` in all the test/experiment code, which could be a separate PR.

Let's merge this as is, and we can incubate a bit on `DefaultEarthParameterSet` and implement that in a later PR.

szy21:
> @LenkaNovak , @szy21 - I could add it to some GCM experiment setup as well? Maybe it would help with testing. If you think it's useful, let me know which one.

You could try to add it to baroclinic-wave.jl, but I don't know what to expect for the result. @LenkaNovak may have better suggestions. I'm happy to merge it without GCM testing.

src/Diagnostics/atmos_les_default.jl
trontrytel: I see a thumbs up. So I'm gonna assume it's fine :)

src/Atmos/Model/source.jl
trontrytel: I had to re-understand it myself!

trontrytel:
Thank you for the reviews. I'll r+ it and leave the cleanup related to `DefaultEarthParameterSet` for later PR. 

@tapios - I'll test and compare the equilibrium vs non-equilibrium Bomex for longer simulations. They should be converging. So either the relaxation time scale should be shorter or the initial condition (with non-zero `q_liq` already at t=0 for equilibrium case vs `q_liq = 0` for non-equilibrium case) is still affecting the results. 

@charleskawczynski - You are right. This could be added as an integration test at some point. 

@szy21 - Sounds good. Let's do it (maybe after the code freeze?). Would be good to see how it works with both liquid and ice phase. I expect some additional stability problems along the freezing level.

src/Atmos/Model/source.jl
charleskawczynski: Not now, but perhaps we can write a `new_thermo_state_eq` convenience wrapper in the case that equilibrium states/phase partitioning is frequently needed in the NonEquil components.

trontrytel: Agreed. Maybe once I add the precipitation interface and all the microphysics source terms are integrated, we can reevaluate what wrappers we need. It will be more clear what functions are used and to what purpose.

PR #1555, opened 2020-09-15
Using implicit solver for soil problems

Zhengyu-Huang:
# Description

Set up an example for using implicit solver for the soil model, this example is in test/Land/Model/haverkamp_test_implicit.jl, which needs to use the implicit solver  [https://github.com/CliMA/ClimateMachine.jl/pull/1483](url)

I have


simonbyrne:
Can you hook it up so it runs as part of the tests?

simonbyrne:
@Zhengyu-Huang Can you add it to buildkite?

Zhengyu-Huang:
> @Zhengyu-Huang Can you add it to buildkite?

Sure

test/Land/Model/haverkamp_implicit_test.jl
```diff
@@ -0,0 +1,200 @@
...
+    timeend = FT(60 * 60 * 24)
+
+    dt = FT(200)
+
+    solver_config = ClimateMachine.SolverConfiguration(
+        t0,
+        timeend,
+        driver_config,
+        ode_dt = dt,
+    )
```
kmdeck, line 122: Let's remove these lines

test/Land/Model/haverkamp_implicit_test.jl
```diff
@@ -0,0 +1,200 @@
...
+        split_explicit_implicit = false,
+        variant = NaiveVariant(),
+    )
+
+    solver_config.solver = ode_solver
+
+    #######################################
+
+    mygrid = solver_config.dg.grid
+    Q = solver_config.Q
```
kmdeck, line 179: I think we can delete, as Q defined above?

test/Land/Model/haverkamp_implicit_test.jl
```diff
@@ -0,0 +1,200 @@
...
+        vdg,
+        NonLinearBackwardEulerSolver(
+            nonlinearsolver;
+            isadjustable = true,
+            preconditioner_update_freq = 100,
+        ),
+        Q;
+        dt = dt,
+        t0 = 0,
+        split_explicit_implicit = false,
```
kmdeck, line 170: I thought vdg was tendency the implicit solver was used with, and dg-vdg would use the explicit solver. Is that true in general, but here, since we only have one column, you can set split_explicit_implicit = false?

kmdeck (commented):
looks good to me, left a couple minor comments. before merging, I think it is convention to squash your commits. The beginning of this goes over that.
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

kmdeck (approved):
approving so you can submit when you want! but should squash commits first I think.

jkozdon:
> but should squash commits first I think.

Yes. Please do.

test/Land/Model/haverkamp_implicit_test.jl
Zhengyu-Huang: " I thought vdg was tendency the implicit solver was used with, and dg-vdg would use the explicit solver. Is that true in general, but here, since we only have one column ", yes
my understanding of "split_explicit_implicit = false",  we use dg-vdg as explicit part instead of defining it https://clima.github.io/ClimateMachine.jl/latest/APIs/Numerics/ODESolvers/ODESolvers/#ClimateMachine.ODESolvers.AdditiveRungeKutta

Zhengyu-Huang:
> looks good to me, left a couple minor comments. before merging, I think it is convention to squash your commits. The beginning of this goes over that.
> https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

I think I messed this up with squash , should I close this PR and do new PR ?

PR #1563, opened 2020-09-17
Remove overloaded thermo in mms test

charleskawczynski:
# Description

This PR removes the overloaded thermodynamics in the mms tests. To ensure that the compared solution is in sync, I've also updated the solution generator to compute `γ` from CLIMAParameters. Unfortunately, `γ` computed from CLIMAParameters cannot be represented as a rational number, so SymPy can no longer simplify the generated solution (which is a bit uglier in this PR). I'm on the fence with these changes, but I thought that I would open a PR showing what's down this road.

I've also removed the duplicate `mms_solution_generated.jl` in the `test/Driver/` folder and just grabbed it from the global path.

I have


tapios (approved):
LGTM. To me, the benefit of having the redundant code removed is worth the minor aesthetic cost.

PR #1564, opened 2020-09-17
Add tests for initializing small planet

charleskawczynski:
# Description

When playing around with the solid-body rotation problem, one thing I was messing around with was changing the planet size. This PR adds a test to demonstrate the range of planet sizes that can be initialized. Therefore, no `@tests` are technically added, but failures should still pop up if initialization fails.

I have

### DIFF
--- a/test/Numerics/DGMethods/test_planet_sizes.jl
+++ b/test/Numerics/DGMethods/test_planet_sizes.jl
@@ -1,0 +1,55 @@
+using ClimateMachine
+using Test
+using ClimateMachine.Atmos
+using ClimateMachine.ConfigTypes
+using ClimateMachine.Diagnostics
+using ClimateMachine.Mesh.Interpolation
+using CLIMAParameters
+using CLIMAParameters.Planet: planet_radius
+import CLIMAParameters
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 1
+function initialize_planet_size(domain_height)
+    FT, n_horz, n_vert, poly_order = Float64, 12, 6, 3
+    model = AtmosModel{FT}(
+        AtmosGCMConfigType,
+        param_set;
+        init_state_prognostic = x -> x,
+    )
+    _planet_radius = FT(planet_radius(param_set))
+    driver_config = ClimateMachine.AtmosGCMConfiguration(
+        "SmallPlanet",
+        poly_order,
+        (n_horz, n_vert),
+        domain_height,
+        param_set,
+        x -> x;
+        model = model,
+    )
+    info = driver_config.config_info
+    boundaries = [
+        FT(-90) FT(-180) _planet_radius
+        FT(90) FT(180) FT(_planet_radius + info.domain_height)
+    ]
+    resolution = (FT(2), FT(2), FT(1000)) # in (deg, deg, m)
+    interpol = ClimateMachine.InterpolationConfiguration(
+        driver_config,
+        boundaries,
+        resolution,
+    )
+    return nothing
+end
+
+let
+    ClimateMachine.init()
+    for planet_rad in 10.0 .^ (0:7)
+        CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = planet_rad
+        for aspect_ratio in 10.0 .^ (-2.33:0.6:0.3)
+            domain_height = aspect_ratio * planet_radius(param_set)
+            initialize_planet_size(domain_height)
+        end
+    end
+end
+
+nothing
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -204,6 +204,15 @@ steps:
   - label: "cpu_vars_test"
     key: "cpu_vars_test"
     command:
+      - "mpiexec julia --color=yes --project test/Numerics/DGMethods/test_planet_sizes.jl "
+    agents:
+      config: cpu
+      queue: central
+      slurm_ntasks: 1
+
+  - label: "cpu_vars_test"
+    key: "cpu_vars_test"
+    command:
       - "mpiexec julia --color=yes --project test/Numerics/DGMethods/vars_test.jl "
     agents:
       config: cpu


src/Numerics/Mesh/Interpolation.jl
```diff
@@ -737,7 +737,7 @@ struct InterpolationCubedSphere{
         qm1 = poly_order + 1
         toler1 = FT(eps(FT) * vert_range[1] * 2.0) # tolerance for unwarp function
         toler2 = FT(eps(FT) * 4.0)                 # tolerance
-        toler3 = FT(eps(FT) * vert_range[1] * 10.0) # tolerance for Newton-Raphson
+        toler3 = 10^5 * eps(FT) / vert_range[1] + 10 * eps(FT) * vert_range[1] # tolerance for Newton-Raphson
```
charleskawczynski, line 740: I divided the tolerance offset by `vert_range[1]` (the radius) so that the offset disappears for large planets and we roughly recover the original tolerance.

tapios: A hardwired 10^5 here seems weird. Any number that is not O(1) should be related to something physical (I don't know what it is in the context here).

charleskawczynski: True. I can change to `sqrt(eps(FT))`:
```julia
julia> 10^5*eps(Float64)
2.220446049250313e-11

julia> sqrt(eps(Float64))
1.4901161193847656e-8
```

tapios: Maybe better yet to understand where these tolerances come from. Is `vert_range` dimensional or non-dimensional? If it is dimensional, l. 740 is not dimensionally consistent.

jkozdon: I don't know what these are used for either, but my guess if you want something related to `eps(maximum(vert_range)))`

jkozdon: I believe this is code that @skandalaCLIMA wrote.

skandalaCLIMA: The tolerance needed to be set in relation to planet radius as there is precision loss when dealing with large numbers(planet radius). This was especially important for Float32 runs. But, we shouldn’t need to hardwire tolerance numbers though.

skandalaCLIMA: The tolerance needed to be set in relation to planet radius as there is precision loss when dealing with large numbers(planet radius). This was especially important for Float32 runs. But, we shouldn’t need to hardwire tolerance numbers though.

tapios: Can you suggest something that is dimensionally correct (i.e., does not have one term of units 1/m, added to a term of units m, as in the current l. 740?

skandalaCLIMA: The dimension of toler3, I believe, is meters. Easiest change would be to reduce the tolerance (multiplication by a factor of 10 or 100) or increase the number of Newton-Raphson iterations. Let me try to replicate this.

skandalaCLIMA: I ran a test (test/Numerics/Mesh/interpolation.jl) with _planet_radius = 1.0 and domain_height/_planet_radius = 30000/6.371e6 = 0.0047088. **The Float64 ran fine with an L_infinity max domain error of ~8E-8.** However, Float32 produced L_infinity max domain error of 0.0104.

skandalaCLIMA: @charleskawczynski : Please advise on the exact parameters (planet radius and domain height) used for the test run.

charleskawczynski: I'll try to set up an MWE

charleskawczynski: @skandalaCLIMA, here's an MWE (or Minimum Breaking Example? is MBE a thing?):
```julia
using ClimateMachine
ClimateMachine.init()
using ClimateMachine.Atmos
using ClimateMachine.ConfigTypes
using ClimateMachine.Diagnostics
using ClimateMachine.Mesh.Interpolation
using CLIMAParameters
using CLIMAParameters.Planet: planet_radius
import CLIMAParameters
struct EarthParameterSet <: AbstractEarthParameterSet end
const param_set = EarthParameterSet()
CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 1
function main()
    FT,n_horz,n_vert,poly_order = Float64,12,6,3
    model = AtmosModel{FT}(AtmosGCMConfigType, param_set; init_state_prognostic = x->x)
    domain_height = FT(30e3)
    _planet_radius = FT(planet_radius(param_set))
    driver_config = ClimateMachine.AtmosGCMConfiguration(
        "SmallPlanet", poly_order, (n_horz, n_vert),
        domain_height, param_set, x->x; model = model,
    )
    info = driver_config.config_info
    boundaries = [ FT(-90) FT(-180) _planet_radius
        FT(90) FT(180) FT(_planet_radius + info.domain_height)]
    resolution = (FT(2), FT(2), FT(1000)) # in (deg, deg, m)
    interpol = ClimateMachine.InterpolationConfiguration(
        driver_config,
        boundaries,
        resolution,
    )
    return nothing
end
main()
```

skandalaCLIMA: @charleskawczynski : The grid in the example above seems to be covering 30,000 m in the z direction for a 1 m radius planet. Can you please confirm if this is intended?

charleskawczynski: Yes, I realize that this is for more pencil-shaped atmospheric columns, which I suppose we should avoid. That said, it still fails at 10x aspect ratio:
```julia
CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 3000
...
    domain_height = FT(3000*10)
```

charleskawczynski: And 4x aspect ratio at `planet_radius = 10^4`:
```julia
CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 10^4
...
    domain_height = FT(10^4*4)
```

jkozdon: Would the whole problem be fixed if a `vert_range[end]` was used. Then you would be using the largest number instead of the smaller.

charleskawczynski: More thoroughly, passing in `domain_height`, and running

```julia
results = Dict()
for pr in 10.0 .^(0:6)
    CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 10^4
    for ar in 10.0 .^(-10:3:10)
        domain_height = ar*planet_radius(param_set)
        results[pr, ar] = false
        try
            main(domain_height)
            results[pr, ar] = true
        catch
        end
    end
end
```
Yields
```julia
julia> count(values(results))/length(values(results))
0.5714285714285714
```

Maybe we can add this as a test as
```julia
@test all(values(results))
```
?

skandalaCLIMA: The parameter toler3 is used in the non-linear solver, that essentially estimates inner coordinates from unwarped exterior coordinates. Toler3 measures how good the result is from expected.

30000x or 15000x could result in mesh that are agressively streched. One way would be to contruct the tolerance using both interior and exterior radius.

A better way would be to have a local toler3 for each spectral element, or one for each direction of the spectral element.

However, I believe the domain height/radius ratios we use are tiny (<<1), I it is not clear if the additional effort is justified.



> Yes, I realize that this is for more pencil-shaped atmospheric columns, which I suppose we should avoid. That said, it still fails at 5x aspect ratio:
> 
> ```julia
> CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 3000
> ...
>     domain_height = FT(3000*5)
> ```

The aspect ratio here would be 15000 right?

charleskawczynski: More thoroughly, passing in `domain_height`, and running

```julia
results = Dict()
for pr in 10.0 .^(0:7)
    CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = 10^4
    for ar in 10.0 .^(-1:0.2:1)
        domain_height = ar*planet_radius(param_set)
        results[pr, ar] = false
        try
            main(domain_height)
            results[pr, ar] = true
        catch
        end
    end
end
```
Yields
```julia
julia> count(values(results))/length(values(results))
0.7272727272727273
```

Maybe we can add this as a test as
```julia
@test all(values(results))
```
?

charleskawczynski: > The aspect ratio here would be 15000 right?
I'm calling `domain_height/planet_radius` the aspect ratio, in which case it would be 5.

charleskawczynski: I'll split off the other part of this PR, since that seems more straightforward.

skandalaCLIMA: The current aspect ratio, used for earth simulations, is 0.0047088. 0.727 seems to be a factor of ~182x higher. Do we have an estimate of what aspect ratios we will be using. I can make toler3 cell specific, if we intend to use more agressive aspect ratios frequently in future.

skandalaCLIMA: > > The aspect ratio here would be 15000 right?
> 
> I'm calling `domain_height/planet_radius` the aspect ratio, in which case it would be 5.

Yes. I agree. I missed that the radius was also changed here.

test/Numerics/DGMethods/test_planet_sizes.jl
```diff
@@ -0,0 +1,55 @@
...
+        resolution,
+    )
+    return nothing
+end
+
+let
+    ClimateMachine.init()
+    for planet_rad in 10.0 .^ (0:7)
+        CLIMAParameters.Planet.planet_radius(::EarthParameterSet) = planet_rad
+        for aspect_ratio in 10.0 .^ (-1:0.2:1)
```
charleskawczynski, line 48: @skandalaCLIMA this effectively establishes the range.

skandalaCLIMA:
The code seems to be working fine, irrespective of planet radius, for an aspect ration <= 0.727. This is way more than what we need (we currently use about 0.0047088). Easiest would be to check if aspect ratio is less than some 0.5, which we will never exceed.

test/Numerics/DGMethods/test_planet_sizes.jl
skandalaCLIMA: The starting point for this test itself (0.1) is well outside our typcial use range (~0.005). I think checking if AS is < 0.5 should suffice.

skandalaCLIMA:
A local (instead of global), cell dependent, tolerance needs to be specified if we need to be more thorough. However, in our case, the GCM simulation domain is very thin compared to the radius (like apple skin on an apple). It might not warrant the extra effort.

charleskawczynski:
Ok, I'll roll back the changes until the tests pass, so that we explicitly know what ranges the code is valid for.

test/Numerics/DGMethods/test_planet_sizes.jl
charleskawczynski: Updated.

charleskawczynski:
Alright @skandalaCLIMA, looks like this is passing. I think that this is a more reasonable range.

charleskawczynski:
bors try

charleskawczynski:
bors try

jkozdon:
@charleskawczynski  can you close this? Seems pretty stale.

PR #1566, opened 2020-09-18
Use dry ref state for dry moisture model

charleskawczynski:
# Description

This PR makes the local thermo state in `src/Atmos/Model/ref_state.jl` a dry thermo state if we have a dry model, so that we don't run into sat adjust issues when we shouldn't need to.

I have


tapios:
Sure, makes sense.

szy21 (approved):

PR #1568, opened 2020-09-18
move to per build init / depot

jakebolewski:
# Description
Move to an to an explicit per build depot (depot per agent config) as a workaround for permissions issues with loading a precompiled shared depot.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,6 +3,33 @@ env:
   CLIMATEMACHINE_SETTINGS_FIX_RNG_SEED: "true"
 
 steps:
+  - label: "init cpu env"
+    key: "init_cpu_env"
+    command:
+      - "echo $JULIA_DEPOT_PATH"
+      - "julia --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
LGTM

.buildkite/pipeline.yml
```diff
@@ -3,6 +3,33 @@ env:
...
+    command:
+      - "echo $JULIA_DEPOT_PATH"
+      - "julia --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
+      - "julia --project -e 'using Pkg; Pkg.precompile()'"
+      - "julia --project -e 'using Pkg; Pkg.status()'"
+    agents:
+      config: gpu
+      queue: central
+      slurm_ntasks: 1
+      slurm_gres: "gpu:1"
```
simonbyrne, line 29: so when i tried this last time, you don't actually need a gpu as long as you're not actually running `MPI.Init()`.

jakebolewski: yeah we can change this, the default config for `gpu` tries to output smi information for the job but we can make make that a silent failure

PR #1578, opened 2020-09-23
Make `-` in filenames `_`

jkozdon:
Total bikeshed!

But I noticed that a small number of files used `-` instead of `_` to separate words.

I have


kpamnany (approved):
I like it!

jkozdon:
bors try

szy21:
I like it too! How about those command line arguments? @kpamnany

kpamnany:
> I like it too! How about those command line arguments? @kpamnany

Are you suggesting we use underscores instead of hyphens on the command line? We _could_, but POSIX has a [recommended form](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), and GNU has a "long options" form, which is what we're using.

szy21:
> Are you suggesting we use underscores instead of hyphens on the command line? We _could_, but POSIX has a [recommended form](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), and GNU has a "long options" form, which is what we're using.

Oh, I mean, for example, change `--with-moisture` to `--with_moisture`. But either way would be good.

kpamnany:
Try `julia --help`, for example. That's just the protocol...

szy21:
> Try `julia --help`, for example. That's just the protocol...

Ah I didn't know. Thanks!

PR #1580, opened 2020-09-23
Update Array type when switching from GPU to CPU

blallen:
# Description

`ClimateMachine.init()` would update the `disable_gpu` setting, but not the `array_type` setting when reinitializing with `disable_gpu = true` after having previously initialized with `disable_gpu = false`. This PR fixes it so you can switch between GPU and CPU backend within a single session of the Julia REPL.

I have


blallen:
Should I add a test for 
```julia
ClimateMachine.init(disable_gpu = true)
ClimateMachine.init(disable_gpu = false)
ClimateMachine.init(disable_gpu = true)
```
somewhere, or is that overkill?

kpamnany (approved):
LGTM. Don't think a test is useful here.

PR #1594, opened 2020-09-30
Allow views into MPIStateArrays with nested vars

jkozdon:
# Description

Mainly allows access to nested vars within MPIStateArrays, for example:

```julia-repl
julia> S = @vars(x::Float64, y::@vars(α::Float64, β::SVector{3, Float64}));

julia> Q = MPIStateArray{Float64, S}(MPI.COMM_WORLD, Array, 2, 5, 3);

julia> β = MPIStateArrays.getstateview(Q, "y.β")
2×3×3 view(::Array{Float64,3}, :, 3:5, 1:3) with eltype Float64:
[:, :, 1] =
 0.0  0.0  0.0
 0.0  0.0  0.0

[:, :, 2] =
 0.0  0.0  0.0
 0.0  0.0  0.0

[:, :, 3] =
 0.0  0.0  0.0
 0.0  0.0  0.0
```

Couple other fixes are also included:

- Check that the number of states within an MPIStateArray matches the vars size
- Allow `varsize` to be called on all `Sarray`s (not just `SVector`)

I have


jkozdon:
bors try

src/Arrays/MPIStateArrays.jl
```diff
@@ -169,6 +174,32 @@ function Base.getproperty(Q::MPIStateArray{FT, V}, sym::Symbol) where {FT, V}
...
+# Examples
+```julia-repl
+julia> S = @vars(x::Float64, y::@vars(α::Float64, β::SVector{3, Float64}))
+julia> Q = MPIStateArray{Float64, S}(MPI.COMM_WORLD, Array, 2, 5, 3)
+julia> β = MPIStateArrays.getstateview(Q, "y.β")
+julia> β = MPIStateArrays.getstateview(Q, :(y.β))
+julia> x = MPIStateArrays.getstateview(Q, :x)
+```
+"""
+
```
charleskawczynski, line 194:
```suggestion
```

src/Arrays/MPIStateArrays.jl
```diff
@@ -169,6 +174,32 @@ function Base.getproperty(Q::MPIStateArray{FT, V}, sym::Symbol) where {FT, V}
...
+function getstateview(
+    Q::MPIStateArray{FT, V},
+    fieldname::Union{String, Symbol, Expr},
+) where {FT, V}
+    varrange = varsindices(V, fieldname)
+    return view(realview(Q), :, varrange[1]:varrange[end], :)
```
charleskawczynski, line 200:
```suggestion
    fieldname::Union{String, Symbol, Expr, Tuple},
) where {FT, V}
    vi = varsindex(V, fieldname)
    return view(realview(Q), :, vi, :)
```

Can we use `varsindex` instead? It looks like it might be more appropriate (it already returns a range) and supports `NTuple` types.

src/Arrays/MPIStateArrays.jl
charleskawczynski: (Documenter doesn't allow a space between the doc string and function)

src/Arrays/MPIStateArrays.jl
jkozdon: no, since we need to handle strings and expressions, whereas `varsindex` only handles symbols:
```julia-repl
julia> S = @vars(x::Float64, y::@vars(α::Float64, β::SVector{3, Float64}));

julia> varsindex(S, :(y.β))
ERROR: MethodError: no method matching unval(::Expr)
Closest candidates are:
  unval(::Val{i}) where i at /home/jekozdon/codes/ClimateMachine.jl/src/Utilities/VariableTemplates/VariableTemplates.jl:50
Stacktrace:
 [1] varsindex(::Type{NamedTuple{(:x, :y),Tuple{Float64,NamedTuple{(:α, :β),Tuple{Float64,SArray{Tuple{3},Float64,1,3}}}}}}, ::Expr) at /home/jekozdon/codes/ClimateMachine.jl/src/Utilities/VariableTemplates/VariableTemplates.jl:60
 [2] top-level scope at REPL[12]:1

julia> varsindex(S, "y.β")
ERROR: MethodError: no method matching unval(::String)
Closest candidates are:
  unval(::Val{i}) where i at /home/jekozdon/codes/ClimateMachine.jl/src/Utilities/VariableTemplates/VariableTemplates.jl:50
Stacktrace:
 [1] varsindex(::Type{NamedTuple{(:x, :y),Tuple{Float64,NamedTuple{(:α, :β),Tuple{Float64,SArray{Tuple{3},Float64,1,3}}}}}}, ::String) at /home/jekozdon/codes/ClimateMachine.jl/src/Utilities/VariableTemplates/VariableTemplates.jl:60
 [2] top-level scope at REPL[13]:1
```

src/Arrays/MPIStateArrays.jl
jkozdon: Thanks for the catch!

jkozdon:
bors try

jkozdon:
I've made is so an `MPIStateArray` which has vars set must have the same number of vars as the number of states it holds.

@christophernhill: Can you review commit c5cbaae2fdff38703bc41fbe4fd8e159c810af11

@mwarusz: Can you review commits 8493d72c8df64f71aba3c65edbcb2822f0db7a34 and 09c57bdcb71f093a3d1778f8b7f435c13b2a3650

blallen:
I don't think we use nested vars on the Ocean side; so, if the CI is passing, this should be good for us?

jkozdon:
bors try

jkozdon:
> @christophernhill: Can you review commit [c5cbaae](https://github.com/CliMA/ClimateMachine.jl/commit/c5cbaae2fdff38703bc41fbe4fd8e159c810af11)

I think that I was able to resolve the issue. I was right that it was a print precision issue.

jkozdon:
bors try

jkozdon:
I think this is ready to go if folks this it would be useful...

blallen (approved):
Let's do it!

PR #1608, opened 2020-10-11
Fix Munk setup type typo, so code now compiles

christophernhill:
# Description

Fixes small type typo that stopped current Ocean barotropic gyre code compiling when Munk
style problem was selected.


charleskawczynski (approved):

christophernhill:
bors try

blallen (approved):
Good catch

PR #1616, opened 2020-10-13
Extend the Multirate-Infinitesimal-Step method

vogel-d:
# Description

Extends the existing implementation of the Multirate-Infinitesimal-Step (MIS) method for different fast solvers (LSRK, SSPRK, MRK, ARK). 
Additionally the computation of the inner time step is changed to enable using the biggest possible inner time step and some further MIS methods (MISRK1, MISRK2a, MISRK2b, MISRK3, MISKWRK43) are added. The rising bubble test by Bryan and Fritsch (2002) is used for testing.

I have


src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -34,14 +34,23 @@ fast and slow dynamics respectively, depending on the state `Q`.
...
+- `nsubsteps` (Tuple): Tuple denoting the total number of times
     to substep the fast process.
-    Default: `50`
+    Default: `(50,)`
 - `discrete_splitting` (Boolean): Boolean denoting whether a PDE level or
     discretized level splitting should be used. If `true` then the PDE is
     discretized in such a way that `f_fast + f_slow` is equivalent to
     discretizing the original PDE directly.
     Default: `false`
+- `hivi_splitting` (Boolean): Boolean denoting whether the fast processes need to
```
thomasgibson, line 45: We don't need `hivi_splitting`; there is already a `ClimateMachine.HEVISplitting` type that is passed as an argument to the MIS solver type constructor

src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -34,14 +34,23 @@ fast and slow dynamics respectively, depending on the state `Q`.
...
+    be split into their horizontal and vertical components. For instance this is
+    necessary, when a MIS method is chosen as the fast solver.
+    Default: `false`
+- `numerical_flux_first_order_fast` (NumericalFluxFirstOrder): The type of
```
thomasgibson, line 49: We should try to avoid passing these numerical flux arguments --- much of this could be inferred from the DG model that this solver type has access to.

src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -148,21 +170,66 @@ function solversetup(
     # slower processes (advection and diffusion)
     if ode_solver.discrete_splitting
         numerical_flux_first_order =
-            (dg.numerical_flux_first_order, (dg.numerical_flux_first_order,))
+            (ode_solver.numerical_flux_first_order_slow, (ode_solver.numerical_flux_first_order_slow,))
```
thomasgibson, line 173: I see. So you want to distinguish what fluxes you want to use for the different parts. OK, we should talk with @simonbyrne about some ways we can do this without having to pass so many arguments

src/Numerics/ODESolvers/AdditiveRungeKuttaMethod.jl
```diff
@@ -82,8 +82,10 @@ mutable struct AdditiveRungeKutta{
     rhs!
     "rhs linear operator"
     rhs_implicit!
-    "backwark Euler solver"
+    "backward Euler solver"
```
thomasgibson, line 85: Thank you for fixing this typo :)

src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -34,14 +34,23 @@ fast and slow dynamics respectively, depending on the state `Q`.
     Default: `MIS2`
 - `fast_method` (Function): Function defining the fast solver.
     Default: `LSRK54CarpenterKennedy`
-- `nsubsteps` (Int): Integer denoting the total number of times
+- `nsubsteps` (Tuple): Tuple denoting the total number of times
```
thomasgibson, line 37: We should document how to specify sub stepping in nested solvers. This is also a problem when using the multirate solver type, so this might be an opportunity to finally pin down something concrete!

thomasgibson (commented):
Some initial comments after a quick first sweep. We can discuss more on the details when we meet up again

src/Driver/SolverTypes/MISSolverType.jl
```diff
@@ -148,21 +170,66 @@ function solversetup(
...
+            dg.numerical_flux_gradient,
+            state_auxiliary = dg.state_auxiliary,
+            state_gradient_flux = dg.state_gradient_flux,
+            states_higher_order = dg.states_higher_order,
+            direction = VerticalDirection(),
+        )
+        fast_dg = (fast_dg_h, fast_dg_v)
+        if length(ode_solver.nsubsteps) == 1
+            nsubsteps = getnsubsteps(
+                Symbol(ode_solver.mis_method),
```
simonbyrne, line 209: Pass the object instead of the symbol:
```suggestion
                ode_solver.mis_method,
```

src/Numerics/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -517,3 +829,85 @@ function TVDMISB(
         t0 = t0,
     )
 end
+
+function getnsubsteps(mis::Symbol, ns::Int, RT::DataType)
+    if mis == :MISRK1
+        β = [
+            0 0
+            1 0
+        ]
```
simonbyrne, line 838:
```suggestion
function beta(::typeof(MISRK1)
        β = [
            0 0
            1 0
        ]
end
# etc...

function getnsubsteps(mis, ns)
  d = sum(beta(mis), dims = 2)
  return d ./ ns
end
```

simonbyrne: Then `beta` can also be used in the main `MISRK1` function.

src/Numerics/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -151,6 +190,48 @@ mutable struct MultirateInfinitesimalStep{
     end
 end
 
+function MultirateInfinitesimalStep(
+    method::Symbol,
```
simonbyrne, line 194: can pass the object instead of the symbol
```suggestion
    mis, 
```

thomasgibson:
bors try

thomasgibson:
bors try

src/Numerics/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -151,6 +198,49 @@ mutable struct MultirateInfinitesimalStep{
     end
 end
 
+function MultirateInfinitesimalStep(
```
thomasgibson, line 201: @vogel-d Is this used? I don't see an instance of this being used anywhere

experiments/AtmosLES/rising_bubble_bryan.jl
```diff
@@ -0,0 +1,253 @@
...
+    state.moisture.ρq_tot = ρ * q_pt.tot
+end
+
+function config_risingbubble(FT, N, resolution, xmax, ymax, zmax)
+
+    # Choose explicit solver
+    solver_type = MultirateInfinitesimalStep
+    fast_solver_type = LowStorageRungeKutta2N
+
+    if solver_type == MultirateInfinitesimalStep
```
thomasgibson, line 90: @vogel-d So I would like to try to simplify this script before we merge this. Can we choose a single configuration for this example and just use that? For example, just use MIS with an ARK fast solver?

src/Numerics/ODESolvers/MultirateInfinitesimalStepMethod.jl
vogel-d: It is used when MIS is called as a fast solver, e.g. in the rising bubble test (l. 112-119):
`fast_method = (dg, Q, nsubsteps) ->
                    MultirateInfinitesimalStep(
                        MISKWRK43,
                        dg,
                        (dgi, Qi) -> LSRK54CarpenterKennedy(dgi, Qi),
                        Q,
                        nsubsteps = nsubsteps,
                    ),`

This could probably alternatively be achieved by
`fast_method = (dg, Q, nsubsteps) ->
                    MISKWRK43(
                        dg.rhs![1],
                        dg.rhs![2],
                        (dgi, Qi) -> LSRK54CarpenterKennedy(dgi, Qi),
                        nsubsteps,
                        Q
                    ),`
and then the outer constructor could be avoided. 

I defined similar outer constructors for MRK and ARK as fast solvers to make it cleaner and to avoid confusion regarding the data type of dg and TimeScaledRHS.

experiments/AtmosLES/rising_bubble_bryan.jl
vogel-d: Of course! I just used this for testing and to have an overview over the different fast solvers, while coding. 

I however noticed that after merging the master, rising_bubble_bryan.jl does not run anymore. 
In line 44, as far as I can tell, instead of 
`function init_risingbubble!(problem, bl, state, aux, (x, y, z), t)`
the coordinates have to be specified as
`function init_risingbubble!(problem, bl, state, aux, localgeo, t)`
`(x, y, z) = localgeo.coord`

Additionaly in line 69
`ts = LiquidIcePotTempSHumEquil(bl.param_set, θ, ρ, q_tot)`
LiquidIcePotTempSHumEquil does not seem to be defined anymore. But I don't know, with what it has to be replaced.

thomasgibson: OK great! I'm fixing this now

thomasgibson:
bors try

src/Numerics/ODESolvers/MultirateInfinitesimalStepMethod.jl
```diff
@@ -186,27 +277,39 @@ function dostep!(Q, mis::MultirateInfinitesimalStep, p, time)
...
+        # instead we just update the slow tendency
+        if abs(d[i]) < 1.e-10
+            Q .+= dt .* offset
+        else
+            fastrhs!.a = time + c̃[i] * dt
+            fastrhs!.b = (c[i] - c̃[i]) / d[i]
+
+            τ = zero(FT)
+            nsubstepsLoc = ceil(Int, nsubsteps * d[i])
+            dτ = d[i] * dt / nsubstepsLoc
```
simonbyrne, line 296: So this looks like the culprit for the test failure: previously we computed it as
```julia
dτ = d[i] * dt / mis.nsubsteps
```
This change probably makes more sense though.

simonbyrne:
bors try

tutorials/Atmos/dry_rayleigh_benard.jl
```diff
@@ -169,7 +169,7 @@ function config_problem(::Type{FT}, N, resolution, xmax, ymax, zmax) where {FT}
         splitting_type = ClimateMachine.SlowFastSplitting(),
         mis_method = MIS2,
         fast_method = LSRK144NiegemannDiehlBusch,
-        nsubsteps = 10,
+        nsubsteps = (40,),
```
simonbyrne, line 172: I've increased the number of steps accordingly.

thomasgibson (approved):

Issue #1510, opened 2020-09-01
Heartbeat with synchronised error handling


simonbyrne:
# Description

We want to be able to dump the full state when an error occurs inside the DG kernels. The challenge is determining when this occurs and flagging this on all ranks. A simple idea is to do something like
```julia
try
   # do stuff
   err = MPI.Allreduce(false, max, comm)
catch e
   err = MPI.Allreduce(true, max, comm)
finally
   if err
      # write state to disk
      error("now quit")
   end
end
```

The problem with this approach is that if `# do stuff` involved communication, we might end up in a state where ranks without errors are stuck waiting on communication from the errored ranks.

Since our errors will mostly occur inside events, one idea is to build this into an event wrapper
```julia
struct HeartbeatEvent <: Event
   ev::Event
   callback
end
function wait(h::HeartbeatEvent)
   try
     wait(h.ev)
     err = MPI.Allreduce(false, max, comm)
   catch e
     err = MPI.Allreduce(true, max, comm)
   finally
     if err
      h.callback()
      error("now quit")
     end
   end
end
```
cc: @kpamnany @vchuravy 

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

simonbyrne:
Alternatively, could we use the MPI error handling machinery?
https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node217.htm#Node217

kpamnany:
Done in #2074.

PR #1630, opened 2020-10-15
Apply abs to temperature in sat adjust

charleskawczynski:
# Description

At the moment, `saturation_adjustment` failures can occur because `find_zero` cannot find the root, or because `find_zero` searches for `T` for `T<0`, which results in `DomainError` (`saturation_vapor_pressure(T) ~ (T / _T_triple)^(Δcp / _R_v)`). Rather than implementing bracketing tooling in RootSolvers.jl, this PR effectively brackets temperature by applying `abs` to temperature in `find_zero` (and wraps the result in `abs`). The upside of this PR is that, IIUC, ThermodynamicState input parameters will now always be reported when convergence fails. Before this PR, ThermodynamicState input parameters are reported so long `DomainError` does not occur. This should help us more gracefully handle, and collect data for, non-converged saturation adjustment cases.

This is one more small step towards fixing #263. This also opens some doors with #1510, in particular, we could store `converged::Bool` in the `PhaseEquil` state, and query whether any states are not converged, rather than doing `try-catch`.

I have

### DIFF
--- a/src/Common/Thermodynamics/relations.jl
+++ b/src/Common/Thermodynamics/relations.jl
@@ -41,6 +41,8 @@ export specific_enthalpy, total_specific_enthalpy
 export moist_static_energy
 export saturated
 
+heavisided(x) = (x > 0) * x
+
 """
     gas_constant_air(param_set, [q::PhasePartition])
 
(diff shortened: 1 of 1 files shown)


tapios:
Reflecting temperatures about zero with `abs` seems risky to be. It would be better to truncate: replace `T` by `T * H(T)` with Heaviside function `H`.

charleskawczynski:
> Reflecting temperatures about zero with `abs` seems risky to me. It would be better to truncate: replace `T` by `T * H(T)` with Heaviside function `H`.

I'm not sure I see any risk since `T` is the only variable in the root finding algorithm. What's the concern? a Heaviside function will zero out gradients and be problematic for gradient-based search algorithms.

tapios:
> > Reflecting temperatures about zero with `abs` seems risky to me. It would be better to truncate: replace `T` by `T * H(T)` with Heaviside function `H`.
> 
> I'm not sure I see any risk since `T` is the only variable in the root finding algorithm. What's the concern? a Heaviside function will zero out gradients and be problematic for gradient-based search algorithms.

`abs` has a discontinuous derivative at 0 too. The problem is that projecting `T<0` to `T>0` is just completely physically wrong. Projecting `T<0` to `0` is more acceptable (and is how we deal with these situations, e.g., in positivity preserving ODE solvers). The gradients are analytical for the Newton method anyway, so truncating does not present a special problem as far as I can see.

charleskawczynski:
I see. Thinking out loud, on second thought this should be fine because we'll likely not converge regardless of `abs`/`heaviside` so, even if the gradients are zero, we're really only trying to avoid `DomainError`. I'll go update.

charleskawczynski:
Updated.

charleskawczynski:
@tapios should we merge this? And if so, can you approve?

tapios (approved):

PR #1631, opened 2020-10-15
Improve thermo names

charleskawczynski:
# Description

I know that verbosity is good, but the names of thermo state constructors have gotten a bit out of hand. Thoughts on these changes? The proposed convention is: `PhaseName_{inputvariables}`. Without the suffix implies that you're calling the default constructor. We could write wrappers for these, too, but the ultimate solution here is to use kwargs, once this proves to be performant. And then we'll only have 3 names: `PhaseDry`, `PhaseEquil`, `PhaseNonEquil` with kwargs specifying the inputs. The other nice side effect of this is that users can easily locally check that the order of the input variables are correct.

Old names:

```
PhaseDry
PhaseDry_given_pT
PhaseDry_given_pθ
PhaseDry_given_ρT
PhaseEquil
PhaseNonEquil
TemperatureSHumEquil
TemperatureSHumNonEquil
TemperatureSHumEquil_given_pressure
LiquidIcePotTempSHumEquil
LiquidIcePotTempSHumNonEquil
LiquidIcePotTempSHumNonEquil_given_pressure
LiquidIcePotTempSHumEquil_given_pressure
```
New names (not necessarily in the same order)
```
PhaseDry
PhaseDry_pT
PhaseDry_pθ
PhaseDry_ρT
PhaseEquil
PhaseEquil_ρTq
PhaseEquil_pTq
PhaseEquil_pθq
PhaseEquil_ρθq
PhaseNonEquil
PhaseNonEquil_ρTq
PhaseNonEquil_ρθq
PhaseNonEquil_pθq
```

I also moved around the constructors to group the 3 fundamental types. This PR also swaps the order of the arguments to be of the form (ρ/p,T/θ,q).

I have

### DIFF
--- a/docs/src/APIs/Common/Thermodynamics.md
+++ b/docs/src/APIs/Common/Thermodynamics.md
@@ -14,18 +14,18 @@ PhasePartition
 PhasePartition_equil
 ThermodynamicState
 PhaseDry
-PhaseDry_given_pT
-PhaseDry_given_pθ
-PhaseDry_given_ρT
+PhaseDry_pT
+PhaseDry_pθ
+PhaseDry_ρT
 PhaseEquil
+PhaseEquil_ρTq
+PhaseEquil_pTq
+PhaseEquil_pθq
+PhaseEquil_ρθq
 PhaseNonEquil
-TemperatureSHumEquil
-TemperatureSHumNonEquil
-TemperatureSHumEquil_given_pressure
-LiquidIcePotTempSHumEquil
-LiquidIcePotTempSHumNonEquil
-LiquidIcePotTempSHumNonEquil_given_pressure
-LiquidIcePotTempSHumEquil_given_pressure
+PhaseNonEquil_ρTq
+PhaseNonEquil_ρθq
+PhaseNonEquil_pθq
 ```
 
 ## Thermodynamic state methods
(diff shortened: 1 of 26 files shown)


trontrytel:
I like it alot. As you know ShumShams were never my favourite! I think it improves the code readability.

trontrytel (approved):

yairchn:
Hi Charlie, 
Great idea. 
The last two new-names repeat:
"
PhaseNonEquil_θρq
PhaseNonEquil_θpq
"
Maybe you meant 
"
PhaseNonEquil_θρq
PhaseEquil_θpq
"
Also would it be possible to replace θ with 
"θ_liq, θl or θˡ" to indicate it is the liquid ice potential temperature?

charleskawczynski:
> The last two new-names repeat:

They don't repeat, one uses density and another pressure. This is one downside, it's a bit difficult to visually discern the two. On the other hand, you easily know what you're passing in when you actually go to use these methods, so it should be clear during usage.

> Also would it be possible to replace θ in these cases with
> "θ_liq, θl or θˡ" to indicate it is the liquid ice potential temperature?

I thought about that, but I think it's a bit redundant wit the phase (e.g., `PhaseDry_pT_dry` and `PhaseDry_pθ_dry`). FWIW, the docs are explicit about the input variables.

jiahe23 (approved):

PR #1632, opened 2020-10-16
Typo in StateCheck output

jkozdon:
one of the `row pass count` should be `row failure counts`


christophernhill (approved):

blallen (approved):
wow surprised I never noticed this

Issue #1633, opened 2020-10-16
Decompose RHS/tendency evaluation


charleskawczynski:
# Description

Decomposing the RHS/tendency evaluation has been discussed so that we can
 - write more granular unit tests
 - monitor budget terms
 - ease time splitting capabilities

## Additional context

Add any other reasons why this should be addressed.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

Issue #1527, opened 2020-09-08
Budget terms in the diagnostics


szy21:
# Description

It would be very helpful to have the budget terms (all source/sink/tendency terms of the equations) in the diagnostics. Separated from #214 as a new issue.

Some of us have discussed about this a few months ago. @akshaysridhar @thomasgibson @simonbyrne @LenkaNovak please add your comments.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

Issue #1036, opened 2020-04-30
Extract fluxes / sources for diagnostics and budget evaluations


akshaysridhar:
As we're moving towards implementation of parameterizations within the model soon, a general debugging and analysis tool at the upper level (`SubModel` <: `BalanceLaw`) is needed to be able to evaluate individual contributions to the source and flux terms. 

# Description
e.g. save output from 

```
flux_turbulence!(...)
flux_moisture!(...)
flux_radiation!(...)
flux_other!(...)
source_{n}(...) for n == 1:number_of_sources
```

Currently it seems we can hack this into the `auxiliary state` through the `nodal update`, is there a more systematic way of handling this ? 

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

akshaysridhar:
See also : #395, #1459

Issue #395, opened 2019-08-24
Solution to composability of source terms


charleskawczynski:
We need a solution to adding different combinations of source terms. How about we use something like this:
```
abstract type Case end
struct Dycoms <: Case end
struct HeldsSuarez <: Case end

abstract type Source end
struct DycomsSrc1 <: Source end
struct DycomsSrc2 <: Source end
struct DycomsSrc3 <: Source end

struct HeldsSuarezSrc1 <: Source end
struct HeldsSuarezSrc2 <: Source end
struct HeldsSuarezSrc3 <: Source end

abstract type SourceSet{case <: Case} end

SourceSet(::Dycoms) = (DycomsSrc1(), DycomsSrc2(), DycomsSrc3())
SourceSet(::HeldsSuarez) = (HeldsSuarezSrc1(), HeldsSuarezSrc2(), HeldsSuarezSrc3())

function source!(source::Vars, state::Vars, aux::Vars, t::Real, ::DycomsSrc1)
  source.ρ+= ...
end
function source!(source::Vars, state::Vars, aux::Vars, t::Real, ::DycomsSrc2)
  source.ρ+= ...
end
...
source!(source::Vars, state::Vars, aux::Vars, t::Real, ::HeldsSuarezSrc1) = ...
source!(source::Vars, state::Vars, aux::Vars, t::Real, ::HeldsSuarezSrc2) = ...
...

# In AtmosModel.jl
function source!(m::AtmosModel, source::Vars, state::Vars, aux::Vars, t::Real)
  for src in m.source
    source!(source, state, aux, t, src)
  end
end

# In driver
model = AtmosModel(FlatOrientation(),
                   ConstantViscosityWithDivergence(DFloat(μ_exact)),
                   MMSDryModel(),
                   NoRadiation(),
                   SourceSet(Dycoms()),
                   InitStateBC(),
                   mms2_init_state!)
```

This way, we can specify custom and standard lists of source terms. Plus, we can reuse `Dycoms()` for boundary conditions, or any other necessary parameters.

charleskawczynski:
One issue with this approach is that there will exist potentially many `source!` functions, and perhaps dispatch will be slow.

vchuravy:
Dynamic dispatch won't be that slow, given that each source term is "costly". Running this on the GPU is trickier since we need to precompute the dispatch there.

charleskawczynski:
I think it's doable, yes.

simonbyrne:
[...]

PR #1634, opened 2020-10-16
Add source type hierarchy

charleskawczynski:
# Description

This is a first step towards fixing #395, #1527, #1036, and #1633.

It seems that all sources could be written this way. I think this has a few advantages:
 - Users can overload `eq_tends` and add / remove terms in a hierarchical fashion, while not affecting the other flux terms
 - `flux` and `source` can be used by diagnostics to export/monitor the budget.
 - This could also be used for specifying time-explicit/implicit source terms more easily via types
 - We can remove the generated functions in `Atmos/Model/source.jl`

This is just a prototype for now. I figured I'd open this PR to get feedback. This seems to work fine for `experiments/AtmosLES/bomex_single_stack.jl` (both on CPU and GPU)

One potential concern is, can we continue to unroll the tendencies if the number of sources gets large. @vchuravy? I can't imagine we'll reach over 100 (ever), but I could imagine reaching/needing granularity of, say 20.

I have

### DIFF
--- a/src/Atmos/Model/tendencies_energy.jl
+++ b/src/Atmos/Model/tendencies_energy.jl
@@ -1,0 +1,33 @@
+##### Energy tendencies
+
+#####
+##### First order fluxes
+#####
+
+function flux(::Advect{Energy}, m, state, aux, t, ts, direction)
+    return (state.ρu / state.ρ) * state.ρe
+end
+
+function flux(::Pressure{Energy}, m, state, aux, t, ts, direction)
+    return state.ρu / state.ρ * air_pressure(ts)
+end
+
+#####
+##### Sources
+#####
+
+function source(
+    s::Subsidence{Energy},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    z = altitude(m, aux)
+    w_sub = subsidence_velocity(s, z)
+    k̂ = vertical_unit_vector(m, aux)
+    return -state.ρ * w_sub * dot(k̂, diffusive.∇h_tot)
+end
--- a/docs/src/APIs/BalanceLaws/BalanceLaws.md
+++ b/docs/src/APIs/BalanceLaws/BalanceLaws.md
@@ -10,6 +10,29 @@ CurrentModule = ClimateMachine.BalanceLaws
 BalanceLaw
 ```
 
+## Tendency types and methods
+
+```@docs
+PrognosticVariable
+AbstractOrder
+AbstractTendencyType
+TendencyDef
+eq_tends
+prognostic_vars
+fluxes
+sources
+show_tendencies
+```
+
+## Methods for fluxes and sources
+
+```@docs
+flux
+source
+Σfluxes
+Σsources
+```
+
 ## State variable types
 
 ```@docs
--- a/experiments/AtmosLES/dycoms.jl
+++ b/experiments/AtmosLES/dycoms.jl
@@ -297,7 +297,7 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
     source = (
         Gravity(),
         rayleigh_sponge,
-        Subsidence{FT}(D_subsidence),
+        Subsidence(D_subsidence)...,
         geostrophic_forcing,
     )
 
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -37,6 +37,7 @@ using ..Mesh.Grids:
     Direction
 
 using ClimateMachine.BalanceLaws
+import ClimateMachine.BalanceLaws: flux, source
 using ClimateMachine.Problems
 
 import ClimateMachine.BalanceLaws:
@@ -368,6 +369,12 @@ gravitational_potential(bl, aux) = gravitational_potential(bl.orientation, aux)
 turbulence_tensors(atmos::AtmosModel, args...) =
     turbulence_tensors(atmos.turbulence, atmos, args...)
 
+include("declare_prognostic_vars.jl") # declare prognostic variables
+include("multiphysics_types.jl")      # types for multi-physics tendencies
+include("tendencies_mass.jl")         # specify mass tendencies
+include("tendencies_momentum.jl")     # specify momentum tendencies
+include("tendencies_energy.jl")       # specify energy tendencies
+include("tendencies_moisture.jl")     # specify moisture tendencies
 
 include("problem.jl")
 include("ref_state.jl")
@@ -380,6 +387,9 @@ include("tracers.jl")
 include("linear.jl")
 include("courant.jl")
 include("filters.jl")
+
+include("atmos_tendencies.jl")        # specify atmos tendencies
+include("get_prognostic_vars.jl")     # get tuple of prognostic variables
 
 """
     flux_first_order!(
@@ -401,25 +411,15 @@ equations.
     t::Real,
     direction,
 )
-    ρ = state.ρ
-    ρinv = 1 / ρ
-    ρu = state.ρu
-    u = ρinv * ρu
-
-    # advective terms
-    flux.ρ = ρ * u
-    flux.ρu = ρ * u .* u'
-    flux.ρe = u * state.ρe
+    ρu_pad = SVector(1, 1, 1)
+    ts = recover_thermo_state(m, state, aux)
+    tend = Flux{FirstOrder}()
+    args = (m, state, aux, t, ts, direction)
+    flux.ρ = Σfluxes(eq_tends(Mass(), m, tend), args...)
+    flux.ρu = Σfluxes(eq_tends(Momentum(), m, tend), args...) .* ρu_pad
+    flux.ρe = Σfluxes(eq_tends(Energy(), m, tend), args...)
 
     # pressure terms
-    ts = recover_thermo_state(m, state, aux)
-    p = air_pressure(ts)
-    if m.ref_state isa HydrostaticState
-        flux.ρu += (p - aux.ref_state.p) * I
-    else
-        flux.ρu += p * I
-    end
-    flux.ρe += u * p
     flux_radiation!(m.radiation, m, flux, state, aux, t)
     flux_moisture!(m.moisture, m, flux, state, aux, t)
     flux_tracers!(m.tracers, m, flux, state, aux, t)
@@ -740,6 +740,18 @@ function source!(
     t::Real,
     direction,
 )
+    ρu_pad = SVector(1, 1, 1)
+    tend = Source()
+    ts = recover_thermo_state(m, state, aux)
+    args = (m, state, aux, t, ts, direction, diffusive)
+    source.ρ = Σsources(eq_tends(Mass(), m, tend), args...)
+    source.ρu = Σsources(eq_tends(Momentum(), m, tend), args...) .* ρu_pad
+    source.ρe = Σsources(eq_tends(Energy(), m, tend), args...)
+    if !(m.moisture isa DryModel)
+        source.moisture.ρq_tot =
+            Σsources(eq_tends(TotalMoisture(), m, tend), args...)
+    end
+
     atmos_source!(m.source, m, source, state, diffusive, aux, t, direction)
 end
 
--- a/src/Atmos/Model/atmos_tendencies.jl
+++ b/src/Atmos/Model/atmos_tendencies.jl
@@ -1,0 +1,37 @@
+#####
+##### Tendency specification
+#####
+
+import ..BalanceLaws: eq_tends
+
+# --------- Some of these methods are generic or
+#           temporary during transition to new specification:
+filter_source(pv::PrognosticVariable, s) = nothing
+# Sources that have been added to new specification:
+filter_source(pv::PV, s::Subsidence{PV}) where {PV <: PrognosticVariable} = s
+
+# Filter sources / empty elements
+filter_sources(t::Tuple) = filter(x -> !(x == nothing), t)
+filter_sources(pv::PrognosticVariable, srcs) =
+    filter_sources(map(s -> filter_source(pv, s), srcs))
+
+# Entry point
+eq_tends(pv::PrognosticVariable, m::AtmosModel, ::Source) =
+    filter_sources(pv, m.source)
+# ---------
+
+# Mass
+eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Mass} =
+    (Advect{PV}(),)
+
+# Momentum
+eq_tends(pv::PV, m::AtmosModel, ::Flux{FirstOrder}) where {PV <: Momentum} =
+    (Advect{PV}(), PressureGradient{PV}())
+
+# Energy
+eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Energy} =
+    (Advect{PV}(), Pressure{PV}())
+
+# TotalMoisture
+eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: TotalMoisture} =
+    ()
--- a/src/Atmos/Model/declare_prognostic_vars.jl
+++ b/src/Atmos/Model/declare_prognostic_vars.jl
@@ -1,0 +1,13 @@
+##### Prognostic variable
+
+export Mass, Momentum, Energy
+export TotalMoisture, LiquidMoisture, IceMoisture
+export Tracers
+
+struct Mass <: PrognosticVariable end
+struct Momentum <: PrognosticVariable end
+struct Energy <: PrognosticVariable end
+struct TotalMoisture <: PrognosticVariable end
+struct LiquidMoisture <: PrognosticVariable end
+struct IceMoisture <: PrognosticVariable end
+struct Tracers <: PrognosticVariable end
--- a/src/Atmos/Model/get_prognostic_vars.jl
+++ b/src/Atmos/Model/get_prognostic_vars.jl
@@ -1,0 +1,10 @@
+##### Get prognostic variable list
+
+import ..BalanceLaws: prognostic_vars
+
+prognostic_vars(::DryModel) = ()
+prognostic_vars(::EquilMoist) = (TotalMoisture(),)
+prognostic_vars(::NonEquilMoist) =
+    (TotalMoisture(), LiquidMoisture(), IceMoisture())
+prognostic_vars(m::AtmosModel) =
+    (Mass(), Momentum(), Energy(), prognostic_vars(m.moisture)...)
--- a/src/Atmos/Model/multiphysics_types.jl
+++ b/src/Atmos/Model/multiphysics_types.jl
@@ -1,0 +1,21 @@
+##### Multi-physics types
+
+export Subsidence
+struct Subsidence{PV, FT} <: TendencyDef{Source, PV}
+    D::FT
+end
+
+# Subsidence includes tendencies in Mass, Energy and TotalMoisture equations:
+Subsidence(D::FT) where {FT} = (
+    Subsidence{Mass, FT}(D),
+    Subsidence{Energy, FT}(D),
+    Subsidence{TotalMoisture, FT}(D),
+)
+
+subsidence_velocity(subsidence::Subsidence{PV, FT}, z::FT) where {PV, FT} =
+    -subsidence.D * z
+
+struct PressureGradient{PV <: Momentum} <: TendencyDef{Flux{FirstOrder}, PV} end
+struct Pressure{PV <: Energy} <: TendencyDef{Flux{FirstOrder}, PV} end
+
+struct Advect{PV} <: TendencyDef{Flux{FirstOrder}, PV} end
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -4,7 +4,6 @@ using CLIMAParameters.Planet: Omega, e_int_i0, cv_l, cv_i, T_0
 export AbstractSource,
     Gravity,
     RayleighSponge,
-    Subsidence,
     GeostrophicForcing,
     Coriolis,
     RemovePrecipitation,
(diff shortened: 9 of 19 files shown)


kpamnany:
This looks like a good way to break things down, but not sure I have useful input.

charleskawczynski:
bors try

thomasgibson:
This is a great start! I will spend some time reading over this carefully --- this is closely related to questions many of us have about how we abstract the various terms in Clima.

thomasgibson (commented):
So this PR basically captures two main things: (1) an abstraction for tendencies (which I think we should explicitly define in the documentation, e.g. tendency vs term); and (2) sums of tendencies (not as an explicitly types object, but rather as a wrapper).

I think helps simplify how the AtmosModel (and similar subtypes of BalanceLaw) specifies the specification of fluxes. I am in favor of proceeding in this fashion towards something which exposes the tendencies in a more clear way.

My only concern is how this can be leveraged by the rest of clima. Right now, these changes are mostly cosmetic. But does this typing help us factorize out specific terms in a given balance law? I would be very interested to explore this more with you @charleskawczynski --- as I think this helps us get closer to what we really want at the higher-level interfaces

src/Atmos/Model/tendencies.jl
```diff
@@ -0,0 +1,15 @@
+#####
+##### Tendency specification
+#####
```
thomasgibson, line 3: We should be very clear what we define as a tendency --- I think there is a lot of confusion between the various labels we commonly use (tendency, term, etc)

charleskawczynski: I agree that we should have a clear and documented terminology. I've been using "tendency" to mean any RHS term, be it a flux or non-conservative source, in a spatially/temporally-continuous sense.

blallen:
@charleskawczynski do you think you can mock up an example where you use this structure to actually pick out one of the terms? What you have looks good, but I feel like the real test is if adding all this structure actually lets you do the selection you want

charleskawczynski:
> So this PR basically captures two main things: (1) an abstraction for tendencies (which I think we should explicitly define in the documentation, e.g. tendency vs term); and (2) sums of tendencies (not as an explicitly types object, but rather as a wrapper).

Agreed

> I think helps simplify how the AtmosModel (and similar subtypes of BalanceLaw) specifies the specification of fluxes. I am in favor of proceeding in this fashion towards something which exposes the tendencies in a more clear way.

🎉 

> My only concern is how this can be leveraged by the rest of clima.

How so? This same pattern can be applied to any balance law, AFAICT.

> But does this typing help us factorize out specific terms in a given balance law? I would be very interested to explore this more with you @charleskawczynski --- as I think this helps us get closer to what we really want at the higher-level interfaces!

Perhaps I'm unclear about what is needed for the multirate bits. I imagine that this level of granularity would make specifying balance laws easier via types/tuples.

charleskawczynski:
> @charleskawczynski do you think you can mock up an example where you use this structure to actually pick out one of the terms? What you have looks good, but I feel like the real test is if adding all this structure actually lets you do the selection you want

I can try to mock something up, is there an example that I can look at? Also, I have some ideas that might help trim some code.

charleskawczynski:
bors try

charleskawczynski:
bors try

blallen:
> > @charleskawczynski do you think you can mock up an example where you use this structure to actually pick out one of the terms? What you have looks good, but I feel like the real test is if adding all this structure actually lets you do the selection you want
> 
> I can try to mock something up, is there an example that I can look at? Also, I've not trimmed out everything that can be removed to help make the diff more readable.

I mean actually finding something like 'mass tendency' from a constructed `AtmosModel` struct. Just have a short example where you initialize an `AtmosModel` and then check the mass tendency somehow

charleskawczynski:
> I mean actually finding something like 'mass tendency' from a constructed `AtmosModel` struct. Just have a short example where you initialize an `AtmosModel` and then check the mass tendency somehow

Meaning just print the tendencies for a given atmos model? I was thinking we could add and call something like
```julia
function show_atmos_tendencies(m)
  T = (Flux1ˢᵗOrder(), Flux2ⁿᵈOrder(), NonConservative())
  @show ρ_tend.(Ref(m), T)
  @show ρu_tend.(Ref(m), T)
  @show ρe_tend.(Ref(m), T)
  @show ρq_tot_tend.(Ref(m), T)
end
```
in the driver.

charleskawczynski:
bors try

blallen:
> > I mean actually finding something like 'mass tendency' from a constructed `AtmosModel` struct. Just have a short example where you initialize an `AtmosModel` and then check the mass tendency somehow
> 
> Meaning just print the tendencies for a given atmos model? I was thinking we could add and call something like 
> 
> ```julia
> function show_atmos_tendencies(m)
>   T = (Flux1ˢᵗOrder(), Flux2ⁿᵈOrder(), NonConservative())
>   @show ρ_tend.(Ref(m), T)
>   @show ρu_tend.(Ref(m), T)
>   @show ρe_tend.(Ref(m), T)
>   m.moisture isa EquilMoist || m.moisture isa NonEquilMoist && @show ρq_tot_tend.(Ref(m), T)
> end
> ```
> 
> in the driver.

Yep, something like this exactly. Would this show them split into 1st, 2nd, and source terms? Also is it possible to pull out just `ρeAdvect` or `ρePressure`?

charleskawczynski:
> Yep, something like this exactly.

Will do.

> Would this show them split into 1st, 2nd, and source terms?

Yep, each of the 3 lines should print a tuple of tuples-- the outer being one of `Flux1ˢᵗOrder(), Flux2ⁿᵈOrder(), NonConservative()`, and the inner being the particular tendency.

> Also is it possible to pull out just `ρeAdvect` or `ρePressure`?

What exactly do you mean by pull out? I think we should be able to remove types from the tuple (using `filter / isa` and potentially abstract types), and identify if a particular type exists (just with `isa`).

blallen:
> > Yep, something like this exactly.
> 
> Will do.
> 
> > Would this show them split into 1st, 2nd, and source terms?
> 
> Yep, each of the 3 lines should print a tuple of tuples-- the outer being one of `Flux1ˢᵗOrder(), Flux2ⁿᵈOrder(), NonConservative()`, and the inner being the particular tendency.
> 
> > Also is it possible to pull out just `ρeAdvect` or `ρePressure`?
> 
> What exactly do you mean by pull out? I think we should be able to remove types from the tuple (using `filter / isa` and potentially abstract types), and identify if a particular type exists (just with `isa`).

Ah, will the `Flux1ˢᵗOrder()` tuple contain `ρeAdvect` and `ρePressure`? And for the latter question, I mean something like `energy_advect = return_individual_tendency(model, ρeAdvect)` and it returns how this term is specified (or the value at a nodal point, or an average or something if you want to be fancy for diagnostics).

charleskawczynski:
> Ah, will the `Flux1ˢᵗOrder()` tuple contain `ρeAdvect` and `ρePressure`? And for the latter question, I mean something like `energy_advect = return_individual_tendency(model, ρeAdvect)` and it returns how this term is specified (or the value at a nodal point, or an average or something if you want to be fancy for diagnostics).

Yes, it will contain a tuple of both. Oh, the latter point would be cool! I think we could do something like that if, for each source term, we include a doc string of the precise term (perhaps in LaTeX/Unicode). Here's what it currently shows (for `tutorials/Atmos/heldsuarez.jl` and not all source terms added):
```julia
┌ Info: Tendency list:
│     ρ (Flux1ˢᵗOrder) = (ρAdvect())
│     ρ (NonConservative) = ()
│     ρu (Flux1ˢᵗOrder) = (ρuAdvect(), ρuHydroPress())
│     ρu (NonConservative) = ()
│     ρe (Flux1ˢᵗOrder) = (ρeAdvect(), ρePressure())
└     ρe (NonConservative) = ()
```

charleskawczynski:
bors try

kmdeck:
> Yep, that list looks great. And in addition to the string documentation, I think it would be nice if you could use these types to return the nodal values for diagnostics, but that's probably beyond the scope of this PR.

hey Charlie, if I am understanding the general point, this lets us treat the RHS in a more granular way. it looks like this code makes it so all RHS terms are treated like we treat sources now. so we would need to define e.g. a type `SecondOrderHeatFlux` for the sole RHS term in the simple heat equation, and then we have a method of a function `_flux_` which evaluates this, and eventually this gets added to the sum of tendencies for second order fluxes, which then gets added to the total RHS sum. 

If down the line different RHS terms get treated differently (for ODE advancement, eg.), it makes sense that they each have a type that other functions can dispatch on.

More granularity is helpful for debugging too - I'm not sure if we can do this currently with sources, or if this is what Brandon is getting at with his comment above, but it would be helpful to be able to call these functions in a callback, i.e. to get a particular RHS term evaluated every X time steps. 

my personal preference would also be to call them e.g. FluxFirstOrder and not Flux1^{st}Order :)

kmdeck (approved):

blallen:
> my personal preference would also be to call them e.g. FluxFirstOrder and not Flux1^{st}Order :)

yes please

charleskawczynski:
> hey Charlie, if I am understanding the general point, this lets us treat the RHS in a more granular way. it looks like this code makes it so all RHS terms are treated like we treat sources now. so we would need to define e.g. a type `SecondOrderHeatFlux` for the sole RHS term in the simple heat equation, and then we have a method of a function `_flux_` which evaluates this, and eventually this gets added to the sum of tendencies for second order fluxes, which then gets added to the total RHS sum.

Yes, exactly. That said, we don't necessarily need to use this pattern everywhere (e.g., tutorials), since it's compatible with the existing interface. But I imagine that this will be useful in the Atmos/Ocean/Land/Ice models, since we'll likely have a long tendency list per model.

> If down the line different RHS terms get treated differently (for ODE advancement, eg.), it makes sense that they each have a type that other functions can dispatch on.

Yes, this should provide an alternative way to hook specific terms based on the tendency types.

> More granularity is helpful for debugging too - I'm not sure if we can do this currently with sources, or if this is what Brandon is getting at with his comment above, but it would be helpful to be able to call these functions in a callback, i.e. to get a particular RHS term evaluated every X time steps.

yes, I believe that's the idea behind optionally exporting a budget, which should be possible in an automated way using this tendency specification.

> my personal preference would also be to call them e.g. FluxFirstOrder and not Flux1^{st}Order :)

Agreed, I was unsure what to use for some names.

charleskawczynski:
bors try

src/Atmos/Model/tendencies.jl
valeriabarra: This would be super helpful for me, too - as I have been a bit puzzled by this terminology to be honest, but I am not familiar with other climate codes, so I thought that it was due to my unfamiliarity

charleskawczynski:
I spoke a bit with @kpamnany about adding a callback somewhere to print the tendency names/values, but I'm actually not sure which test/driver to use. I'll drop what I have here for now:

```julia
function tend_tuple(var_tend, m, args_f1, args_f2, args_s)
    return (
    ((Atmos.format_tend(tend),Atmos._flux_(tend, args_f1...)) for tend in var_tend(m, Flux1ˢᵗOrder()))...,
    ((Atmos.format_tend(tend),Atmos._flux_(tend, args_f2...)) for tend in var_tend(m, Flux2ⁿᵈOrder()))...,
    ((Atmos.format_tend(tend),Atmos._source_(tend, args_s...)) for tend in var_tend(m, NonConservative()))...,
    )
end

tend_cb = GenericCallbacks.EveryXSimulationTime(every_x_simulation_time) do
    dg = solver_config.dg
    m = dg.balance_law
    direction = dg.direction
    vs_s = vars_state(m, Prognostic(), eltype(dg.Q))
    vs_a = vars_state(m, Auxiliary(), eltype(dg.Q))
    vs_d = vars_state(m, GradientFlux(), eltype(dg.Q))
    vs_hd = vars_state(m, Hyperdiffusive(), eltype(dg.Q))
    loc = (1,:,1)                                                                                        # print tendencies at a specific point in space
    state = Vars{vs_s}(dg.Q[loc...])
    aux = Vars{vs_a}(dg.state_auxiliary[loc...])
    diffusive = Vars{vs_d}(dg.state_gradient_flux[loc...)
    hyperdiff = Vars{vs_hd}(dg.states_higher_order[2][loc...])

    args_f1 = (m, state, aux, t, ts)
    args_f2 = (m, state, aux, t, ts, diffusive, hyperdiff)
    args_s = (m, state, aux, t, ts, direction)

    ρ_tends = tend_tuple(Atmos.ρ_tend, m, args_f1, args_f2, args_s)
    ρu_tends = tend_tuple(Atmos.ρu_tend, m, args_f1, args_f2, args_s)
    ρe_tends = tend_tuple(Atmos.ρe_tend, m, args_f1, args_f2, args_s)
    @show ρ_tends
    @show ρu_tends
    @show ρe_tends

    if m.moisture isa EquilMoist || m.moisture isa NonEquilMoist
        ρq_tot_tends = tend_tuple(Atmos.ρq_tot_tend, m, args_f1, args_f2, args_s)
        @show ρq_tot_tends
    end

    nothing
end;
```

kpamnany:
> I spoke a bit with @kpamnany about adding a callback somewhere to print the tendency names/values

This will go into a diagnostics group so the callback is really an example for developing that rather than an example for use by an experiment.

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/Model/tendencies.jl
charleskawczynski: IIUC, tendencies of `Y_j` are individual terms on the RHS of the time-stepped governing equation for `Y_j`, which have units `units(Y_j)/time`, i.e., `∂_t Y_j = Σ_i tendency_i`. Maybe others can chime in? I believe it's a temporally-continuous concept, so there shouldn't be ambiguities regarding the implicitness of the way each tendency is treated.

charleskawczynski: If people prefer "term", I think that would be fine, but I think "tendency" is actually more specific. It not only restricts the units of the quantity, but also the context (a single term in a time-stepped equation). I was a bit skeptical of this terminology at first (we usually use RHS terms in Mech Eng), but I've come around to like it.

charleskawczynski: Perhaps there's ambiguity with the time-dependent term and sign, however. I realize the definition I gave is vague to disambiguate `∂_t Y_j = Σ_i tendency_i` from `Σ_i tendency_{i,j} = 0`. So, we'd need a sign convention with this definition.

charleskawczynski: I'll move the `tendencies_sum.jl` to `BalanceLaws/` and add some documentation there.

blallen: How about both? `tendency term`? `tendency` can refer to both a `tendency_i` and `∂_t Y_j`, which might be ambiguous, and `term` can refer to any type of term in an equation. combining the two seems to be specific enough

charleskawczynski:
Alright, I
 - Renamed `Flux1ˢᵗOrder` to `FluxFirstOrder` and `Flux2ⁿᵈOrder` to `FluxSecondOrder`
 - Renamed the existing `Atmos.Source` to `Atmos.AbstractSource` (which will eventually be phased out) and `BalanceLaws.NonConservative` to `BalanceLaws.Source`
 - Moved `tendencies_sum.jl` to `BalanceLaws/` and realized that, for the interface to work with the Atmos/Ocean/Land, we'll likely want the tendencies sum to be pretty generic, so I just changed the sum to slurp/splat all args after the tendency type.
 - After moving/generalizing the interface, I realized that `Σfluxes` and `Σsources` had exactly the same signature, and just combined it to `Σtendencies`, which simply calls `tendency`.

I'm pleased with how things have evolved. I think the only thing left is to do a final pass on documentation. @thomasgibson @valeriabarra @blallen do you have thoughts on improvements?

charleskawczynski:
bors try

src/Atmos/Model/tendencies.jl
valeriabarra: Thanks @charleskawczynski for the explanation. So from what I gather, these are the spatial terms, as in everything else that equals the time derivative, as you wrote above `∂_t Y_j = Σ_i tendency_i` . But you also specified RHS terms, meaning when you use an explicit time stepper. What about when you use an implicit one? Are the spatial terms still called tendencies?

charleskawczynski: I apologize, I've abused the phrase "RHS term" a bit. My understanding is that tendencies are temporally-continuous. So, whether a term is treated explicitly / implicitly is orthogonal. I.e., we could label any tendency to be treated explicitly / implicitly in time.

valeriabarra: Okay, so basically I gather that is the totality of all spatial terms that equal the time variation in a conservation form, does not matter if/how discretized

simonbyrne:
How about something like this

```julia
abstract type PrognosticVariable end
struct Mass <: PrognosticVariable end
struct Momentum <: PrognosticVariable end
struct Energy <: PrognosticVariable end

# something you can  "add" to the RHS of an equation set that affects one or more variables
abstract type Tendency end
struct Advection <: Tendency end
struct PressureGradient <: Tendency end
struct Subsidence <: Tendency end
struct Total{TS} <: Tendency
    tends::TS
end

# something that we need to compute
abstract type Term end
struct FluxFirstOrder{V<:PrognosticVariable,T<:Tendency,R<:Rate} <: Term
    var::V
    tend::T
end
struct FluxSecondOrder{V<:PrognosticVariable,T<:Tendency} <: Term end
struct Source{V<:PrognosticVariable,T<:Tendency} <: Term end

# e.g.
compute(::FluxFirstOrder{Advection,Mass}, state) = state.ρu
compute(::FluxFirstOrder{Advection,Momentum}, state) = state.ρu ⊗ (state.ρu ./ state.ρ)
compute(::FluxFirstOrder{Advection,Energy}, state) = (state.ρu ./ state.ρ) .* state.ρe


function compute(::Source{Subsidence,Mass}, state)
    z = altitude(m, aux)
    w_sub = subsidence_velocity(s.subsidence, z)
    k̂ = vertical_unit_vector(m, aux)
    return -state.ρ * w_sub * dot(k̂, diffusive.moisture.∇q_tot)
end
function compute(::Source{Subsidence,Energy}, state)
    z = altitude(m, aux)
    w_sub = subsidence_velocity(s.subsidence, z)
    k̂ = vertical_unit_vector(m, aux)
    return -state.ρ * w_sub * dot(k̂, diffusive.∇h_tot)
end

# could define extra terms to force common subexpression elimination?
struct SubsidenceVelocity <: Term end
compute(::SubsidenceVelocity, state) = ...

# sum up all tendencies for a given variable
compute(f1::FluxFirstOrder{V,Total{TS}}, args...) where {V,TS} =
    sum(tend -> compute(FluxFirstOrder(f1.tend, term),args...), f1.tend.tends)
FluxFirsrOrder(var::Momentum) = FluxFirstOrder(var,Total((Advection,PressureGradient)))
```

For multirate schemes, we could either add an extra parameter to each `FluxFirstOrder`, `FluxSecondOrder`, etc, or we could wrap each `Tendency`
```julia
struct Rate{R, T<: Tendency} <: Tendency
    rate::R
    tend::T
end

compute(::FluxFirstOrder{Rate{Linear, Advection},Mass}, state) = state.ρu
```

simonbyrne:
Actually it occurs to me that in the above, `PrognosticVariable` could just be a `Term` as well.

src/Atmos/Model/tendencies_multiphysics.jl
```diff
@@ -0,0 +1,20 @@
...
+Subsidence(D::FT) where {FT} = (
+    Subsidence{Mass, FT}(D),
+    Subsidence{Energy, FT}(D),
+    Subsidence{TotalMoisture, FT}(D),
+)
+
+subsidence_velocity(subsidence::Subsidence{PV, FT}, z::FT) where {PV, FT} =
+    -subsidence.D * z
+
+struct Pressure{PV} <: FluxFirstOrderTendency{PV} end
```
simonbyrne, line 18: `Pressure` is a quantity so I don't think we should call the tendency that as well. How about:
```suggestion
struct PressureGradient{PV} <: FluxFirstOrderTendency{PV} end
```
since it is the tendency contribution of the gradient of the pressure.

src/Atmos/Model/tendencies_energy.jl
```diff
@@ -0,0 +1,33 @@
+##### Energy tendencies
+
+#####
+##### First order fluxes
+#####
+
+function tendency(::Advect{Energy}, m, state, aux, t, ts, direction)
```
simonbyrne, line 7: I don't think we should use `tendency` here, since it isn't applied directly as a tendency (its divergence is). How about:
```suggestion
function flux(::Advect{Energy}, m, state, aux, t, ts, direction)
```

charleskawczynski:
Ok, I've renamed a bunch of things and added some docs to hopefully clarify precisely what each newly added piece is/does. Most of these docs were added to `src/BalanceLaws/tendency_types.jl`. To avoid terminology confusion `tendency` was split back into `flux` and `source`. The other recent major change was introducing an `AbstractPrognosticVariable`, which allows us to basically dispatch on the rows (equations) and columns (flux^1, flux^2, source) for each of the methods.

I think I'm happy with the way things are looking now. I'm going to look around and try to add a few more flux tendencies and source tendencies (TM 😄) that may cover important edge cases that I don't want to overlook. Then, I'll be happy to merge this thing, woohoo! 🎉

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/Model/tendencies_multiphysics.jl
charleskawczynski: @simonbyrne I'm realizing that `PressureGradient` may only be appropriate for the momentum equation, whereas the pressure term for the energy equation has no gradients (it's a scalar equation). Should we just leave this as `Pressure`? @ilopezgp maybe you can weigh in? Alternatively, we could leave `PressureGradient{PV<:Momentum}` and just add `Pressure{Energy}`. I have no strong opinions on this, so long the names are appropriate.

charleskawczynski: For now, I've added `Pressure` and `PressureGradient` separately, but we can combine them if it's preferred--it's a simple switch.

charleskawczynski:
Alright, I made some changes that I think simplify some things that were previously bugging me. I think we're ready to go!

PR #1635, opened 2020-10-16
Simplify the PR and issue templates

kpamnany:
### Description

The present issue and PR templates are unnecessarily complex. For PRs in particular, the checkboxes are often ignored. This is an attempt to make the templates reasonable so that we can _require_ that PR checkboxes be filled appropriately by the author.


simonbyrne (approved):

leios:
If they are not useful to this project, then we can remove them.

If we are removing them, I think a new checklist for reviewing criteria in CLIMA is necessary. Also, no reason to have all the issue templates. Just leave the general one.

jkozdon:
We could turn the checklist into blocking bors labels, then something has to clicked before they PR is merged.

jkozdon:
This would also allow others to add the labels back if they feel they are needed still

leios:
Yup. I honestly think the big problem is the interaction between the templates and bors because bors puts the full description into the git log. If there is a borsy solution, then we can just remove them entirely without too much lost.

Again, I find these really useful for other projects, but we don't get many fly-by PRs, so it's not really effective here.

kpamnany:
> We could turn the checklist into blocking bors labels, then something has to clicked before they PR is merged.

This seemed like a great idea to me, but then I saw that automatic labels are [not yet supported for PR templates](https://github.com/isaacs/github/issues/1252). :-(

kpamnany:
> If they are not useful to this project, then we can remove them.

They'll be more useful when we have many of those fly-by PRs.

> If we are removing them, I think a new checklist for reviewing criteria in CLIMA is necessary. EDIT: Just realized a checklist was left for CLIMA reviewers only on the PR template. I think that's fair.

I thought of that checklist as being for the author -- the reviewer should verify each of those points before approving.

> Also, no reason to have all the issue templates. Just leave the general one. In fact, you can just remove the issue templates entirely if no one is finding them useful.

Reduced to a single one.

leios (approved):
Looks good!

PR #1642, opened 2020-10-20
Limit exported thermodynamic methods

### LINKED ISSUE #1612: Limit exporting thermo-internal methods
# Description

Some functions in Thermodynamics are exported, e.g., `air_temperature_from_ideal_gas_law`, and they probably shouldn't be. Users should call `air_temperature(::ThermodynamicState)`, so we should probably limit what is exported. Perhaps we could even change internal method names to avoid exporting overloaded methods that are really internal methods.

## Additional context

# For CLIMA Developers

charleskawczynski:
### Description

This PR
 - Removes several methods from being exported in Thermodynamics.jl (highlighted in red in `Thermodynamics.md`).
 - Shortens some thermodynamics-internal method names, and argument orders, to follow the style used in #1631.
 - Fixes a `p` vs `ρ` bug 👊🏻! Thanks @trontrytel!

Closes #1612.

### DIFF
--- a/src/Common/Thermodynamics/relations.jl
+++ b/src/Common/Thermodynamics/relations.jl
@@ -1,43 +1,54 @@
 # Atmospheric equation of state
-export air_pressure,
-    air_temperature, air_density, specific_volume, soundspeed_air
-export total_specific_humidity,
-    liquid_specific_humidity, ice_specific_humidity, vapor_specific_humidity
+export air_pressure
+export air_temperature
+export air_density
+export specific_volume
+export soundspeed_air
+export total_specific_humidity
+export liquid_specific_humidity
+export ice_specific_humidity
+export vapor_specific_humidity
 
 # Energies
-export total_energy, internal_energy, internal_energy_sat
+export total_energy
+export internal_energy
+export internal_energy_sat
 
 # Specific heats and gas constants of moist air
 export cp_m, cv_m, gas_constant_air, gas_constants
 
 # Latent heats
-export latent_heat_vapor,
-    latent_heat_sublim, latent_heat_fusion, latent_heat_liq_ice
+export latent_heat_vapor
+export latent_heat_sublim
+export latent_heat_fusion
+export latent_heat_liq_ice
 
 # Saturation vapor pressures and specific humidities over liquid and ice
 export Liquid, Ice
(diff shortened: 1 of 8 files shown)


src/Common/Thermodynamics/relations.jl
```diff
@@ -1532,7 +1543,7 @@ function saturation_adjustment_q_tot_θ_liq_ice_given_pressure(
     if unsaturated && T_1 > _T_min
         return T_1
     else
-        T_2 = air_temperature_from_liquid_ice_pottemp(
+        T_2 = air_temperature_given_θρq(
```
trontrytel, line 1546: My eyes are bad. But maybe I see rho in function name but pressure in argument list?

trontrytel: If only the compiler could check units of our arguments. This would be a compiler error!

trontrytel (changes requested):
Left one comment. Looks good otherwise!

src/Common/Thermodynamics/relations.jl
charleskawczynski: Good catch! Looks like a bug to me! Look at that, including `ρ` in the name came in handy!

charleskawczynski: Wow, not sure how that's been passing CI 🤔. 😮 This branch is potentially hit in BOMEX/Dycoms initialization!

charleskawczynski: Updated. 😬 Alright, merge merge merge! 😆

trontrytel (approved):

src/Common/Thermodynamics/relations.jl
charleskawczynski: Maybe this is responsible for the strange spike in `q_tot` in the BOMEX init. Could be impacting EDMF, too! @yairchn

charleskawczynski: Ah, actually, I think we want to leave pressure but call `air_temperature_given_θpq`. Fixing..

charleskawczynski: Ah, good news, this was only used for computing a temperature bound, so not terrible.

PR #1645, opened 2020-10-21
Better spatial/temporal discretization separation for SplitExplicitModel

blallen:
### Description

Moves all timestepping related parameters into the SplitExplicitLSRK2nMethod and has the IVDCModel store parameters extracted from its parent model. The first change is for easier composability of models and timesteppers, while the second change is to enable the IVDCModel to be used with the explicit HBModel as well (will actually implement in a future PR). 

No numerics or physics should change as a result of this PR.


blallen:
bors try

PR #1646, opened 2020-10-21
GPUify surface fluxes

charleskawczynski:
### Description

This PR GPUifies SurfaceFluxes with several steps:
 - Isolates the closure `f!` that is passed into `nlsolve`
 - Concertizes the properties of `SurfaceFluxConditions`, so that instances of `SurfaceFluxConditions` are `isbits`
 - Changes `for` loops to `ntuple`, which will unroll on the GPU and return a similar type to `x_star`
 - Converts arrays to `Tuple`s so that we can effectively `getindex` inside the nonlinear function
 - Use `NonlinearSolvers.jl` to solve the nonlinear system.


charleskawczynski:
@akshaysridhar I think this is ready to go.

akshaysridhar:
@charleskawczynski As a standalone GPU-ify PR this seems good, lgtm! Thanks for following this through with NLSolve. The next immedidate step is to include these in the integrated tests / experiments with the LES configuration (e.g. stable / unstable BL checks) + iterations on the user-facing component.

akshaysridhar (approved):

PR #1647, opened 2020-10-21
Make `recover_thermo_state` call `new_thermo_state`

charleskawczynski:
### Description

See #1648.

This PR should make thermodynamics consistent on the boundaries.

### DIFF
--- a/docs/src/HowToGuides/Atmos/MoistureModelChoices.md
+++ b/docs/src/HowToGuides/Atmos/MoistureModelChoices.md
@@ -48,6 +48,12 @@ Because the assumed relaxation timescale for condensation/evaporation and
 
 ## Implementation notes
 
+!!! warn
+    While `recover_thermo_state` is an ideal long-term solution,
+    right now we are directly calling new_thermo_state to avoid
+    inconsistent aux states in kernels where the aux states are
+    out of sync with the boundary state.
+
 The moisture models rely on the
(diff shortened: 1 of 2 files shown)


trontrytel:
This is very sad. But you gotta do what you gotta do. 

Could you update the docs as well? https://github.com/CliMA/ClimateMachine.jl/blob/master/docs/src/HowToGuides/Atmos/MoistureModelChoices.md#implementation-notes

trontrytel:
Also, would it be better to get rid of the `recover_thermo_state` function? It is the same as `new_thermo_state`. This will be confusing for people.

charleskawczynski:
> Could you update the docs as well?... Also, would it be better to get rid of the `recover_thermo_state` function? It is the same as `new_thermo_state`. This will be confusing for people.

I'm hesitant to make too many changes. I don't know to what degree this will impact performance.

trontrytel:
> > Could you update the docs as well?... Also, would it be better to get rid of the `recover_thermo_state` function? It is the same as `new_thermo_state`. This will be confusing for people.
> 
> I'm hesitant to make too many changes. I don't know to what degree this will impact performance.

How do you want to test that then? 
Keeping the docs as they are and keeping two functions that do the same thing is dangerous.

charleskawczynski:
Updated.

trontrytel (approved):
Thanks!

PR #1559, opened 2020-09-16
1D and 2D energy spectra for GCM

LenkaNovak:
# Description

Co-authored by @jiahe23 , @kpamnany 

This code calculates 1d spectra (Fourier transform at each latitude and height) and 2d spectra (spherical harmonic transformation at each height) for the GCM on the fly, or it can be ported to analysis scripts for offline calculations. 

It is currently set up to be interpolated using the same output timestep and Gaussian grid as the `atmos_default_diagnostics` group, but these can be set separately (e.g. for less severe wavenumber truncations). 

Energy spectra are currently calculated using the meridional wind only.

Output is a separate NetCDF file, containing:
"spectrum_1d"[zonal wavenumber, latitude, height, time]
"spectrum_2d"[truncated zonal wavenumber, total wavenumber, height, time]
 
TODO in a separate PR
- add mass weighting and vertical integration

_The code structure follows Yassine’s 3d turbulence LES spectra, and the spherical harmonic transformation is based on Daniel’s Julian FMS code._

I have

- [x] Written and run all necessary tests with CLIMA by including `tests/runtests.jl`
- [x] Followed all necessary [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and run `julia .dev/climaformat.jl .`
- [x] Updated the documentation to reflect changes from this PR.

# For review by CLIMA developers

- [ ] There are no open pull requests for this already
- [ ] CLIMA developers with relevant expertise have been assigned to review this submission
- [ ] The code conforms to the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) and has consistent naming conventions. `julia .dev/format.jl` has been run in a separate commit.
- [ ] This code does what it is technically intended to do (all numerics make sense physically and/or computationally)

PR #1654, opened 2020-10-22
Add DGModel outer constructor

charleskawczynski:
### Description

This PR adds a `DGModel` outer constructor that accepts a `DriverConfiguration`.

### DIFF
--- a/src/Driver/driver_configs.jl
+++ b/src/Driver/driver_configs.jl
@@ -583,3 +583,21 @@ Establishing single stack configuration for %s
         SingleStackSpecificInfo(),
     )
 end
+
+import ..DGMethods: DGModel
+
+"""
+    DGModel(driver_config; kwargs...)
+
+Initialize a [`DGModel`](@ref) given a
+[`DriverConfiguration`](@ref) and keyword
+arguments supported by [`DGModel`](@ref).
+"""
+DGModel(driver_config; kwargs...) = DGModel(
+    driver_config.bl,
+    driver_config.grid,
+    driver_config.numerical_flux_first_order,
+    driver_config.numerical_flux_second_order,
+    driver_config.numerical_flux_gradient;
+    kwargs...,
+)
--- a/src/Driver/solver_configs.jl
+++ b/src/Driver/solver_configs.jl
@@ -101,9 +101,6 @@ function SolverConfiguration(
 
     bl = driver_config.bl
     grid = driver_config.grid
-    numerical_flux_first_order = driver_config.numerical_flux_first_order
-    numerical_flux_second_order = driver_config.numerical_flux_second_order
-    numerical_flux_gradient = driver_config.numerical_flux_gradient
 
     # Create the DG model and initialize the ODE state. If we're restarting,
     # use state data from the checkpoint.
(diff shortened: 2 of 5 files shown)


kpamnany:
I see the convenience, but this messes up what I see as the implicit module hierarchy by making `DGModel` dependent on the definition of `DriverConfiguration` which is external to and "above" `DGModel` in the hierarchy. In general, the module dependencies should form a tree, not a graph (a somewhat related [principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). IMO the added convenience is not worth the cost.

charleskawczynski:
> I see the convenience, but this messes up what I see as the implicit module hierarchy by making `DGModel` dependent on the definition of `DriverConfiguration` which is external to and "above" `DGModel` in the hierarchy. In general, the module dependencies should form a tree, not a graph (here's a somewhat related [principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle)). IMO the added convenience is not worth the cost.

Ahh, very true and I agree. Closing.

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -15,6 +15,15 @@ struct DGModel{BL, G, NFND, NFD, GNF, AS, DS, HDS, D, DD, MD}
...
 
+DGModel(driver_config; kwargs...) = DGModel(
+    driver_config.bl,
+    driver_config.grid,
+    driver_config.numerical_flux_first_order,
+    driver_config.numerical_flux_second_order,
+    driver_config.numerical_flux_gradient;
+    kwargs...,
+)
+
```
blallen, line 26: If we move this to belong in the Driver module instead of the DGMethods module, wouldn't that solve anything? I thought one of the advantages of Julia and multiple dispatch was that you can extend a module from the outside.

kpamnany: You're quite right, I see no issue with that.

charleskawczynski: Ah, good point! I'll reopen and move it

kpamnany (commented):
LGTM. Suggest adding a docstring to the new constructor for [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/docs/src/APIs/Driver/index.md).

kpamnany:
Duplicate docstring? But clearly `ClimateMachine.DGModel` is different than `DGMethods.DGModel`... I wonder if there's a way around this.

charleskawczynski:
There's an [upstream issue](https://github.com/JuliaDocs/Documenter.jl/issues/396) about multiple doc strings, but I think a temporary work-around is to just remove `DGModel` from the driver config API for now.

charleskawczynski:
Hmm. Some spurious formatting complaints (I did not modify `src/Diagnostics/atmos_les_spectra.jl`)

blallen:
> Hmm. Some spurious formatting complaints (I did not modify `src/Diagnostics/atmos_les_spectra.jl`)

I got that too on #1645

kpamnany:
Ditto in #1559. I just added a separate commit for misc formatting. 🤷‍♂️

charleskawczynski:
Rebased.

kpamnany (approved):

Issue #1466, opened 2020-08-25
collect all moisture sources, add them to density once


trontrytel:
As suggested by @tapios below. @charleskawczynski, I don't think we have a mechanism for it right now?

My preference would be to collect all mass sources owing moisture sources and add them as a density source where we specify the equations of motion, not piece-by-piece where we define parameterizations.

_Originally posted by @tapios in https://github.com/CliMA/ClimateMachine.jl/pull/1440#discussion_r470810399_

charleskawczynski:
> My preference would be to collect all mass sources owing moisture sources and add them as a density source where we specify the equations of motion, not piece-by-piece where we define parameterizations.

This is how the tendency specification layer works, which is now complete for the atmoshpere.

PR #1657, opened 2020-10-23
add water mass source from subsidence

szy21:
### Description

Right now in `source.jl` the water mass source from subsidence is not included. This PR adds this term.

### DIFF
--- a/experiments/AtmosLES/bomex_model.jl
+++ b/experiments/AtmosLES/bomex_model.jl
@@ -276,10 +276,12 @@ function atmos_source!(
     end
 
     # Collect Sources
-    source.moisture.ρq_tot += ρ∂qt∂t
     source.ρe += cvm * ρ∂θ∂t * exner(TS) + _e_int_v0 * ρ∂qt∂t
     source.ρe -= ρ * w_s * dot(k̂, diffusive.∇h_tot)
+    source.moisture.ρq_tot += ρ∂qt∂t
     source.moisture.ρq_tot -= ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot)
+    source.ρ += ρ∂qt∂t
+    source.ρ -= ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot)
     return nothing
 end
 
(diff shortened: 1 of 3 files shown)


szy21:
This brings up #1466 which suggests that we collect all mass sources owing moisture sources and add them as a density source where we specify the equations of motion. I think it will help prevent issues like this.

tapios (approved):
Looks good!

szy21:
bors try

szy21:
@charleskawczynski @yairchn I changed the source term from subsidence in bomex_model.jl but this breaks the bomex_edmf test. Could you please take a look? Thanks!

charleskawczynski:
It looks like the error is off by a small margin, so I think it would be okay to modify `s` such that `atol = 0.00015`. @yairchn can you check that the results look somewhat okay?

szy21:
> It looks like the error is off by a small margin, so I think it would be okay to modify `s` such that `atol = 0.00015`. @yairchn can you check that the results look somewhat okay?

The difference is from a change in the bomex source term. I wonder if you want to change the ref_data in the regression test instead?

charleskawczynski:
> I wonder if you want to change the ref_data in the regression test instead?

We could do this, and it would be the more elegant solution, but to do it in a backward-compatible way requires a bit of work.

szy21:
> We could do this, and it would be the more elegant solution, but to do it in a backward-compatible way requires a bit of work.

I see. Either way looks good to me. I will leave the decision to you :)

charleskawczynski:
Once @yairchn confirms that things look okay, I'd be okay with just bumping the tolerance for now.

charleskawczynski:
I looked at the results, and they seem fine AFAICT. I checked, and the change you'll need to make is in `test/Atmos/EDMF/bomex_edmf_regression_test.jl` from
```julia
@test isapprox(norm(absΔdata), 0, atol = tol * 0.01 * s) # norm
```
to
```julia
@test isapprox(norm(absΔdata), 0, atol = tol * 0.015 * s) # norm
```

szy21:
> I looked at the results, and they seem fine AFAICT. I checked, and the change you'll need to make is in `test/Atmos/EDMF/bomex_edmf_regression_test.jl` from
> 
> ```julia
> @test isapprox(norm(absΔdata), 0, atol = tol * 0.01 * s) # norm
> ```
> 
> to
> 
> ```julia
> @test isapprox(norm(absΔdata), 0, atol = tol * 0.015 * s) # norm
> ```

Thanks! I will make the change.

Issue #1092, opened 2020-05-12
Diagnostics - element interface handling


akshaysridhar:
# Description

Currently `AtmosDefault` output regards element interfaces as separate nodes, such that two instances of the same physical coordinate / variable are logged (one from each adjacent element). 
To avoid confusion at the output level, this is perhaps best addressed at the Diagnostics module level. 

## Additional context
Example (choose an arbitrary `AtmosDefault` output `.nc` file and examine, for instance, the z coordinate)
```
julia> z[:]
15-element Array{Float64,1}:
    0.0
  143.8943038716762
  416.66666666666634
  689.4390294616572
----- (e.g. 833 appears 2x)
  833.3333333333322
  833.3333333333322
  977.2276372050085
 1250.0000000000002
```

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

jkozdon:
@akshaysridhar: We talked about this on slack, but figured it was worth summarizing on the issue

You could project the solution to a continuous, piecewise polynomial. In 1-D this would just be averaging the values across the faces. In multi-D the averaging need to be done properly across element corners and edges; the connectivity information needed for this is not currently constructed in the code.

Also keep in mind that even if the solution is made continuous, the derivative will still be discontinuous.

If the multi-D polynomial is made continuous and you dump data in minimal storage, you will also need a map from global degree of freedom to element local degree of freedom (or dump the data still in DG storage, which would not get around the problem noted above).

In general, I think that any diagnostics should take into account the fact that we have a piecewise polynomial, not just nodal points. For instance if an average and moments are needed these should be computed with an integral approximation not just by averaging the points; I believe we currently use the integral approximation in the diagnostics.

jkozdon:
> You could project the solution to a continuous, piecewise polynomial

[...]

PR #1660, opened 2020-10-24
Interpolate duplicate lgl points in `get_vars_from_nodal_stack`

### LINKED ISSUE #1556: Add interpolation step to dict_of_nodal_states
# Description
@charleskawczynski and I would like to add the interpolation helper functions here:
https://github.com/CliMA/ClimateMachine.jl/blob/f5798ef909c03b15363f2ead410979ccc4c3e887/tutorials/Land/Soil/interpolation_helper.jl
into the `dict_of_nodal_states` action. 

# Additional context

Many users who plot output from the `SingleStackConfiguration` will want to smooth over the discontinuities at element boundaries. 

# For CLIMA Developers

charleskawczynski:
### Description

This PR
 - Adds a kwarg `rm_dupes` to `get_z`, to remove duplicate lgl points at the element faces.
 - Adds a kwarg `interp` to `get_vars_from_nodal_stack` to vertically average variables at the duplicate lgl points at the element faces.
 - Swaps some positional args and kwargs in `dict_of_nodal_states`
 - Adds `@inbounds` to `get_vars_from_nodal_stack` loops

This PR will help with two issues at the moment
 - Exported plots currently included duplicate points at the element faces, which visually appear as jagged oscillations.
 - We cannot create contour plots with duplicate points in space

This new feature is exercised in the heat tutorial.

Closes #1556

### DIFF
--- a/src/Utilities/SingleStackUtils/SingleStackUtils.jl
+++ b/src/Utilities/SingleStackUtils/SingleStackUtils.jl
@@ -28,6 +28,7 @@ using ..VariableTemplates
         i::Int = 1,
         j::Int = 1,
         exclude::Vector{String} = String[],
+        interp = false,
     ) where {T, dim, N}
 
 Return a dictionary whose keys are the `flattenednames()` of the variables
@@ -47,12 +48,14 @@ function get_vars_from_nodal_stack(
     i::Int = 1,
     j::Int = 1,
     exclude::Vector{String} = String[],
+    interp = false,
 ) where {T, dim, N}
 
     # extract grid information and bring `Q` to the host if needed
     FT = eltype(Q)
     Nq = N + 1
     Nqk = dimensionality(grid) == 2 ? 1 : Nq
+    Np = dofs_per_element(grid)
     state_data = array_device(Q) isa CPU ? Q.realdata : Array(Q.realdata)
 
     # set up the dictionary to be returned
@@ -60,20 +63,51 @@ function get_vars_from_nodal_stack(
     stack_vals = OrderedDict()
     num_vars = varsize(vars)
     vars_wanted = Int[]
-    for vi in 1:num_vars
+    @inbounds for vi in 1:num_vars
         if !(var_names[vi] in exclude)
             stack_vals[var_names[vi]] = FT[]
             push!(vars_wanted, vi)
         end
     end
-
+    vmap⁻ = grid.vmap⁻
+    vmap⁺ = grid.vmap⁺
+    vgeo = grid.vgeo
     # extract values from `state_data`
-    for ev in vrange
-        for k in 1:Nqk
-            ijk = i + Nq * ((j - 1) + Nq * (k - 1))
-            for v in vars_wanted
-                push!(stack_vals[var_names[v]], state_data[ijk, v, ev])
+    @inbounds for ev in vrange, k in 1:Nqk, v in vars_wanted
+        if interp && k == 1
+            # Get face degree of freedom number
+            n = i + Nq * ((j - 1))
+            # get the element numbers
+            ev⁻ = ev
+            # Get neighboring id data
+            id⁻, id⁺ = vmap⁻[n, 5, ev⁻], vmap⁺[n, 5, ev⁻]
+            ev⁺ = ((id⁺ - 1) ÷ Np) + 1
+            # get the volume degree of freedom numbers
+            vid⁻, vid⁺ = ((id⁻ - 1) % Np) + 1, ((id⁺ - 1) % Np) + 1
+
+            J⁺, J⁻ = vgeo[vid⁻, Grids._M, ev⁻], vgeo[vid⁺, Grids._M, ev⁺]
+            state_local = J⁻ * state_data[vid⁻, v, ev⁻]
+            state_local += J⁺ * state_data[vid⁺, v, ev⁺]
+            state_local /= (J⁻ + J⁺)
+            push!(stack_vals[var_names[v]], state_local)
+        elseif interp && k == Nqk
+            # Get face degree of freedom number
+            n = i + Nq * ((j - 1))
+            # get the element numbers
+            ev⁻ = ev
+            # Get neighboring id data
+            id⁻, id⁺ = vmap⁻[n, 6, ev⁻], vmap⁺[n, 6, ev⁻]
+            # periodic and need to handle this point (otherwise handled above)
+            if id⁺ == id⁻
+                vid⁻ = ((id⁻ - 1) % Np) + 1
+
+                state_local = state_data[vid⁻, v, ev⁻]
+                push!(stack_vals[var_names[v]], state_local)
             end
+        else
+            ijk = i + Nq * ((j - 1) + Nq * (k - 1))
+            state_local = state_data[ijk, v, ev]
+            push!(stack_vals[var_names[v]], state_local)
         end
     end
 
@@ -87,6 +121,7 @@ end
         vars;
         vrange::UnitRange = 1:size(Q, 3),
         exclude::Vector{String} = String[],
+        interp = false,
     ) where {T, dim, N}
 
 Return an array of [`get_vars_from_nodal_stack()`](@ref)s whose dimensions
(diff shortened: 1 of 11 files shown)


christophernhill (commented):
@charleskawczynski I have a comment around ```solver_config``` which may be slightly controversial! 

It feels like ```solver_config``` as used here (and in other places) may be a bit of an _anti-pattern_ as far as composition goes i.e. it precludes using lower level routines when you don't have a setup that matches the ```solver_config``` abstraction as envisioned. 

In the specific code here the ```solver_config``` abstraction use looks like it could get in the way of using ```export_state_plots``` in more flexible ways. For example for a set up that has a bunch of DG entities and grids, rather than just one etc.... We already have entities like this in the ocean and in the coupling stuff I am doodling with I think there may be more emerging. 

What do you think of an interface that looks more like
```
function export_state_plots(
    solver_config,
    all_data,
     time_data,
     output_dir;
     state_types = (Prognostic(), Auxiliary()),
     z = get_z(solver_config.dg.grid),
 )
foo_tuple = ... things I need from solver config ...

export_state_plots( foo_tuple , )
```

would that be practical and make sense as a design pattern here and maybe elsewhere?

charleskawczynski:
> It feels like `solver_config` as used here (and in other places) may be a bit of an _anti-pattern_ as far as composition goes i.e. it precludes using lower level routines when you don't have a setup that matches the `solver_config` abstraction as envisioned.

I agree. That said, `export_state_plots` is not actually used anywhere, this function was added into master so that we can debug/interpret EDMF code as we're developing the model. We could refactor this, but I don't think it's a high priority thing to do

> In the specific code here the `solver_config` abstraction use looks like it could get in the way of using `export_state_plots` in more flexible ways. For example for a set up that has a bunch of DG entities and grids, rather than just one etc.... We already have entities like this in the ocean and in the coupling stuff I am doodling with I think there may be more emerging.

Also, this routine was written to work with results returned from `dict_of_nodal_states`, which utilizes single stack routines, and is likely not practical for LES/GCM runs.

> What do you think of an interface that looks more like
> 
> ```
> function export_state_plots(
>     solver_config,
>     all_data,
>      time_data,
>      output_dir;
>      state_types = (Prognostic(), Auxiliary()),
>      z = get_z(solver_config.dg.grid),
>  )
> foo_tuple = ... things I need from solver config ...
> 
> export_state_plots( foo_tuple , )
> ```
> 
> would that be practical and make sense as a design pattern here and maybe elsewhere?

I'm not sure I follow your suggestion. Would the `export_state_plots` interface stay the same? or use lower level pieces and use a wrapper that collects them given `solver_config`?

kpamnany (commented):
This looks fine to me and I like that `["z"]` has been removed from the interface, but...

> Adds a kwarg interp to get_vars_from_nodal_stack to vertically average variables at the duplicate lgl points at the element faces.

Is this correct? Per @jkozdon (from [this](https://github.com/CliMA/ClimateMachine.jl/issues/1092#issuecomment-627503632) comment):

> > > In 1-D this would just be averaging the values across the faces
> 
> Actually I think that this isn't quite correct, I believe that you also need to account for the Jacobians ([eqn latex](https://www.codecogs.com/eqnedit.php?latex=%5Chat%7Bp%7D&space;=&space;%5Cfrac%7BJ%5E%7B+%7D&space;p%5E%7B+%7D&space;+&space;J%5E%7B+%7D&space;p%5E%7B-%7D%7D%7BJ%5E%7B+%7D&space;+&space;J%5E%7B-%7D%7D))
> 
> latex

charleskawczynski:
> Is this correct? ...

Good question, I'm not sure. Do we need the metric terms?

Also, @jkozdon, I was a bit surprised by how some of the filtered results change between [this PR](https://clima.github.io/ClimateMachine.jl/previews/PR1660/generated/Numerics/DGMethods/showcase_filters/) and [master](https://clima.github.io/ClimateMachine.jl/latest/generated/Numerics/DGMethods/showcase_filters/). Perhaps the interpolation is wrong and this is why some of the results do not appear to have smoothed out (e.g., the `CutoffFilter(grid, Nc=1) with central numerical flux` case)?

jkozdon:
> > Is this correct? ...
> 
> Good question, I'm not sure. Do we need the metric terms?

My old self (and the current @kpamnany) tell the truth about needing the metrics.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -60,20 +62,31 @@ function get_vars_from_nodal_stack(
...
+
+            ev⁺ = ev - 1
+            ijk⁺ = i + Nq * ((j - 1) + Nq * (Nqk - 1))
+            state_local = state_data[ijk⁻, v, ev⁻]
+            state_local += state_data[ijk⁺, v, ev⁺]
+            state_local /= 2
+            push!(stack_vals[var_names[v]], state_local)
+        elseif k == Nqk && ev ≠ last(vrange) && interp
+            # Duplicate, already caught in above branch
+        else
```
jkozdon, line 86: Something like this would work for non-periodic as well (if that's needed)

```suggestion
        if interp && k == 1
            # Get face degree of freedom number
            n = i + Nq * ((j - 1))
            # Get neighboring id data
            id⁻, id⁺ = vmap⁻[n, 5, ev⁻], vmap⁺[n, 5, ev⁻]
            # get the element numbers
            ev⁻ = ev
            ev⁺ = ((id⁺ - 1) ÷ Np) + 1
            # get the volume degree of freedom numbers
            vid⁻, vid⁺ = ((id⁻ - 1) % Np) + 1, ((id⁺ - 1) % Np) + 1

            J⁺, J⁻ = vgeo[vid⁻, _J, ev⁻], vgeo[vid⁺, _J, ev⁺]
            state_local = J⁻ * state_data[vid⁻, v, ev⁻]
            state_local += J⁺ * state_data[vid⁺, v, ev⁺]
            state_local /= (J⁻ + J⁺)
            push!(stack_vals[var_names[v]], state_local)
        elseif interp && k == Nqk
            # Get face degree of freedom number
            n = i + Nq * ((j - 1))
            # Get neighboring id data
            id⁻, id⁺ = vmap⁻[n, 6, ev⁻], vmap⁺[n, 6, ev⁻]
            # periodic and need to handle this point (otherwise handled above)
            if id⁺ == id⁻
                # get the element numbers
                ev⁻ = ev
                vid⁻ = ((id⁻ - 1) % Np) + 1

                state_local = state_data[vid⁻, v, ev⁻]
                push!(stack_vals[var_names[v]], state_local)
            end
        else
```

jkozdon:
> Also, @jkozdon, I was a bit surprised by how some of the filtered results change between [this PR](https://clima.github.io/ClimateMachine.jl/previews/PR1660/generated/Numerics/DGMethods/showcase_filters/) and [master](https://clima.github.io/ClimateMachine.jl/latest/generated/Numerics/DGMethods/showcase_filters/). Perhaps the interpolation is wrong and this is why some of the results do not appear to have smoothed out (e.g., the `CutoffFilter(grid, Nc=1) with central numerical flux` case)?

I don't know enough about the plotting to comment.

That said, I'm not sure you want to remove the dups in that tutorial. Since you want to see the differences between the filters which are pretty relevant to the DG numerics (@trontrytel)

trontrytel:
> I don't know enough about the plotting to comment.
> 
> That said, I'm not sure you want to remove the dups in that tutorial. Since you want to see the differences between the filters which are pretty relevant to the DG numerics (@trontrytel)

Agreed. For the filtering tutorial it would be best to keep the plots as "raw" as possible.

charleskawczynski:
> Agreed. For the filtering tutorial it would be best to keep the plots as "raw" as possible.

I thought that might be the case (I turned it on there because I figured this would highlight the added behavior). I'll revert it.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
charleskawczynski: Thanks! I'll try this out

ilopezgp (approved):

charleskawczynski:
Now that the metric terms properly interpolate the data, we end up with small precision errors which yield poor `xlims` defaults in Plots (see [Plots #3100](https://github.com/JuliaPlots/Plots.jl/issues/3100)). So, I added a hack solution+warning for now.

PR #1662, opened 2020-10-26
Add GCM forced LES case

szy21:
Co-authored-by: @akshaysridhar 
### Description
This PR:

- Adds the ability to read initial conditions and forcing from a netcdf file and use them to force the LES
- Adds a GCM-forced LES driver

### DIFF
--- a/experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
+++ b/experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
@@ -1,0 +1,704 @@
+# General Julia modules
+using ArgParse
+using Dierckx
+using Distributions
+using DocStringExtensions
+using LinearAlgebra
+using NCDatasets
+using Pkg.Artifacts
+using Printf
+using Random
+using StaticArrays
+using Test
+using UnPack
+
+# ClimateMachine Modules
+using ClimateMachine
+
+using ClimateMachine.ArtifactWrappers
+using ClimateMachine.Atmos
+using ClimateMachine.ConfigTypes
+using ClimateMachine.GenericCallbacks
+using ClimateMachine.DGMethods.NumericalFluxes
+using ClimateMachine.Diagnostics
+using ClimateMachine.Mesh.Filters
+using ClimateMachine.Mesh.Grids
+using ClimateMachine.Orientations
+using ClimateMachine.Thermodynamics
+using ClimateMachine.BalanceLaws
+using ClimateMachine.TurbulenceClosures
+using ClimateMachine.ODESolvers
+using ClimateMachine.VariableTemplates
+using ClimateMachine.Writers
+
+# Planet parameters
+using CLIMAParameters
+using CLIMAParameters.Planet: e_int_v0, grav, day
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+# Physics specific imports
+using ClimateMachine.Atmos: altitude, recover_thermo_state
+import ClimateMachine.Atmos: source!, atmos_source!, filter_source
+import ClimateMachine.BalanceLaws: source, eq_tends
+
+# Citation for problem setup
+## CMIP6 Test Dataset - cfsites
+## [Webb2017](@cite)
+
+"""
+    GCMRelaxation{PV, FT} <: TendencyDef{Source, PV}
+
+"""
+struct GCMRelaxation{PV <: Union{Mass, TotalMoisture}, FT} <:
+       TendencyDef{Source, PV}
+    τ_relax::FT
+end
+GCMRelaxation(::Type{FT}, args...) where {FT} = (
+    GCMRelaxation{Mass, FT}(args...),
+    GCMRelaxation{TotalMoisture, FT}(args...),
+)
+
+function source(
+    s::GCMRelaxation{Mass},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # TODO: write correct tendency
+    return 0
+end
+
+function source(
+    s::GCMRelaxation{TotalMoisture},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # TODO: write correct tendency
+    return 0
+end
+
+"""
+    LargeScaleProcess{PV <: Union{Mass,Energy,TotalMoisture}} <: TendencyDef{Source, PV}
+
+# Energy tendency ∂_t ρe
+
+Temperature tendency for the LES configuration based on quantities
+from a GCM. Quantities are included in standard CMIP naming format.
+Tendencies included here are
+
+    tntha = temperature tendency due to horizontal advection
+    tntva = temperature tendency due to vertical advection
+    tntr = temperature tendency due to radiation fluxes
+    ∂T∂z = temperature vertical gradient from GCM values
+
+# Moisture tendency ∂_t ρq_tot
+
+Moisture tendency for the LES configuration based on quantities
+from a GCM. Quantities are included in standard CMIP naming format.
+Tendencies included here are
+
+    tnhusha = moisture tendency due to horizontal advection
+    tnhusva = moisture tendency due to vertical advection
+    ∂qt∂z = moisture vertical gradient from GCM values
+"""
+struct LargeScaleProcess{PV <: Union{Mass, Energy, TotalMoisture}} <:
+       TendencyDef{Source, PV} end
+
+LargeScaleProcess() = (
+    LargeScaleProcess{Mass}(),
+    LargeScaleProcess{Energy}(),
+    LargeScaleProcess{TotalMoisture}(),
+)
+
+function source(
+    s::LargeScaleProcess{Energy},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # Establish problem float-type
+    FT = eltype(state)
+    # Establish vertical orientation
+    k̂ = vertical_unit_vector(m, aux)
+    _e_int_v0 = e_int_v0(m.param_set)
+    # Unpack vertical gradients
+    ∂qt∂z = diffusive.lsforcing.∇ᵥhus
+    ∂T∂z = diffusive.lsforcing.∇ᵥta
+    w_s = aux.lsforcing.w_s
+    cvm = cv_m(ts)
+    # Compute tendency terms
+    # Temperature contribution
+    T_tendency = aux.lsforcing.Σtemp_tendency + ∂T∂z * w_s
+    # Moisture contribution
+    q_tot_tendency =
+        compute_q_tot_tend(m, state, aux, t, ts, direction, diffusive)
+
+    return cvm * state.ρ * T_tendency + _e_int_v0 * state.ρ * q_tot_tendency
+end
+
+function compute_q_tot_tend(m, state, aux, t, ts, direction, diffusive)
+    # Establish problem float-type
+    FT = eltype(state)
+    k̂ = vertical_unit_vector(m, aux)
+    # Establish vertical orientation
+    ∂qt∂z = diffusive.lsforcing.∇ᵥhus
+    w_s = aux.lsforcing.w_s
+    # Compute tendency terms
+    return aux.lsforcing.Σqt_tendency + ∂qt∂z * w_s
+end
+
+function source(
+    s::LargeScaleProcess{Mass},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    q_tot_tendency =
+        compute_q_tot_tend(m, state, aux, t, ts, direction, diffusive)
+    return state.ρ * q_tot_tendency
+end
+
+function source(
+    s::LargeScaleProcess{TotalMoisture},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    q_tot_tendency =
+        compute_q_tot_tend(m, state, aux, t, ts, direction, diffusive)
+    return state.ρ * q_tot_tendency
+end
+
+# Large-scale subsidence forcing
+"""
+    Subsidence <: AbstractSource
+
+Subsidence tendency, given a vertical velocity at the large scale,
+obtained from the GCM data.
+
+    wap = GCM vertical velocity [Pa s⁻¹]. Note the conversion required
+"""
+struct SubsidenceTendency <: AbstractSource end
+
+struct LargeScaleSubsidence{PV <: Union{Mass, Energy, TotalMoisture}} <:
+       TendencyDef{Source, PV} end
+
+LargeScaleSubsidence() = (
+    LargeScaleSubsidence{Mass}(),
+    LargeScaleSubsidence{Energy}(),
+    LargeScaleSubsidence{TotalMoisture}(),
+)
+
+function source(
+    s::LargeScaleSubsidence{Mass},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # Establish vertical orientation
+    k̂ = vertical_unit_vector(m, aux)
+    # Establish subsidence velocity
+    w_s = aux.lsforcing.w_s
+    return -state.ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot)
+end
+function source(
+    s::LargeScaleSubsidence{Energy},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # Establish vertical orientation
+    k̂ = vertical_unit_vector(m, aux)
+    # Establish subsidence velocity
+    w_s = aux.lsforcing.w_s
+    return -state.ρ * w_s * dot(k̂, diffusive.∇h_tot)
+end
+function source(
+    s::LargeScaleSubsidence{TotalMoisture},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # Establish vertical orientation
+    k̂ = vertical_unit_vector(m, aux)
+    # Establish subsidence velocity
+    w_s = aux.lsforcing.w_s
+    return -state.ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot)
+end
+
+# Sponge relaxation
+"""
+    LinearSponge{PV <: Momentum, FT} <: TendencyDef{Source, PV}
+
+Two parameter sponge (α_max, γ) for velocity relaxation to a reference
+state.
+    α_max = Sponge strength (can be interpreted as timescale)
+    γ = Sponge exponent
+    z_max = Domain height
+    z_sponge = Altitude at which sponge layer starts
+"""
+struct LinearSponge{PV <: Momentum, FT} <: TendencyDef{Source, PV}
+    "Maximum domain altitude (m)"
+    z_max::FT
+    "Altitude at with sponge starts (m)"
+    z_sponge::FT
+    "Sponge Strength 0 ⩽ α_max ⩽ 1"
+    α_max::FT
+    "Sponge exponent"
+    γ::FT
+end
+
+LinearSponge(::Type{FT}, args...) where {FT} =
+    LinearSponge{Momentum, FT}(args...)
+
+function source(
+    s::LinearSponge{Momentum},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    #Unpack sponge parameters
+    FT = eltype(state)
+    @unpack z_max, z_sponge, α_max, γ = s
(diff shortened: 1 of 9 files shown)


src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -1518,6 +1518,9 @@ end
         @unroll for s in 1:num_state_prognostic
             state[n, s, e] = l_state[s]
         end
+        @unroll for s in 1:num_state_auxiliary
+            state_auxiliary[n, s, e] = local_state_auxiliary[s]
+        end
```
charleskawczynski, line 1523: I'm not sure that this is a good idea. Looping in @kpamnany.

kpamnany: This is clearly telling us that we need to go ahead with that split of `state_auxiliary` into a prognostic state independent auxiliary state and a prognostic state dependent auxiliary state. The very first thing to do is come up with suitable names for these two auxiliary states.

I think it is fine for `init_state_prognostic` itself to initialize the prognostic state dependent auxiliary state (and it is probably more efficient as well). If we follow this, then we can use `init_state_auxiliary` for the initialization of the prognostic state independent auxiliary state and don't have to add any new interface calls -- just rename them appropriately.

charleskawczynski: `AuxStatic` and `AuxDynamic`?

kpamnany: They aren't really "static" or "dynamic" though? @simonbyrne, opinion?

charleskawczynski: I think the ones that don't depend on the prognostic state are only computed once, whereas ones that do depend on the prognostic state are continually updated.

kpamnany: This split is best left to another PR. I think this change is okay in the meantime.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -134,6 +135,8 @@ struct AtmosModel{FT, PS, PR, O, RS, T, TC, HD, VS, M, P, R, S, TR, DC} <:
     source::S
     "Tracer Terms (Equations for dynamics of active and passive tracers)"
     tracers::TR
+    "GCM Information"
+    gcminfo::GI
```
charleskawczynski, line 139: `gcminfo`? Why is this a submodel of the `AtmosModel`? This is not a physics model, and is counter-intuitive if we're using a single stack / LES.

szy21: The `gcminfo` is used to store external source terms to force the LES / single stack. Maybe a better name would be `forcing`? (But that's also confusing because radiation, source, etc. are forcings too.)

charleskawczynski: Why can't we just add sources like we already have?

szy21: I think it's because we would like to read the netcdf during the initialization, store the forcing terms in `aux`, and compute the source terms (e.g. subsidence tendency) at each timestep. And in the future the forcing may be a function of time. @akshaysridhar can correct me. Is there an easy way to use the current `source` structure?

akshaysridhar: Currently we're using this when computing DG-gradients of the GCM state; maybe `gcminfo` is a misnomer and it should be `modeldata` or similar, but it's being used as a way to store / pass around the gcm-states [which eventually will be time dependent]

szy21: @charleskawczynski does this sound good? or is there a way to use `source` to do this?

charleskawczynski: I need to take a closer look at what’s being done. It may be that GCMInfo is just a bad name

src/Atmos/Model/gcminfo.jl
```diff
@@ -0,0 +1,273 @@
...
+function flux_second_order!(
+    gcminfo::HadGEM,
+    flux::Grad,
+    state::Vars,
+    diffusive::Vars,
+    aux::Vars,
+    t::Real,
+    D_t,
+)
+    nothing
```
charleskawczynski, line 184: Do any GCMInfo types add second order fluxes? If not we should just remove this (same for first order fluxes/sources.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: This Atmos sub-model isn’t adding any tendencies . It kind of looks like this is just adding GCM-specific diagnostics. @kpamnany can we add the GCM-info diagnostics without adding all of these aux/grad fields?

szy21: These are not GCM diagnostics. These are information from an external forcing file, and the tendencies are added in the driver `experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl`. See the `EnergyTendency`, `MoistureTendency`, and `SubsidenceTendency` in the driver for example.

src/Atmos/Model/gcminfo.jl
```diff
@@ -0,0 +1,187 @@
...
+    integ::Vars,
+    state::Vars,
+    aux::Vars,
+) end
+function reverse_integral_load_auxiliary_state!(
+    ::GCMModel,
+    aux::Vars,
+    integ::Vars,
+) end
+function flux_radiation!(
```
charleskawczynski, line 37: Why is this called `flux_radiation!`?

src/Atmos/Model/gcminfo.jl
```diff
@@ -0,0 +1,187 @@
...
+    diffusive::Vars,
+    ∇transform::Grad,
+    state::Vars,
+    aux::Vars,
+    t::Real,
+)
+    nothing
+end
+
+function flux_first_order_gcm!(
```
charleskawczynski, line 76: Why is this empty? Do we need to define `flux_first_order_gcm!`?

src/Atmos/Model/gcminfo.jl
```diff
@@ -0,0 +1,187 @@
...
+    atmos::AtmosModel,
+    flux::Grad,
+    state::Vars,
+    aux::Vars,
+    t::Real,
+)
+    nothing
+end
+
+function flux_second_order!(
```
charleskawczynski, line 87: Why is this empty? Do we need to define `flux_second_order!`?

src/Atmos/Model/gcminfo.jl
```diff
@@ -0,0 +1,187 @@
...
+    flux::Grad,
+    state::Vars,
+    diffusive::Vars,
+    aux::Vars,
+    t::Real,
+    D_t,
+)
+    nothing
+end
+function flux_second_order!(gcminfo::GCMModel, flux::Grad, state::Vars, d_q_tot)
```
charleskawczynski, line 98: Same, do we need this?

szy21: We probably don't need any of these. Akshay and I will check and remove them if they are not needed.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Ok, then we should probably just come up with a better name for `gcminfo` if it's introducing tendencies.

src/Atmos/Model/lsforcing.jl
```diff
@@ -0,0 +1,81 @@
...
+
+abstract type LSForcingModel end
+
+vars_state(::LSForcingModel, ::Prognostic, FT) = @vars()
+vars_state(::LSForcingModel, ::Auxiliary, FT) = @vars()
+vars_state(::LSForcingModel, ::Gradient, FT) = @vars()
+vars_state(::LSForcingModel, ::GradientFlux, FT) = @vars()
+vars_state(::LSForcingModel, ::UpwardIntegrals, FT) = @vars()
+vars_state(::LSForcingModel, ::DownwardIntegrals, FT) = @vars()
+
```
charleskawczynski, line 11:
```suggestion
```

szy21:
@charleskawczynski @kpamnany @akshaysridhar Thanks for your comments! This is ready for review. Could you please take a look again? Thanks!

src/Atmos/Model/lsforcing.jl
szy21: If I remove these lines the tests will fail with the error message "no method matching vars_state(::NoLSForcing, ::ClimateMachine.BalanceLaws.Auxiliary, ::Type{Int64})".

charleskawczynski: Ah, you need to make `LSForcingModel` a subtype of `BalanceLaw`

thomasgibson:
Why is this change included here https://github.com/CliMA/ClimateMachine.jl/pull/1662/commits/345e74267d94dee2a7cd06ca982148d6d7c01d99? Did you guys actually see a noticeable difference in the results with and without this change?

szy21:
> Why is this change included here [345e742](https://github.com/CliMA/ClimateMachine.jl/commit/345e74267d94dee2a7cd06ca982148d6d7c01d99)? Did you guys actually see a noticeable difference in the results with and without this change?

We reverted `345e742`. This change was added before for testing. I didn't really see a noticeable difference, but I also didn't make an apple-to-apple comparison.

src/Atmos/Model/lsforcing.jl
kpamnany: 🤔  I think this is right for a sub-component? But maybe [do what `MoistureModel` does](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/moisture.jl#L6) instead.

src/Atmos/Model/lsforcing.jl
```diff
@@ -0,0 +1,81 @@
...
+    hus::FT,
+    ua::FT,
+    va::FT,
+    tntΣhava::FT,
+    Σtemp_tendency::FT,
+    Σqt_tendency::FT,
+    w_s::FT,
+)
+
+vars_state(::HadGEMVertical, ::Prognostic, FT) = @vars()
```
kpamnany, line 56:
```suggestion
```
Can let this fall back to the parent type.

akshaysridhar: Noted, thanks !

src/Atmos/Model/lsforcing.jl
```diff
@@ -0,0 +1,81 @@
...
+
+vars_state(m::HadGEMVertical, ::Auxiliary, FT) = @vars(
+    ta::FT,
+    hus::FT,
+    ua::FT,
+    va::FT,
+    tntΣhava::FT,
+    Σtemp_tendency::FT,
+    Σqt_tendency::FT,
+    w_s::FT,
```
charleskawczynski, line 53: Can/should we give these more verbose names? Also, what are the ones with `Σ`?

akshaysridhar: lumped tendencies from the forcing netcdf file; collapses the number of aux vars we carry

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,645 @@
...
+##    AUTHOR = {Webb, M. J. and Andrews, T. and Bodas-Salcedo, A. and Bony, S. and Bretherton, C. S. and Chadwick, R. and Chepfer, H. and Douville, H. and Good, P. and Kay, J. E. and Klein, S. A. and Marchand, R. and Medeiros, B. and Siebesma, A. P. and Skinner, C. B. and Stevens, B. and Tselioudis, G. and Tsushima, Y. and Watanabe, M.},
+##    TITLE = {The Cloud Feedback Model Intercomparison Project (CFMIP) contribution to CMIP6},
+##    JOURNAL = {Geoscientific Model Development},
+##    VOLUME = {10},
+##    YEAR = {2017},
+##    NUMBER = {1},
+##    PAGES = {359--384},
+##    URL = {https://www.geosci-model-dev.net/10/359/2017/},
+##    DOI = {10.5194/gmd-10-359-2017},
+##}
```
charleskawczynski, line 55: Let's move this to `docs/bibliography.bib` and cite it using `[Webb2017](@cite)`

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,645 @@
...
+    N,
+    resolution,
+    xmax,
+    ymax,
+    zmax,
+    hfls,
+    hfss,
+    ts,
+    groupid,
+)
```
charleskawczynski, line 418:
```suggestion
function config_cfsites(
    ::Type{FT},
    N,
    resolution,
    xmax,
    ymax,
    zmax,
    hfls,
    hfss,
    ts,
    grouped,
) where {FT}
```

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,644 @@
...
+        tntva,
+        tntr,
+        tnhusha,
+        tnhusva,
+        wap,
+        hfls,
+        hfss,
+        ts,
+        alpha,
+    )
```
charleskawczynski, line 353: Should we return a `NamedTuple` here?

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,645 @@
...
+
+# Define the get_gcm_info function
+"""
+    get_gcm_info(group_id)
+
+For a specific global site, establish and store the GCM state
+for each available vertical level. `group_id` refers to the integer
+index of the specific global site that we are interested in.
+"""
+const forcingfile = "HadGEM2-A_amip.2004-2008.07"
```
charleskawczynski, line 272:
```suggestion
const forcingfile = "HadGEM2-A_amip.2004-2008.07"
"""
    get_gcm_info(group_id)

For a specific global site, establish and store the GCM state
for each available vertical level. `group_id` refers to the integer
index of the specific global site that we are interested in.
"""
```

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,645 @@
...
+Temperature tendency for the LES configuration based on quantities 
+from a GCM. Quantities are included in standard CMIP naming format. 
+Tendencies included here are 
+
+    tntha = temperature tendency due to horizontal advection
+    tntva = temperature tendency due to vertical advection
+    tntr = temperature tendency due to radiation fluxes
+    ∂T∂z = temperature vertical gradient from GCM values
+"""
+
```
charleskawczynski, line 88:
```suggestion
```

src/Atmos/Model/lsforcing.jl
szy21: > Can/should we give these more verbose names? Also, what are the ones with `Σ`?

These names adopt the cmip conventions (e.g. https://pcmdi.llnl.gov/mips/cmip5/requirements.html). This might not be the best option but is widely used I think.

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,631 @@
...
+import ClimateMachine.Atmos: source!, atmos_source!
+
+# Citation for problem setup
+## CMIP6 Test Dataset - cfsites
+## [Webb2017](@cite)
+
+struct GCMRelaxation{FT} <: AbstractSource
+    τ_relax::FT
+end
+function atmos_source!(
```
charleskawczynski, line 48: Should we use the new tendency specification? If so, I can help by adding a commit or two

szy21: Yes, please feel free to add commits! Thanks!

szy21:
bors try

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,706 @@
...
+    t,
+    ts,
+    direction,
+    diffusive,
+)
+    # TODO: write correct tendency
+    return 0
+end
+
+# Temperature tendency term, ∂T∂t
```
charleskawczynski, line 89: Sorry if I left this hanging, let's delete/move this stranded comment

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,706 @@
...
+    w_s = aux.lsforcing.w_s
+    cvm = cv_m(ts)
+    # Compute tendency terms
+    # Temperature contribution
+    T_tendency = aux.lsforcing.Σtemp_tendency + ∂T∂z * w_s
+    # Moisture contribution
+    q_tot_tendency =
+        compute_q_tot_tend(m, state, aux, t, ts, direction, diffusive)
+
+    # TODO: Better names for these? Should we separate them?
```
charleskawczynski, line 151: Can we address this TODO?

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,706 @@
...
+    # Establish vertical orientation
+    k̂ = vertical_unit_vector(m, aux)
+    # Establish subsidence velocity
+    w_s = aux.lsforcing.w_s
+    return -state.ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot)
+end
+
+# Sponge relaxation
+"""
+  LinearSponge (AbstractSource)
```
charleskawczynski, line 269:
```suggestion
    LinearSponge{PV <: Momentum, FT} <: TendencyDef{Source, PV}
```

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
```diff
@@ -0,0 +1,706 @@
...
+    add_arg_group!(cfsite_args, "HadGEM2-A_SiteInfo")
+    @add_arg_table! cfsite_args begin
+        "--group-id"
+        help = "Specify cfSite-ID for forcing data"
+        metavar = "site<number>"
+        arg_type = String
+        default = "site17"
+    end
+    cl_args =
+        ClimateMachine.init(parse_clargs = true, custom_clargs = cfsite_args)
```
charleskawczynski, line 598: Since we're using `rand` in initialization, can we use `fix_rng_seed = true` here?

charleskawczynski (approved):
Aside from my last few small comments which we may want to address (+ squash first!), LGTM

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
szy21: I don't think we need to separate them. I will combine them together.

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
szy21: Yes, should we do this for all the experiments using `rand` (in a separate PR)?

szy21:
> Aside from my last few small comments which we may want to address (+ squash first!), LGTM

Thanks Charlie!

experiments/AtmosLES/cfsite_hadgem2-a_07_amip.jl
kpamnany: If it is useful to run the experiment with a truly random seed at times, then `fix_rng_seed = true` should not be specified here. That way, you can run the experiment interactively (or in some other way) with true randomness without having to edit this file. For CI though, we don't want a random number possibly breaking things sometimes, so we specify this on the command line (in the buildkite file).

For experiments where it doesn't matter if it's truly random or not (such as ones run on the QA pipeline maybe?), you can add this parameter in the `init` parameter list.

szy21: It doesn't really matter for this driver, although I will almost always use `fix_rng_seed = true` in my runscript for scientifically useful results (we compare several simulations and don't want a random number to be the reason of the difference). So maybe I will change it back?

kpamnany: 🤷‍♂️ That's up to you, I just wanted to point out the implications. :)

szy21: If we set `fix_rng_seed = true` here, can we still use `fix_rng_seed = False` in the run script to run the experiment with true randomness? Or we will have to change the file?

kpamnany: Yes, command line arguments override `init` parameters (if `parse_clargs = true` is specified, which it is, here).

szy21: ok I will keep it as it is then, thanks Kiran!

PR #1666, opened 2020-10-27
Remove support for `nothing` source

charleskawczynski:
### Description

#1634 requires sources to be specified via Tuples, and supporting mixed tuple types seems a bit more complicated than what it is worth. This PR removes the use of and reliance on `source = nothing` and `source = (nothing,....)`.

### DIFF
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -24,16 +24,7 @@ function atmos_source!(
 )
     f(atmos, source, state, diffusive, aux, t, direction)
 end
-function atmos_source!(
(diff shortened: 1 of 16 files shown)


kpamnany (commented):
LGTM. @christophernhill: will this affect any of MIT's branches?

jkozdon (approved):
Seems fine to me. This seems to be a relic of the old approach

charleskawczynski:
I took a look, and the ocean doesn't use this machinery for specifying sources. So this should not impact the ocean group at all.

PR #1667, opened 2020-10-27
Remove source function support

charleskawczynski:
### Description

Similar to #1666, #1634 requires sources to be specified via Tuples. This PR removes the use of and reliance on supporting functions inside the `atmos.source` tuple, for example, `source = held_suarez_forcing!` and `source = (held_suarez_forcing!,....)`. Instead a new subtype of `Source` is made and we overload `atmos_source!` for the given subtype. Will wait for #1666 to avoid merge conflicts.

This PR also reuses what was `held_suarez_forcing!`, which seemed to be defined in multiple places. I left the `held_suarez_forcing!` defined in the tutorial to keep it stand-alone.

### DIFF
--- a/experiments/AtmosGCM/heldsuarez.jl
+++ b/experiments/AtmosGCM/heldsuarez.jl
@@ -15,6 +15,7 @@ parsed_args = ClimateMachine.init(parse_clargs = true, custom_clargs = s)
 const number_of_tracers = parsed_args["number-of-tracers"]
 
 using ClimateMachine.Atmos
+import ClimateMachine.Atmos: atmos_source!
 using ClimateMachine.Orientations
 using ClimateMachine.ConfigTypes
 using ClimateMachine.Diagnostics
@@ -42,7 +43,7 @@ const param_set = EarthParameterSet()
 function init_heldsuarez!(problem, bl, state, aux, localgeo, t)
     FT = eltype(state)
 
-    # parameters 
+    # parameters
     _a::FT = planet_radius(bl.param_set)
 
     z_t::FT = 15e3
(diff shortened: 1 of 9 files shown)


experiments/AtmosGCM/heldsuarez.jl
```diff
@@ -97,65 +97,7 @@ function init_heldsuarez!(problem, bl, state, aux, localgeo, t)
...
-    T_equil = max(T_min, T_equil)
-    k_T = k_a + (k_s - k_a) * height_factor * cos(φ)^4
-    k_v = k_f * height_factor
-
-    # Apply Held-Suarez forcing
-    source.ρu -= k_v * projection_tangential(bl, aux, ρu)
-    source.ρe -= k_T * ρ * _cv_d * (T - T_equil)
-    return nothing
-end
+include(joinpath("GCMDriver", "gcm_sources.jl"))
```
kpamnany, line 100: I think this was kept separate intentionally, correct @LenkaNovak?

kpamnany (commented):
LGTM, but not sure about removing `held_suarez_forcing!` from `experiments/AtmosGCM/heldsuarez.jl` so let's see what Lenka has to say?

experiments/AtmosGCM/heldsuarez.jl
LenkaNovak: All the changes look good to me! Thanks, guys!

szy21: > I think this was kept separate intentionally, correct @LenkaNovak?

Yes, I think this was kept separate intentionally. I haven't run GCM tests for a while and don't have a strong preference. @jiahe23 any thoughts?

jiahe23: I guess move it to the TestCase directory is not an option ...

charleskawczynski: We don't have to remove it if it's helpful to keep separate--defining a separate type is easy.

kpamnany: I suggest reverting this one -- Held-Suarez should be moving to `TestCase` as Jia says, so best to keep this independent of the GCMDriver.

jkozdon (approved):
I have no problem with this.

experiments/AtmosGCM/heldsuarez.jl
charleskawczynski: Ok, I left things as is.

PR #1674, opened 2020-10-28
Remove plot helpers hack

charleskawczynski:
### Description

The recently added `plothelpers.jl` hack is already causing issues with some upstream branches 🙄.

This PR
 - Removes the plothelpers hack and just add `xlims` as a kwarg, which should default to Plots.jl default (`:auto`). 
 - Makes `time_data` a positional arg, to unify the interface with the other plotting functions.
 - Fix condition for when `single_var` should sometimes be `true`.
 - Adds a kwarg for `time_units`
 - Propagates `ylabel` in `export_state_plots`

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #1675, opened 2020-10-28
Use `dict_of_nodal_states` in land tutorials

charleskawczynski:
### Description

This PR utilizes the `dict_of_nodal_states` with interpolation in the land tutorials.

Will wait on #1674, which has some things we need.

### DIFF
--- a/tutorials/Land/Soil/Coupled/equilibrium_test.jl
+++ b/tutorials/Land/Soil/Coupled/equilibrium_test.jl
@@ -1,7 +1,8 @@
 # # Coupled heat and water equations tending towards equilibrium
 
 # Other tutorials, such as the [soil heat tutorial](../Heat/bonan_heat_tutorial.md)
-# and -link water tutorial here- demonstrate how to solve the heat
+# and [Richards equation tutorial](../Water/equilibrium_test.md)
+# demonstrate how to solve the heat
 # equation or Richard's equation without considering
 # dynamic interactions between the two. As an example, the user could
 # prescribe a fixed function of space and time for the liquid water content,
@@ -121,15 +122,11 @@ const param_set = EarthParameterSet();
 # Initialize and pick a floating point precision:
 ClimateMachine.init()
 FT = Float64;
-# Load a function that will interpolate output of simulation:
+
+# Load plot helpers:
 const clima_dir = dirname(dirname(pathof(ClimateMachine)));
-include(joinpath(
-    clima_dir,
-    "tutorials",
-    "Land",
-    "Soil",
-    "interpolation_helper.jl",
-));
+include(joinpath(clima_dir, "docs", "plothelpers.jl"));
+
 # Set soil parameters to be consistent with sand.
 # Please see e.g. the [soil heat tutorial](../Heat/bonan_heat_tutorial.md)
 # for other soil type parameters, or [2].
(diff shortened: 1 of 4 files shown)


kmdeck:
woohoo!

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -424,101 +391,74 @@ boundaries = [
...
-        "ϑ_l" => ϑ_l,
-        "T" => T,
-        "K∇h_vert" => K∇h_vert,
-        "κ∇T_vert" => κ∇T_vert,
-    )
-    all_data[step[1]] = all_vars
 
-    step[1] += 1
+callback = GenericCallbacks.EveryXSimulationTime(every_x_simulation_time) do
+    dons = dict_of_nodal_states(solver_config, state_types; interp = true)
```
kmdeck, line 396: question re: how this does interpolations...
before we were doing something to conform with this discussion (what is used in Diagnostics)
https://github.com/CliMA/ClimateMachine.jl/issues/1092

I assume this is creating an interpolation function and then evaluating that at the unique z values, but wondering if that interpolation function is different from what we were using before. I think it boiled down to a simple mean in the previous case, but that from the conversation on that issue I gathered it could be more complex for other geometries.

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -424,101 +391,74 @@ boundaries = [
...
-plot!(all_data[3]["T"], iz, label = "t = 2.25 days")
-plot!(all_data[4]["T"], iz, label = "t = 3 days")
-plot!(legend = :bottomright)
-
-κ∇T_z_plot = plot(
-    all_data[1]["κ∇T_vert"],
-    iz,
-    label = "0.75 days",
-    xlabel = "κ∇T_z (W/m^2)",
+export_plot(
```
kmdeck, line 430: This plot looks incorrect, not sure why?

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -424,101 +391,74 @@ boundaries = [
     FT(1) FT(1) zmax
```
kmdeck, line 391: we can remove these lines now, since they were specific to the interpolation grid.

kmdeck:
If we make a tutorial in the future with a 3D grid, we would have to use something like what we had before, right? (`dict_of_nodal_states` is single stack specific). Is there a goal of extending `dict_of_nodal_states` to non `SingleStack` problems?

tutorials/Land/Soil/Coupled/equilibrium_test.jl
```diff
@@ -424,101 +391,74 @@ boundaries = [
...
-κ∇T_z_plot = plot(
-    all_data[1]["κ∇T_vert"],
-    iz,
-    label = "0.75 days",
-    xlabel = "κ∇T_z (W/m^2)",
+export_plot(
+    z,
+    all_data[2:end],
+    ("soil.water.K∇h[3]",),
+    joinpath(output_dir, "eq_enthalpy_plot.png");
```
kmdeck, line 434: h = hydraulic head :)

kmdeck:
would it be desirable to split `export_plot` into two steps? something like state_plot = create_plot(args...) and then the user can e.g. alter the x and y limits, add this plot as a subplot to a different set of plots, or choose to save it directly? (using plot! commands)

On the one hand, this is a great tool for making plots quickly & for doing sanity checks, and we should highlight it in the tutorials when applicable...but on the other hand, i think once a user needs to do anything different it feels a little inflexible. I guess the idea is that  in that case they have the `dons` variable and can plot whatever they want as they like.

tutorials/Land/Soil/Coupled/equilibrium_test.jl
kmdeck: oops, maybe comment on the wrong line - the gradient of the head one looks funny

charleskawczynski: I think I need to rebase / retrigger the build to get some recent fixes.

tutorials/Land/Soil/Coupled/equilibrium_test.jl
charleskawczynski: This only interpolates the face elements, it does not interpolate onto a regular grid.

charleskawczynski:
> would it be desirable to split export_plot into two steps? something like state_plot = create_plot(args...) and then the user can e.g. alter the x and y limits, add this plot as a subplot to a different set of plots, or choose to save it directly? (using plot! commands)

Yes, I like this idea :) We can easily write a wrapper that returns the plot handle.

> On the one hand, this is a great tool for making plots quickly & for doing sanity checks, and we should highlight it in the tutorials when applicable...but on the other hand, i think once a user needs to do anything different it feels a little inflexible. I guess the idea is that in that case they have the dons variable and can plot whatever they want as they like.

Agreed. `dict_of_nodal_states` really just works with single stacks. For LES/3D, we probably want to use the diagnostics module because both the existing approach and `dons` will be inefficient. Also, I think it makes sense to utilize `dons` since it's already used in a bunch of other places.

tutorials/Land/Soil/Coupled/equilibrium_test.jl
charleskawczynski: @kmdeck I think the plots are looking as expected, now.

charleskawczynski:
I just realized that `tutorials/Land/Soil/Water/equilibrium_test.jl` is not included in `docs/list_of_tutorials.jl`. Should we add it?

kmdeck:
> I just realized that `tutorials/Land/Soil/Water/equilibrium_test.jl` is not included in `docs/list_of_tutorials.jl`. Should we add it?

hi, yes! it was nearly there but I kept getting an error somewhere. but, it looks like it runs now, so we can include it!

there is one other tutorial, in the Soil/Heat directory, which could use `dict_of_nodal_states`. If you want I can make that change.

charleskawczynski:
bors try

charleskawczynski:
bors try

kmdeck:
hey Charlie - this looks good. Monday morning I will make a few aesthetic changes (not showing the intermediate plots in the Soil/Water/equilibrium_test.jl file), reference that in the Soil/Coupled/ tutorial, and then add dict_of_nodal_states to our soil heat tutorial. Then I think it is ready to go! Does that sound ok?

kmdeck:
looks good to me, @charleskawczynski

kmdeck (approved):

PR #1676, opened 2020-10-29
Added modularized convective BL LES case (model+driver)

yairchn:
### Description

This PR modularize the Convective Boundary Layer experiment, separating the previous unstable_bl_kitamura.jl into a model file and an LES configuration script. This follows the convention of the change that was made for the Bomex LES and stable BL cases, and it is a necessary step towards creating a Single Column Model configuration for the stable boundary layer experiment.

I have checked that it yields the same results as unstable_bl_kitamura.jl, and I have added an additional conservation check.

### DIFF
--- a/experiments/AtmosLES/convective_bl_model.jl
+++ b/experiments/AtmosLES/convective_bl_model.jl
@@ -1,39 +1,37 @@
 #!/usr/bin/env julia --project
+# This experiment file establishes the initial conditions, boundary conditions,
+# source terms and simulation parameters (domain size + resolution) for the
 
-## ### Unsteady Boundary Layer LES
-## @article{Kitamura2016,
-## author={Kitamura, Yuji
-## and Ito, Junshi},
-## title={Revisiting the Bulk Relation for Heat Flux in the Free Convection Limit},
-## journal={Boundary-Layer Meteorology},
-## year={2016},
-## month={Jan},
-## day={01},
-## volume={158},
-## number={1},
-## pages={93-103},
-## issn={1573-1472},
-## doi={10.1007/s10546-015-0075-z},
-## url={https://doi.org/10.1007/s10546-015-0075-z}
+# Convective Boundary Layer LES case (Kitamura et al, 2016).
+
+## ### Convective Boundary Layer LES
+## @article{Nishizawa2018,
+## author={Nishizawa, S and Kitamura, Y},
+## title={A Surface Flux Scheme Based on the Monin-Obukhov Similarity for Finite Volume Models},
+## journal={Journal of Advances in Modeling Earth Systems},
+## year={2018},
+## volume={10},
+## number={12},
+## pages={3159-3175},
+## doi={10.1029/2018MS001534},
+## url={https://doi.org/10.1029/2018MS001534}
 ## }
+#
+# To simulate the experiment, type in
+#
+# julia --project experiments/AtmosLES/convective_bl_les.jl
 
-##  @article{doi:10.1029/2018MS001534,
-##  author = {Nishizawa, S. and Kitamura, Y.},
-##  title = {A Surface Flux Scheme Based on the Monin-Obukhov Similarity for Finite Volume Models},
-##  journal = {Journal of Advances in Modeling Earth Systems},
-##  volume = {10},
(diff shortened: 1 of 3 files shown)


kpamnany:
Looks like `unstable_bl_kitamura.jl` is still there. Also, you probably want to change `.buildkite/pipeline.yml` to run this instead.

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env julia --project
+# This experiment file establishes the initial conditions, boundary conditions,
+# source terms and simulation parameters (domain size + resolution) for the
```
ilopezgp, line 3: This should be changed

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -0,0 +1,322 @@
...
+## number={1},
+## pages={93-103},
+## issn={1573-1472},
+## doi={10.1007/s10546-015-0075-z},
+## url={https://doi.org/10.1007/s10546-015-0075-z}
+## }
+#
+# To simulate the experiment, type in
+#
+# julia --project experiments/AtmosLES/stable_bl_les.jl
```
ilopezgp, line 25: This should be changed to the convective bl

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -0,0 +1,322 @@
...
+    e_pot = _grav * z
+    ρe_tot = ρ * total_energy(e_kin, e_pot, TS)
+
+    # Assign initial conditions for prognostic state variables
+    state.ρ = ρ
+    state.ρu = SVector(ρu, ρv, ρw)
+    state.ρe = ρe_tot
+    state.moisture.ρq_tot = ρ * q_tot
+
+    if z <= FT(400) # Add random perturbations to bottom 50m of model
```
ilopezgp, line 197: comment should be changed to bottom 400m

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -0,0 +1,322 @@
...
+    # Problem floating point precision
+    FT = eltype(state)
+    R_gas::FT = R_d(bl.param_set)
+    c_p::FT = cp_d(bl.param_set)
+    c_v::FT = cv_d(bl.param_set)
+    p0::FT = MSLP(bl.param_set)
+    _grav::FT = grav(bl.param_set)
+    γ::FT = c_p / c_v
+    # Initialise speeds [u = Eastward, v = Northward, w = Vertical]
+    u::FT = 8
```
ilopezgp, line 166: This is also in the former file, but I think it is a typo. If you look at the forcing below, the geostrophic velocity is 4 m/s, not 8. Could you check with the paper @yairchn ?

yairchn: that is correct - good catch - I think the "8" stay there from using the Bomex file and changing

experiments/AtmosLES/convective_bl_model.jl
yairchn: done

experiments/AtmosLES/convective_bl_model.jl
yairchn: done

experiments/AtmosLES/convective_bl_model.jl
yairchn: done

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -164,14 +159,14 @@ function init_problem!(problem, bl, state, aux, localgeo, t)
     _grav::FT = grav(bl.param_set)
     γ::FT = c_p / c_v
     # Initialise speeds [u = Eastward, v = Northward, w = Vertical]
-    u::FT = 8
+    u::FT = 4
```
charleskawczynski, line 162: Was this velocity change intended?

charleskawczynski (commented):
LGTM. Let's squash (and resolve the git conflict).

charleskawczynski (approved):

PR #1677, opened 2020-10-29
Fix index in reduce_nodal_stack (SingleStackUtils)

jkozdon:


charleskawczynski (approved):
Good catch! Will this require changing the test, also?

kpamnany (approved):
Good catch!

jkozdon:
> Good catch! Will this require changing the test, also?

Not sure. We'll find out :-)

PR #1681, opened 2020-10-29
Prep support for dry turbconv

charleskawczynski:
### Description

This PR preps support for running with a dry `turbconv` model. More is coming, but we don't want to add too much untested code at a time (the DryModel branches, in this case).

### DIFF
--- a/test/Atmos/EDMF/bomex_edmf.jl
+++ b/test/Atmos/EDMF/bomex_edmf.jl
@@ -46,8 +46,13 @@ function init_state_prognostic!(
     # a moist_thermo state is used here to convert the input θ,q_tot to e_int, q_tot profile
     e_int = internal_energy(m, state, aux)
 
-    q_tot = state.moisture.ρq_tot / state.ρ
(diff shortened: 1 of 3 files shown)


test/Atmos/EDMF/bomex_edmf.jl
```diff
@@ -46,8 +46,13 @@ function init_state_prognostic!(
     # a moist_thermo state is used here to convert the input θ,q_tot to e_int, q_tot profile
     e_int = internal_energy(m, state, aux)
 
-    q_tot = state.moisture.ρq_tot / state.ρ
-    ts = PhaseEquil(m.param_set, e_int, state.ρ, q_tot)
+    if m.moisture isa DryModel
+        ρq_tot = 0
```
ilopezgp, line 50: Is this type stable?

charleskawczynski: We'd have to be aware of promotion rules downstream. I'll wrap it in `FT` so we won't need to think about it.

ilopezgp (approved):

PR #1687, opened 2020-11-02
Extract remaining citations

charleskawczynski:
### Description

Extracts remaining references/citations to `docs/bibliography.bib`.

### DIFF
--- a/docs/bibliography.bib
+++ b/docs/bibliography.bib
@@ -1,9 +1,803 @@
-@article{smagorinsky1963,
-  title={General circulation experiments with the primitive equations: I. The basic experiment},
-  author={Smagorinsky, Joseph},
-  journal={Monthly weather review},
-  volume={91},
-  number={3},
-  pages={99--164},
-  year={1963}
+# The citation keys have been formatted as:
+#    Last author name (titlecase), followed by
+#    (no characters in-between) the year.
+
+@article{Arabas2015,
+  title = {libcloudph++ 1.0: a single-moment bulk, double-moment bulk, and particle-based warm-rain microphysics library in C++},
+  author = {Arabas, Sylwester and Jaruga, Anna and Pawlowska, Hanna and Grabowski, Wojciech W},
+  journal = {Geoscientific Model Development},
+  volume = {8},
+  number = {6},
+  url = {https://www.geosci-model-dev.net/8/1677/2015/},
+  doi = {10.5194/gmd-8-1677-2015},
+  pages = {1677--1707},
+  year = {2015},
+  publisher = {Copernicus GmbH}
 }
+
+@article{Baer1972,
+  title = {An alternate scale representation of atmospheric energy spectra},
+  author = {Baer, Ferdinand},
+  journal = {Journal of the Atmospheric Sciences},
+  volume = {29},
+  number = {4},
+  url = {https://doi.org/10.1175/1520-0469(1972)029<0649:AASROA>2.0.CO;2},
+  pages = {649--664},
+  year = {1972}
+}
+
+@article{Bank1985,
+  title = {Transient simulation of silicon devices and circuits},
+  author = {R. E. Bank and W. M. Coughran and W. Fichtner and E. H. Grosse and D. J. Rose and R. K. Smith},
+  journal = {IEEE Transactions on Computer-Aided Design of Integrated Circuits and Systems},
+  volume = {4},
+  number = {4},
+  pages = {436-451},
+  year = {1985},
+  publisher = {IEEE},
+  doi = {10.1109/TCAD.1985.1270142}
+}
+
+@article{Beare2006,
+  title = {An intercomparison of large-eddy simulations of the stable boundary layer},
+  author = {Beare, Robert J and Macvean, Malcolm K and Holtslag, Albert AM and Cuxart, Joan and Esau, Igor and Golaz, Jean-Christophe and Jimenez, Maria A and Khairoutdinov, Marat and Kosovic, Branko and Lewellen, David and others},
+  journal = {Boundary-Layer Meteorology},
+  volume = {118},
+  number = {2},
+  doi = {10.1175/1520-0469(2000)057<1052:ALESSO>2.0.CO;2},
+  url = {https://doi.org/10.1175/1520-0469(2000)057<1052:ALESSO>2.0.CO;2},
+  pages = {247--272},
+  month = {04},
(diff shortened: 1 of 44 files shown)


trontrytel:
This is great. However I think there is something wrong in how the References page is rendered. The entries are listed based on the labels given to them in the code and not by the first author name or year. 

Also, before this change, clicking on the links in my docs would lead to the relevant paper available on the journal page. Now it just leads to the References page. And the links in the Reference page are placed weirdly and in general lead back to the References page and not the paper.

charleskawczynski:
> However I think there is something wrong in how the References page is rendered.

I agree that it doesn't look great right now, but I think we can improve things in `DocumenterCitations.jl` and the reference page will eventually look better as we upgrade.

> The entries are listed based on the labels given to them in the code and not by the first author name or year.

I'm not sure I understand. Can you give an example?

> Also, before this change, clicking on the links in my docs would lead to the relevant paper available on the journal page. Now it just leads to the References page. And the links in the Reference page are placed weirdly and in general lead back to the References page and not the paper.

That's a good point. Ideally, each entry would be placed inside an HTML tag with an `active` attribute so that when someone clicks on one of the references, it could activate the HTML attribute so that the entry is highlighted. I think that this is something that can also be improved on the `DocumenterCitations.jl` side.

trontrytel:
> > The entries are listed based on the labels given to them in the code and not by the first author name or year.
> 
> I'm not sure I understand. Can you give an example?
> 
> >

Screenshot 2020-11-02 at 09 25 11

trontrytel:
> > Also, before this change, clicking on the links in my docs would lead to the relevant paper available on the journal page. Now it just leads to the References page. And the links in the Reference page are placed weirdly and in general lead back to the References page and not the paper.
> 
> That's a good point. Ideally, each entry would be placed inside an HTML tag with an `active` attribute so that when someone clicks on one of the references, it could activate the HTML attribute so that the entry is highlighted. I think that this is something that can also be improved on the `DocumenterCitations.jl` side.

I'm not sure I understand. My point is that it's way more useful for the links to point to the actual paper than to a reference list based on which you have to find the paper.

charleskawczynski:
> I'm not sure I understand. My point is that it's way more useful for the links to point to the actual paper than to a reference list based on which you have to find the paper.

Wouldn't it still be useful if we included direct links to the paper in the reference entry? I think that between including direct links and highlighting the clicked entry, navigating citations would be very easy.

trontrytel:
> > I'm not sure I understand. My point is that it's way more useful for the links to point to the actual paper than to a reference list based on which you have to find the paper.
> 
> Wouldn't it still be useful if we included direct links to the paper in the reference entry? Right now they're broken (and there are obviously formatting issues), but that's a `DocumenterCitation` problem. I think that navigating citations would be very easy between including direct links and highlighting the clicked entry.

I agree. If the links in the reference page work I'm fine with it. 

I guess we want the reference page to safeguard against when the link stops working and you still want to be able to find the paper by yourself.

charleskawczynski:
> I guess we want the reference page to safeguard against when the link stops working and you still want to be able to find the paper by yourself.

That's a good point, we should make sure that Documenter still checks the links after `References.md` has been generated.

jkozdon:
We should make sure we use doi numbers and links when possible, this will safeguard against links breaking.

charleskawczynski:
> I don't see any dois listed in the references page, can we add that?

Yes, most of the refs I grabbed were from Google Scholar, but not all of them had doi links, so we'll need a second pass to add them where possible.

simonbyrne:
The links are based on the doi when it is available. We should add them to the bibtex entries.

jkozdon:
> The links are based on the doi when it is available. We should add them to the bibtex entries.

I see that now. I just didn't mouse over the ones that were.

charleskawczynski:
bump

trontrytel:
> bump

Are the links pointing to the actual papers working now in the docs?

trontrytel:
Also, the references are still not sorted in any order?

charleskawczynski:
> Are the links pointing to the actual papers working now in the docs? Some of them seem to work, but most do not.

Ahh, I can see that not all entries I used included URLs, I'll go and fix the ones that we previously had.

> Also, the references are still not sorted in any order?

No, IMO, that's a problem that we can ship off for `DocumenterCitations` to solve-- and we should be able to fix it through an upgrade to a newer version which solves this problem.

trontrytel:
> > Are the links pointing to the actual papers working now in the docs? Some of them seem to work, but most do not.
> 
> Ahh, I can see that not all entries I used included URLs, I'll go and fix the ones that we previously had.
> 
> > Also, the references are still not sorted in any order?
> 
> No, IMO, that's a problem that we can ship off for `DocumenterCitations` to solve-- and we should be able to fix it through an upgrade to a newer version which solves this problem.

Sounds good. IMHO I don't care about formatting of the reference list. But I think that links should end up pointing to papers

charleskawczynski:
Updated.

charleskawczynski:
> Looks like now it's timing out?

I think I just haven't tried rebuilding now that the docs are run on the cluster

charleskawczynski:
Bump

simonbyrne (approved):

simonbyrne:
Thanks!

trontrytel:
For me the links in the references are still mostly not working. So in my book it's change for worse. But if everyone else disagrees, please merge

charleskawczynski:
> For me the links in the references are still mostly not working. So in my book it's change for worse. But if everyone else disagrees, please merge

Which link(s)?

simonbyrne:
https://clima.github.io/ClimateMachine.jl/previews/PR1687/References/#Smagorinsky1963

trontrytel:
> > For me the links in the references are still mostly not working. So in my book it's change for worse. But if everyone else disagrees, please merge
> 
> Which link(s)?

For example the first two Toro2013 and Rutledge1983 entries have links that for me point back to the reference page

charleskawczynski:
Ok, I think the links that point to the references page can be fixed upstream.

The smagorinsky entry is wrong, I see that I accidentally copied entries of Lilly model and put them into the Smagorinsky model. I’ll fix that tomorrow.

Perhaps we can merge after that and we can try to tackle some of the upstream issues (I’ll make sure to open an upstream issue regarding the links that aren’t working).

charleskawczynski:
Thanks for reviewing!

PR #1688, opened 2020-11-02
Add continuous data interp to single stack utils

### LINKED ISSUE #1410: option to SingleStackUtils to sample output on finer grid
# Description
As suggested by @lcw and @jkozdon here: https://github.com/CliMA/ClimateMachine.jl/pull/1375#issuecomment-662762117

It would be good to add an option to sample the solution on a fine grid when doing output from SingleStack. This is how its done for the vtk: https://github.com/CliMA/ClimateMachine.jl/commit/4514acad6b696ef240c85a888599991bbacd5b71# 

I think it would help with debugging. Leaving it here so that we don't forget about it @charleskawczynski 

# For CLIMA Developers

charleskawczynski:
### Description

The EDMF regression test needs to be made more robust. We're already running into needing to increase the tolerance as physics is being updated, and more changes are in the works which will certainly change the EDMF results. Ideally, we should compare with LES, which requires interpolating to a common grid.

This PR adds `ContinuousData` to `SingleStackUtils`. `ContinuousData` is a wrapper struct `interpolate`/`extrapolate` in `Interpolations.jl`, which will allow us to easily compare solutions on different grids for single stack problems. I believe this could be used in one of the land tutorials. Closes #1410.


trontrytel:
LGTM. I thought we were supposed to use `Dierckx` for interpolating things - it is already in our dependencies. Why do we need another package for it?

trontrytel (approved):

charleskawczynski:
> LGTM. I thought we were supposed to use `Dierckx` for interpolating things - it is already in our dependencies. Why do we need another package for it?

I think that was because `Interpolations.jl` had stopped being maintained, but it looks like someone has picked it back up. That said, we can try to use `Dierckx` if we can avoid adding another dep!

charleskawczynski:
Ah, `Dierckx`'s `Spline1D` effectively behaves the same way. Will just use this instead. Thanks!

PR #1697, opened 2020-11-03
Adding an optional arg to pass extra modeldata to pass from top-level

christophernhill:
Small PR to pass an arbitrary extra modeldata tuple into ocean configs in ocean/dev, for use in trying out 
new config approaches. 

Also a test of making a PR against the ocean/dev branch.

### DIFF
--- a/experiments/OceanBoxGCM/simple_box.jl
+++ b/experiments/OceanBoxGCM/simple_box.jl
@@ -50,6 +50,7 @@ function run_simple_box(
     BC = nothing,
     Δt = nothing,
     refDat = (),
+    mymodeldata = (),
 )
     if imex
         solver_type =
@@ -68,7 +69,7 @@ function run_simple_box(
     grid = driver_config.grid
     vert_filter = CutoffFilter(grid, polynomialorder(grid) - 1)
     exp_filter = ExponentialFilter(grid, 1, 8)
-    modeldata = (vert_filter = vert_filter, exp_filter = exp_filter)
+    modeldata = (vert_filter = vert_filter, exp_filter = exp_filter, mymodeldata = mymodeldata )
 
     timestart, timeend = timespan
     solver_config = ClimateMachine.SolverConfiguration(
--- a/test/Ocean/HydrostaticBoussinesq/test_ocean_gyre_short.jl
+++ b/test/Ocean/HydrostaticBoussinesq/test_ocean_gyre_short.jl
@@ -48,5 +48,8 @@ const FT = Float64
         BC = BC,
         Δt = 120,
         refDat = refVals.short,
+        mymodeldata = ( my_component_tag = "oceanExp2Lal0",
+                        ny_name = "chris" 
+                      ),
     )
 end


experiments/OceanBoxGCM/simple_box.jl
```diff
@@ -68,7 +69,7 @@ function run_simple_box(
     grid = driver_config.grid
     vert_filter = CutoffFilter(grid, polynomialorder(grid) - 1)
     exp_filter = ExponentialFilter(grid, 1, 8)
-    modeldata = (vert_filter = vert_filter, exp_filter = exp_filter)
+    modeldata = (vert_filter = vert_filter, exp_filter = exp_filter, mymodeldata = mymodeldata )
```
glwagner, line 72: What is `modeldata` and it's purpose? It's hard for me to interpret whether `mymodeldata` is the right name for this arg (I'm suspicious of putting the prefix `my` in front of anything) but otherwise looks fine.

glwagner (approved):

experiments/OceanBoxGCM/simple_box.jl
jkozdon: You could just all it `modeldata` and splat it onto the tuple (though then you could end up with naming problems if duplicates are used)
```suggestion
    modeldata = (vert_filter = vert_filter, exp_filter = exp_filter, modeldata... )
```

christophernhill: @jkozdon I was thinking it might be good to experiment with having a private namespace. I am not sure exactly yet! 

@glwagner modeldata is a tuple that holds a bunch of metadata that is then carried around with the DG abstraction. It is mean to used/abused as someone sees fit. Right now I want to use it to pass an option into writevtk ( https://github.com/CliMA/ClimateMachine.jl/blob/6c97b3f16e238893bb218f0036a5e1908ba468ee/src/InputOutput/VTK/writevtk.jl#L63  ) that would let me add a feature so that this range here 

https://github.com/CliMA/ClimateMachine.jl/blob/6c97b3f16e238893bb218f0036a5e1908ba468ee/src/InputOutput/VTK/writevtk.jl#L133

could be customized to run from say -0.9 to 0.9 or whatever range I am interested in (and/or some other meta things like only doing VTK for some elements for a particular callback etc....). 

I am thinking of this as a convenience for little things that may not become standard, but can be useful.

glwagner: I like @jkozdon's suggestion!

jkozdon: > @jkozdon I was thinking it might be good to experiment with having a private namespace. I am not sure exactly yet!

Makes sense.

src/InputOutput/VTK/writevtk.jl
```diff
@@ -69,11 +69,20 @@ function writevtk(
...
     (h_vgeo, h_Q, h_aux) =
         device isa CPU ? (vgeo, Q.data, state_auxiliary.data) :
         (Array(vgeo), Array(Q), Array(state_auxiliary))
+
+    rnglo = -1
+    rnghi =  1
+    if haskey(md,:mymodeldata)
+     rnglo = Main.get_parm(md.mymodeldata,:my_vtk_range_lo,rnglo)
+     rnghi = Main.get_parm(md.mymodeldata,:my_vtk_range_hi,rnghi)
+    end
```
glwagner, line 84: Perhaps a comment might be useful here to explain what "problem" this piece of code is trying to solve?

To summarize the user interface that this design is implementing, if one wants to use the "rnglo" and "rnghi" parameters in VTK output, then one must configure their `modeldata` object to have the fields

```
modeldata.mymodeldata.my_vtk_range_low
```

and

```
modeldata.mymodeldata.my_vtk_range_hi
```

I'd just like to point out that it's probably unsustainable to keep trying to use `modeldata` to hold parameters for VTK output. Is there no object called, say, `VTKOutputWriter` that might be a good place to keep track and interface with these kinds of parameters?

I'm also a bit confused about `Main.get_parm`. Does this mean that a user needs to define the function `get_parm` in the global scope for this functionality to work?

christophernhill: @glwagner my_vtk_range_low etc... is sort of just a specific case of a general pattern. It seems likely there are always going to be things where someone wants to tweak a setting, but the data structures don't expose a convenient short cut. This is meant to be a short cut that provides a convenient pattern for getting experimental things done quickly and succinctly that can help with knowing how to evolve API/interfaces etc... 

get_parm could be in a module. Its in Main. at moment just to get the discussion going. If we think this is something that would be useful to have, at least for some workflows, then it can go in a module. 

So this PR is somewhat about deciding if we want to use ocean/dev to experiment with some tools to support rapid shortcuts for configuring ad hoc stuff that colors outside the lines, or if we want to try and get everyone to only color within opinionated interface lines - and leave it to the after market to look after the unofficial add ons?

glwagner: I don't draw a distinction between experimenting and development. There are no lines to color outside of! I am simply trying to provide feedback on this new feature.

I don't know much about the code, which limits my ability to review this contribution. Perhaps this is the best way to achieve your goals. I also don't really understand the problem that's being solved. It looks like your parameter changes `ξdst`. But since I don't know what `ξdst` is, or what "range low" might refer to, I'm pretty much stuck at trying to help further.

More generally it looks like there is a function `writevtk` but no object that represents VTK output. This is presumably the main problem. Since there's no object to carry output specification, extending the code to provide new features re: output is difficult, and multiple dispatch cannot be used to modularize the functionality of `writevtk`. This is probably the main issue --- extending the functionality of the code should not require editing the source code as is done in this PR, but rather should involve defining new functions in your script that extend or change the behavior of existing ones. 

I'd be up for taking a broader look at how VTK output is specified, and designing an abstraction that has the flexibility you need to be productive. To start this discussion, I think an issue is appropriate so that we can review the existing design. If you'd rather just merge this and move on, that's fine too.

christophernhill:
Once IVP PR is merged I can add this on top of that. It probably makes more
sense to discuss then, since there is some overlap.

PR #1699, opened 2020-11-04
Formatting base files in CLiMA style

christophernhill:
This PR is only formatting changes and changes to make branch work with CI

It updates base files that were added to ocean/dev so that the files are in line with CLiMA coding style
and so that subsequent PRs won't have formatter errors from these files. 

The PR also includes fixes to pipeline.yml to make CI work against latest scripts and one change to
API docs to include time stepper doctoring that had been missed.


charleskawczynski (approved):

christophernhill:
bors try

christophernhill:
bors try

charleskawczynski (changes requested):
This seems to be now changing the pipeline too. cc @jakebolewski

christophernhill:
> This seems to be now changing the pipeline too. cc @jakebolewski

@charleskawczynski @jakebolewski suggested changing to latest pipeline, so trying that.

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
bors try

christophernhill:
@jm-c or @glwagner can you see if you can approve. I am getting 

```
 Rejected by too few approved reviews
```

src/Ocean/SplitExplicit01/Communication.jl
```diff
@@ -11,8 +11,11 @@ using Printf
     dgFast,
     Qfast,
     S_fast,
-    slow_dt, rkC, rkW,
-    s, nStages,
+    slow_dt,
+    rkC,
+    rkW,
+    s,
+    nStages,
```
glwagner, line 18: Happy to approve this. I do have a question about style. Most of the code is written using snakecase, but here some variables appear in camelcase. Should we pick one style and stick with it?

src/Ocean/SplitExplicit01/OceanBoundaryConditions.jl
```diff
@@ -131,9 +131,9 @@ apply no penetration boundary for temperature
     A⁻,
     t,
 )
-#   D⁺.ν∇u = -D⁻.ν∇u
```
glwagner, line 134: I guess we really shouldn't have commented out lines in commits. Anyways, if we do note for the future that's better to put the comment in the place of code, ie, we shouldn't add comments at the very beginning of the line (like you'd do with fortran) since we'll end up with double indents after formatting.

glwagner (approved):

src/Ocean/SplitExplicit01/Communication.jl
christophernhill: We should fix this in a separate PR to make sure @jm-c is OK with it and to keep PRs clean.

PR #1701, opened 2020-11-04
Working Squall line with NonEquilMoist

slifer50:
### Description

Added a working Squall line experiment using NonEquilMoist

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #1706, opened 2020-11-05
warm rain interface

trontrytel:
### Description

Adds interface for warm rain (i.e. no ice or snow) precipitation. 

Right now I was using dycoms setup for testing. But Ideally once https://github.com/CliMA/ClimateMachine.jl/pull/1701 is merged I'll switch to that. 

@charleskawczynski - LMK if this looks reasonable! Snow is soon to follow once I get feedback on my equations.


experiments/AtmosLES/dycoms.jl
```diff
@@ -115,7 +141,8 @@ function integral_load_auxiliary_state!(
     aux::Vars,
 )
     FT = eltype(state)
-    integrand.radiation.attenuation_coeff = state.ρ * m.κ * aux.moisture.q_liq
+    integrand.radiation.attenuation_coeff =
+        state.ρ * m.κ * state.moisture.ρq_liq / state.ρ
```
charleskawczynski, line 145:
```suggestion
    integrand.radiation.attenuation_coeff =
        m.κ * state.moisture.ρq_liq
```

src/Atmos/Model/precipitation.jl
```diff
@@ -63,54 +63,18 @@ Precipitation model with rain only.
...
 
-vars_state(::Rain, ::Prognostic, FT) = @vars(ρq_rain::FT)
-vars_state(::Rain, ::Gradient, FT) = @vars(q_rain::FT)
-vars_state(::Rain, ::GradientFlux, FT) = @vars(ρd_q_rain::SVector{3, FT})
-vars_state(::Rain, ::Auxiliary, FT) =
-    @vars(terminal_velocity::FT, src_q_rai_tot::FT)
+vars_state(::Rain, ::Prognostic, FT) = @vars(ρq_rai::FT)
+vars_state(::Rain, ::Gradient, FT) = @vars(q_rai::FT)
+vars_state(::Rain, ::GradientFlux, FT) = @vars(∇q_rai::SVector{3, FT})
+vars_state(::Rain, ::Auxiliary, FT) = @vars()
```
charleskawczynski, line 69:
```suggestion
```
I think we default to `@vars()`

experiments/AtmosLES/dycoms.jl
trontrytel: I will prolly keep dycoms without rain and add rain to squall line case when it is merged. I just wanted to have something to compile for.

experiments/AtmosLES/dycoms.jl
```diff
@@ -268,7 +322,21 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
 
     F_0 = FT(70)
     F_1 = FT(22)
-    radiation = DYCOMSRadiation{FT}(κ, α_z, z_i, ρ_i, D_subsidence, F_0, F_1)
+    if moisture_model == "equilibrium"
+        equilibrium_moisture_model = true
+    else
+        equilibrium_moisture_model = false
+    end
```
charleskawczynski, line 329:
```suggestion
    equilibrium_moisture_model = moisture_model == "equilibrium"
```

trontrytel: I like the verbose version better

trontrytel:
@charleskawczynski - I think this is good enough. Could you take another look though? I thought this is a small PR but it kind of grew bigger than expected.

@kpamnany, @szy21 , @akshaysridhar , @tapios: Here are results of 1-hour dycoms simulation with drizzle. The lines are roughly t in (0, 20min, 40min, 60min). Bad things happen later and I'm working on it. But it is an improvement from the current state. Plus, I don't want to block the Revamp Efforts ;) 

dycoms_drizzle

kpamnany:
Does `rwp` look right?

trontrytel:
> Does `rwp` look right?

I would say good enough for now. 

Here are the simulation results from the 4 hour run. Different lines come from 0min, 80min, 160min and 240min in the simulation time. I think the results are as good as our current numerics can make them. The time-series of LWP and RWP look reasonable and so do the profiles of cloud water and rain. There are some serious problems around the inversion level that lead to changes in the profiles above the cloud layer. This shouldn't be happening, but it is not caused by adding drizzle in this PR, I think. 

@szy21 - you prolly have more experience with the quality of our dycoms runs so far?

dycoms_3

dycoms_1

dycoms_2

szy21:
The results look reasonable to me. And I think @trontrytel is correct that the problems near the inversion are due to numerics, not this PR.

src/Diagnostics/atmos_les_default.jl
```diff
@@ -348,6 +446,7 @@ function atmos_les_default_init(dgngrp::DiagnosticsGroup, currtime)
         vars["cld_base"] = ((), FT, Variables["cld_base"].attrib)
         vars["cld_cover"] = ((), FT, Variables["cld_cover"].attrib)
         vars["lwp"] = ((), FT, Variables["lwp"].attrib)
+        vars["rwp"] = ((), FT, Variables["lwp"].attrib)
```
trontrytel, line 449: @kpamnany - shouldnt it be `Variables["rwp"].attrib` ?

docs/src/HowToGuides/Atmos/PrecipitationModelChoices.md
```diff
@@ -0,0 +1,32 @@
...
+## NoPrecipitation
+
+The `NoPrecipitation` model assumes there is no precipitating water present
+  in any form in the simulation.
+It does not add any precipitation related variables to state variables.
+This model can be used (i) without defining any precipitation related source terms,
+  (ii) or using the [0-moment microphysics scheme](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/Microphysics_0M/).
+In the first case `ρ q_tot` (total water specific humidity) is not removed
+  during the simulation.
+In the second case `ρ q_tot` is removed if it exceeds certain threshold.
```
tapios, line 16:
```suggestion
In the second case `ρ q_tot` is removed if it exceeds a threshold.
```

docs/src/HowToGuides/Atmos/PrecipitationModelChoices.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+## Rain
+
+The `Rain` model assumes that precipitating water is present but only in the
+  form of rain (liquid phase precipitation).
+It adds `ρ q_rai` (air density times total rain water specific humidity)
+  to state variables.
+It uses a subset of source terms from the [1-moment microphysics scheme](https://clima.github.io/ClimateMachine.jl/latest/Theory/Atmos/Microphysics/)
+  that parameterize processes relevant to liquid phase clouds
```
tapios, line 26:
```suggestion
  that parameterize processes relevant to liquid-phase clouds
```

docs/src/HowToGuides/Atmos/PrecipitationModelChoices.md
```diff
@@ -0,0 +1,32 @@
...
+  (namely autoconverion, accretion and rain evaporation).
```
tapios, line 27:
```suggestion
  (autoconverion, accretion, and rain evaporation).
```

docs/src/Theory/Atmos/Microphysics.md
```diff
@@ -676,6 +676,44 @@ If ``T > T_{freeze}``:
...
+```math
+\begin{equation}
+\left. \sum_{j\in\{v,l,i\}}(I_j + \Phi)  \rho C(q_j \rightarrow q_p) \right|_{precip} =
+  (I_l + \Phi) \rho \, \mathcal{S}_{q_{tot}}
+\end{equation}
+```
+where:
+ - ``I_l = c_{vl} (T - T_0)`` is the internal energy of liquid water,
+ - ``T`` is the temperature,
+ - ``T_0`` is the thermodynamic reference temperature (which is unrelated to the reference temperature used in hydrostatic reference states used in the momentum equations),
```
tapios, line 713:
```suggestion
```
`T_0` does not seem to be used here.

src/Atmos/Model/boundaryconditions.jl
```diff
@@ -12,6 +12,7 @@ export AtmosBC,
     PrescribedEnergyFlux,
     BulkFormulaEnergy,
     Impermeable,
+    ImpermeablePrecipitation,
```
tapios, line 15: I'm confused here by what this means. Precipitation should have an outflow condition, which means that the surface is permeable to it. (Maybe it's just a matter of terminology.)

src/Atmos/Model/precipitation.jl
```diff
@@ -63,115 +63,79 @@ Precipitation model with rain only.
...
+        v_term = terminal_velocity(
+            atmos.param_set,
+            atmos.param_set.microphys.rai,
+            state.ρ,
+            q_rai,
+        )
+    end
+
+    flux.precipitation.ρq_rai +=
+        state.precipitation.ρq_rai * (u - SVector(FT(0), FT(0), v_term))
```
tapios, line 124: I think it would be better to define the terminal velocity as a downward pointing vertical vector from the outset, and then just add that vector velocity here. It seems to me that what is written here only works in a Cartesian box but not on a sphere (where the third vector component is not necessarily vertical).

trontrytel:
@tapios - Thank you for taking a look. 

In this PR there is a similar issue as in the snow sublimation question I asked. The coupling between the energy and precipitation sources in this PR is coded as:

https://clima.github.io/ClimateMachine.jl/previews/PR1706/Theory/Atmos/Microphysics/#Coupling-to-the-state-variables

i.e., in my derivations the `rho I` source term from rain microphysics is `(S_evap + S_acnv + S_accr) I_l`. But if I understand the notation from the DDs correctly it should be `S_evap I_v + (S_acnv + S_accr) I_l`

src/Atmos/Model/boundaryconditions.jl
trontrytel: I admit I'm confused by boundary conditions. I think the implemented BC would result in having `rho q_rai = 0` in the ghost points and no additional conditions on the velocity field. So this would mean free outflow at the bottom and potentially inflow of zeros at the top and sides? @charleskawczynski - am I reading it right?

In general, I'm not sure if I should do anything more to make the BC consistent between what we do for air velocity and what happens for rain. Rain has its own velocity equal to the sum of air velocity and terminal velocity.

src/Atmos/Model/precipitation.jl
trontrytel: Replaced with: 

https://github.com/CliMA/ClimateMachine.jl/blob/47a3c983568896118ba6afba7c9b26f3d9ad2329/src/Atmos/Model/precipitation.jl#L123-L124

src/Atmos/Model/boundaryconditions.jl
charleskawczynski: Mathematically, what is the intended BC? When I read `Impermeable`, I think zero Neumann BCs. I think it'd be a good idea to loop in those who better understand `boundary_state!`. cc @jkozdon @blallen

tapios: This is a BC on an advective flux. We want an outgoing condition on the falling precipitation. Usually outgoing conditions are hard to implement, but here it should be reasonably straightforward: The normal component of the general velocity field is zero at the surface; the normal component of the terminal velocity, which advects the rain relative to the air, should have no gradients at the surface (hence homogeneous Neumann conditions on this component).

charleskawczynski: Based on my understanding, and using the heat tutorial as an example, we should be setting `diff⁺.∇q_rai = 0` in `atmos_precipitation_normal_boundary_flux_second_order!` which I think should result in `n̂⋅∇q_rai = 0`. That said, I don't see the same thing done in `atmos_moisture_normal_boundary_flux_second_order!`. @jkozdon can you weigh in?

tapios: We should do this too, but it is not the boundary condition in question here. Setting the diffusive fluxes to zero at the boundary just sets the normal SGS flux to zero. The question here is about the advective fluxes (first-order fluxes).

charleskawczynski (approved):
Looks good. It sounds like there are a few things that need to be worked on-top of this PR. Let's merge now so that this doesn't hold up the tendency specification PRs (there are still a bunch and several depend on this PR), and we can work out the long sims and nail down BCs in a separate PR. I'll open issues so that we don't lose track.

tapios:
> Looks good. It sounds like there are a few things that need to be worked on-top of this PR. Let's merge now so that this doesn't hold up the tendency specification PRs (there are still a bunch and several depend on this PR), and we can work out the long sims and nail down BCs in a separate PR. I'll open issues so that we don't lose track.

Hang on just for the rest of the day. I wanted to get back to @trontrytel's questions above, and hope to be able to do so tonight.

tapios:
> @tapios - Thank you for taking a look.
> 
> In this PR there is a similar issue as in the snow sublimation question I asked. The coupling between the energy and precipitation sources in this PR is coded as:
> 
> https://clima.github.io/ClimateMachine.jl/previews/PR1706/Theory/Atmos/Microphysics/#Coupling-to-the-state-variables
> 
> i.e., in my derivations the `rho I` source term from rain microphysics is `(S_evap + S_acnv + S_accr) I_l`. But if I understand the notation from the DDs correctly it should be `S_evap I_v + (S_acnv + S_accr) I_l`

What you have here looks correct to me. We should update the DD's (please feel free to do it if you have a chance). The expression in the DD's only consider the situation of in-phase conversion of cloud condensate to precipitation (i.e., cloud liquid to rain, cloud ice to snow). Your formulation for rain evaporation, it seems to me, assumes a two-step process: first convert rain to cloud condensate (and incorporate the energy change in the working fluid); then evaporate that newly converted condensate (assuming no further energy transformation). So the latent cooling occurs in the working fluid, as it should. 

It might be useful to make this clearer in the documentation (i.e., the lhs Eq. (30) is a bit misleading).

szy21:
> What you have here looks correct to me. We should update the DD's (please feel free to do it if you have a chance). The expression in the DD's only consider the situation of in-phase conversion of cloud condensate to precipitation (i.e., cloud liquid to rain, cloud ice to snow). Your formulation for rain evaporation, it seems to me, assumes a two-step process: first convert rain to cloud condensate (and incorporate the energy change in the working fluid); then evaporate that newly converted condensate (assuming no further energy transformation). So the latent cooling occurs in the working fluid, as it should.

Sorry, I'm still confused. Should the energy source term from rain to vapor transition be the same as the one from vapor to rain transition, just of opposite signs? If so, consider the transition from vapor to rain, and assume the latent heating occurs in the working fluid, then the energy source term would be I_v * S_evap? I think I miss something here.

(You can merge the PR before the question is answered.)

trontrytel:
> > What you have here looks correct to me. We should update the DD's (please feel free to do it if you have a chance). The expression in the DD's only consider the situation of in-phase conversion of cloud condensate to precipitation (i.e., cloud liquid to rain, cloud ice to snow). Your formulation for rain evaporation, it seems to me, assumes a two-step process: first convert rain to cloud condensate (and incorporate the energy change in the working fluid); then evaporate that newly converted condensate (assuming no further energy transformation). So the latent cooling occurs in the working fluid, as it should.
> 
> Sorry, I'm still confused. Should the energy source term from rain to vapor transition be the same as the one from vapor to rain transition, just of opposite signs? If so, consider the transition from vapor to rain, and assume the latent heating occurs in the working fluid, then the energy source term would be I_v * S_evap? I think I miss something here.
> 
> (You can merge the PR before the question is answered.)

Thanks for thinking about those things along with me!

1) I think it's enough to say that the energy change is always happening at the cost of working fluid. We don't have to specify the order at which it happens. The specific heat from the temperature time derivative drops out.

2) `vapor -> rain` and `rain -> vapor` should have the same value and opposite sign. 

Let's consider first transition between vapor and cloud liquid:
`dI/dt = 0 = S_evap (I_v - L_v + R_v T - I_l)`
The terms in the brackets stand from change in the nominator of `q_v` (i.e. mass change on the vapor side), temperature change, mass change on the `q_l` side.

Now for the transition between vapor and rain:
`dI/dt = S_evap (I_v -L_v + R_v T - I) = S_evap (I_l - I)`
The mass gain in `q_v` and temperature change contributions are the same as in the previous case. But now they are not balanced by the `q_l` changes. Instead there now appears an `I` term that comes from the changes in the denominators of different `q_`. But this term drops out once you compute the `d rho I /dt` derivative. 
So the same as in in the working fluid phase change we are left with `(I_v - L_v + R_v T) * S_evap = I_l * S_evap` contribution.

I'll update the DD's and also spend more text on it in the documentation of the microphysics scheme.

tapios:
> > > What you have here looks correct to me. We should update the DD's (please feel free to do it if you have a chance). The expression in the DD's only consider the situation of in-phase conversion of cloud condensate to precipitation (i.e., cloud liquid to rain, cloud ice to snow). Your formulation for rain evaporation, it seems to me, assumes a two-step process: first convert rain to cloud condensate (and incorporate the energy change in the working fluid); then evaporate that newly converted condensate (assuming no further energy transformation). So the latent cooling occurs in the working fluid, as it should.
> > 
> > 
> > Sorry, I'm still confused. Should the energy source term from rain to vapor transition be the same as the one from vapor to rain transition, just of opposite signs? If so, consider the transition from vapor to rain, and assume the latent heating occurs in the working fluid, then the energy source term would be I_v * S_evap? I think I miss something here.
> > (You can merge the PR before the question is answered.)
> 
> Thanks for thinking about those things along with me!
> 
> 1. I think it's enough to say that the energy change is always happening at the cost of working fluid. We don't have to specify the order at which it happens. The specific heat from the temperature time derivative drops out.
> 2. `vapor -> rain` and `rain -> vapor` should have the same value and opposite sign.
> 
> Let's consider first transition between vapor and cloud liquid:
> `dI/dt = 0 = S_evap (I_v - L_v + R_v T - I_l)`
> The terms in the brackets stand from change in the nominator of `q_v` (i.e. mass change on the vapor side), temperature change, mass change on the `q_l` side.
> 
> Now for the transition between vapor and rain:
> `dI/dt = S_evap (I_v -L_v + R_v T - I) = S_evap (I_l - I)`
> The mass gain in `q_v` and temperature change contributions are the same as in the previous case. But now they are not balanced by the `q_l` changes. Instead there now appears an `I` term that comes from the changes in the denominators of different `q_`. But this term drops out once you compute the `d rho I /dt` derivative.
> So the same as in in the working fluid phase change we are left with `(I_v - L_v + R_v T) * S_evap = I_l * S_evap` contribution.
> 
> I'll update the DD's and also spend more text on it in the documentation of the microphysics scheme.

Thank you both, @szy21 and @trontrytel, for your thoughts. I'll need to see this written out cleanly to be sure. If you, @trontrytel , can take a stab at updating the DD's, perhaps the three of us can talk about it next week?

trontrytel:
> 
> Thank you both, @szy21 and @trontrytel, for your thoughts. I'll need to see this written out cleanly to be sure. If you, @trontrytel , can take a stab at updating the DD's, perhaps the three of us can talk about it next week?

Sounds good!

docs/src/APIs/Atmos/AtmosModel.md
```diff
@@ -82,4 +83,7 @@ ClimateMachine.Atmos.average_density
 ```@docs
 ClimateMachine.Atmos.RemovePrecipitation
 ClimateMachine.Atmos.CreateClouds
+ClimateMachine.Atmos.Rain_1M
+
+
```
charleskawczynski, line 88:
```suggestion
```
Not sure if Documenter will like these blank spaces

charleskawczynski: Looks like it was fine.

szy21:
> I'll update the DD's and also spend more text on it in the documentation of the microphysics scheme.

Ah, I think I know what's wrong in my previous comment. Part of the internal energy of the vapor contributes to the temperature change in the working fluid. Now the only thing I don't understand is why the 1-step and 2-step derivations on your overleaf give different answers. Maybe we could talk about that next week?

PR #1708, opened 2020-11-05
change from fractional to rates

yairchn:
### Description

This draft PR changes the fractional entrainment and detrainment (ε, δ) to be written as entrainment and detrainment rates (E=ρawε , Δ=ρawδ) which should be positive definite regardless of updraft vertical velocity (w) to maintain physical sources.

### DIFF
--- a/test/Atmos/EDMF/closures/entr_detr.jl
+++ b/test/Atmos/EDMF/closures/entr_detr.jl
@@ -44,78 +44,76 @@ function entr_detr(
     N_up = n_updrafts(m.turbconv)
     ρ_inv = 1 / gm.ρ
     a_up_i = up[i].ρa * ρ_inv
+    if a_up_i<FT(0)#m.turbconv.subdomains.a_min
+        E_dyn = FT(0)
+        Δ_dyn = FT(0)
+        E_trb = FT(0)
+    else
+        lim_E = entr.lim_ϵ
+        lim_amp = entr.lim_amp
+        w_min = entr.w_min
+        # precompute vars
+        w_up_i = up[i].ρaw / up[i].ρa
+        sqrt_tke = sqrt(max(en.ρatke, 0) * ρ_inv / env.a)
+        # ensure far from zero
+        Δw = filter_w(w_up_i - env.w, w_min)
+        w_up_i = filter_w(w_up_i, w_min)
+        Δb = up_aux[i].buoyancy - en_aux.buoyancy
+        D_E, D_δ, M_δ, M_E =
+            nondimensional_exchange_functions(m, entr, state, aux, t, ts, env, i)
 
-    lim_ϵ = entr.lim_ϵ
-    lim_amp = entr.lim_amp
-    w_min = entr.w_min
-    # precompute vars
-    w_up_i = up[i].ρaw / up[i].ρa
-    sqrt_tke = sqrt(max(en.ρatke, 0) * ρ_inv / env.a)
-    # ensure far from zero
-    Δw = filter_w(w_up_i - env.w, w_min)
-    w_up_i = filter_w(w_up_i, w_min)
+        # I am commenting this out for now, to make sure there is no slowdown here
+        Λ_w = abs(Δb / Δw)
+        Λ_tke = entr.c_λ * abs(Δb / (max(en.ρatke * ρ_inv, 0) + w_min))
+        λ = lamb_smooth_minimum(
+            SVector(Λ_w, Λ_tke),
+            m.turbconv.mix_len.smin_ub,
+            m.turbconv.mix_len.smin_rm,
+        )
 
-    Δb = up_aux[i].buoyancy - en_aux.buoyancy
+        # compute entrainment/detrainment components
+        E_trb =
+            2 * up[i].ρa * entr.c_t * sqrt_tke /
+            max(m.turbconv.pressure.H_up, entr.εt_min)
+        E_dyn = up[i].ρa*λ*(D_E + M_E)
+        Δ_dyn = up[i].ρa*λ*(D_δ + M_δ)
 
-    D_ε, D_δ, M_δ, M_ε =
-        nondimensional_exchange_functions(m, entr, state, aux, t, ts, env, i)
-
-    # I am commenting this out for now, to make sure there is no slowdown here
-    Λ_w = abs(Δb / Δw)
-    Λ_tke = entr.c_λ * abs(Δb / (max(en.ρatke * ρ_inv, 0) + w_min))
-    λ = lamb_smooth_minimum(
-        SVector(Λ_w, Λ_tke),
-        m.turbconv.mix_len.smin_ub,
-        m.turbconv.mix_len.smin_rm,
-    )
-
-    # compute limiters
-    εt_lim = εt_limiter(w_up_i, lim_ϵ, lim_amp)
-    ε_lim = ε_limiter(a_up_i, lim_ϵ, lim_amp)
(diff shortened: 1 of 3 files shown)


test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -91,9 +91,9 @@ function mixing_length(
     w_up = vuntuple(i -> up[i].ρaw / up[i].ρa, N_up)
     b = sum(
         ntuple(N_up) do i
-            a_up[i] * w_up[i] * δ[i] / env.a *
+            a_up[i] * w_up[i] * Δ[i]/max(up[i].ρaw,eps(FT)) / env.a *
```
yairchn, line 94: we need to decide how to better handle the case of Δ=ρawδ=0 when w=0. (w=0 is an allowed value, in contrast to a=0 which should be circumvented elsewhere in the code).

ilopezgp:
@yairchn do you have an idea about why this fails the regression tests? From a theoretical point of view, an implementation of E or ε should not matter, only in limiting cases where either a or w are zero.

test/Atmos/EDMF/closures/entr_detr.jl
```diff
@@ -44,78 +44,43 @@ function entr_detr(
...
+            SVector(Λ_w, Λ_tke),
+            m.turbconv.mix_len.smin_ub,
+            m.turbconv.mix_len.smin_rm,
+        )
 
-    Δb = up_aux[i].buoyancy - en_aux.buoyancy
+        # compute entrainment/detrainment components
+        E_trb =
+            2 * up[i].ρa * entr.c_t * sqrt_tke /
+            max(m.turbconv.pressure.H_up, entr.εt_min)
```
ilopezgp, line 77: This is something I don't understand fully either in the previous or current formulation. Why is there a `entr.εt_min`, but it does not have units of `εt`? Is this the minimum updraft height? If so, we should definitely change the name of this variable.

test/Atmos/EDMF/closures/mixing_length.jl
ilopezgp: In the entr_detr function, we could return both E_dyn, Δ_dyn, E_trb and ε_dyn, δ_dyn, ε_trb (or a subset of that). That way, we can deal with 0/0 problems at the source of the problem, and pass the fractional rates here.

ilopezgp, line 94: Should be: ` Δ[i] / gm.ρ/ env.a *`

test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -91,9 +91,9 @@ function mixing_length(
...
             ((w_up[i] - env.w) * (w_up[i] - env.w) / 2 - tke_en) -
-            a_up[i] * w_up[i] * (w_up[i] - env.w) * εt[i] * env.w / env.a
+            a_up[i] * w_up[i] * (w_up[i] - env.w) * Et[i]/max(up[i].ρaw,eps(FT)) * env.w / env.a
```
ilopezgp, line 96: Same here, you are multiplying and dividing by `a_up[i] * w_up[i]`

ilopezgp (changes requested):

test/Atmos/EDMF/closures/entr_detr.jl
```diff
@@ -44,78 +44,43 @@ function entr_detr(
     N_up = n_updrafts(m.turbconv)
     ρ_inv = 1 / gm.ρ
     a_up_i = up[i].ρa * ρ_inv
+    if a_up_i<FT(0)#m.turbconv.subdomains.a_min
+        E_dyn = FT(0)
```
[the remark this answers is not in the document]
yairchn: its a good question. the problem is that the filters that prevent negative area are not in place in master, so I was feeling that I should include it for now to remove once all pieces are in master

test/Atmos/EDMF/closures/entr_detr.jl
```diff
@@ -44,78 +44,43 @@ function entr_detr(
...
-entrainment needed to ensure positiveness
-of the updraft vertical velocity, given
- - `w_up_i`, the updraft vertical velocity
- - `ϵ`, a minimum threshold value
-"""
-εt_limiter(w_up_i::FT, ϵ::FT, lim_amp::FT) where {FT} =
-    FT(1) + lim_amp * exp(-w_up_i^2 / (2 * ϵ))
+        E_dyn = min(max(E_dyn, FT(0)), FT(1))
+        Δ_dyn = min(max(Δ_dyn, FT(0)), FT(1))
+        E_trb = min(max(E_trb, FT(0)), FT(1))
```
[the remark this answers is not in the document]
yairchn: The ideal would be only to take " E_dyn = max(E_dyn, FT(0))"
to limit to non negative values. But I left it as is for now. what do you think?

test/Atmos/EDMF/closures/entr_detr.jl
yairchn: it should be changed I will make a PR to change the name of this parameter

test/Atmos/EDMF/closures/mixing_length.jl
yairchn: by definition  δ[I]=Δ[i]/up[i].ρaw

test/Atmos/EDMF/closures/mixing_length.jl
yairchn: by definition εt[I]=Et[I]/up[i].ρaw

test/Atmos/EDMF/closures/entr_detr.jl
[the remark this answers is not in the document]
ilopezgp: I would take out the min condition... We did not have it in SCAMPy and it worked like this

test/Atmos/EDMF/closures/entr_detr.jl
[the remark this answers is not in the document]
ilopezgp: This is something worth checking with @charleskawczynski.

test/Atmos/EDMF/closures/mixing_length.jl
ilopezgp: I know, but you have ` a_up[i] * w_up[I]` in front, which was used because `up[i].ρaw*δ[I]=Δ[i].` If you look at equation (19) in my paper you will see that the strange I is b: The formulation is valid in term of Δ, we do not need updraft area fraction or vertical velocity in the denominator.

test/Atmos/EDMF/closures/mixing_length.jl
ilopezgp: Same as above, if you look at equation (19) you will see that we do not need it.

test/Atmos/EDMF/closures/entr_detr.jl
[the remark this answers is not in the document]
charleskawczynski: The filter would disallow this. We can include the filter in a callback in the driver in this PR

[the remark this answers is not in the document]
yairchn: in that case I will remove. this if command

test/Atmos/EDMF/closures/entr_detr.jl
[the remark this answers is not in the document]
yairchn: ok removed

ilopezgp (approved):

ilopezgp:
LGTM, just need to fix the formatter.

PR #1711, opened 2020-11-05
Fix the ordering of the state in the AtmosModel

### LINKED ISSUE #1494: Implicit time marching method question
# Description

I found some behavior I've been struggling to explain recently, and I wanted to open an issue to track it / allow others to take a look. On [this branch](https://github.com/CliMA/ClimateMachine.jl/compare/ck/EDMFImplicitTest), running

```julia
include("test/Atmos/EDMF/bomex_edmf_demo.jl")
```
runs a matrix of tests:

 - Solver type
   - Explicit timestepper
   - Implicit timestepper

 - `turbconv` submodel prognostic initialization
   - Initialize with 0s
   - Initialize with `NaN`s

What I've found is that the dycore, which receives no fluxes / sources from the `turbconv` model, successfully runs to completion for all cases _except_ with the implicit stepper and `NaN` initialization. The issue I see with this is that, unless I'm somehow misusing the implicit stepper (which I very well may be), the dycore should not see the turbconv model no matter what time stepper is being used. When the dycore breaks, you'll see failed saturation adjustment in the dycore (in `src/Atmos/Model/moisture.jl`).

Thanks to @jkozdon and @thomasgibson for the discussion on this, and suggestions on boiling things down--I'll continue trying to simplify this branch down further, but it's pretty simple as is (all compute kernels are empty).

## Additional context

# For CLIMA Developers

jkozdon:
The AtmosLinearModel must use a contiguous subset of the AtmosModel,
since `turbconv` was not in the `AtmosLinearModel` it needs to be
moved later in the state list.

closes #1494 

co-authored-by: Charles Kawczynski <kawczynski.charles@gmail.com>

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -237,16 +237,23 @@ end
     vars_state(m::AtmosModel, ::Prognostic, FT)
 
 Conserved state variables (prognostic variables).
+
+!!! warning
+
+    The order of the fields for `AtmosModel` needs to match the one for
+    `AtmosLinearModel` since a shared state is used
 """
 function vars_state(m::AtmosModel, st::Prognostic, FT)
     @vars begin
+        # start of inclusion in `AtmosLinearModel`
         ρ::FT
         ρu::SVector{3, FT}
         ρe::FT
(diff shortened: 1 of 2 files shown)


charleskawczynski (approved):
🚀🚀🚀

PR #1713, opened 2020-11-06
Fix bugs in the preconditioner construction step in the fully implicit solver

Zhengyu-Huang:
### Description

The vertical discretization might depend on the time and parameters(which is the case for EDMF test), but the parameter and time are set to be nothing and NaN when we construct the preconditioner


thomasgibson (approved):

Zhengyu-Huang:
bor r+

PR #1703, opened 2020-11-05
Reorganizes OceanProblems module and adds InitialValueProblem

glwagner:
### Description

This PR adds `InitialValueProblem <: AbstractSimpleBoxProblem` and subsidiary type `InitialConditions`.

`InitialConditions` is a container for initial conditions functions of `x, y, z`.

Usage looks like:

```julia
a = 0.01 # m
L = 1e5 # m

ηᵢ(x, y, z) = a * exp(-x^2 / 2L^2)

initial_conditions = InitialConditions(η=ηᵢ)

problem = InitialValueProblem(dimensions = (Lx, Ly, Lz), initial_conditions = initial_conditions)
```

`problem` is then passed to `HydrostaticBoussinesqModel`, which is one of the `BalanceLaw`s implemented in `ClimateMachine.Ocean`.

`InitialValueProblem` is a convenience type intended to be a temporary addition to the code that make it slightly easier to write simple tutorials and facilitate experimentation. It may be refactored in the near future, along with much of the interface associated with `HydrostaticBoussinesqModel`.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1716, opened 2020-11-06
Experimental CartesianField abstractions for plotting and simple analysis, plus two new Ocean tutorials

glwagner:
### Description

This PR adds an experimental `CartesianDomain` and `CartesianField` abstraction, plus two new ocean tutorials.

Eventually we would use `CartesianDomain` to idealized rectangular domains with both regular and stretched elements, and `CartesianDomain` will manage the construction of the `DiscontinuousSpectralElementGrid`.

However, in the spirit of incremental changes, the current constructor for `CartesianDomain` does the opposite, by inverting the data in `DiscontinuousSpectralElementGrid` assuming an underlying Cartesian structure to build a Cartesian abstraction.

Usage is illustrated in the tutorials, which produce these animations:

geostrophic_adjustment

radiating_eddy

# `CartesianDomain` usage

Currently we can't build a `DiscontinuousSpectralElementGrid` all that easily outside of the `DriverConfiguration` / `SolverConfiguration` interface. So my example asumes one has a `SolverConfiguration` object in hand. Then we'd write

```julia
julia> domain = CartesianDomain(solver_configuration.dg.grid, Ne)
CartesianDomain{Float64, ClimateMachine.Mesh.Grids.DiscontinuousSpectralElementGrid}:
    Np = 4, Ne = (x = 100, y = 1, z = 1)
    x = (0.00e+00, 1.00e+06), y = (0.00e+00, 1.00e+06), z = (-4.00e+02, 0.00e+00)
    Lx = 1.00e+06, Ly = 1.00e+06, Lz = 4.00e+02
```

where `Ne` is a tuple corresponding to the number of elements in each Cartesian direction `x, y, z`.

# `CartesianField` usage

`CartesianField` provides a Cartesian `view` into data in `MPIStateArray`. The current interface (which hopefully will be improved) is

```
julia> u = CartesianField(domain, solver_configuration.Q, 1);
```

where `domain::CartesianDomain` and `1` is the index of the field in `solver_configuration.solver.Q.realdata`. `CartesianField` keeps track of `CartesianField.domain` and `CartesianField.elements` is an `Array{RectangularElement, 3}`:

```julia
julia> u.elements
100×1×1 Array{ClimateMachine.Ocean.CartesianDomains.RectangularElement... # goes on
```

which is indexed into via

```julia
julia> u[1, 1, 1]
RectangularElement{SubArray} with data ∈ [-1.29e-05, 7.50e-06]
    x ∈ [0.00e+00, 1.00e+04]
    y ∈ [0.00e+00, 1.00e+06]
    z ∈ [-4.00e+02, 0.00e+00]
```

which have properties `x, y, z, data`:

```julia
julia> u[1, 1, 1].x
5-element view(::Array{Float64,4}, :, 1, 1, 1) with eltype Float64:
     0.0
  1726.731646460114
  5000.0
  8273.268353539885
 10000.0

julia> u[1, 1, 1].data
5×5×5 view(reshape(view(::Array{Float64,3}, :, 1, 1:100), 5, 5, 5, 100), :, :, :, 1) with eltype Float64:
# much output...
```

Note that `CartesianField` does not assume ~~anything~~ too much about the structure of `MPIStateArray.realdata` (except that data is stored in an array with indices `(node, field, element)`), but uses a linear coordinate to sort elements:

https://github.com/CliMA/ClimateMachine.jl/blob/c7a48d20604b6ef0148aec0b273126471d58fd53/src/Ocean/CartesianDomains/cartesian_field.jl#L57-L75

However, we do assume that all coordinates are 1D, despite that in the underlying `DiscontinuousSpectralElement` grid the nodal coordinates can differ by factors close to machine precision.

# Key function: `assemble`

Using this abstraction we are `assemble` the elements of a field into 

https://github.com/CliMA/ClimateMachine.jl/blob/860a801baff7edd0fdcc9328a8374b217ca7c8b7/src/Ocean/CartesianDomains/rectangular_element.jl#L88-L103

which in the wild looks like

```julia
julia> assembly = assemble(u)
RectangularElement{Array} with data ∈ [-1.93e-03, 1.82e-03]
    x ∈ [0.00e+00, 1.00e+06]
    y ∈ [0.00e+00, 1.00e+06]
    z ∈ [-4.00e+02, 0.00e+00]
```

which, crucially, averages data on shared nodes and returns a giant `RectangularElement` with all the data. This is necessary for creating visualizations using tools like `Plots.contourf`, for example.

When adapting this object for `CuArray`, we will have to provide an option specifying whether to transfer data before `assemble`ing or to perform all computations on the GPU.  @christophernhill has mentioned we would also like a function that does the same thing but avoids averaging. Perhaps extending `cat` for `RectangularElement`s will work for this.

This abstraction may require more development to work with memory-distributed problems. I don't think its crucial that this abstraction works for memory-distributed problems, because its mostly useful for quick visualization needed for fast turnover while designing experiments, producing tutorials, and post-processing data.

If the tutorials are to run without external dependencies, we'd have to depend on `Plots`.

Note: the tutorials in this PR depend on #1703 .


src/Ocean/CartesianDomains/cartesian_field.jl
```diff
@@ -0,0 +1,128 @@
...
+#####
+
+struct K
+    elem :: Int
+    node :: Int
+end
+
+using Plots
+
+import Plots: contourf
```
charleskawczynski, line 99: We don't have `Plots` in ClimateMachine's dependencies. `Plots` is part of the docs environment, maybe we can move this to `docs/plothelpers.jl`?

src/Ocean/CartesianDomains/CartesianDomains.jl
```diff
@@ -0,0 +1,9 @@
+module CartesianDomains
```
charleskawczynski, line 1: I'm wondering if this belongs in `src/Common/` in case other groups may find it useful.

src/Ocean/CartesianDomains/cartesian_field.jl
glwagner: Ah yes, very WIP! This does not belong here; I'm going to remove the recipe entirely. It's not necessary since we have `glue` anyways. We can add it back if need be.

src/Ocean/CartesianDomains/CartesianDomains.jl
glwagner: Perhaps they would --- that's more up to you (and others) than me. This functionality is still a bit experimental but I wouldn't mind if others wanted to try it out and gave feedback.

glwagner:
Hmm, I'm noticing that `glue` is fundamentally flawed and won't work as-is on the GPU. We need to create a data structure that represents the data movement / averaging, allocate memory for the result, and then launch a kernel to perform the computation. It might be appropriate to leave that improvement for a future PR.

For the near future, we can just hack a solution that transfers the input array of `elements` to the CPU before `glue`ing.

glwagner:
For reference, the same problem in `Oceananigans.jl` (which uses the rigid lid approximation and therefore has an infinite gravity wave speed) looks like

radiating_eddy

jkozdon:
So what you are calling the "glue" is commonly referred to as the "direct stiffness summation" in the continuous Galerkin (at least the averaging part). On warped domains, to get it "right" will require a bit more work, but I think that averaging ends up being the right thing when you have uniform, square elements.

Also, this is the line
https://github.com/CliMA/ClimateMachine.jl/blob/97d0e51be28483a463ceb544511bd3083525e1dc/src/Numerics/Mesh/Topologies.jl#L392
that introduces the space filling curve partitioning, and without it a cartesian mesh will be more regularly partitioned (you wouldn't want to do this for a parallel run, but would be fine for a single rank run). If you do not partition, the elements will have a more standard Cartesian layout; see here
https://github.com/CliMA/ClimateMachine.jl/blob/97d0e51be28483a463ceb544511bd3083525e1dc/src/Numerics/Mesh/BrickMesh.jl#L157-L272
Note, to get this to work, you will want to add `[]` to the last argument of the returned tuple here
https://github.com/CliMA/ClimateMachine.jl/blob/97d0e51be28483a463ceb544511bd3083525e1dc/src/Numerics/Mesh/BrickMesh.jl#L346

glwagner:
> On warped domains, to get it "right" will require a bit more work, but I think that averaging ends up being the right thing when you have uniform, square elements.

That will be exciting when we are ready to develop a similar abstraction for warped domains! I will open an issue independent from this PR so we can discuss that.

glwagner:
> So what you are calling the "glue" is commonly referred to as the "direct stiffness summation"

I'm not attached to `glue` (I changed it from `join` to `glue` this morning. I am willing to change the name of this function to `direct_stiffness_summation`. However, I'm worried that this name is too jargon-y and not descriptive; I don't quite understand why what we are doing should be described as a "summation".

A key aspect of this operation is to "flatten" the data --- eg, data that's expressed hierarchically as a 3D array of elements, each of which has a 3D array of nodal data, is flattened into a single 3D data structure. In addition to flattening, `glue` also averages data on overlapping nodes (that's the "summation"?). @jkozdon, is there a name that is closer to the technical description you've mentioned, but also descriptive for people that don't know what "direct stiffness summation" is? 

I'm happy to read a reference if you have one to help come up with a descriptive but technically accurate name for this crucial function.

glwagner:
> that averaging ends up being the right thing when you have uniform, square elements.

This was merely intended to be an aid for visualization but I'm glad that there is a technical basis for averaging shared nodes. The best abstractions are both simple and technically correct.

glwagner:
Hmm, possibly we could call this function `assemble` with an option `merge=true`. So if we write

```julia
assembled_u = assemble(u.elements)
```

we obtained a flattened data structure that contains shared notes (essentially a concatenation), and if we use

```julia
assembled_u = assemble(u.elements, merge=true)
```

we obtain the behavior of what's called `glue` in this PR.

This seems to reflect language that's used on Wikipedia:

https://en.wikipedia.org/wiki/Direct_stiffness_method#Assembly

but of course I'll defer to experts...

glwagner:
> averaging ends up being the right thing when you have uniform, square elements.

The elements are _rectangular_ rather than square --- hopefully it's still ok.

If we develop this further, I can write documentation that explains the theory a la continuous Galerkin approximations.

christophernhill:
@glwagner what do you think about using  LS3NRK33Heuns  in the examples? That is what we are using for 2d:3d split explicit. The LSRK144NiegemannDiehlBusch has some better CFL properties, but not sure they are significant in these experiments. For these problems LS3NRK33Heuns may be quite a bit faster without significant loss in solution quality.

jkozdon:
Not sure I have a better name. Also, in hind sight I don’t know how universal DSS is. I think the idea originates with Strang, but not 100% (Paul Fischer’s work is where I know the term from, referring to direct summation of the stiffness in continuous Galerkin methods).

christophernhill:
@glwagner should we have some ```StateCheck``` ( https://clima.github.io/ClimateMachine.jl/latest/generated/Diagnostics/Debug/StateCheck/ ) stuff in the tutorials. It might be good to have some way to test if they change, and also show how to get a few basic statistics?  What do you think? I can add code if you think it makes sense?

glwagner:
@christophernhill I'm worried that `StateCheck` is too verbose for the tutorials --- it will produce voluminous output that will make the literate files difficult to read. You'll note that I have written a more concise kind of state checker for the tutorial that just measures the magnitude of a single field (the surface displacement in this case) so that I know if the simulation NaN'd or not. Do you think that's ok?

I think tests for `StateCheck` itself might be the best way to ensure that it doesn't change?

PS how do I merge PRs? I know there's some magic incantation that I must whisper...

glwagner:
bors try

christophernhill:
bors r+

merges

christophernhill (approved):

charleskawczynski (commented):
We need to add the tutorials to `docs/list_of_tutorials.jl` so that they run and show up in the docs. It would be nice to review the previews of these newly added tutorials.

test/Ocean/runtests.jl
```diff
@@ -1,7 +1,9 @@
 using MPI, Test
+
 include("../testhelpers.jl")
 
 @testset "Ocean" begin
     runmpi(joinpath(@__DIR__, "HydrostaticBoussinesq/test_ocean_gyre_short.jl"))
     runmpi(joinpath(@__DIR__, "SplitExplicit/test_spindown_short.jl"))
+    include("OceanProblems/test_initial_value_problem.jl")
```
charleskawczynski, line 8:
```suggestion
    include(joinpath("OceanProblems","test_initial_value_problem.jl"))
```

src/Ocean/CartesianDomains/cartesian_domain.jl
```diff
@@ -0,0 +1,78 @@
...
+
+    # Check number of elements
+    prod(Ne) === size(volume_geometry, 3) ||
+    error("prod(Ne) must match the total number of grid elements.")
+
+    Np = polynomialorder(grid)
+
+    x = view(volume_geometry, :, 13, :)
+    y = view(volume_geometry, :, 14, :)
+    z = view(volume_geometry, :, 15, :)
```
charleskawczynski, line 67: Should we be grabbing these indexes from `Grids`?

src/Ocean/CartesianDomains/cartesian_field.jl
```diff
@@ -0,0 +1,109 @@
...
+    volume_geometry = domain.grid.vgeo
+
+    Ne = domain.Ne
+    Te = prod(domain.Ne)
+    Np = domain.Np
+
+    # Extract coordinate arrays with size (xnode, ynode, znode, element)
+    x = reshape(volume_geometry[:, 13, :], Np + 1, Np + 1, Np + 1, Te)
+    y = reshape(volume_geometry[:, 14, :], Np + 1, Np + 1, Np + 1, Te)
+    z = reshape(volume_geometry[:, 15, :], Np + 1, Np + 1, Np + 1, Te)
```
charleskawczynski, line 71: Can we grab these indexes from `Grids`?

src/Ocean/OceanProblems/initial_value_problem.jl
```diff
@@ -0,0 +1,99 @@
...
+
+Returns an `InitialValueProblem` with `dimensions = (Lˣ, Lʸ, H)`, `initial_conditions`,
+and `boundary_conditions`.
+
+The default `initial_conditions` are resting with no temperature perturbation;
+the default `boundary_conditions` are horizontally-periodic with `Insulating()`
+and `FreeSlip()` conditions at the top and bottom.
+"""
+function InitialValueProblem(
+    FT = Float64;
```
charleskawczynski, line 26: Can we make `FT` non-optional. This could hide type instabilities.

src/Ocean/OceanProblems/SimpleBoxProblem.jl
```diff
@@ -1,435 +0,0 @@
-module OceanProblems
```
charleskawczynski, line 1: Does it make sense to rename this file instead of deleting?

src/Ocean/CartesianDomains/cartesian_field.jl
christophernhill: @charleskawczynski and @jkozdon do you have an opinion which is the authoritative variable out of ```_x1``` and ```x1id```?  ```src/Numerics/Mesh/Grids.jl``` has a loop for ```_``` prefix things and then defines some suffix ```id``` variables . I think an originalist view could be that the framers of the original code meant x1id to be authoritative. But _x1 and x1id seem to get used somewhat equally in under more recent rulings.  The original framers in their wisdom do not seem to have left too much guidance in their writings that I could find. 

```
const _nvgeo = 16
const _ξ1x1,
_ξ2x1,
_ξ3x1,
_ξ1x2,
_ξ2x2,
_ξ3x2,
_ξ1x3,
_ξ2x3,
_ξ3x3,
_M,
_MI,
_MH,
_x1,
_x2,
_x3,
_JcV = 1:_nvgeo
const vgeoid = (
    ξ1x1id = _ξ1x1,
    ξ2x1id = _ξ2x1,
    ξ3x1id = _ξ3x1,
    ξ1x2id = _ξ1x2,
    ξ2x2id = _ξ2x2,
    ξ3x2id = _ξ3x2,
    ξ1x3id = _ξ1x3,
    ξ2x3id = _ξ2x3,
    ξ3x3id = _ξ3x3,
    Mid = _M,
    MIid = _MI,
    MHid = _MH,
    x1id = _x1,
    x2id = _x2,
    x3id = _x3,
    JcVid = _JcV,
)
```

test/Numerics/DGMethods/horizontal_integral_test.jl uses ClimateMachine.Mesh.Grids.vgeoid.x1id.

test/Diagnostics/diagnostic_fields_test.jl uses import ClimateMachine.Mesh.Grids: _x1, _x2, _x3

charleskawczynski: I'm not sure. @jkozdon?

src/Ocean/OceanProblems/initial_value_problem.jl
glwagner: I can do that here. I am submitting a PR soon that may relieve the user from concerns like these by introducing a single location for specifying the type of participants in the Ocean's `BalanceLaw`, however. If the design proposed by that PR is accepted, `InitialValueProblem` may eventually become obsolete.

src/Ocean/CartesianDomains/cartesian_field.jl
glwagner: If I may, I would prefer to raise an issue to address this, given its experimental nature. I am going to submit a PR soon that may render this function obsolete.

src/Ocean/CartesianDomains/cartesian_domain.jl
glwagner: Perhaps. But actually, a better design would use the `CartesianDomain` constructor to construct the grid itself rather than analyzing it after it has been created. This the pattern that I'm working on now and will submit a PR for soon.

src/Ocean/OceanProblems/SimpleBoxProblem.jl
glwagner: I think I did do this --- I created a new eponymous `OceanProblems.jl` to correspond to the `OceanProblems` module (which I understand is convention --- though it is not followed in `ClimateMachine`?) I moved the "simple box" code into it's own file, `simple_box_problem.jl`.

src/Ocean/CartesianDomains/cartesian_domain.jl
glwagner: I'll update this function anyways in the spirit of a cleaner PR.

src/Ocean/CartesianDomains/cartesian_field.jl
glwagner: I made the requisite changes to resolve this conversation, but I'll wait for @jkozdon's input.

glwagner:
> We need to add the tutorials to `docs/list_of_tutorials.jl` so that they run and show up in the docs.

Okay, I've done that.

src/Ocean/OceanProblems/initial_value_problem.jl
glwagner: @charleskawczynski the type here is probably unnecessary because they only specify the type of the domain limits. The domain limits are used here:

https://github.com/CliMA/ClimateMachine.jl/blob/4a3f08659b23e5be5d0a83ce72ad8c33df824ea2/src/Driver/driver_configs.jl#L309-L327

and it looks like a variable called `FloatType` is passed to `DiscontinuousSpectralElementGrid`. Is `FloatType` respected regardless of the eltype of `brickrange`?

If it doesn't affect type instability, is it okay to relieve the user of the need to explicitly type this object?

glwagner: I made the changes anyways.

kpamnany:
> This abstraction may require more development to work with memory-distributed problems. I don't think its crucial that this abstraction works for memory-distributed problems, because its mostly useful for quick visualization needed for fast turnover while designing experiments, producing tutorials, and post-processing data.

I'm uncertain about this -- I think we've avoided introducing abstractions that do not extend to multi-process. My concern is that `assemble` _cannot_ be so extended and it is better to do the work now to figure out a multi-process friendly abstraction.

glwagner:
> I'm uncertain about this -- I think we've avoided introducing abstractions that do not extend to multi-process. My concern is that `assemble` _cannot_ be so extended and it is better to do the work now to figure out a multi-process friendly abstraction.

It may be important to distinguish between abstractions that are used in core numerical algorithms, and abstractions that improve UX / are useful for post-processing. I agree that we can't rely on non-parallelizable abstractions for core numerical algorithms.

`CartesianField` is not designed or intended to be used in core algorithms; it's a tool for plotting and analyzing data. Since plotting is usually single node I think restricting its use to those cases is ok?

kpamnany:
> it's a tool for plotting and analyzing data

If you clearly specify it as such, I'm okay with it -- my concern arose from the suggestion that you might want to extend it to multi-process.

glwagner:
> If you clearly specify it as such, I'm okay with it -- my concern arose from the suggestion that you might want to extend it to multi-process.

I see why that's confusing! To clarify, the advantage of having a multi-process-capable `CartesianField` lies in the (potential, future) possibility of doing doing distributed post-processing analysis on ClimateMachine data. This is not a necessary feature for us --- just a pretty plum to pick that @christophernhill might benefit from (who has a knack for creating absurdly large data sets that can't be analyzed on a single node).

From where I sit it looks like the core numerics is pretty powerful. What this PR is working towards are tools that make working with the data a bit easier. A key requirement of this development is that it does not touch the core numerics; it just works with what's provided.

src/Ocean/CartesianDomains/cartesian_field.jl
jkozdon: I think that we mostly use the underscore versions in the kernels.

A lot of this code was written when @lcw and I were just learning Julia, trying out different ideas, etc.. I don't remember for sure, but maybe we originally planned on using the `vgeoid` everywhere, but this didn't work with our initial GPU code so we fell back to the underscore. But I really don't know for sure!

We should probably just use one and delete the other.

charleskawczynski:
I definitely like the idea of adding more tools specifically for visualization of small/simple models

glwagner:
Do I need another review for this PR? @charleskawczynski

christophernhill:
> Do I need another review for this PR? @charleskawczynski

@charleskawczynski it would be great to get this merged, so we can move to some next steps on ocean-dev. Apart from squashing is there anything else that looks like it is needed?

glwagner:
@christophernhill what is there to squash? It looks like there's only 1 commit.

christophernhill:
> @christophernhill what is there to squash? It looks like there's only 1 commit.

Cool - sorry that was just a reflex response, I didn't look at the commits! @charleskawczynski anything else on your list?

src/Ocean/CartesianDomains/cartesian_field.jl
glwagner: Sounds like cleaning up the ambiguity between `x1id` and `_x1` deserves a new issue? Since I addressed the original question I'm going to resolve this conversation.

PR #1719, opened 2020-11-06
Add stable_bl_edmf_implicit.jl example, which uses fully implicit solver to solve EDMF

Zhengyu-Huang:
This is an example which uses fully implicit solver to solve EDMF(stable boundary layer) for 6h within 15 mins.
This example only demonstrates that we can use fully implicit solver here, but I have not visualized/compared the results

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -237,16 +237,23 @@ end
     vars_state(m::AtmosModel, ::Prognostic, FT)
 
 Conserved state variables (prognostic variables).
+
+!!! warning
+
+    The order of the fields for `AtmosModel` needs to match the one for
+    `AtmosLinearModel` since a shared state is used
 """
 function vars_state(m::AtmosModel, st::Prognostic, FT)
     @vars begin
+        # start of inclusion in `AtmosLinearModel`
         ρ::FT
         ρu::SVector{3, FT}
         ρe::FT
         turbulence::vars_state(m.turbulence, st, FT)
-        turbconv::vars_state(m.turbconv, st, FT)
         hyperdiffusion::vars_state(m.hyperdiffusion, st, FT)
         moisture::vars_state(m.moisture, st, FT)
+        # end of inclusion in `AtmosLinearModel`
+        turbconv::vars_state(m.turbconv, st, FT)
         radiation::vars_state(m.radiation, st, FT)
         tracers::vars_state(m.tracers, st, FT)
     end
--- a/src/Numerics/SystemSolvers/SystemSolvers.jl
+++ b/src/Numerics/SystemSolvers/SystemSolvers.jl
@@ -124,7 +124,7 @@ function nonlinearsolve!(
         update_Q!(jvp!, Q, args...)
 
         # update preconditioner based on finite difference, with jvp!
-        preconditioner_update!(jvp!, rhs!.f!, preconditioner, nothing, FT(NaN))
+        preconditioner_update!(jvp!, rhs!.f!, preconditioner, args...)
 
         # do newton iteration with Q^{n+1} = Q^{n} - dF/dQ(Q^n)⁻¹ (rhs!(Q) - Qrhs)
         residual_norm, linear_iterations = donewtoniteration!(
@@ -136,7 +136,7 @@ function nonlinearsolve!(
             solver,
             args...,
         )
-        @info "Linear solver converged in $linear_iterations iterations"
+        # @info "Linear solver converged in $linear_iterations iterations"
         iters += 1
 
         preconditioner_counter_update!(preconditioner)
@@ -151,7 +151,7 @@ function nonlinearsolve!(
         # ||Delta Q|| / ||Q|| ?
         relresidual = residual_norm / initial_residual_norm
         if relresidual < tol || residual_norm < tol
-            @info "Newton converged in $iters iterations!"
+            # @info "Newton converged in $iters iterations!"
             converged = true
         end
     end
--- a/test/Atmos/EDMF/stable_bl_edmf_implicit.jl
+++ b/test/Atmos/EDMF/stable_bl_edmf_implicit.jl
@@ -1,0 +1,310 @@
+using ClimateMachine
+using ClimateMachine.SystemSolvers
+using ClimateMachine.ODESolvers
+using ClimateMachine.MPIStateArrays
+
+ClimateMachine.init(;
+    parse_clargs = true,
+    output_dir = get(ENV, "CLIMATEMACHINE_SETTINGS_OUTPUT_DIR", "output"),
+    fix_rng_seed = true,
+)
+using ClimateMachine.SingleStackUtils
+using ClimateMachine.Checkpoint
+using ClimateMachine.BalanceLaws: vars_state
+const clima_dir = dirname(dirname(pathof(ClimateMachine)));
+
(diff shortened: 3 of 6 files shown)


ilopezgp:
This test should be added to the `.buildkite/pipeline.yml`.

ilopezgp (approved):
LGTM, I think the test has to be buildkite.

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
```diff
@@ -0,0 +1,316 @@
...
+        end
+
+    cb_check_cons = GenericCallbacks.EveryXSimulationSteps(3000) do
+        Q = solver_config.Q
+        δρ = (sum(Q.ρ .* M) - Σρ₀) / Σρ₀
+        δρe = (sum(Q.ρe .* M) .- Σρe₀) ./ Σρe₀
+        @show (abs(δρ))
+        @show (abs(δρe))
+        @warn (abs(δρ) <= 0.001)
+        @warn (abs(δρe) <= 0.0025)
```
charleskawczynski, line 288: Shouldn't these be `@show` s?

src/Numerics/SystemSolvers/SystemSolvers.jl
```diff
@@ -151,7 +151,7 @@ function nonlinearsolve!(
         # ||Delta Q|| / ||Q|| ?
         relresidual = residual_norm / initial_residual_norm
         if relresidual < tol || residual_norm < tol
-            @info "Newton converged in $iters iterations!"
+            # @info "Newton converged in $iters iterations!"
```
charleskawczynski, line 154: Should we just delete these `@info`s?

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
```diff
@@ -0,0 +1,316 @@
...
+            FT(1 - z / 250.0) *
+            FT(1 - z / 250.0) *
+            FT(1 - z / 250.0)
+        en.ρaθ_liq_cv =
+            gm.ρ *
+            FT(0.4) *
+            FT(1 - z / 250.0) *
+            FT(1 - z / 250.0) *
+            FT(1 - z / 250.0)
+    else
```
charleskawczynski, line 79:
```suggestion
    if z <= FT(250)
        coeff = gm.ρ *
            FT(0.4) *
            FT(1 - z / 250.0) *
            FT(1 - z / 250.0) *
            FT(1 - z / 250.0)
        en.ρatke = coeff
        en.ρaθ_liq_cv = coeff
    else
```

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
Zhengyu-Huang: I am not sure about this, originally these are error(), I do not know the purpose @ilopezgp

src/Numerics/SystemSolvers/SystemSolvers.jl
Zhengyu-Huang: Still plan to leave these commented out info here, and we need to use them to help tune parameters for new cases

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
```diff
@@ -0,0 +1,316 @@
...
+    Q = solver_config.Q
+    # Volume geometry information
+    vgeo = driver_config.grid.vgeo
+    M = vgeo[:, Grids._M, :]
+    # Unpack prognostic vars
+    ρ₀ = Q.ρ
+    ρe₀ = Q.ρe
+    # DG variable sums
+    Σρ₀ = sum(ρ₀ .* M)
+    Σρe₀ = sum(ρe₀ .* M)
```
charleskawczynski, line 252: Can we use the new way of adding conservation test callbacks? (you can check other experiments/drivers, they do not use `vgeo`)

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
charleskawczynski: No, originally they were `@test`, which ensures that it passes the test. When running this earlier, I saw that this fails the conservation tests.

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
jkozdon: [`weightedsum`](https://github.com/CliMA/ClimateMachine.jl/blob/0e9f0b985eedc1e2168a2a24a126265555424a6b/src/Arrays/MPIStateArrays.jl#L630-L638) should probably be used here.

This is an example of usage: https://github.com/CliMA/ClimateMachine.jl/blob/7d034daf7bae3e9a7131546517dca54903d26806/src/Diagnostics/atmos_mass_energy_loss.jl#L65-L68

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
Zhengyu-Huang: Oh got it, yes, the energy has has relative error larger than 0.0025

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
```diff
@@ -0,0 +1,311 @@
...
+        arg_type = String
+        default = "bulk"
+    end
+
+    cl_args = ClimateMachine.init(parse_clargs = true, custom_clargs = sbl_args)
+
+    surface_flux = cl_args["surface_flux"]
+
+    # DG polynomial order
+    N = 1
```
charleskawczynski, line 101: We should really set `N = 3` at least. I know that it's `N = 1` in master, but we shouldn't have that, and we should fix it here.

charleskawczynski: We can reduce `nelem_vert` to offset this a bit.

Zhengyu-Huang: will revisit the vertical spatial discretization after the horizontal-vertical splitting is done, I think a stabler spatial discretization will help the convergence

src/Atmos/Model/AtmosModel.jl
```diff
@@ -237,16 +237,23 @@ end
     vars_state(m::AtmosModel, ::Prognostic, FT)
 
 Conserved state variables (prognostic variables).
+
+!!! warning
+
+    The order of the fields for `AtmosModel` needs to match the one for
+    `AtmosLinearModel` since a shared state is used
```
charleskawczynski, line 244: There is a rebase conflict

charleskawczynski (changes requested):
Need to rebase

src/Atmos/Model/AtmosModel.jl
Zhengyu-Huang: just rebased

charleskawczynski: This code already exists in master, so I think something is still wrong.

test/Atmos/EDMF/stable_bl_edmf_implicit.jl
Zhengyu-Huang: Does this commit  introduce "Fix the ordering of the state in the AtmosModel" and "pass param and time to preconditioner_update" ?

src/Atmos/Model/AtmosModel.jl
Zhengyu-Huang: I will close this one and do a new PR

PR #1720, opened 2020-11-07
Updates section in Contributing docs about using the formatter

glwagner:
The new section looks like

image

### DIFF
--- a/docs/src/Contributing.md
+++ b/docs/src/Contributing.md
@@ -78,14 +78,15 @@ by the formatter implemented in the script `.dev/climaformat.jl`.
 To use the script `.dev/climaformat.jl` to reformat every file in
 `ClimateMachine.jl/` run
 
-```
-julia --project .dev/climaformat.jl
+```bash
+.dev/climaformat.jl
 ```
 
-from `ClimateMachine.jl`'s root directory. To format individual files, type
+in a terminal from `ClimateMachine.jl`'s root directory. To format individual files,
+type
 
-```
-julia --project .dev/climaformat.jl src/individual_file.jl
+```bash
+.dev/climaformat.jl src/individual_file.jl
 ```
 
 instead.
@@ -95,8 +96,8 @@ instead.
     1. `git add` your unformatted contributions to the git index with
        commands like `git add unformatted_contribution.jl` before running
        the formatter.
-    2. Reformat your code by running `julia --project .dev/climaformat.jl`.
-    3. Use `git diff --cached` to view changes made by the formatter, and
+    2. Reformat your code by running `.dev/climaformat.jl`.
+    3. Use `git diff` to view changes made by the formatter, and
        `git checkout :/` to throw them away if you don't like the way they
        look.
 
--- a/docs/src/Contributing.md
+++ b/docs/src/Contributing.md
@@ -70,7 +70,37 @@ in `docs/`.
 When your PR is ready for review, clean up your commit history by squashing
 and make sure your code is current with `ClimateMachine` master by rebasing.
 
-### Squash and rebase
+## Reformatting your code with `.dev/climaformat.jl`
+
+Your PR cannot be merged unless it conforms to the format style specified
+by the formatter implemented in the script `.dev/climaformat.jl`.
+
+To use the script `.dev/climaformat.jl` to reformat every file in
+`ClimateMachine.jl/` run
+
+```
+julia --project .dev/climaformat.jl
+```
+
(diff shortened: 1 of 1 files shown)


glwagner:
bors try

docs/src/Contributing.md
```diff
@@ -70,7 +70,35 @@ in `docs/`.
...
+```
+julia --project .dev/climaformat.jl src/individual_file.jl
+```
+
+instead.
+
+!!! warn "Reformatting edits your code"
+    If you would like to be able to revert changes made by the formatter,
+    1. Create a commit that contains your unformatted contributions before running
+       the formatter.
```
simonbyrne, line 96: Technically it is sufficient to add them to the index (i.e. have been `git add`ed). This is probably advisable anyway, since it will only format files that are in the index (so if you create a new .jl file, the formatter won't be applied until you `git add` it).

docs/src/Contributing.md
```diff
@@ -70,7 +70,35 @@ in `docs/`.
...
+    2. Reformat your code by running `julia --project .dev/climaformat.jl`.
+    3. Use `git diff` to view changes made by the formatter, and `git stash`
```
simonbyrne, line 98: You can just `git checkout :/`, which will reset it to the state of the index.

glwagner: Ok, I updated the recommendation, let me know you like.

charleskawczynski: My 2 cents: Leave it vaguer and direct users to the git docs page. e.g., "Reformatting may make changes which should be committed [and squashed/rebased](https://clima.github.io/ClimateMachine.jl/latest/Contributing/#Squash-and-rebase)".

docs/src/Contributing.md
```diff
@@ -70,7 +70,37 @@ in `docs/`.
...
+## Reformatting your code with `.dev/climaformat.jl`
+
+Your PR cannot be merged unless it conforms to the format style specified
+by the formatter implemented in the script `.dev/climaformat.jl`.
+
+To use the script `.dev/climaformat.jl` to reformat every file in
+`ClimateMachine.jl/` run
+
+```
+julia --project .dev/climaformat.jl
```
kpamnany, line 82: You don't need the `--project`, `climaformat.jl` uses `Pkg.activate` on its own `Project.toml`.

docs/src/Contributing.md
```diff
@@ -70,7 +70,37 @@ in `docs/`.
...
+```
+
+from `ClimateMachine.jl`'s root directory. To format individual files, type
+
+```
+julia --project .dev/climaformat.jl src/individual_file.jl
```
kpamnany, line 88: Ditto here.

docs/src/Contributing.md
```diff
@@ -70,7 +70,37 @@ in `docs/`.
...
+```
+
+instead.
+
+!!! warn "Reformatting edits your code"
+    If you would like to be able to revert changes made by the formatter,
+    1. `git add` your unformatted contributions to the git index with
+       commands like `git add unformatted_contribution.jl` before running
+       the formatter.
+    2. Reformat your code by running `julia --project .dev/climaformat.jl`.
```
kpamnany, line 98: And here.

docs/src/Contributing.md
glwagner: Isn't it friendly to provide recipes / explicit information? I usually find link-based explanations frustrating and energy-consuming, but maybe I'm alone.

docs/src/Contributing.md
simonbyrne: in fact it's executable, so you can just do `.dev/climaformat.jl`

docs/src/Contributing.md
```diff
@@ -70,7 +70,37 @@ in `docs/`.
...
+    3. Use `git diff --cached` to view changes made by the formatter, and
```
simonbyrne, line 99: you just want `git diff` (`git diff --cached` shows he difference between `HEAD` and the index, `git diff` will show the difference between the index and the working tree)

docs/src/Contributing.md
glwagner: Ah nice ok.

kpamnany (approved):

glwagner:
I don't think I ever got the updated recommendations for using the formatter to work for me (which is why I didn't feel I could merge this PR). I'll close this now but someone else should feel free to open and merge if they would like to.

PR #1722, opened 2020-11-07
Update compatability bound for JuliaFormatter in .dev/Project.toml

### LINKED ISSUE #1721: Formatter violates order-of-operations rules for julia syntax
I recently discovered this diff after applying `.dev/climaformat.jl` to my code:

image

Notice the change from

```julia
Ψ(x, L) = exp(-x^2 / 2L^2) # a Gaussian
```

to

```julia
Ψ(x, L) = exp(-x^2 / 2 * L^2) # a Gaussian
```

But these are not equivalent, since

```julia
julia> L = 2
2

julia> 1 / 2L^2
0.125

julia> 1 / 2 * L^2
2.0
```

Should we change the formatter so that it only changes the way code _looks_ (not functions)? This is a potential source of subtle and devious bugs in scientific code.

I personally find the ability to write constructs like `2a` to be a beautifully literate aspect of julia syntax and it would be a shame to throw it away.

glwagner:
🇺🇸

Resolves #1721


christophernhill:
@glwagner and @simonbyrne did you mean for this PR to update 47 files? That seems to make this less of a clean PR? 

Is there a way just to update compat, and then let files get updated later or is that not practical?

christophernhill:
@glwagner and @simonbyrne did you mean for this PR to update 47 files? That seems to make this less of a clean PR? 

Is there a way just to update compat, and then let files get updated later or is that problematic?

simonbyrne:
I did it, as I wanted to see what would be changed.

I thought it would make sense to switch the formatter so that it is applied to all tracked files, rather than just changed files, which should help address some of the issues people have had when merging changes.

Most of these changes aren't in files people touch very often, so shouldn't cause conflicts. If people are opposed though, I'm happy to revert.

src/Ocean/SplitExplicit01/OceanModel.jl
```diff
@@ -243,9 +243,9 @@ end
     ν = viscosity_tensor(m)
     #   D.ν∇u = ν * G.u
     D.ν∇u = @SMatrix [
-        m.νʰ * G.ud[1, 1] m.νʰ * G.ud[1, 2]
-        m.νʰ * G.ud[2, 1] m.νʰ * G.ud[2, 2]
-        m.νᶻ * G.u[3, 1] m.νᶻ * G.u[3, 2]
+        m.νʰ*G.ud[1, 1] m.νʰ*G.ud[1, 2]
```
charleskawczynski, line 246: Interesting that spaces between `*` aren't disappearing virtually everywhere.

charleskawczynski:
Looks fine to me.

simonbyrne (approved):

PR #1723, opened 2020-11-09
A proposal for subcomponent experiment specification

glwagner:
### A proposal for subcomponent experiment specification

This PR provides a proof-of-concept implementation of a new pattern for specifying experiments using `ClimateMachine` subcomponents. This PR does not address the possibly different problem of specifying coupled model / Earth system model experiments. The proposed pattern is inspired by modeling software such as Fenix, Firedrake, dedalus, Oceananigans, and more.

The core idea is that experiment specification generally consists of three independent steps:

1. Domain / geometry / mesh specification
2. Model / physics / equation / numerics specification
3. Simulation / diagnostics / output specification (and run! / invoke!)

When mature, the proposed interface will have types corresponding to each of these three steps, plus a suite of functions for interacting each type's data.

This PR adds two new `tutorials/Ocean/internal_wave.jl` and `tutorials/Ocean/shear_instability.jl`. Shortcuts are taken in order to implement the new design with minimal changes to the existing codebase. I made one change to the existing code base to make `SolverConfiguration` mutable. This is not crucial and can be reverted, but since `SolverConfiguration` is not concretely typed it seems that mutability is not a concern.

`shear_instability.jl` is intentionally concise so that I could post it in this PR:

```julia
# # Shear instability of a free-surface flow
#
# This script simulates the instability of a sheared, free-surface
# flow using `ClimateMachine.Ocean.HydrostaticBoussinesqSuperModel`.

using Printf
using Plots
using ClimateMachine

ClimateMachine.init()

using ClimateMachine.Ocean
using ClimateMachine.Ocean.CartesianDomains

using ClimateMachine.GenericCallbacks: EveryXSimulationTime
using ClimateMachine.Ocean: steps, Δt, current_time
using CLIMAParameters: AbstractEarthParameterSet, Planet

# We begin by specifying the domain and mesh,

domain = CartesianDomain(
    elements = (24, 24, 1),
    polynomialorder = 4,
    x = (-3π, 3π),
    y = (-3π, 3π),
    z = (0, 1),
    periodicity = (true, false, false),
    boundary = ((0, 0), (1, 1), (1, 2)),
)

# Note that the default solid-wall boundary conditions are free-slip and
# insulating on tracers. Next, we specify model parameters and the sheared
# initial conditions

struct NonDimensionalParameters <: AbstractEarthParameterSet end
Planet.grav(::NonDimensionalParameters) = 1

initial_conditions = InitialConditions(
    u = (x, y, z) -> tanh(y) + 0.1 * cos(x / 3) + 0.01 * randn(),
    v = (x, y, z) -> 0.1 * sin(y / 3),
    θ = (x, y, z) -> x,
)

model = Ocean.HydrostaticBoussinesqSuperModel(
    domain = domain,
    time_step = 0.1,
    initial_conditions = initial_conditions,
    parameters = NonDimensionalParameters(),
    turbulence_closure = (νʰ = 1e-2, νᶻ = 1e-2, κʰ = 1e-2, κᶻ = 1e-2),
    rusanov_wave_speeds = (cʰ = 0.1, cᶻ = 1),
)

# We prepare a callback that periodically fetches the horizontal velocity and
# tracer concentration for later animation,

u, v, η, θ = model.fields
fetched_states = []

data_fetcher = EveryXSimulationTime(1) do
    @info "Step: $(steps(model)), t: $(current_time(model)), max|u|: $(maximum(abs, u))"

    push!(
        fetched_states,
        (u = assemble(u), θ = assemble(θ), time = current_time(model)),
    )
end

# and then run the simulation.

model.solver_configuration.timeend = 100.0

result = ClimateMachine.invoke!(
    model.solver_configuration;
    user_callbacks = [data_fetcher],
)

# Finally, we make an animation of the evolving shear instability.

animation = @animate for (i, state) in enumerate(fetched_states)
    @info "Plotting frame $i of $(length(fetched_states))..."

    kwargs = (xlim = domain.x, ylim = domain.y, linewidth = 0, aspectratio = 1)

    x, y, = state.u.x, state.u.y

    u_plot = contourf(x, y, state.u.data[:, :, 1]'; color = :balance, kwargs...)
    θ_plot = contourf(x, y, state.θ.data[:, :, 1]'; color = :thermal, kwargs...)

    u_title = @sprintf("u at t = %.2f", state.time)
    θ_title = @sprintf("θ at t = %.2f", state.time)

    plot(u_plot, θ_plot, title = [u_title θ_title], size = (1200, 500))
end

gif(animation, "shear_instability.gif", fps = 8)
```

which produces the animation:

shear_instability

The script should run in a few minutes on a laptop.

A few notes:

1. The floating point type is given to `domain`, and propagated through to the `BalanceLaw` inside the `Model` constructor. The default is `Float64`.

2. Getting the field abstraction to work for distributed models may be difficult. However, it is still useful for single node runs and post-processing. The function `assemble`, which is crucial for visualization, will require a bit of work to use with `CuArray` `MPIStateArray`s.

3. The script above is missing the key step of outputting data, and then loading it and plotting it. This requires the ability to load data from file into `CartesianField`.

4. Hopefully we can specify boundary conditions entirely in the model constructor without having to use `domain`.

5. It's useful to have a function `set!(model, initial_conditions)`, rather than requiring initial conditions in the model constructor. This can probably be achieved using the functionality of `init_state_prognostic!` but I have to look into it more.

This PR depends on #1716, so it might make sense to save detailed review for after that's merged.


kpamnany:
I like `CartesianDomain`. I think it should replace `DriverConfiguration` in entirety.

src/Ocean/SuperModels.jl
```diff
@@ -84,6 +84,10 @@ function HydrostaticBoussinesqSuperModel(;
     filters = nothing,
     modeldata = NamedTuple(),
     init_on_cpu = true,
+    boundary_conditions = (
```
christophernhill, line 87: @glwagner is there a reason to have some defaults here, instead of

```
boundary_conditions = (),
```

src/Ocean/OceanProblems/initial_value_problem.jl
```diff
@@ -0,0 +1,100 @@
...
+    aux,
+    coords,
+    time,
+)
+
+    ics = ivp.initial_conditions
+    x, y, z = coords
+
+    state.u = @SVector [ics.u(x, y, z), ics.v(x, y, z)]
+    state.θ = ics.θ(x, y, z)
```
christophernhill, line 96: @glwagner would it be better to have

```
state.θ = ics.θ(x, y, z, ivp)
```

instead of 

```
state.θ = ics.θ(x, y, z)
```

and for other ```ics.``` funcs?

For example - what if I want to do an initial condition that is

```
θᵢ(x,y,z,p) = ( 20. * ( 1 - z/p.H ) )
```

src/Ocean/SuperModels.jl
glwagner: It depends on what we want the default model to be. If our default domain / geometry has boundaries, I think we also need to include default boundary conditions somewhere, somehow.

I think we can make specifying boundary conditions a bit easier for most of the problems we are interested in by constraining the input a bit. We can generally use the terms "east, west, north, south, top, bottom" to refer to the 6 end points of a domain (2 for each of the 3 directions). This works for simple `CartesianDomain`s and `SphericalDomain`s. More complicated boundary conditions for unstructured meshes would require the more flexible interface that boundary condition "tag"s permits, so we can keep that as a fallback when we need it, too.

src/Ocean/OceanProblems/initial_value_problem.jl
glwagner: What's the advantage of being able to reference a parameter as a local variable versus a global variable in the initial conditions function?

If we want functionality like this, I think it's better to attach parameters to `ics.θ` rather than baking this assumption into the function signature. This way we can provide a simple easy-to-use interface, but _support_ more complicated interaction, if desired.

This can be achieved with callable objects:

```julia
struct ParameterizedInitialCondition{P, C}
    condition :: C
    parameters :: P
end

(ivp::ParameterizedInitialCondition)(x, y, z) = condition(x, y, z, parameters)
(ivp::ParameterizedInitialCondition{<:Nothing})(x, y, z) = condition(x, y, z)
```

This way we don't have to change the source code of the function you're referencing; instead we introduce an object that carries around these optional parameters.

On the user-side, this is easy to achieve through dispatch because we already have an interface for `InitialConditions`. So we write add a kwarg `parameters` to `InitialConditions`:

```julia
function InitialConditions(; parameters=nothing, u=default, ...)
    u = ParameterizedInitialCondition(u, parameters)
    # etc
end
```

This way the default is a function signature `x, y, z`, but if you need to pass in parameters, you can.

This pattern is extremely useful for forcing functions and boundary conditions and used extensively in `Oceananigans`. However, generally speaking we haven't found the need to use it for initial conditions --- mainly because initial conditions don't have to run on the GPU and thus don't have special compilation requirements.

src/Ocean/SuperModels.jl
christophernhill: Definitely BC abstractions (and implementation of current abstraction) need work.

I think default of periodic in X and Y makes sense, not sure about Z. Periodic is nice for symmetry. 
Z as periodic by default its a bit too obscure turbulence/CFD for most people. Most ocean users would expect there to be an ocean floor or an open boundary. Also with g it doesn't usually work very well, things fall out of the bottom......

glwagner: Doesn't our balance law sort of require there to be a free surface at the top? Does vertically periodic even make sense for us?

src/Ocean/OceanProblems/initial_value_problem.jl
christophernhill: That works. I don't really need it right now, but we will need something like that - so was just checking.

src/Ocean/SuperModels.jl
christophernhill: Vertically periodic does not make sense for the model - I was thinking more topologically. At the model level it doesn't. I think periodic in X and Y is a good default though. For vertical we need to choose between NoSlip and FreeSlip at bottom for Kinematic - both are legitimate defaults. For Surface, something inert e.g. NoSlip, NoFlux etc... would be my preference.

christophernhill: @jm-c any preference on default BC's?

glwagner: I think no flux boundary conditions for solid boundaries make the most sense because they simplest model from the standpoint of the numerics. A NoSlip might not be inert, since the flux of momentum associated with it must be modeled by a flux reconstruction of some kind. 

In addition, NoSlip boundary conditions are problematic because they require a certain amount of viscous friction and, depending on the numerical methods used, may require resolving a boundary layer of unknown thickness. Thus as a default this can create problems for users who just want to spin up some simple initial condition to start out.

For some problems (depending on numerics) we can run without viscosity at all. This might be the simplest of all defaults --- numerical diffusion only with free slip and insulating conditions at solid boundaries.

kpamnany:
As discussed with @glwagner, I like the `CartesianDomain` abstraction (introduced [here](https://github.com/CliMA/ClimateMachine.jl/blob/04ca07bbbe03d9f030e6862236f50391acdf3592/src/Ocean/CartesianDomains/cartesian_domain.jl)) and will replace `DriverConfiguration` with it at some point in the future.

The `SuperModel` abstraction he's proposing also looks like something that will simplify experiment specification -- it essentially bundles the domain, the model (now 'equations'?), the state and the time stepper into a single container. This also seems reasonable and 

Cc: @simonbyrne, @charleskawczynski

src/Ocean/SuperModels.jl
christophernhill: That sounds good to me. I think going with default as the "simplest sensible thing possible, but no simpler" is a good principle. "Simplest sensible thing possible" is of course a little subjective, but that seems OK - it allows some opinionated element into the mix to guide people.

jm-c: I agree with you @glwagner , this default BC makes sense. It's also the type of default BC that is/was there in src/Ocean/OceanBC.jl if I am right.

christophernhill:
@glwagner some types were missing in src/Ocean/SuperModels.jl 

Would it make sense for OceanBC.jl to be a module, this

https://github.com/CliMA/ClimateMachine.jl/blob/38cdf196be3ba5d2d83ea34fba3b6957af947625/src/Ocean/SuperModels.jl#L11

is a bit ugly?

glwagner:
@christophernhill it could make sense to add them in the `HydrostaticBoussinesqEquations` module; they're specific to that BalanceLaw it would seem.

christophernhill:
> @christophernhill it could make sense to add them in the `HydrostaticBoussinesqEquations` module; they're specific to that BalanceLaw it would seem.

I think they are also in ShallowWater e.g. 

https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/test/Ocean/ShallowWater/GyreDriver.jl#L88

glwagner:
@christophernhill let's discuss refactoring the `radiating_eddy.jl` tutorial to simulate a baroclinic ocean gyre here.

The first change we need is arbitrary boundary conditions, resolving #1704 

What is the kinematic stress function that we need to use, domain limits, and reduced gravity?

src/Ocean/Domains/rectangular_domain.jl
```diff
@@ -0,0 +1,194 @@
...
+    @sprintf(
+        "    x = (%.2e, %.2e), y = (%.2e, %.2e), z = (%.2e, %.2e)",
+        domain.x[1],
+        domain.x[2],
+        domain.y[1],
+        domain.y[2],
+        domain.z[1],
+        domain.z[2]
+    )
+)
```
charleskawczynski, line 49: Is it worth battling the formatter a bit here?

glwagner: Might be, since I'm not as wily as you I submit to the formatter's dominance much more quickly

src/Ocean/Domains/rectangular_domain.jl
```diff
@@ -0,0 +1,194 @@
...
+        L,
+        (west, east),
+        (south, north),
+        (bottom, top),
+    )
+end
+
+array_type(domain::RectangularDomain) = Array #array_type(domain.grid)
+eltype(::RectangularDomain{FT}) where {FT} = FT
+communicator(args...) = MPI.COMM_WORLD
```
charleskawczynski, line 194: This looks like it's grabbing `MPI` globally--we probably don't want to do that. cc @kpamnany

src/Ocean/Domains/rectangular_domain.jl
```diff
@@ -0,0 +1,194 @@
...
+using ClimateMachine.Mesh.Grids:
+    DiscontinuousSpectralElementGrid, polynomialorder
+
+using ClimateMachine.Mesh.Topologies: StackedBrickTopology
+
+#####
+##### RectangularDomain
+#####
+
+struct RectangularDomain{FT, G} <: AbstractDomain
```
charleskawczynski, line 12: Docs?

src/Ocean/Domains/rectangular_domain.jl
glwagner: For sure, that's why I put this place holder function here. Please advise.

src/Ocean/Domains/rectangular_domain.jl
glwagner: Do we put docs on types and constructors or just constructors? Over at Oceananigans we put docs on everything (types and constructors) but my thought these days is that docs on constructors really pollutes the help messages and makes it harder to use stuff. Happy to doc this all up though.

src/Ocean/Domains/rectangular_domain.jl
```diff
@@ -0,0 +1,194 @@
...
+    elements,
+    polynomialorder,
+    x::Tuple{<:Number, <:Number},
+    y::Tuple{<:Number, <:Number},
+    z::Tuple{<:Number, <:Number},
+    periodicity = (true, true, false),
+    boundary = ((0, 0), (0, 0), (1, 2)),
+    array_type = Array,
+    message_communicator = MPI.COMM_WORLD,
+)
```
charleskawczynski, line 147:
```suggestion
function RectangularDomain(
    ::Type{FT};
    elements,
    polynomialorder,
    x::Tuple{<:Number, <:Number},
    y::Tuple{<:Number, <:Number},
    z::Tuple{<:Number, <:Number},
    periodicity = (true, true, false),
    boundary = ((0, 0), (0, 0), (1, 2)),
    array_type = Array{FT},
    message_communicator = MPI.COMM_WORLD,
) where {FT}
```
Or, we could grab `FT` from `eltype` of `array_type`?

glwagner: That seems like a good idea --- more compact specification is always nice. And just to be clear, we use `array_type` to distinguish between CPU / GPU computations?

src/Ocean/SuperModels.jl
```diff
@@ -0,0 +1,209 @@
...
+
+using CLIMAParameters: AbstractEarthParameterSet
+
+struct EarthParameters <: AbstractEarthParameterSet end
+
+#####
+##### It's super good
+#####
+
+struct HydrostaticBoussinesqSuperModel{D, E, S, F, N, T, C}
```
charleskawczynski, line 38: Docs?

glwagner:
Notes:

The problem-specific `kinematic_stress` functions are defined in `Ocean/OceanProblems`. The trivial case is here:

https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L300

The "gyre" case is here:

https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L482-L483

@christophernhill please advise if the "gyre" kinematic stress pattern is what you want in the new tutorial.

src/Ocean/SuperModels.jl
```diff
@@ -0,0 +1,209 @@
...
+
+import ClimateMachine: SolverConfiguration
+
+#####
+##### Build default EarthParameters
+#####
+
+using CLIMAParameters: AbstractEarthParameterSet
+
+struct EarthParameters <: AbstractEarthParameterSet end
```
charleskawczynski, line 32: Please read [the CLIMAParameters.jl docs](https://clima.github.io/CLIMAParameters.jl/dev/#Implementation-notes). These structs must live at the driver level to ensure that we don't accidentally have multiple (and potentially inconsistent) implementations of parameters. We can define and instantiate the structs at the top level and pass them down into the driver wherever they may be needed.

src/Ocean/Domains/rectangular_domain.jl
kpamnany: When it gets really outrageous, you can [skip formatting](https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/) but we really don't want to abuse that.

src/Ocean/SuperModels.jl
glwagner: We need a centralized default. Where can I grab it from?

glwagner: If we move forward with `AbstractModels` the logical place for a default parameter set will indeed be in the submodule top-level `Model.jl`. I'm happy to move things around until then if a different solution is better.

charleskawczynski: Why can't you pass it down? It's done like that everywhere else.

src/Ocean/Domains/rectangular_domain.jl
kpamnany: We use the name `mpicomm` everywhere. Not sure what this `communicator` call is used for... the communicator is ideally passed into the constructor for internal use by the module.

src/Ocean/Domains/rectangular_domain.jl
kpamnany: Yes. The array type is a [global setting](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Driver.jl#L92) -- the default should come from there [like so](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/driver_configs.jl#L123).

However, there is no global setting for the floating point precision selection (but there should be) so we require that it [be specified](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/driver_configs.jl#L140).

src/Ocean/SuperModels.jl
kpamnany: Yes. But we did, I think, agree on adding some convenience constructors to `CLIMAParameters.jl`, particularly to help with code like [this](https://github.com/CliMA/ClimateMachine.jl/blob/aj/precip_interface/experiments/AtmosLES/dycoms.jl#L28-L56)?

src/Ocean/Domains/rectangular_domain.jl
kpamnany: My immediate reaction is that you want to document the constructors, not necessarily the type. I'm not entirely sure though, because Greg's point makes sense. So, when in doubt, I look at what the Julia docs do. [Here](https://github.com/JuliaLang/julia/blob/788b2c77c10c2160f4794a4d4b6b81a95a90940c/stdlib/Sockets/src/IPAddr.jl#L31-L41) is one example where they have a docstring on one constructor only. But [here](https://github.com/JuliaLang/julia/blob/788b2c77c10c2160f4794a4d4b6b81a95a90940c/stdlib/SparseArrays/src/sparsevector.jl#L12-L16), they document the type.

So I guess the answer is: use your best judgement? @simonbyrne: thoughts?

blallen:
> > @christophernhill it could make sense to add them in the `HydrostaticBoussinesqEquations` module; they're specific to that BalanceLaw it would seem.
> 
> I think they are also in ShallowWater e.g.
> 
> https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/test/Ocean/ShallowWater/GyreDriver.jl#L88

They can also be used for `OceanModel` and `BarotropicModel` with a few changes. The plan is for these types in `OceanBC.jl` to be used to specify all types of boundary conditions for any ocean model (or equations), that's why they are in `ClimateMachine.Ocean` instead of `ClimateMachine.Ocean.HydrostaticBoussinesq`.

christophernhill:
> > > @christophernhill it could make sense to add them in the `HydrostaticBoussinesqEquations` module; they're specific to that BalanceLaw it would seem.
> > 
> > 
> > I think they are also in ShallowWater e.g.
> > https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/test/Ocean/ShallowWater/GyreDriver.jl#L88
> 
> They can also be used for `OceanModel` and `BarotropicModel` with a few changes. The plan is for these types in `OceanBC.jl` to be used to specify all types of boundary conditions for any ocean model (or equations), that's why they are in `ClimateMachine.Ocean` instead of `ClimateMachine.Ocean.HydrostaticBoussinesq`.



> Notes:
> 
> The problem-specific `kinematic_stress` functions are defined in `Ocean/OceanProblems`. The trivial case is here:
> 
> https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L300
> 
> The "gyre" case is here:
> 
> https://github.com/CliMA/ClimateMachine.jl/blob/102d15ef61a5b6ae9f69b0bcb96ee6a6880cfdd0/src/Ocean/OceanProblems/SimpleBoxProblem.jl#L482-L483
> 
> @christophernhill please advise if the "gyre" kinematic stress pattern is what you want in the new tutorial.

Here is what I think I want

```
Lx=4e6
Ly=6e6
H=-3000
domain = CartesianDomain(
    elements = (10, 15, 2),
    polynomialorder = 4,
    polynomialorder = 2,
    x = (0, Lx),
    y = (0, Ly),
    z = (H, 0),
    periodicity = (false, false, false),
    boundary = ((1, 1), (1, 1), (2, 3)),
)

# Note that the default solid-wall boundary conditions are free-slip and
# insulating on tracers. Next, we specify model parameters and the sheared
# initial conditions

struct ParamSet <: AbstractEarthParameterSet end

initial_conditions = InitialConditions(
    u = (x, y, z) -> -0,
    v = (x, y, z) -> -0,
    θ = (x, y, z) -> 20.,
    # θ(x,y,z) = ( 20. * ( 1 - z/H ) ),
)

ρₒ=1000
τₒ=1e-1
wind_stress = (
    τ¹¹ = ( x, y, t, p) -> (τₒ / ρₒ) * cos(2 * y * π / Lʸ),
)
# wind_stress = ()


model = Ocean.HydrostaticBoussinesqSuperModel(
    domain = domain,
    time_step = 30,
    initial_conditions = initial_conditions,
    parameters = ParamSet(),
    turbulence_closure = (νʰ = 5e3, νᶻ = 5e-3, κʰ = 1e3, κᶻ = 1e-4),
    rusanov_wave_speeds = (cʰ = 1, cᶻ = 1.e-3),
    buoyancy = ( αᵀ = 0., ρₒ=ρₒ ),
    coriolis = (f₀=1.e-4, β=1.e-11),
    boundary_conditions = (
      OceanBC(Impenetrable(NoSlip()), Insulating()),
      OceanBC(Impenetrable(FreeSlip()), Insulating()),
      OceanBC(Penetrable(KinematicStress(stress=wind_stress)), Insulating()),
    ),
)

```

blallen:
@christophernhill, I think the above specification is good, but I think you will need to specify both components of the wind stress, i.e.

```
wind_stress = (
    τ¹¹ = ( x, y, t, p) -> (τₒ / ρₒ) * cos(2 * y * π / Lʸ),
    τ²² = ( x, y, t, p) -> 0,
)
```
but that also probably depends on the discussion in #1740.

christophernhill:
> @christophernhill, I think the above specification is good, but I think you will need to specify both components of the wind stress, i.e.
> 
> ```
> wind_stress = (
>     τ¹¹ = ( x, y, t, p) -> (τₒ / ρₒ) * cos(2 * y * π / Lʸ),
>     τ²² = ( x, y, t, p) -> 0,
> )
> ```
> 
> but that also probably depends on the discussion in #1740.

I am hoping that we will end up with default functions that returns 0.  i.e. if you ask for a stress bc but don't specify the stress or one of the components, then we take that to mean zero stress for the unspecified piece. Not really sure what a typical user will expect?

blallen:
> I am hoping that we will end up with default functions that returns 0. i.e. if you ask for a stress bc but don't specify the stress or one of the components, then we take that to mean zero stress for the unspecified piece. Not really sure what a typical user will expect?

Ah, I was just going off the use pattern from `InitialConditions`. Could in theory shorten
```julia
initial_conditions = InitialConditions(
    u = (x, y, z) -> -0,
    v = (x, y, z) -> -0,
    η = (x, y, z) -> -0,
    θ = (x, y, z) -> 20.,
)
```
to
```julia
initial_conditions = InitialConditions(
    θ = (x, y, z) -> 20.,
)
```
??

Also just realized you explicitly set `u` and `v` to zero, but not `η`. I think zero defaulting is smart, but it's not currently how things behave.

glwagner:
> Also just realized you explicitly set u and v to zero, but not η. I think zero defaulting is smart, but it's not currently how things behave.

I might be confused, but the defaults are all "resting":

https://github.com/CliMA/ClimateMachine.jl/blob/f82b25684a7143a26c2be0dc8c8df160659f79fe/src/Ocean/OceanProblems/initial_value_problem.jl#L73-L74

where `resting(x, y, z) = 0`.

glwagner:
> I am hoping that we will end up with default functions that returns 0. i.e. if you ask for a stress bc but don't specify the stress or one of the components, then we take that to mean zero stress for the unspecified piece. Not really sure what a typical user will expect?

I think we should discuss a general boundary conditions interface on #1740 and implement it in a different PR.

src/Ocean/Domains/rectangular_domain.jl
glwagner: I don't want to spend time on this now so I will battle the formatter in a later PR when I have more tricks up my sleeve.

src/Ocean/Domains/rectangular_domain.jl
glwagner: Is "mpicomm" an abbreviation for "message_communicator"? Is there a function "mpicomm" that I can use?

As a student, I was eternally confused and frustrated reading FORTRAN codes and running into the name "MPI_COMM_WORLD" (or whatever) and having no idea what it meant or what it did. It's short for something --- but what?!?! It seems important but is so inscrutable. Bundle that experience with more general anxiety about learning new things and it becomes a toxic mess.

Now that I know that it literally "communicates messages" I feel `message_communicator` is the right name. I'm trying to help future Greg's...

glwagner: The reason why it's needed here is because `DiscontinuousSpectralElementGrid` needs a communicator. But the grid is built before the model, so to use the same communicator in the model and the grid, I need to query the grid to get its communicator.

Rather than hard code a reference to the communicator via domain/grid, I wrote this function, anticipating that the design would change shortly...

I added a naive, simple default that would work in the cases I knew I would use it in, with the plan of extending / changing it in the future when I had examples of different communicators being used.

I'm happy to change this design.

src/Ocean/SuperModels.jl
glwagner: I removed the default, so I'm resolving this.

src/Ocean/Domains/rectangular_domain.jl
glwagner: Hmm.

Thinking about it more, I think it's cumbersome to have to write `Array{Float64}` rather than `Array`. The vast majority of users will use double precision; it's only specialty applications that need to changing the floating point type (that we know of). Thus I think it makes sense to specify the floating point precision and the array type with independent kwargs.

I'll configure this constructor to use the global setting as a default.

src/Ocean/SuperModels.jl
glwagner: I put docs on the constructor, so I'm going to leave docs off this so that `help? HydrostaticBoussinesqSuperModel` just returns the constructor docstring.

src/Ocean/Domains/rectangular_domain.jl
kpamnany: > Is "mpicomm" an abbreviation for "message_communicator"? Is there a function "mpicomm" that I can use?

It means "MPI communicator". It isn't a function -- I meant that the name of the variable used to hold an MPI communicator (in all of the existing ClimateMachine code) is `mpicomm`.

We could use something besides MPI to do HPC ([GASNet](https://gasnet.lbl.gov/), [Global Arrays](https://hpc.pnl.gov//globalarrays/), etc.) but we're not -- this is an MPI code.
> 
> As a student, I was eternally confused and frustrated reading FORTRAN codes and running into the name "MPI_COMM_WORLD" (or whatever) and having no idea what it meant or what it did. It's short for something --- but what?!?! It seems important but is so inscrutable. Bundle that experience with more general anxiety about learning new things and it becomes a toxic mess.

Take deep breaths! Then read at least the beginning of [this](https://pages.tacc.utexas.edu/~eijkhout/pcse/html/mpi-comm.html).

> Now that I know that it literally "communicates messages" I feel `message_communicator` is the right name. I'm trying to help future Greg's...

I think `communicator` is fine for this -- see next comment.

kpamnany: > The reason why it's needed here is because `DiscontinuousSpectralElementGrid` needs a communicator. But the grid is built before the model, so to use the same communicator in the model and the grid, I need to query the grid to get its communicator.

Makes excellent sense. I suggest the implementation be:
```julia
communicator(domain::RectangularDomain) = communicator(domain.grid)
```
And add:
```julia
communicator(grid::DiscontinuousSpectralElementGrid) = communicator(grid.topology)
```
Somewhere [here](https://github.com/CliMA/ClimateMachine.jl/blob/ocean/glw/super-hydrostatic-boussinesq/src/Numerics/Mesh/Grids.jl#L277), and add:
```julia
communicator(topology::BoxElementTopology) = topology.mpicomm
```
Somewhere [here](https://github.com/CliMA/ClimateMachine.jl/blob/ocean/glw/super-hydrostatic-boussinesq/src/Numerics/Mesh/Grids.jl#L277) as well as a few other similar definitions for the hierarchy of topologies (primarily [these](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/Mesh/Topologies.jl#L7-L10) I think).

kpamnany: My original comment should have been targeted at [this line](https://github.com/CliMA/ClimateMachine.jl/blob/ocean/glw/super-hydrostatic-boussinesq/src/Ocean/Domains/rectangular_domain.jl#L146) -- please use `mpicomm` for the argument name there for consistency with the rest of the code.

kpamnany: Cc: @jkozdon for any thoughts.

glwagner: Great, thanks. I'm certainly not suggesting to use something other than MPI, I hope I didn't imply that. I was really just pointing out that long ago this was a confusion for me. I'm using my past experiences as a guide for how other, similar people might feel when reading my code.

Looking into the code it looks like `mpicomm` is a property of `grid.topology`... ? I didn't know exactly where it was before, hence the fallback / function. Is `grid.topology` reliable for inferring the `communicator`? How would a user specify a different `communicator` if they needed to?

glwagner: Ok, I will use the name `mpicomm` for the MPI communicator.

glwagner: It's probably not appropriate for me to put these deep changes into this PR, which is rather experimental. Can we do something simple temporarily in order to make progress quickly, and open an issue about writing an infrastructure for using consistent communicators? I'm happy to implement the PR that resolves the issue too, I just think this PR is growing, and growing...

kpamnany: No, it's a reasonable concern, hence I've pulled Jeremy in to comment. Previously, we were using the global communicator exclusively (your favorite `MPI_COMM_WORLD`). Then, [this](https://github.com/CliMA/ClimateMachine.jl/blob/ocean/glw/super-hydrostatic-boussinesq/src/Arrays/MPIStateArrays.jl#L128-L131) was added.

Wait... when you say:
> The reason why it's needed here is because DiscontinuousSpectralElementGrid needs a communicator. But the grid is built before the model, so to use the same communicator in the model and the grid, I need to query the grid to get its communicator.

What's the model now? The `*Equations` shouldn't need a communicator, and the `SuperModel` is constructing the grid, so it should already have the communicator...

glwagner: @jkozdon @kpamnany @charleskawczynski I'm resolving this issue by removing the `communicator` function and using COMM_WORLD when building DriverConfiguration (wihch doesnt matter becuaase we will nuke DriverConfiguration anyways). We can discuss communicator ju-jitsu when the time comes!

src/Ocean/Domains/rectangular_domain.jl
charleskawczynski: That's fine with me. I think we've been using mixed styles, but perhaps it's context dependent.

glwagner:
bors try

glwagner:
Superceded by #1806 .

PR #1724, opened 2020-11-09
Move Gravity to new tendency specification

charleskawczynski:
### Description

This PR moves `Gravity` to the new tendency specification. I can confirm that, in `experiments/AtmosGCM/heldsuarez.jl`, the tendency table is properly updated:
Screen Shot 2020-11-09 at 6 09 37 AM

### DIFF
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -2,7 +2,6 @@ using ..Microphysics_0M
 using CLIMAParameters.Planet: Omega, e_int_i0, cv_l, cv_i, T_0
 
 export AbstractSource,
-    Gravity,
     RayleighSponge,
     GeostrophicForcing,
     Coriolis,
@@ -38,7 +37,6 @@ end
 
 abstract type AbstractSource end
 
-struct Gravity <: AbstractSource end
 function atmos_source!(
     ::Gravity,
     atmos::AtmosModel,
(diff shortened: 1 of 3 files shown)


src/Atmos/Model/source.jl
```diff
@@ -49,11 +47,7 @@ function atmos_source!(
     t::Real,
     direction,
 )
-    if atmos.ref_state isa HydrostaticState
-        source.ρu -= (state.ρ - aux.ref_state.ρ) * aux.orientation.∇Φ
-    else
-        source.ρu -= state.ρ * aux.orientation.∇Φ
-    end
+    # Migrated to Σsources
```
thomasgibson, line 50: So is the idea that, once all other sources are converted (i.e. Coriolis, etc), these functions will become unnecessary? This function is left here for API compatibility?

thomasgibson (commented):
Looks good, just a clarifying comment:

src/Atmos/Model/source.jl
charleskawczynski: Yes, we have to leave it for now until all of the sources have been converted (`atmos_source!` sums over all source types and our source specification at the driver layer is not really changing).

thomasgibson (approved):
My only concern is the documenter, but I think this change is good. Once the docs are sorted, go ahead and r+!

PR #1725, opened 2020-11-09
Move first order moisture flux to new tendency specification

charleskawczynski:
### Description

This PR

 - moves the first order moisture flux to the new tendency specification. I can confirm that the tendency table is updated, for `experiments/AtmosLES/bomex_model.jl` with `moisture_model == "nonequilibrium"`, but I haven't checked the results.
 - Renames `flux_moisture!` to `flux_first_order!` in efforts to standardize some of the AtmosModel method calls
 - Adds `Moisture <: PrognosticVariable` to combine specification for the moisture variables.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -434,7 +434,7 @@ equations.
 
     # pressure terms
     flux_radiation!(m.radiation, m, flux, state, aux, t)
-    flux_moisture!(m.moisture, m, flux, state, aux, t)
+    flux_first_order!(m.moisture, m, flux, state, aux, t, direction)
     flux_precipitation!(m.precipitation, m, flux, state, aux, t)
     flux_tracers!(m.tracers, m, flux, state, aux, t)
     flux_first_order!(m.turbconv, m, flux, state, aux, t)
--- a/src/Atmos/Model/atmos_tendencies.jl
+++ b/src/Atmos/Model/atmos_tendencies.jl
@@ -37,4 +37,5 @@ eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Energy} =
     (Advect{PV}(), Pressure{PV}())
 
 # Moisture
-eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Moisture} = ()
+eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Moisture} =
+    (Advect{PV}(),)
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1079,6 +1079,16 @@ steps:
       slurm_ntasks: 1
       slurm_gres: "gpu:1"
 
+  - label: "gpu_bomex_single_stack_nonequil"
+    key: "gpu_bomex_single_stack_nonequil"
+    command:
+      - "mpiexec julia --color=yes --project experiments/AtmosLES/bomex_single_stack.jl --moisture-model nonequilibrium --diagnostics=default --fix-rng-seed"
+    agents:
+      config: gpu
+      queue: central
+      slurm_ntasks: 1
+      slurm_gres: "gpu:1"
+
   - label: "gpu_bomex_edmf"
     key: "gpu_bomex_edmf"
     command:
--- a/experiments/AtmosLES/bomex_single_stack.jl
+++ b/experiments/AtmosLES/bomex_single_stack.jl
@@ -6,6 +6,11 @@ function main()
     bomex_args = ArgParseSettings(autofix_names = true)
     add_arg_group!(bomex_args, "BOMEX")
     @add_arg_table! bomex_args begin
+        "--moisture-model"
+        help = "specify cloud condensate model"
+        metavar = "equilibrium|nonequilibrium"
+        arg_type = String
+        default = "equilibrium"
         "--surface-flux"
         help = "specify surface flux for energy and moisture"
         metavar = "prescribed|bulk"
@@ -17,6 +22,7 @@ function main()
         ClimateMachine.init(parse_clargs = true, custom_clargs = bomex_args)
 
     surface_flux = cl_args["surface_flux"]
+    moisture_model = cl_args["moisture_model"]
 
     FT = Float64
     config_type = SingleStackConfigType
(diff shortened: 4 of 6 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -425,10 +425,19 @@ equations.
...
+    if (m.moisture isa DryModel)
+        flux.moisture.ρq_tot =
+            Σfluxes(eq_tends(TotalMoisture(), m, tend), args...)
+    end
+    if (m.moisture isa NonEquilMoist)
+        flux.moisture.ρq_liq =
+            Σfluxes(eq_tends(LiquidMoisture(), m, tend), args...)
+        flux.moisture.ρq_ice =
+            Σfluxes(eq_tends(IceMoisture(), m, tend), args...)
+    end
```
kpamnany, line 437: I'd prefer we continue to use `m.moisture` as a dispatch type rather than test it like this.

src/Atmos/Model/atmos_tendencies.jl
```diff
@@ -34,4 +34,15 @@ eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Energy} =
 
 # TotalMoisture
 eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: TotalMoisture} =
```
kpamnany, line 36: Perhaps sub-component types like `MoistureModel` should be added into these methods?

kpamnany (commented):
> it seems that NonEquilMoist models are not tested

Maybe add `bomex_les.jl --moisture-model=nonequilibrium` also to the pipeline?

src/Atmos/Model/AtmosModel.jl
charleskawczynski: How so? Do you mean leave
```julia
    flux_moisture!(m.moisture, m, flux, state, aux, t)
```
And then call the sum inside `flux_moisture!`? If so, I'm fine with that.

src/Atmos/Model/atmos_tendencies.jl
charleskawczynski: I think we can do that, but the moment that `TotalMoisture` and `IceMoisture` or `LiquidMoisture` have different fluxes, they'll need their own methods. I'll combine them for now, and we can split them if we need.

src/Atmos/Model/AtmosModel.jl
kpamnany: I think that's cleaner, yeah.

Speaking more generally, perhaps having a more concrete relationship between model and components, and thereby tendencies, would help make all this neater. Maybe `MoistureModel` etc. should all be subtypes of an `AbstractAtmosComponent` or something like that? Not too sure how best to do that and for a separate PR anyway.

charleskawczynski:
> Maybe add bomex_les.jl --moisture-model=nonequilibrium also to the pipeline?

I'll add the single stack one, since this is cheaper but still exercises the kernels.

src/Atmos/Model/atmos_tendencies.jl
charleskawczynski: I've added `abstract type Moisture <: PrognosticVariable end`

trontrytel: We will need them splitted very soon. Like this week - with the precipitation + 1-moment microphysics interface I am working on.

Would it make sense to merge that first so that you can see the more complicated example of moisture and precipitation sources?

trontrytel:
> > it seems that NonEquilMoist models are not tested
> 
> Maybe add `bomex_les.jl --moisture-model=nonequilibrium` also to the pipeline?

The plan was to have a full bomex run with both equilibrium and nonequilibrium mositure as a part of weekly/nightly tests. Once we have them

src/Atmos/Model/atmos_tendencies.jl
trontrytel: This is what I have for rain now - https://github.com/CliMA/ClimateMachine.jl/pull/1706 It has its own additional velocity that gets computed in the flux. 

I think it's mostly done. I got stuck adding precipitation to diagnostics. But should be done soon (I hope)

trontrytel:
In general I would say that moisture variables (q_tot, q_liq and q_ice) will have the same flux but different source terms. And then the upcoming precipitation variables (q_rai and q_sno) will have different fluxes and source terms.

charleskawczynski:
> In general I would say that moisture variables (q_tot, q_liq and q_ice) will have the same flux but different source terms. And then the upcoming precipitation variables (q_rai and q_sno) will have different fluxes and source terms.

That's great, we can group them as we need, then.

charleskawczynski:
> The plan was to have a full bomex run with both equilibrium and nonequilibrium mositure as a part of weekly/nightly tests. Once we have them

Sounds good. I'll leave it for now so that we know that the kernels are syntactically correct, but I agree that we should hook in QA tests.

glwagner (approved):
Unfortunately my limited understanding of the balance law framework hinders my ability to review this PR. Nothing egregious pops out to me so I'm approving.

A higher-level question is --- does it make sense to modularize the microphysics implementation a bit more? Right now it looks like terms like "mass, momentum, energy" prognostic variables are specified alongside moisture terms. A more modular design might separate these implementations a bit more (I can't tell if they just _happen_ to be nearby in the code, or if their proximity reflects a deeper entanglement).

Can you use independent balance laws to different aspects of the equations (one for "primary" variables like mass, momentum, and energy, another for microphysics variables, another for passive chemical species) to enforce / ease this modularization?

charleskawczynski:
> A higher-level question is --- does it make sense to modularize the microphysics implementation a bit more? Right now it looks like terms like "mass, momentum, energy" prognostic variables are specified alongside moisture terms. A more modular design might separate these implementations a bit more (I can't tell if they just _happen_ to be nearby in the code, or if their proximity reflects a deeper entanglement). Can you use independent balance laws to different aspects of the equations (one for "primary" variables like mass, momentum, and energy, another for microphysics variables, another for passive chemical species) to enforce / ease this modularization?

They are modularized in the sense that we have fallback methods for (for example) `flux_first_order!(atmos.moisture, ...)` which do nothing when `atmos.moisture isa DryModel`, and do moisture / microphysics things for moist models.

glwagner:
> They are modularized in the sense that we have fallback methods for (for example) `flux_first_order!(atmos.moisture, ...)` which do nothing when `atmos.moisture isa DryModel`, and do moisture / microphysics things for moist models.

For sure, my questions is about "more" modularization (not necessarily functionally but just by writing the code differently, eg by using different files), which makes it easier for future developers to build further without impacting the readability of core stuff.

PR #1735, opened 2020-11-11
Update installation guide: use CLIMATEMACHINE_SETTINGS_DISABLE_GPU

valeriabarra:
This is to avoid the following warning:
```
Warning: CLIMATEMACHINE_GPU will be deprecated; use CLIMATEMACHINE_SETTINGS_DISABLE_GPU
```

### Description

### DIFF
--- a/docs/src/GettingStarted/Installation.md
+++ b/docs/src/GettingStarted/Installation.md
@@ -78,4 +78,15 @@ $ julia --project tutorials/Atmos/dry_rayleigh_benard.jl
 ```
 
 The `ClimateMachine` is CUDA-enabled and will use GPU(s) if available. To run
-on the CPU, set the environment variable `CLIMATEMACHINE_GPU` to `false`.
+on the CPU, set the environment variable `CLIMATEMACHINE_SETTINGS_DISABLE_GPU` to `true`.
+This can either be done inline with the Julia launch command using
+
+```
+CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true julia --project
+```
+
+or for the whole shell session, for example with `bash` this would be
+
+```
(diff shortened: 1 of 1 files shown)


docs/src/GettingStarted/Installation.md
```diff
@@ -78,4 +78,8 @@ $ julia --project tutorials/Atmos/dry_rayleigh_benard.jl
 ```
 
 The `ClimateMachine` is CUDA-enabled and will use GPU(s) if available. To run
-on the CPU, set the environment variable `CLIMATEMACHINE_GPU` to `false`.
+on the CPU, set the environment variable `CLIMATEMACHINE_SETTINGS_DISABLE_GPU` to `true` by running
+
+```
+$ export CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true
+```
```
jkozdon, line 85: Since this won't work in all shell (e.g, fish) might be better to say

> on the CPU, set the environment variable
> `CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true`. This can either be done by done
> inline with the Julia launch command
> ```
> CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true julia --project
> ```
> or for the whole shell session, for example with `bash` this would be
> ```
> export CLIMATEMACHINE_SETTINGS_DISABLE_GPU=true
> ```

(Suggestion wasn't displaying right)

jkozdon:
Also might be worth updating title of the PR and description so the git commit message looks ok.

Thanks for catching this!

docs/src/GettingStarted/Installation.md
valeriabarra: Oooh, thank you so much for the great suggestion!

jkozdon (approved):
Looks good to me once the commits are squashed.

Issue #1638, opened 2020-10-19
Sponge modifier is broken


charleskawczynski:
### Description

The [sponge modifier](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L1093-L1109) assumes that arguments `ν,D_t,τ` are passed by reference but, they're not. So, this `sponge_viscosity_modifier!` doesn't actually do or modify anything.

PR #1742, opened 2020-11-12
Add some second order fluxes to new tend spec

charleskawczynski:
### Description

This PR adds some of the second-order fluxes to the new tendency specification.

@simonbyrne one thing I had thought about was, rather than splatting in `args = (atmos, state, aux, t, ts, diffusive, hyperdiffusive)` to the `flux` and `source` functions, we could pass in a `NamedTuple`:
```julia
    nt = (atmos=atmos, state=state, aux=aux, t=t, ts=ts, diffusive=diffusive, hyperdiff=hyperdiffusive)
```
and then have:
```julia
function flux(::ViscousProduction{Energy}, nt)
    @unpack diffusive, aux, atmos, t = nt
    ν, D_t, τ = turbulence_tensors(atmos, state, diffusive, aux, t)
    ν, D_t, τ = sponge_viscosity_modifier(m, m.viscoussponge, ν, D_t, τ, aux)
    return τ * state.ρu
end
```

Pros of using `NamedTuple`s:
 - Compactify signature
 - Makes changing the effective arg list much easier

Cons of using `NamedTuple`s:
 - Need to look where `nt` was constructed to know what's available / what name is used

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -552,11 +552,18 @@ function. Contributions from subcomponents are then assembled (pointwise).
     aux::Vars,
     t::Real,
 )
+    ρu_pad = SVector(1, 1, 1)
+    ts = recover_thermo_state(atmos, state, aux)
+    tend = Flux{SecondOrder}()
+    args = (atmos, state, aux, t, ts, diffusive, hyperdiffusive)
+    flux.ρ = Σfluxes(eq_tends(Mass(), atmos, tend), args...) .* ρu_pad
+    flux.ρu = Σfluxes(eq_tends(Momentum(), atmos, tend), args...) .* ρu_pad
+    flux.ρe = Σfluxes(eq_tends(Energy(), atmos, tend), args...)
+
     ν, D_t, τ = turbulence_tensors(atmos, state, diffusive, aux, t)
     ν, D_t, τ =
         sponge_viscosity_modifier(atmos, atmos.viscoussponge, ν, D_t, τ, aux)
-    d_h_tot = -D_t .* diffusive.∇h_tot
-    flux_second_order!(atmos, flux, state, τ, d_h_tot)
+
     flux_second_order!(atmos.moisture, flux, state, diffusive, aux, t, D_t)
     flux_second_order!(atmos.precipitation, flux, state, diffusive, aux, t, D_t)
     flux_second_order!(
@@ -570,19 +577,6 @@ function. Contributions from subcomponents are then assembled (pointwise).
     )
     flux_second_order!(atmos.tracers, flux, state, diffusive, aux, t, D_t)
     flux_second_order!(atmos.turbconv, atmos, flux, state, diffusive, aux, t)
-end
-
-#TODO: Consider whether to not pass ρ and ρu (not state), foc BCs reasons
-@inline function flux_second_order!(
-    atmos::AtmosModel,
-    flux::Grad,
-    state::Vars,
-    τ,
-    d_h_tot,
-)
-    flux.ρu += τ * state.ρ
-    flux.ρe += τ * state.ρu
-    flux.ρe += d_h_tot * state.ρ
 end
 
 @inline function wavespeed(
--- a/src/Atmos/Model/atmos_tendencies.jl
+++ b/src/Atmos/Model/atmos_tendencies.jl
@@ -4,6 +4,10 @@
 
 import ..BalanceLaws: eq_tends
 
+#####
+##### Sources
+#####
+
 # --------- Some of these methods are generic or
 #           temporary during transition to new specification:
 filter_source(pv::PrognosticVariable, s) = nothing
@@ -24,6 +28,10 @@ eq_tends(pv::PrognosticVariable, m::AtmosModel, ::Source) =
     filter_sources(pv, m.source)
 # ---------
 
+#####
+##### First order fluxes
+#####
+
 # Mass
 eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Mass} =
     (Advect{PV}(),)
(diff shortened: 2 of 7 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -536,11 +536,18 @@ function. Contributions from subcomponents are then assembled (pointwise).
     aux::Vars,
     t::Real,
 )
+    ρu_pad = SVector(1, 1, 1)
+    ts = recover_thermo_state(atmos, state, aux)
+    tend = Flux{SecondOrder}()
+    args = (atmos, state, aux, t, ts, diffusive, hyperdiffusive)
+    flux.ρ = Σfluxes(eq_tends(Mass(), atmos, tend), args...) .* ρu_pad
+    flux.ρu = Σfluxes(eq_tends(Momentum(), atmos, tend), args...) .* ρu_pad
+    flux.ρe = Σfluxes(eq_tends(Energy(), atmos, tend), args...)
```
glwagner, line 545: My English translation of this line is:

> Set the energy flux, `flux.ρe`, to a sum of fluxes, `Σfluxes`, which depend on "`Energy` equation tendencies", `(eq_tends(Energy()...)`

The last bit is non-sensical to me though, so I can't quite figure out what the code is telling me. I guess a tendency typically depends on flux divergences in the sense that a generic equation can be written

```
∂t Q = - ∇ ⋅ F + S
```

So fluxes and tendencies are related. But the tendency doesn't go into the flux; instead, the flux divergence is one of the contributions to a quantity's tendency. Maybe I am missing something?

src/Atmos/Model/atmos_tendencies.jl
```diff
@@ -36,3 +44,29 @@ eq_tends(pv::PV, ::AtmosModel, ::Flux{FirstOrder}) where {PV <: Energy} =
...
     ()
+
+#####
+##### Second order fluxes
+#####
+
+# Mass
+moist_diffusion(pv::PV, ::DryModel) where {PV <: Mass} = ()
+moist_diffusion(pv::PV, ::MoistureModel) where {PV <: Mass} =
+    (MoistureDiffusion{PV}(),)
```
glwagner, line 55: This name was not introduced by this PR, but just a warning that a line like "MoistureDiffusion{PV}" is primed to confuse Earth scientists, because the term "PV" is often used as an abbreviation for ["potential vorticity"](https://en.wikipedia.org/wiki/Potential_vorticity). Most of the places where "PV" shows up there is enough context to infer that it refers to something other than potential vorticity. But in lines like these (and lines like `Advect{PV}` it's not quite so clear. I certainly did a double-take, myself, even though I know julia...

src/Atmos/Model/tendencies_momentum.jl
```diff
@@ -16,6 +16,26 @@ function flux(::PressureGradient{Momentum}, m, state, aux, t, ts, direction)
...
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    diffusive,
+    hyperdiff,
+)
+    ν, D_t, τ = turbulence_tensors(m, state, diffusive, aux, t)
+    ν, D_t, τ = sponge_viscosity_modifier(m, m.viscoussponge, ν, D_t, τ, aux)
```
glwagner, line 35: Does this term mutate `τ`? Could we write `modify_turbulence_tensors!(τ, args...)` if so? It doesn't look like `ν, D_t` are used after this, why does `sponge_viscosity_modifier` return them?

glwagner (approved):
My review is not very high-quality, but everything looks good. I just had a few comments that can be ignored.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I think you're right, `eq_tends` technically returns `TendencyDef{TT}`'s, where `TT <: Union{Flux{AbstractOrder},Source}`, I think part of the naming challenge is that there's not really a good "supertype" name for both a flux, whose tendency is the divergence of the flux, and the source, which is a tendency itself. All that said, this is the way it is because we couldn't come up with better names.

src/Atmos/Model/atmos_tendencies.jl
charleskawczynski: Do you, or anyone else (cc @yairchn, @trontrytel) have a better name for this? What's wrong with `Advect`? I'm happy to change it/them. Regarding `PV`, it's the most sensible short type name for `PrognosticVariable` (the abstract type of `Mass`, `TotalMoisture` etc.) I think, but again, I'm open to changing. Perhaps changing `PV` and `Advect` would be better for a separate PR, as that would need to happen across the board--and it would be nice to see how everything changes at once.

src/Atmos/Model/tendencies_momentum.jl
charleskawczynski: It does, and we used to. But they're scalars, and Julia [does not pass scalars by reference](https://github.com/CliMA/ClimateMachine.jl/issues/1638).

src/Atmos/Model/atmos_tendencies.jl
glwagner: It's just PV that conflicts with other abbreviations "PV". I wouldn't change it now, just something to keep in the back of your mind so if you see an opportunity to change it, you can.

trontrytel: I agree. I also think of vorticity first when I see PV. But I don't have a good 2-letter replacement for this. So I got used to it.

kpamnany: Use `PVAR`, there's no reason it has to be 2 letters.

charleskawczynski: Maybe we can change to `PVAR` across the board in a separate PR?

trontrytel (approved):
LGTM and it works with drizzling dycoms experiment.

PR #1745, opened 2020-11-12
Remove unpack_fields, use UnPack package

charleskawczynski:
### Description

This PR removes `@unpack_fields` and replaces it with `@unpack` from [UnPack.jl](https://github.com/mauro3/UnPack.jl) (which has been added).

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -517,10 +517,10 @@ uuid = "739be429-bea8-5141-9913-cc70e7f3736d"
 version = "1.0.2"
 
 [[MbedTLS_jll]]
-deps = ["Libdl", "Pkg"]
-git-tree-sha1 = "c0b1286883cac4e2b617539de41111e0776d02e8"
+deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
+git-tree-sha1 = "0eef589dd1c26a3ac9d753fe1a8bcad63f956fa6"
 uuid = "c8ffd9c3-330d-5841-b78e-0817d7145fa1"
-version = "2.16.8+0"
+version = "2.16.8+1"
 
 [[MicrosoftMPI_jll]]
 deps = ["Libdl", "Pkg"]
--- a/Project.toml
+++ b/Project.toml
@@ -44,6 +44,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StaticNumbers = "c5e4b96a-f99f-5557-8ed2-dc63ef9b5131"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [compat]
--- a/docs/src/HowToGuides/Common/Thermodynamics.md
+++ b/docs/src/HowToGuides/Common/Thermodynamics.md
@@ -148,6 +148,7 @@ Thermodynamics.jl is tested using a set of profiles specified in `test/Common/Th
 ```@example
 using ClimateMachine.Thermodynamics
 using ClimateMachine.TemperatureProfiles
+using UnPack
 using CLIMAParameters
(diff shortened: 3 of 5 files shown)


src/Utilities/UnpackFields/UnpackFields.jl
```diff
@@ -0,0 +1,45 @@
...
+macro unpack_fields(obj, syms...)
+    thunk = Expr(:block)
+    for sym in syms
+        push!(
+            thunk.args,
+            :($(esc(sym)) = getproperty($(esc(obj)), $(QuoteNode(sym)))),
+        )
+    end
+    push!(thunk.args, nothing)
+    return thunk
```
kpamnany, line 42: Alternatively:
```suggestion
    exs = [
        :($(esc(sym)) = getproperty($(esc(obj)), $(QuoteNode(sym))))
        for sym in syms
    ]
    Expr(:block, exs...)
```
But what you have is perhaps clearer.

test/Utilities/UnpackFields/runtests.jl
```diff
@@ -0,0 +1,15 @@
+using Test
+using ClimateMachine.UnpackFields
+
+@testset "UnpackFields" begin
+    struct Foo
+        a::Any
+        b::Any
+        c::Any
+    end
+    f = Foo(1, 2, 3)
```
kpamnany, line 10:
```suggestion
    f = Foo(1, 2, 3)
    @test_throws UndefVarError a == 1
```
Etc.

kpamnany (commented):
I'm not sure this is a "good" utility because it pollutes the namespace. Maybe document that it isn't for general use?

src/Utilities/UnpackFields/UnpackFields.jl
charleskawczynski: Hm, good point. Let's go with your suggestion.

charleskawczynski:
> I'm not sure this is a "good" utility because it pollutes the namespace. Maybe document that it isn't for general use?

How so? I've put it inside `ClimateMachine.UnpackFields` module so that it's only used when we explicitly use it. Do you think we should move it somewhere else? Or just define it locally when needed? All that said, I'm certainly fine putting a note in the docs that it's not intended for general use--I'd just like to understand the reasoning for why it wouldn't be recommended.

kpamnany:
> How so? I've put it inside `ClimateMachine.UnpackFields` module so that it's only used when we explicitly use it. Do you think we should move it somewhere else? Or just define it locally when needed? All that said, I'm certainly fine putting a note in the docs that it's not intended for general use--I'd just like to understand the reasoning for why it wouldn't be recommended.

So, this is an intentionally [non-hygienic macro](https://en.wikipedia.org/wiki/Hygienic_macro). Julia (and LISP, et al.) specifically provide [hygienic macros](https://docs.julialang.org/en/v1/manual/metaprogramming/#Hygiene). Since this macro is _intentionally_ violating hygiene, it is probably fine but I thought a second opinion would be good...

charleskawczynski:
Ahhh, ok. Thanks for the explanation! I'll add a note in the docs. Also, is there a way to make this macro hygienic?

kpamnany:
> Ahhh, ok. Thanks for the explanation! I'll add a note in the docs. Also, is there a way to make this macro hygienic?

No -- you're introducing new variables with the macro that take their names from the field names. They could easily overwrite existing variables in the macro calling scope. You _could_ check if those variable names already exist and throw an error, but that's getting a bit silly for a small helper like this...

simonbyrne:
Would we be better off just using https://github.com/mauro3/UnPack.jl (which used to be part of https://github.com/mauro3/Parameters.jl)? I like its notation a bit better (`@unpack a,c = f` vs `@unpack_fields f a c`) as it's clearer what variables are being defined.

charleskawczynski:
Ok, I added a warning to the docs.

charleskawczynski:
> Would we be better off just using https://github.com/mauro3/UnPack.jl (which used to be part of https://github.com/mauro3/Parameters.jl)? I like its notation a bit better (`@unpack a,c = f` vs `@unpack_fields f a c`) as it's clearer what variables are being defined.

Ah, yes, I'm fine with this :)

charleskawczynski:
Much nicer!

kpamnany (approved):

PR #1748, opened 2020-11-12
Update system image instructions

jkozdon:
- Adds scripts for building the system images for the julia formatter
- Adds examples for how to call the system image script for `ClimateMachine`

PR #1749, opened 2020-11-12
Add `docs` and `.dev` directories to the environment to `CompatHelper`

jkozdon:
Add `docs` and `.dev` directories to the environment to `CompatHelper`

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -16,4 +16,4 @@ jobs:
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: julia -e 'using CompatHelper; CompatHelper.main()'
+        run: julia -e 'using CompatHelper; CompatHelper.main(; subdirs=["", "docs", ".dev"])'


jkozdon:
If this is merged, then the scripts in #1748 should be updated (as they currently deletes the `.dev/Manifest.toml`)

kpamnany:
I clearly still don't grok Pkg. Doesn't the `[compat]` entry in `.dev/Project.toml` pin the version? How is it that there can be formatter version changes?

.github/workflows/CompatHelper.yml
```diff
@@ -16,4 +16,4 @@ jobs:
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: julia -e 'using CompatHelper; CompatHelper.main()'
+        run: julia -e 'using CompatHelper; CompatHelper.main(; subdirs=["", ".dev"])'
```
simonbyrne, line 19:
```suggestion
        run: julia -e 'using CompatHelper; CompatHelper.main(; subdirs=["", "docs", ".dev"])'
```

jkozdon:
> I clearly still don't grok Pkg. Doesn't the `[compat]` entry in `.dev/Project.toml` pin the version? How is it that there can be formatter version changes?

Hmm... I guess you're right. So maybe be don't need the manifest just to add `.dev` to the `Compathelper`? (I guess I don't know enough about how the `CompatHelper` works.

Also, I guess it depends on whether we want to catch minor updates to `JuliaFormatter` or only major updates. Right now our compat is `0.10`, so nothing would change between `0.10.7` and `0.10.8` (which in theory should be fine I suppose).

charleskawczynski:
I'm fine with this solution if it makes sense, or adding documentation that the `.dev/Manifest.toml` must be in sync with master before rebuilding the sys image.

kpamnany:
I'm pretty sure we don't need a manifest but... this is quite strange. `Pkg.instantiate` is maybe not working correctly?
```
login2:~/ClimateMachine.jl$ julia --project=.dev
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.2 (2020-09-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(.dev) pkg> status
Status `/central/home/kpamnany/ClimateMachine.jl/.dev/Project.toml`
  [98e50ef6] JuliaFormatter v0.3.9

(.dev) pkg> instantiate

(.dev) pkg> status
Status `/central/home/kpamnany/ClimateMachine.jl/.dev/Project.toml`
  [98e50ef6] JuliaFormatter v0.3.9

julia>
login2:~/ClimateMachine.jl$ cat .dev/Project.toml
[deps]
JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"

[compat]
JuliaFormatter = "0.10"
```
:eyes:

It should have installed `0.10` something according to this:
```julia
help?> Pkg.instantiate
  Pkg.instantiate(; verbose = false)

  If a Manifest.toml file exists in the active project, download all the packages declared in that manifest.
  Otherwise, resolve a set of feasible packages from the Project.toml files and install them. verbose = true
  prints the build output to stdout/stderr instead of redirecting to the build.log file. If no Project.toml
  exist in the current active project, create one with all the dependencies in the manifest and instantiate
  the resulting project.
```
But [here](https://julialang.github.io/Pkg.jl/v1/compatibility/#Compatibility) it says:
```
After a compatibility entry is put into the project file, `up` can be used to apply it.
```
So now I'm wondering if we need a `Pkg.update()` after [this line](https://github.com/CliMA/ClimateMachine.jl/blob/jek/formatter/.dev/climaformat.jl#L30)...

Cc: @vchuravy

jkozdon:
> I'm pretty sure we don't need a manifest

OK. Made the PR just add `docs` and `.dev` directories to the environment to `CompatHelper`

charleskawczynski (approved):
LGTM

jkozdon:
Do we still want this?

PR #1750, opened 2020-11-12
Move RayleighSponge to new tendency specification

charleskawczynski:
### Description

This PR moves the RayleighSponge to the new tendency specification.

### DIFF
--- a/experiments/AtmosLES/dycoms.jl
+++ b/experiments/AtmosLES/dycoms.jl
@@ -281,7 +281,7 @@ function config_dycoms(FT, N, resolution, xmax, ymax, zmax)
     # Rayleigh damping
     zsponge = FT(1000.0)
     rayleigh_sponge =
-        RayleighSponge{FT}(zmax, zsponge, c_sponge, u_relaxation, 2)
+        RayleighSponge(FT, zmax, zsponge, c_sponge, u_relaxation, 2)
     # Geostrophic forcing
     geostrophic_forcing =
         GeostrophicForcing(FT, f_coriolis, u_geostrophic, v_geostrophic)
--- a/experiments/AtmosLES/schar_scalar_advection.jl
+++ b/experiments/AtmosLES/schar_scalar_advection.jl
@@ -156,7 +156,7 @@ function config_schar(FT, N, resolution, xmax, ymax, zmax)
 
     ## Pass the sponge parameters to the sponge calculator
     rayleigh_sponge =
-        RayleighSponge{FT}(zmax, z_s, sponge_ampz, u_relaxation, 2)
+        RayleighSponge(FT, zmax, z_s, sponge_ampz, u_relaxation, 2)
 
     ## Define the time integrator:
     ## We chose an explicit single-rate LSRK144 for this problem
--- a/src/Atmos/Model/atmos_tendencies.jl
+++ b/src/Atmos/Model/atmos_tendencies.jl
@@ -12,6 +12,7 @@ filter_source(pv::PV, s::Subsidence{PV}) where {PV <: PrognosticVariable} = s
 filter_source(pv::PV, s::Gravity{PV}) where {PV <: Momentum} = s
 filter_source(pv::PV, s::GeostrophicForcing{PV}) where {PV <: Momentum} = s
 filter_source(pv::PV, s::Coriolis{PV}) where {PV <: Momentum} = s
+filter_source(pv::PV, s::RayleighSponge{PV}) where {PV <: Momentum} = s
 
 # Filter sources / empty elements
 filter_sources(t::Tuple) = filter(x -> !(x == nothing), t)
--- a/src/Atmos/Model/source.jl
+++ b/src/Atmos/Model/source.jl
@@ -1,8 +1,7 @@
 using ..Microphysics_0M
 using CLIMAParameters.Planet: Omega, e_int_i0, cv_l, cv_i, T_0
 
-export AbstractSource,
-    RayleighSponge, GeostrophicForcing, RemovePrecipitation, CreateClouds
+export AbstractSource, RemovePrecipitation, CreateClouds
 
 # sources are applied additively
 @generated function atmos_source!(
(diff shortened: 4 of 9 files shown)


glwagner (approved):
This looks good.

Is there a user-interface for specifying forcing functions?

For example a similar interface in Oceananigans implements sponge layers using the existing user interface (rather than defining a new kind of source function, eg `source(s::RayleighSponge, m, ...)`.

User-defined forcing is here:

https://github.com/CliMA/Oceananigans.jl/blob/03a6f855f839504d94cb8cee3c2665b17afbc6d5/src/Forcings/continuous_forcing.jl#L48

which allows a concise implementation of "relaxation" via

```julia
@inline (f::Relaxation)(x, y, z, t, field) =
    f.rate * f.mask(x, y, z) * (f.target(x, y, z, t) - field)
```

charleskawczynski:
> This looks good.
> 
> Is there a user-interface for specifying forcing functions?

Yes. For example, in the Held Suarez experiment:
```julia
struct HeldSuarezForcing <: AbstractSource end

function atmos_source!(
    ::HeldSuarezForcing,
    bl::AtmosModel,
    source::Vars,
    state::Vars,
    diffusive::Vars,
    aux::Vars,
    t::Real,
    direction,
)
    FT = eltype(state)

    # Parameters
    T_ref = FT(255)

    # Extract the state
    ρ = state.ρ
    ρu = state.ρu
    ρe = state.ρe
    ...
    source.ρu -= k_v * projection_tangential(bl, aux, ρu)
    source.ρe -= k_T * ρ * _cv_d * (T - T_equil)
```
And then you pass in the source with `AtmosModel(; source=(HeldSuarezForcing(),...))`

To note `aux` typically has `coords`, which has `x,y,z`. Many of the sources are more complex than simply providing `x,y,z,t`, so our interface is a bit more cumbersome. We might be able to make a cleaner interface for simpler sources, though.

glwagner:
> And then you pass in the source with AtmosModel(; source=(HeldSuarezForcing(),...))

I think what we should go for is a system where source terms are functions that return the source term, rather than thinking of a "source term" as something that mutates a state --- this mirrors what a "source term" is when writing an equation on paper, eg

```julia
heating(φ, θ, z, p) = p.H * exp(-z / p.λ) * cos(φ)
```

It's a bit harder both to use and understand the code if the source functions have to mutate their arguments, since you need to understand more about the underlying structure of the code and naming conventions (and also makes the specification a bit longer).

This is the kind of interface I plan to implement for the ocean model.

charleskawczynski:
> I think what we should go for is a system where source terms are functions that return the source term, rather than thinking of a "source term" as something that mutates a state

That's the direction we're going with this new spec.

```julia
source(
    s::RayleighSponge{Momentum},
    m,
    state,
    aux,
    t,
    ts,
    direction,
    diffusive,
)
```
is a function that returns the source. We just don't yet have fallbacks in `source!`, `flux_first_order!` and `flux_second_order!` to call `flux` and `source` on the tuple returned by `eq_tends` so that users only need to specify the type and overload `source`/`flux`. But, I think (and hope) that's where we're headed.

charleskawczynski:
We're running into loop unrolling problems ☹️. Looks like we'll need the generated functions after all.

charleskawczynski:
Ok, we got this thing working, big thanks to @vchuravy and @leios. We can unroll a loop of at least 209 sources (if needed 😆) 🎉. I took the MWE we worked on, and got things working in this branch but it needs a bit of cleaning/refactoring before we can merge (I had to add a few hard-coded cases which I think should generalize).

charleskawczynski:
# Summary

This has been a roller coaster of debugging, but I want to record the story here. The original PR commit failed on the GPU and, as this PR is adding more sources, I figured that this was a likely suspect since the original PR commit ran fine on the CPU. The motivating example here:
```julia
g6() = sum(ntuple(i->1, 6))
g7() = sum(ntuple(i->1, 7))
@code_typed g6()
@code_typed g7()
```
For an `NTuple` of size 6, the compiler can fully infer the result, but for size 7, it starts calling `Base.add_int`, which I suspected was the issue. We can fully infer much larger `NTuple`s with @vchuravy's magic unrolling:
```julia
@generated function unroll_sum(args::NTuple{N,T}) where {N, T}
  exprs = []
  var = gensym()
  push!(exprs, :($var = args[1]))
  for i in 2:N
     nvar = gensym()
     if T <: Real
         push!(exprs, :($nvar = $var + args[$i]))
     elseif T <: AbstractArray
         push!(exprs, :($nvar = $var .+ args[$i]))
     end
     var = nvar
  end
  push!(exprs, :(return @inbounds $var))
  return Expr(:block, exprs...)
end
f6() = unroll_sum(ntuple(i->1, 6))
f7() = unroll_sum(ntuple(i->1, 7))
@code_typed f6()
@code_typed f7()
```

It turns out that this was not the issue, however, but using this loop unrolling method helped track down the issue because `unroll_sum` requires more strictly typed inputs compared to `sum`. Here are the examples that led to tracking things down.

## Example 1
Using
```julia
pad = similar(state.ρu .* (state.ρu / state.ρ)')
fill!(pad, 0)
```
in `flux(::PressureGradient{Momentum}` produces an `MArray`. This means that `flux(::PressureGradient{Momentum}` yields an `MArray` while `flux(::Advect{Momentum}` yields an `SArray`.

So,
```julia
ntuple(Val(N)) do i
    flux(fluxes[i], args...)
end
```
resulted in a mixed `MArray`/`SArray` `NTuple`. This
 - ran on the CPU with `sum`
 - failed with `unroll_sum` on the CPU (method error, since the result was a `Tuple` and not an `NTuple`)
 - failed on the GPU (dynamic runtime call error).

## Example 2
Similarly, using
```julia
pad = (state.ρu .* (state.ρu / state.ρ)') * 0
...
return (air_pressure(ts) - aux.ref_state.p) * I
```
resulted in a `UniformScaling{FT}` which also
 - ran on the CPU with `sum`
 - failed with `unroll_sum` on the CPU (method error, since the result was a `Tuple` and not an `NTuple`)
 - failed on the GPU (dynamic runtime call error).

## Example 3 (which actually fixed the GPU errors)
Using, in the `else`-case of `source(s::RayleighSponge{Momentum}, ...)`,
```julia
  return SVector(0, 0, 0)
```
resulted in a `SArray{Tuple{3},Int64,1,3}` which
 - ran on the CPU with `sum`
 - failed with `unroll_sum` on the CPU (method error, since the result was a `Tuple` and not an `NTuple`)
 - _ran_ on the GPU.

## Conclusions

It turned out that, IIRC, fixing the type instabilities in examples 1 and 2 are not actually needed to run on the GPU. But, I had to fix the first two examples to find the breaking one (example 3). Tracking this down took more human hours to solve than it should have and, while we don't need this loop unrolling function, I think it may be beneficial to add and use a more restrictive sum function:

```julia
ntuple_sum(nt::NTuple{N,T}) where {N, T} = sum(nt)
```
simply to better synchronize the success/failure of CPU/GPU runs and help catch these errors earlier.

I'll added this function for now, but I'd like feedback before merging.

PR #1741, opened 2020-11-12
Interpolation -> KA kernels + Variable polynomial orders.

skandalaCLIMA:
1). Replaced separate CUDA and CPU kernels with KernelAbstractions kernels.

2). Enabled support for different polynomial orders in different
directions.

### Description

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1702, opened 2020-11-05
Kernel infrastructure for supporting variable-degree data structures

thomasgibson:
### Description

PR which modifies the compute kernels to support variable degree data structures. This is necessary for utilizing Clima in a DG-FV mode, or in a DG configuration with different polynomial orders in the horizontal/vertical directions.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1751, opened 2020-11-13
Driver update: Variable polynomial order 

valeriabarra:
### Description
This PR is part of a series of PRs (see PR #1702 and PR #1741 ) to expand the DG kernels capability to support multiple degree polynomials in the horizontal and vertical directions. 

This PRcan be merged independently of the above mentioned PRs, as it only touches the Driver configurations. 

### DIFF
--- a/src/Driver/driver_configs.jl
+++ b/src/Driver/driver_configs.jl
@@ -38,7 +38,8 @@ struct DriverConfiguration{FT}
     config_type::ClimateMachineConfigType
 
     name::String
-    N::Int
+    # polynomial order tuple (polyorder_horz, polyorder_vert)
+    polyorders::NTuple{2, Int}
     array_type::Any
     solver_type::AbstractSolverType
     #
@@ -63,7 +64,7 @@ struct DriverConfiguration{FT}
     function DriverConfiguration(
         config_type,
         name::String,
-        N::Int,
+        polyorders::NTuple{2, Int},
         FT,
         array_type,
         solver_type::AbstractSolverType,
@@ -79,7 +80,7 @@ struct DriverConfiguration{FT}
         return new{FT}(
             config_type,
             name,
-            N,
+            polyorders,
             array_type,
             solver_type,
             param_set,
@@ -110,7 +111,7 @@ end
 
 function AtmosLESConfiguration(
     name::String,
-    N::Int,
+    (polyorder_horz, polyorder_vert)::NTuple{2, Int},
     (Δx, Δy, Δz)::NTuple{3, FT},
     xmax::FT,
     ymax::FT,
@@ -142,9 +143,9 @@ function AtmosLESConfiguration(
     print_model_info(model)
 
     brickrange = (
-        grid1d(xmin, xmax, elemsize = Δx * N),
-        grid1d(ymin, ymax, elemsize = Δy * N),
-        grid1d(zmin, zmax, elemsize = Δz * N),
+        grid1d(xmin, xmax, elemsize = Δx * polyorder_horz),
+        grid1d(ymin, ymax, elemsize = Δy * polyorder_horz),
+        grid1d(zmin, zmax, elemsize = Δz * polyorder_vert),
     )
     topology = StackedBrickTopology(
         mpicomm,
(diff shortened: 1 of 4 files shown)


src/Driver/driver_configs.jl
```diff
@@ -407,12 +414,7 @@ function OceanSplitExplicitConfiguration(
 
     Q_2D = init_ode_state(dg_2D, FT(0); init_on_cpu = true)
 
-    # XXX: Needs updating for multiple polynomial orders
-    N = polynomialorders(grid_3D)
-    # Currently only support single polynomial order
-    @assert all(N[1] .== N)
-    N = N[1]
-    vert_filter = CutoffFilter(grid_3D, N - 1)
+    vert_filter = CutoffFilter(grid_3D, polyorder_vert - 1)
```
valeriabarra, line 417: @blallen I requested your review because I wasn't extremely sure how to handle this case for the `OceanSplitExplicitConfiguration`, since it appeared to be slightly different than the other driver configurations for the `Atmos` simulations. If we understood correctly (i.e., @thomasgibson and I) this `CutoffFilter` is applied only in the vertical, so I used the `polyorder_vert` degree here. Thanks!

src/Driver/driver_configs.jl
```diff
@@ -110,7 +113,7 @@ end
 
 function AtmosLESConfiguration(
     name::String,
-    N::Int,
+    (polyorder_horz, polyorder_vert)::NTuple{2, Int},
```
jkozdon, line 116: If you left this as `N` or maybe `N::Union{Int, NTuple{2, Int})` and then add after the `where` the following
```julia
(polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
```
then you can support both single `N` and `(polyorder_horz, polyorder_vert)`

jkozdon (commented):
Thanks for taking this on. 

I am wondering if we should support the behavior of specifying a single `N` which would be `N` being the same every

src/Driver/driver_configs.jl
jkozdon: I believe this is correct, but will let @blallen weigh in.

src/Driver/driver_configs.jl
```diff
@@ -76,10 +77,12 @@ struct DriverConfiguration{FT}
         numerical_flux_gradient::NumericalFluxGradient,
         config_info::ConfigSpecificInfo,
     )
+        # FIXME: Once variable degree kernels are merged, remove this assert
+        @assert polyorders[1] === polyorders[2]
```
thomasgibson, line 81: I just realized this is probably not a good idea, since `===` means that the two things you're comparing are the _same_ object (i.e. point to the same location in memory, but what we really want is `==` which compares by value. Is this right @jkozdon ?

jkozdon: In this case it doesn't matter since you are comparing two integers not two arrays. With arrays, it is true that `===` compares the pointers and `==` compares the elements of the arrays

src/Driver/driver_configs.jl
valeriabarra: Thank you so much for the suggestion! I just changed it. Let me know if it looks good now.

src/Driver/driver_configs.jl
valeriabarra: I changed it to `==`. Although it didn't make a difference for this case, I also agree that for integers comparison it's better to use `==`

valeriabarra:
bors try

valeriabarra:
bors try

src/Driver/driver_configs.jl
jkozdon: Should we do this for all the configs?

valeriabarra: Yup, absolutely. I was just making sure I got your suggestion right first. I just added for all configs

valeriabarra:
By the way, I'm a bit puzzled by the failures in this PR. Since tests pass locally for me

src/Driver/driver_configs.jl
blallen: looks good to me!

blallen (approved):
Good job!

valeriabarra:
> By the way, I'm a bit puzzled by the failures in this PR. Since tests pass locally for me

I'll take that back. I've just noticed the failing test is `mms3.jl` which is one I did not update

valeriabarra:
I can squash this down to a minimum number of commits to keep the different authorship (@thomasgibson 's and mine), or, if preferred just squash to one commit only before merge

valeriabarra:
bors try

blallen:
> I can squash this down to a minimum number of commits to keep the different authorship (@thomasgibson 's and mine), or, if preferred just squash to one commit only before merge

You can add co-authors to commits, in this case just add
```
Co-authored-by: Thomas H. Gibson <thomas.gibson@nps.edu>
```
to the commit message.

valeriabarra:
Yes, thanks. I am aware of that. I think that does not add a contribution count to the [contribution counts](https://github.com/CliMA/ClimateMachine.jl/graphs/contributors) though. Some people, especially in early-career stages, do care about that. So I always ask the person in question if they do care to keep their contribution as separate commits in a particular PR, to be polite.

In my opinion, not all merges need to be necessarily squashed to a single commit. Some PRs that are more involved (definitely not this one) may warrant a minimal number of organized commits greater than 1. But I adapt to whatever is the practice in any project :smiley:

jkozdon:
> In my opinion, not all merges need to be necessarily squashed to a single commit. Some PRs that are more involved (definitely not this one) may warrant a minimal number of organized commits greater than 1. But I adapt to whatever is the practice in any project 😃

We generally allow it to be up to the developer of the PR. Most are doing full squashes, but a few of us like too curate into a number of smaller logical commits.

valeriabarra:
> Thanks for taking this on.
> 
> I am wondering if we should support the behavior of specifying a single `N` which would be `N` being the same every

@jkozdon what did you mean with `N` being the same every... ? Sorry, perhaps I did not fully grasp the suggestion.

jkozdon:
Sorry, meant to say:

> I am wondering if we should support the behavior of specifying a single `N` which would be `N` being the same every DIRECTION

essentially what you ad to support `N`  becomes `(N, N)`

jkozdon (approved):

PR #1752, opened 2020-11-13
Support for metrics with polynomial order `N = 0` elements

jkozdon:
Add support for `N = 0` geometry. This is a necessary step for support of finite
volume methods; finite volumes will be in a seperate PR once #1702 lands.

The metrics terms are constructed by first constructing the metrics with `N = 1`
and then these metric terms are interpolated / selected to the `N = 0` case in a
way that ensures conservation and constant preserving properties.

Tests have been added to ensure the correctness and constant preserving
constraints are satisfied.

Inluded in this PR are a number refactors of the metrics which enables this
within the current `DiscontinuousSpectralElementGrid`.


jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try-

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
VTK doesn't work yet to `N=0`. Need to save a special mesh in this case (or something like this) for plotting.

jkozdon:
This PR now sits on top of #1791. This is needed to support VTK with N = 0 (which is the only outstanding issue for the PR)

jkozdon:
I believe that this is now done, and ready for review.

thomasgibson (approved):
Thanks for doing this! I looked over the main changes in `Grids.jl` and it looks consistent with what you wrote in your notes! Also thanks for cleaning up some of the stale code in the metric file.

jkozdon:
> Thanks for doing this! I looked over the main changes in `Grids.jl` and it looks consistent with what you wrote in your notes! Also thanks for cleaning up some of the stale code in the metric file.

No problem.

Since this is needed for #1759 (FVM reconstructions), should we merge this or wait until you've tested against #1702 (multiple polynomial order DG kernels)

src/Numerics/Mesh/Grids.jl
```diff
@@ -678,13 +835,42 @@ function computegeometry(
...
+
+    (
+        #! format: off
+        ξ1x1, ξ2x1, ξ3x1, ξ1x2, ξ2x2, ξ3x2, ξ1x3, ξ2x3, ξ3x3,
+        MJ, MJI, MHJH,
+        x1, x2, x3,
+        JcV,
+       #! format: on
+    ) = ntuple(j -> (@view vgeo[:, j, :]), _nvgeo)
+    J = MJ ./ M[:]
```
Zhengyu-Huang, line 867: What is MJ, is it updated before in the computegeometry_fvm?

src/Numerics/Mesh/Grids.jl
```diff
@@ -697,14 +883,10 @@ function computegeometry(
...
-            ξ2x1,
-            ξ2x2,
-            ξ2x3,
+            #! format: off
+            JcV, J,
+            ξ1x1, ξ1x2, ξ1x3, ξ2x1, ξ2x2, ξ2x3,
+            #! format: on
         ) do J, ξ1x1, ξ1x2, ξ1x3, ξ2x1, ξ2x2, ξ2x3
             x1ξ3 = J * (ξ1x2 * ξ2x3 - ξ2x2 * ξ1x3)
             x2ξ3 = J * (ξ1x3 * ξ2x1 - ξ2x3 * ξ1x1)
```
Zhengyu-Huang, line 892: What is this map! (...) do ...  ?  Why is there JcV in map!(...) but not in do(...) ?

src/Numerics/Mesh/Grids.jl
```diff
@@ -570,106 +591,242 @@ function commmapping(N, commelems, commfaces, nabrtocomm)
...
+        num_vgeo_handled += 2
+
+        # Make sure we handled all the vgeo terms
+        @assert _nvgeo == num_vgeo_handled
+    end
+
+    # Sort out the sgeo terms
+    @views begin
+        sgeo = zeros(FT, _nsgeo, maximum(Nfp), nface, nelem)
+
```
Zhengyu-Huang, line 669: Is Nfp the number of points in each face?

src/Numerics/Mesh/Grids.jl
```diff
@@ -570,106 +591,242 @@ function commmapping(N, commelems, commfaces, nabrtocomm)
...
+        elseif dim == 3
+            fmask = (
+                p[1, :, :][:],
+                p[Nq[1], :, :][:],
+                p[:, 1, :][:],
+                p[:, Nq[2], :][:],
+                p[:, :, 1][:],
+                p[:, :, Nq[3]][:],
+            )
+        end
```
Zhengyu-Huang, line 686: are these fmask, ordered as x direction flux faces, y direction flux faces, z direction flux faces? For example, I have N = [2;3;0], I have fmask ([1, 4, 7, 10], [3, 6, 9, 12], [1, 2, 3], [10, 11, 12], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]), what does this mean? is fmask used to map the face points to the volume points to correct MI?

src/Numerics/Mesh/Grids.jl
```diff
@@ -570,106 +591,242 @@ function commmapping(N, commelems, commfaces, nabrtocomm)
...
+                if Nfp[d] == Nfp_N1[d]
+                    sgeo[:, 1:Nfp[d], f, :] .= sgeo_N1[:, 1:Nfp[d], f, :]
+
+                    # Volume inverse mass will be wrong so reset it
+                    sgeo[_vMI, 1:Nfp[d], f, :] .= MI[fmask[f], :]
+                else
+                    # Counter to make sure we got all the sgeo terms
+                    num_sgeo_handled = 0
+
+                    # sum to get sM
```
Zhengyu-Huang, line 700: What are sM and normal.

are they related to  n ds = det(J) * J^-T * N dS,  J =∂x/∂ξ and N is the outward normal in the reference element? 

Is it because det(J) is linear and J^-T and N are constant with respect to the p0 direction coordinate, we can do summation ?

Zhengyu-Huang (approved):
That looks great!!!! Thank you for the effort! I agree this should be merged first, and the tests with multiple polynomial order can come later!!! Then we have a p0 for all direction scheme, GCM test team can run some GCM tests.

src/InputOutput/VTK/writevtk.jl
```diff
@@ -132,10 +132,33 @@ function writevtk_helper(
...
 
+    # If any dimension are N = 0 we mirror these out to the boundaries for viz
+    # purposed
+    if any(N .== 0)
+        Nq_N1 = max.(Nq, 2)
+        Np_N1 = prod(Nq_N1)
+        ind = ntuple(i -> N[i] == 0 ? [1, 1] : Colon(), dim)
+        fields = ntuple(length(fields)) do i
+            fld = reshape(fields[i], Nq..., nelem)
+            fld2 = @view fld[ind..., :]
```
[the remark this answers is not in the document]
jkozdon: yes (this is just for viz purposes).

src/InputOutput/VTK/writevtk.jl
```diff
@@ -132,10 +132,33 @@ function writevtk_helper(
...
     if number_sample_points > 0
         FT = eltype(Q)
-        ξ = referencepoints(grid)
+        # If any dimension are N = 0 we manual set (-1, 1) grids
+        ξ = ntuple(
+            i -> N[i] == 0 ? FT.([-1, 1]) : referencepoints(grid)[i],
+            dim,
+        )
         ξdst = range(FT(-1); length = number_sample_points, stop = 1)
         I1d = ntuple(i -> interpolationmatrix(ξ[dim - i + 1], ξdst), dim)
```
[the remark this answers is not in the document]
jkozdon: Yes. This is needed for the high-order viz type in VTK.

src/Numerics/Mesh/Grids.jl
```diff
@@ -225,7 +232,12 @@ struct DiscontinuousSpectralElementGrid{
...
 
         # Create element operators for each polynomial order
-        ξω = ntuple(j -> Elements.lglpoints(FloatType, N[j]), dim)
+        ξω = ntuple(
+            j ->
+                N[j] == 0 ? Elements.glpoints(FloatType, N[j]) :
+                Elements.lglpoints(FloatType, N[j]),
+            dim,
+        )
         ξ, ω = ntuple(j -> map(x -> x[j], ξω), 2)
```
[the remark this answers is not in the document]
jkozdon: Yeah, this is needed because you cannot broadcast to multiple tuples.

src/Numerics/Mesh/Grids.jl
```diff
@@ -570,106 +591,242 @@ function commmapping(N, commelems, commfaces, nabrtocomm)
...
+        vgeo[:, _MI, :] = 1 ./ vgeo[:, _M, :]
+        num_vgeo_handled += 1
+
+        # coordinates should just be averages
+        avg_den = 2 .^ sum(Nq .== 1)
+        for fld in (_x1, _x2, _x3)
+            vgeo[:, fld, :] =
+                sum(vgeo_N1_flds[fld], dims = findall(Nq .== 1))[:] / avg_den
+            num_vgeo_handled += 1
+        end
```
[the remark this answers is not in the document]
jkozdon: In the mesh yes. Not in the kernels though.

src/Numerics/Mesh/Grids.jl
jkozdon: Yes.

src/Numerics/Mesh/Grids.jl
```diff
@@ -570,106 +591,242 @@ function commmapping(N, commelems, commfaces, nabrtocomm)
...
+        # due to the quadrature weights)
+        M_N1 = vgeo_N1_flds[_M]
+        MI = vgeo[:, _MI, :]
+        for fld in
+            (_ξ1x1, _ξ2x1, _ξ3x1, _ξ1x2, _ξ2x2, _ξ3x2, _ξ1x3, _ξ2x3, _ξ3x3)
+            vgeo[:, fld, :] =
+                sum(M_N1 .* vgeo_N1_flds[fld], dims = findall(Nq .== 1))[:] .*
+                MI[:]
+            num_vgeo_handled += 1
+        end
```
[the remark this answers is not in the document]
jkozdon: This is to compute `∂ξ_i/∂x_k` in the cell center. It is `J * ∂ξ_i/∂x_k` that is linear (which is not what we store), so we transition back to it average, and then remove the Jacobian. To get the average, we use the fact that for linear LGL the weights are `(1, 1)` and for constant LG the weight is `(2,)`, so in effect you get an average:

```julia
sum(M_N1 .* _ξ1x1 ) *  MI
```

becomes:

```julia
((J * ξ1x1)_{top} + (J * ξ1x1)_{bottom}) / (2 * J_{center})
```

src/Numerics/Mesh/Grids.jl
jkozdon: `fmask` (face mask) is to select the volume points associated with each face in the reference element. In this case it is being used to pull out the volume mass matrix on the face (which is necessary for the "lift" in DG

The faces are numbered in z-ordering, which I think is what you have

- face 1: `ξ1 = -1`
- face 2: `ξ1 = +1`
- face 3: `ξ2 = -1`
- face 4: `ξ2 = +1`
- face 5: `ξ3 = -1`
- face 6: `ξ3 = +1`

src/Numerics/Mesh/Grids.jl
jkozdon: > What are sM and normal.

`sM` is the surface mass matrix (with the surface jacobians scaled in), e.g., the sum off all these terms on a single face is the surface area of the face.

`normal` is the outward unit normal to a face in the physical space (the reference outward normal doesn't need to be stored)

> are they related to n ds = det(J) * J^-T * N dS, J =∂x/∂ξ and N is the outward normal in the reference element?

Yes. The normal is `n` and `ds` is almost `sM` except it includes the quadrature weights as we..

> Is it because det(J) is linear and J^-T and N are constant with respect to the p0 direction coordinate, we can do summation ?

`det(J)` isn't necessarily linear, but this is what you need to do so that the volume and surface metrics are consistently calculated (which is needed for conservation and constant preserving).

src/Numerics/Mesh/Grids.jl
charleskawczynski: It's Julia's [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) syntax

src/Numerics/Mesh/Grids.jl
jkozdon: `MJ` is the mass matrix with the jacobian determinant scaled in, and `M` is the reference matrix.

(This is a bit different than in the DG kernels where `M` is the same as `MJ`)

It is computed in the FVM geometry computation before this is called (so this is the `N = 0` mass matrix)

src/Numerics/Mesh/Grids.jl
jkozdon: `JcV` is the destination of the call, see the docs for [`map!`](https://docs.julialang.org/en/v1/base/collections/#Base.map!)

PR #1754, opened 2020-11-13
Fix ordering of mass matrix elements for multiple polynomial orders

mwarusz:
Fixes ordering of mass matrix elements for multiple polynomial orders and adds a test.

CC: @thomasgibson 

### DIFF
--- a/src/Numerics/Mesh/Grids.jl
+++ b/src/Numerics/Mesh/Grids.jl
@@ -659,7 +659,7 @@ function computegeometry(
         )
     end
 
-    M = kron(1, ntuple(j -> ω[j], dim)...)
+    M = kron(1, reverse(ω)...)
     MJ .= M .* J
     MJI .= 1 ./ MJ
     for d in 1:dim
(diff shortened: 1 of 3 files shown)


src/Numerics/Mesh/Grids.jl
```diff
@@ -659,19 +659,24 @@ function computegeometry(
...
     sM = fill!(similar(sJ, maximum(Nfp), nface), NaN)
     for d in 1:dim
         for f in (2d - 1):(2d)
-            sM[1:Nfp[d], f] =
-                dim > 1 ?
-                kron(1, ntuple(j -> ω[mod1(d + j, dim)], dim - 1)...) : one(FT)
+            ωf = ntuple(j -> ω[mod1(d + j, dim)], dim - 1)
+            if !(dim == 3 && d == 2)
+                ωf = reverse(ωf)
+            end
```
jkozdon, line 678: Had to think about this a minute, so would like a comment
```suggestion
            # Because of the `mod1` this face is already flipped
            if !(dim == 3 && d == 2)
                ωf = reverse(ωf)
            end
```

jkozdon (approved):
You are 100% right! Good catch. Sorry for the bug!

src/Numerics/Mesh/Grids.jl
```diff
@@ -659,19 +659,24 @@ function computegeometry(
         )
     end
 
-    M = kron(1, ntuple(j -> ω[j], dim)...)
+    M = kron(1, reverse(ω)...)
```
jkozdon, line 662:
```suggestion
    # since `ξ1` is the fastest dimension and `ξdim` the slowest the tensor product order is reversed
   M = kron(1, reverse(ω)...)
```

jkozdon:
In my revamp #1752 I am reworking the interface a bit for the metrics test. I will add a volume integration test there too (as I don't think we every test this).

PR #1761, opened 2020-11-16
modify the isothermal zonal flow test case and move it to TestCase folder

jiahe23:
### Description

This PR updates the isothermal zonal flow in:
1. replacing the default Rusanov flux by Roe flux;
2. adding a test assert to ensure the flow stay identical to its initial state;
3. moving the test driver into `TestCase`

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -992,7 +992,7 @@ steps:
   - label: "gpu_isothermal_zonal_flow"
     key: "gpu_isothermal_zonal_flow"
     command:
-      - "mpiexec julia --color=yes --project experiments/AtmosGCM/isothermal_zonal_flow.jl "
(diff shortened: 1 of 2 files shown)


szy21:
It seems the initial state of this test is not in discrete hydrostatic balance? I'm not sure though.

jiahe23:
> It seems the initial state of this test is not in discrete hydrostatic balance? I'm not sure though.

Good catch! I need to double check the calc for pressure and density.

kpamnany (approved):
LGTM but can't comment on the balance.

jiahe23:
> It seems the initial state of this test is not in discrete hydrostatic balance? I'm not sure though.

I've consulted Maciej and Thomas. The initial condition is balanced analytically but not discretely. For now, this is the best we could do [we cannot simply copy the way the ref state is set up because there's flow]. 

The fact that the state var stays identical indicates that the discrepancies are small enough to be ignored.

PR #1763, opened 2020-11-17
Add density-pressure thermo constructors

charleskawczynski:
### Description

This PR adds density-pressure thermo constructors

 - `PhaseDry_ρp`
 - `PhaseEquil_ρpq`
 - `PhaseNonEquil_ρpq`

These constructors are tested along with all of the others, and they're also tested against a newly added method `total_energy_given_ρp`, which is used in the constructors to compute total energy by first computing temperature from the ideal gas law. I believe that this is needed so that we can translate between conservative and primitive variables. This allows users to, instead of calling `total_energy_given_ρp` directly:

```julia
ts = PhaseDry_ρp.(param_set, ρ, p)
ts = PhaseEquil_ρpq.(param_set, ρ, p, q_tot)
ts = PhaseNonEquil_ρpq.(param_set, ρ, p, q_pt)

e_tot = total_energy.(e_kin, e_pot, ts)
```

At the moment, these constructors are failing the consistency tests (for pressure) in the moist constructors (the dry passes the consistency tests, however), but passing the `total_energy_given_ρp` test. I'd like to get to the bottom of this so that we can properly document what's going on.

### DIFF
--- a/src/Common/Thermodynamics/relations.jl
+++ b/src/Common/Thermodynamics/relations.jl
@@ -591,6 +591,31 @@ function total_energy(
 end
 
 """
+    total_energy_given_ρp(param_set, ρ, p, e_kin, e_pot[, q::PhasePartition])
+
+The total energy per unit mass, given
+
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `e_kin` kinetic energy per unit mass
+ - `e_pot` potential energy per unit mass
+ - `ρ` (moist-)air density
+ - `p` pressure
+and, optionally,
+ - `q` [`PhasePartition`](@ref). Without this argument, the results are for dry air.
+"""
+function total_energy_given_ρp(
+    param_set::APS,
+    ρ::FT,
+    p::FT,
+    e_kin::FT,
+    e_pot::FT,
+    q::PhasePartition{FT} = q_pt_0(FT),
+) where {FT <: Real}
+    T = air_temperature_from_ideal_gas_law(param_set, p, ρ, q)
+    return total_energy(param_set, e_kin, e_pot, T, q)
+end
+
+"""
     soundspeed_air(param_set, T[, q::PhasePartition])
 
 The speed of sound in unstratified air, where
@@ -1351,6 +1376,96 @@ function saturation_adjustment(
             return sol.root
         end
     end
+end
+
+"""
+    saturation_adjustment_ρpq(
+        param_set,
+        ρ,
+        p,
+        q_tot,
+        phase_type,
+        maxiter,
+        temperature_tol
+    )
+Compute the temperature that is consistent with
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `ρ` (moist-)air density
+ - `p` pressure
+ - `q_tot` total specific humidity
+ - `phase_type` a thermodynamic state type
+ - `maxiter` maximum iterations for non-linear equation solve
+ - `temperature_tol` temperature tolerance
+by finding the root of
+
+```
+T - air_temperature_from_ideal_gas_law(
+        param_set,
+        p,
+        ρ,
+        PhasePartition_equil(param_set, T, ρ, q_tot, phase_type),
+    )
+```
+using Newtons method using ForwardDiff.
+See also [`saturation_adjustment`](@ref).
+"""
+function saturation_adjustment_ρpq(
+    param_set::APS,
+    ρ::FT,
+    p::FT,
+    q_tot::FT,
+    phase_type::Type{<:PhaseEquil},
+    maxiter::Int,
+    temperature_tol::FT = eps(FT),
+) where {FT <: Real}
+    tol = SolutionTolerance(temperature_tol)
+    q_pt = PhasePartition(q_tot)
+    T_init = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)
+    # Use `oftype` to preserve diagonalized type signatures:
+    sol = find_zero(
+        T ->
+            T - air_temperature_from_ideal_gas_law(
+                param_set,
+                oftype(T, p),
+                oftype(T, ρ),
+                PhasePartition_equil(
+                    param_set,
+                    T,
+                    oftype(T, ρ),
+                    oftype(T, q_tot),
+                    phase_type,
+                ),
+            ),
+        NewtonsMethodAD(T_init),
+        CompactSolution(),
+        tol,
+        maxiter,
+    )
+    if !sol.converged
+        if print_warning()
+            @print("-----------------------------------------\n")
+            @print("maxiter reached in saturation_adjustment_ρpq:\n")
+            @print(
+                "    ρ=",
+                ρ,
+                ", p=",
+                p,
+                ", q_tot=",
+                q_tot,
+                ", T = ",
+                sol.root,
+                ", maxiter=",
+                maxiter,
+                ", tol=",
+                tol.tol,
+                "\n"
+            )
+        end
+        if error_on_non_convergence()
+            error("Failed to converge with printed set of inputs.")
+        end
+    end
+    return sol.root
 end
 
 """
--- a/src/Common/Thermodynamics/states.jl
+++ b/src/Common/Thermodynamics/states.jl
@@ -6,15 +6,18 @@ export ThermodynamicState,
     PhaseDry_pT,
     PhaseDry_ρθ,
     PhaseDry_pθ,
+    PhaseDry_ρp,
     PhaseEquil,
     PhaseEquil_ρTq,
     PhaseEquil_pTq,
     PhaseEquil_ρθq,
     PhaseEquil_pθq,
+    PhaseEquil_ρpq,
     PhaseNonEquil,
     PhaseNonEquil_ρTq,
     PhaseNonEquil_ρθq,
-    PhaseNonEquil_pθq
+    PhaseNonEquil_pθq,
+    PhaseNonEquil_ρpq
 
 """
     ThermodynamicState{FT}
@@ -146,6 +149,21 @@ Constructs a [`PhaseDry`](@ref) thermodynamic state from:
  - `T` temperature
 """
 function PhaseDry_ρT(param_set::APS, ρ::FT, T::FT) where {FT <: Real}
+    e_int = internal_energy(param_set, T)
+    return PhaseDry{FT, typeof(param_set)}(param_set, e_int, ρ)
+end
+
+"""
+    PhaseDry_ρp(param_set, ρ, p)
+
+Constructs a [`PhaseDry`](@ref) thermodynamic state from:
+
+ - `param_set` an `AbstractParameterSet`, see the [`Thermodynamics`](@ref) for more details
+ - `ρ` density
+ - `p` pressure
+"""
+function PhaseDry_ρp(param_set::APS, ρ::FT, p::FT) where {FT <: Real}
+    T = air_temperature_from_ideal_gas_law(param_set, p, ρ)
     e_int = internal_energy(param_set, T)
     return PhaseDry{FT, typeof(param_set)}(param_set, e_int, ρ)
 end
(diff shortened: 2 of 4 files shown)


tapios:
This looks ok to me. But what is the advantage of using the constructors over simply computing `T(p, \rho, q)` and then computing energy from that? We'd get what we need with two simple function calls. With the constructors, it seems we first would have to construct a thermodynamic state from the state variables (with case distinction of equilibrium thermodynamics, non-equilibrium etc.), then call the functions, unpack the state etc. Seems more complicated to me than the simple function calls with the primitive variables that are available to us in the numerics code. Am I missing something?

src/Common/Thermodynamics/states.jl
```diff
@@ -450,3 +491,24 @@ function PhaseNonEquil_pθq(
...
+ - `p` pressure
+ - `q_pt` phase partition
+"""
+function PhaseNonEquil_ρpq(
+    param_set::APS,
+    ρ::FT,
+    p::FT,
+    q_pt::PhasePartition{FT},
+) where {FT <: Real}
+    T = air_temperature_from_ideal_gas_law(param_set, p, ρ)
```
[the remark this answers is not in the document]
charleskawczynski: This fixed consistency for the non-equilibrium thermo state.

charleskawczynski:
> This looks ok to me. But what is the advantage of using the constructors over simply computing `T(p, \rho, q)` and then computing energy from that? We'd get what we need with two simple function calls. With the constructors, it seems we first would have to construct a thermodynamic state from the state variables (with case distinction of equilibrium thermodynamics, non-equilibrium etc.), then call the functions, unpack the state etc. Seems more complicated to me than the simple function calls with the primitive variables that are available to us in the numerics code. Am I missing something?

The advantage is that we can do consistency tests on the incoming variables if we construct a thermo state. The correction you suggested fixed the consistency test for the non-equilibrium case, I'd rather handle the validation logic in thermodynamics than rely on users to validate everywhere else. Also, the distinction between dry, equilibrium, and non-equilibrium is already hard-coded in the arguments (which are different for all three), so I don't think this is any more complex as calling the function directly.

As it is right now, the dry and non-equilibrium consistency tests fully pass, but the moist equilibrium consistency test fails on pressure:
```julia
        ts = PhaseEquil_ρpq.(param_set, ρ, p, q_tot)
        @test all(air_density.(ts) .≈ ρ)
        # @test all(air_pressure.(ts) .≈ p) # Fails
```
Printing the amount that it's failing by reveals that it's relatively small:
```julia
        @show FT, max(abs.(air_pressure.(ts) .- p)...)
        (FT, max(abs.(air_pressure.(ts) .- p)...)) = (Float32, 2299.039f0)
        (FT, max(abs.(air_pressure.(ts) .- p)...)) = (Float64, 2299.031082502159)
```
For reference, our pressure tolerance is `MSLP(param_set) * 2e-2 = 2026.5`. I think we have a few options:

 - [ ] Spend some time to see if this test is expected to have some error
 - [ ] Bump up the tolerance and use `isapprox` to allow this to pass
 - [ ] Use a custom, and slightly larger tolerance for this test
 - [ ] Leave the failed test commented for now and try to tackle it later.

tapios:
If pressure is a state variable, as it is here, then reconstructing pressure from the state should be exact. Something is wrong.

charleskawczynski:
Ok, I'll take a closer look.

charleskawczynski:
The problem is that, in the translation:
```julia
    q_pt = PhasePartition(q_tot)
    T = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)
```
, the `PhasePartition` is not in sync with what would be computed after calling
```julia
    q_pt = PhasePartition_equil(param_set, T, ρ, q_tot, phase_type)
```
In other words, these two functions are out of sync:
```julia
PhasePartition_equil(param_set, T, ρ, q_tot, phase_type)
air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)
```

The reason that this matters w.r.t. pressure is due to the chain of calls: `air_pressure(::ThermodynamicState)` ➡️ `PhasePartition` ➡️ `PhasePartition_equil`.

I confirmed this by trying
```julia
    maxiter = 200
    temperature_tol = eps(FT)
    phase_type = PhaseEquil
    q_pt_0 = PhasePartition(q_tot)
    # We can probably do something better with the bounds, or use NewtonMethod
    # with Jacobians.
    T_1 = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt_0) - 50
    T_2 = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt_0) + 50
    sol = find_zero(
        T -> T - air_temperature_from_ideal_gas_law(
            param_set, p, ρ,
            PhasePartition_equil(param_set, T, ρ, q_tot, phase_type))
        ,
        SecantMethod(T_1, T_2),
        CompactSolution(),
        SolutionTolerance(temperature_tol),
        maxiter,
    )
    if !sol.converged
        error("Not converged")
    else
        T = sol.root
    end

    q_pt = PhasePartition_equil(param_set, T, ρ, q_tot, phase_type)
    e_int = internal_energy(param_set, T, q_pt)
```
Which yields effectively precision errors for `air_pressure`:
```julia
(FT, max(abs.(air_pressure.(ts) .- p)...)) = (Float32, 0.0078125f0)
(FT, max(abs.(air_pressure.(ts) .- p)...)) = (Float64, 2.9103830456733704e-11)
```
I'm seeing the same discrepancy for total energy (which is slightly different when we iterate vs not iterate to ensure `PhasePartition` and temperature are in sync).

I think our options moving forward depend on how we want to address this. Fortunately, we know precisely what's going on.

tapios:
If you want to get temperature given pressure, density, and q_t, of course you need to solve saturation adjustment (i.e., iterate). But my point is that you should not compute pressure from temperature if pressure is given. `air_pressure` in that case should just return the given input pressure, not compute anything. Same for density.

charleskawczynski:
> But my point is that you should not compute pressure from temperature if pressure is given. `air_pressure` in that case should just return the given input pressure, not compute anything. Same for density.

Ah, I see.. Hmmm. The formulation conversion approach allows us to minimize the number of methods with only 3 thermo state types. That said, I agree that this is perhaps inefficient, especially if this is a performant-critical edge case. I've got an idea, I think we can do what you want. Going to open a separate PR.

charleskawczynski:
In a separate branch ([here](https://github.com/CliMA/ClimateMachine.jl/compare/ck/thermo_new_state_formulation)), I added custom thermo states which, given pressure/density, return these variables back exactly. However, I'm still running into the same problem when trying to add the method `total_energy_given_ρp` since `total_energy` ➡️ `internal_energy` ➡️ `air_temperature`. So, we still need to decide if this temperature should be in thermodynamic equilibrium or not. @tapios, perhaps we can zoom to discuss this in more detail.

tapios:
> In a separate branch ([here](https://github.com/CliMA/ClimateMachine.jl/compare/ck/thermo_new_state_formulation)), I added custom thermo states which, given pressure/density, return these variables back exactly. However, I'm still running into the same problem when trying to add the method `total_energy_given_ρp` since `total_energy` ➡️ `internal_energy` ➡️ `air_temperature`. So, we still need to decide if this temperature should be in thermodynamic equilibrium or not. @tapios, perhaps we can zoom to discuss this in more detail.

For the numerical fluxes, we have `ρ`, `p`, `q` at the faces. From that, we just need to compute temperature using `T=air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)`. With this `T`, we get total energy. This will suffice for the FV reconstructions. We'd want to avoid an additional saturation adjustment in this flux calculation.

charleskawczynski:
Ok, I've updated things. Now, for the moist case, we can call `ts = PhaseEquil_ρpq(param_set, ρ, p, q_tot[, perform_sat_adjust = false])`. If for some reason we want to run this with machine precision and perform saturation adjustment, we can just flip the switch of the last argument to true: `ts = PhaseEquil_ρpq(param_set, ρ, p, q_tot, true)`.

src/Common/Thermodynamics/relations.jl
```diff
@@ -590,6 +590,31 @@ function total_energy(
...
+    ρ::FT,
+    p::FT,
+    e_kin::FT,
+    e_pot::FT,
+    q::PhasePartition{FT} = q_pt_0(FT),
+) where {FT <: Real}
+    T = air_temperature_from_ideal_gas_law(param_set, p, ρ, q)
+    return total_energy(param_set, e_kin, e_pot, T, q)
+end
+
```
Zhengyu-Huang, line 617: Does this hold for both for Dry/Moist air ?

src/Common/Thermodynamics/relations.jl
```diff
@@ -1353,6 +1378,96 @@ function saturation_adjustment(
...
+                "\n"
+            )
+        end
+        if error_on_non_convergence()
+            error("Failed to converge with printed set of inputs.")
+        end
+    end
+    return sol.root
+end
+
```
Zhengyu-Huang, line 1470: What is the different between T_0 and the root T? Is oftype(T, p) change the type of p the same as T?

src/Common/Thermodynamics/states.jl
```diff
@@ -450,3 +508,24 @@ function PhaseNonEquil_pθq(
...
+function PhaseNonEquil_ρpq(
+    param_set::APS,
+    ρ::FT,
+    p::FT,
+    q_pt::PhasePartition{FT},
+) where {FT <: Real}
+    T = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)
+    e_int = internal_energy(param_set, T, q_pt)
+    return PhaseNonEquil{FT, typeof(param_set)}(param_set, e_int, ρ, q_pt)
+end
```
Zhengyu-Huang, line 531: so you don't need saturation adjustment to get T for NonEquil ?

Zhengyu-Huang (approved):
Look great for me, thank you for the work!

I am just a little bit confused about these thermo stuffs, why sometimes you need saturation adjustment to compute T sometimes you only need to call idea gas law.

src/Common/Thermodynamics/relations.jl
charleskawczynski: `T` is te solution, and `T_0` is the initial guess.

> Is oftype(T, p) change the type of p the same as T?

Basically, yes. `oftype` is used to convert the `Float` type of `p` to a `Dual` for `ForwardDiff`.

src/Common/Thermodynamics/states.jl
charleskawczynski: Nope, no saturation adjustment required for `PhaseNonEquil`.

src/Common/Thermodynamics/relations.jl
charleskawczynski: I believe so. Moisture is accounted for in `q::Phasepartition`

src/Common/Thermodynamics/relations.jl
```diff
@@ -1353,6 +1378,96 @@ function saturation_adjustment(
...
+                oftype(T, ρ),
+                PhasePartition_equil(
+                    param_set,
+                    T,
+                    oftype(T, ρ),
+                    oftype(T, q_tot),
+                    phase_type,
+                ),
+            ),
+        NewtonsMethodAD(T_0),
```
tapios, line 1439: If we actually want to use it, I suspect the secant method would be considerably faster for plausible tolerances. Did you test this?

src/Common/Thermodynamics/relations.jl
```diff
@@ -1353,6 +1378,96 @@ function saturation_adjustment(
...
+    ρ::FT,
+    p::FT,
+    q_tot::FT,
+    phase_type::Type{<:PhaseEquil},
+    maxiter::Int,
+    temperature_tol::FT = eps(FT),
+) where {FT <: Real}
+    tol = SolutionTolerance(temperature_tol)
+    q_pt = PhasePartition(q_tot)
+    T_0 = air_temperature_from_ideal_gas_law(param_set, p, ρ, q_pt)
```
tapios, line 1423: I would avoid `T_0` here. We have reserved that for the thermodynamic reference temperature. Overloading it here creates confusion. `T_init` or something similar may be better.

charleskawczynski: Good point. Will update

src/Common/Thermodynamics/relations.jl
charleskawczynski: I wasn't sure what bounds to use for SecantMethod. `NewtonsMethodAD` converges in 5 iterations for all thermo profiles, so I figured that is reasonable.

tapios: Ok. We'll not actually use it, I don't think. But the AD itself will also add cost per iteration. 

For the secant method, you could use the same kinds of bounds we had in other places for SA (use the unsaturated solution as one bound).

tapios: Did you try Newton AD vs Newton with analytical gradients for the standard SA? I'm wondering about the timing. If AD is fast enough, it would be nice because it would allow us to use arbitrary (smooth) phase partitioning functions.

charleskawczynski: > Ok. We'll not actually use it, I don't think. But the AD itself will also add cost per iteration.

Agreed, I figured that this will be a nice test to have, and provide transparency for the approximation that we're making.

> For the secant method, you could use the same kinds of bounds we had in other places for SA (use the unsaturated solution as one bound).

I can give that a try.

> Did you try Newton AD vs Newton with analytical gradients for the standard SA? I'm wondering about the timing.

I haven't tried yet, but that's a great point! Maybe that'll be better for a separate PR? I can open an issue to track it.

> If AD is fast enough, it would be nice because it would allow us to use arbitrary (smooth) phase partitioning functions.

This would be cool to try out also!

charleskawczynski:
> why sometimes you need saturation adjustment to compute T sometimes you only need to call idea gas law.

Saturation adjustment is only needed for equilibrium phases (so, `PhaseEquil_ρpq` in this case). We're including the option to not perform saturation adjustment because, based on the given variables (ρpq), the consistency error (the error resulting from recovering the input variables from the resulting thermodynamic state) is very small for the tested profiles, or it may not be worth the extra effort and we may want to settle with an approximation. Ultimately, the logic around performing saturation adjustment should be encapsulated inside Thermodynamics.jl, and users should simply have an interface (e.g., `PhaseEquil_ρpq`) so that they can compute whatever they may need from the resulting thermo state. This allows us to internally optimize how things are done.

src/Common/Thermodynamics/relations.jl
charleskawczynski: Opened #1831.

charleskawczynski: Using `BenchmarkTools`'s `@btime` on the thermo profiles, I'm seeing
```julia
[Float32, Newton AD]  1.206 ms (26298 allocations: 513.67 KiB)
[Float32, Secant   ]  1.429 ms (39854 allocations: 784.30 KiB)
[Float64, Newton AD]  1.492 ms (31110 allocations: 957.42 KiB)
[Float64, Secant   ]  1.894 ms (47694 allocations: 1.01 MiB)
```
So, it looks like Newton method with AD is performing better, which makes sense because Secant method seems to be requiring > 10 iterations to converge on the entire set of thermo profiles. I'll leave Newton method AD as the default for now.

PR #1656, opened 2020-10-23
Decompose discretization kernels into horizontal and vertical calls

mwarusz:
### Description

This PR documents the core DG kernels used in the numerics (volume/interface kernels for the prognostic and auxiliary equations in the LDG formulation). Hyperdiffusion kernels will be documented in a later PR. All DG kernels have been split in the main tendency evaluation function into separate horizontal and vertical calls.

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #1775, opened 2020-11-19
Poly order splitting in Filters

valeriabarra:
### Description

This is just an initial draft of the support for multiple polynomial orders in Filters, so that I can easily get people's feedback and guidance. 

### DIFF
--- a/src/Numerics/Mesh/Filters.jl
+++ b/src/Numerics/Mesh/Filters.jl
@@ -140,28 +140,37 @@ value of the filter function.
 struct ExponentialFilter <: AbstractSpectralFilter
     "filter matrix"
     filter::Any
+    "direction in which you're applying the filter"
+    direction::Direction
 
     function ExponentialFilter(
         grid,
         Nc = 0,
         s = 32,
         α = -log(eps(eltype(grid))),
+        direction = HorizontalDirection(),
     )
         AT = arraytype(grid)
-        # XXX: Needs updating for multiple polynomial orders
+        # Tuple of polynomial degrees (N₁, N₂, N₃)
         N = polynomialorders(grid)
-        # Currently only support single polynomial order
-        @assert all(N[1] .== N)
-        N = N[1]
-        ξ = referencepoints(grid)[1]
+        # In 2D, we assume same polynomial order in the horizontal
+        dim = dimensionality(grid)
+        @assert dim == 2 || N[1] == N[2]
+        # NOTE: assumes same polynomial order in both horizontal directions
+        if direction isa HorizontalDirection
+            N = N[1]
+            ξ = referencepoints(grid)[1]
+        else
+            @assert direction isa VerticalDirection
+            N = N[end]
+            ξ = referencepoints(grid)[end]
+        end
 
         @assert iseven(s)
         @assert 0 <= Nc <= N
-
         σ(η) = exp(-α * η^s)
         filter = spectral_filter_matrix(ξ, Nc, σ)
-
-        new(AT(filter))
+        new(AT(filter), direction)
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/Numerics/Mesh/Filters.jl
```diff
@@ -141,27 +141,62 @@ struct ExponentialFilter <: AbstractSpectralFilter
...
+            @assert iseven(s)
+            @assert 0 <= Nc <= vert_polyorder
+
+            σ(η) = exp(-α * η^s)
+
+            #TODO: think about how this matrix changes in different directions
+            filter = spectral_filter_matrix(vert_ξ, Nc, σ)
+
+            new(AT(filter))
+        end
```
valeriabarra, line 199: @mwarusz if I understoon your suggestion here, was to base this similar to what was done in #1656 . In that PR, the direction was queried from the diffusion direction. But I am afraid this will be wrong here.

valeriabarra:
bors try

src/Numerics/Mesh/Filters.jl
```diff
@@ -140,28 +140,37 @@ value of the filter function.
 struct ExponentialFilter <: AbstractSpectralFilter
     "filter matrix"
     filter::Any
+    "direction in which you're applying the filter"
+    direction::Direction
```
thomasgibson, line 144: So we opted to add `direction` as an attribute to the filter. The reason for this is that the filtering matrix you want to use is different depending on what direction you're applying it.

test/Numerics/Mesh/filter.jl
```diff
@@ -193,13 +202,33 @@ end
...
                         dim,
                     )
-                    ClimateMachine.Mesh.Filters.apply!(
-                        Q,
-                        target,
-                        grid,
-                        filter,
-                        direction = direction(),
-                    )
+                    if direction isa EveryDirection
```
thomasgibson, line 205: Applying it in `EveryDirection` now is a bit cumbersome... because we split it into horizontal and vertical filter applications (by basically creating and applying two filters, one for each direction). Not sure if this was the best way, but we don't really know how to handle this case in the filter constructors (what filter matrix do we use?)

src/Numerics/Mesh/Filters.jl
mwarusz: An alternative approach might be to store in the filter object a tuple of filter matrices, one for each direction. This is similar to how the grid now contains tuple of differentiation matrices. Think it would help with application of a filter in `EveryDirection` too.

thomasgibson: @mwarusz Good idea! I like this idea better

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

src/Numerics/Mesh/Filters.jl
```diff
@@ -477,9 +536,7 @@ horizontal and/or vertical reference directions.
     i, j, k = @index(Local, NTuple)
 
     @inbounds begin
-        ijk = i + Nq * ((j - 1) + Nq * (k - 1))
-
-        s_filter[i, j] = filtermatrix[i, j]
+        ijk = i + Nq1 * ((j - 1) + Nq1 * (k - 1))
```
jkozdon, line 539:
```suggestion
        ijk = i + Nq1 * ((j - 1) + Nq2 * (k - 1))
```

src/Numerics/Mesh/Filters.jl
```diff
@@ -525,9 +582,9 @@ horizontal and/or vertical reference directions.
 
         if filterinξ2
             @synchronize
-            @unroll for n in 1:Nq
+            @unroll for n in 1:Nq1
```
jkozdon, line 585:
```suggestion
            @unroll for n in 1:Nq2
```

src/Numerics/Mesh/Filters.jl
```diff
@@ -561,7 +618,7 @@ horizontal and/or vertical reference directions.
         )
 
         # Store result
-        ijk = i + Nq * ((j - 1) + Nq * (k - 1))
+        ijk = i + Nq1 * ((j - 1) + Nq1 * (k - 1))
```
jkozdon, line 621:
```suggestion
        ijk = i + Nq1 * ((j - 1) + Nq2 * (k - 1))
```

src/Numerics/Mesh/Filters.jl
```diff
@@ -278,41 +315,63 @@ function apply!(
...
-        Val(isnothing(state_auxiliary) ? nothing : vars(state_auxiliary)),
-        direction,
-        Q.data,
-        isnothing(state_auxiliary) ? nothing : state_auxiliary.data,
-        target,
-        filtermatrix,
-        ndrange = (nrealelem * Nq, Nq, Nqk),
-        dependencies = (event,),
-    )
+    if direction isa EveryDirection || direction isa HorizontalDirection
```
jkozdon, line 341:
```suggestion
    if direction isa EveryDirection || direction isa HorizontalDirection
        @assert dim == 2 || Nq1 == Nq2
```

jkozdon (commented):
Can we test the TMAR filter with multiple polynomial orders in this PR?

src/Numerics/Mesh/Filters.jl
valeriabarra: Thanks for catching those! Fixed.

valeriabarra:
bors try

valeriabarra:
> Can we test the TMAR filter with multiple polynomial orders in this PR?

Just for future reference, it will be done in #1702

jkozdon (approved):

PR #1781, opened 2020-11-21
Add DOIs to .bib, remove urls

charleskawczynski:
### Description

I went to fix the bib entries whose hyperlinks point to the reference page and realized that these correspond to bib entries with missing DOI/URL/eprint attributes. This PR adds some missing DOIs. Also, in case these links are checked by Documenter, it makes sense to rely on DOIs over other URLs, so this PR also removes some of the URLs so long the DOIs exist.


trontrytel (approved):
Looks great. Thank you for doing this!

PR #1787, opened 2020-11-24
Agl/kd/update soil bc

kmdeck:
### Description
This updates how we handle the soil boundary conditions. Previously, we had in place functionality so that the user could provide arbitrary boundary conditions - either Neumann or Dirichlet, at the top or at the bottom of the domain. This was useful for comparing with textbook examples, for simple sanity checks, and for comparing to lab experiments. These conditions were supplied as functions of space and time.

Moving forwards, however, we need to also handle boundary conditions that are a function of state, that can handle different models for runoff, evaporation, etc. These would be boundary conditions used in comparing with field data, for example.
In this case, the boundary conditions are more structured and are now physically motivated functions at the top and bottom of the domain. They would not be simple functions the user specifies at at driver level, but rather correspond to a particular class of boundary conditions (that the user would opt to use at the driver level).

To account for this, we added a layer of methods that dispatch on AbstractBoundaryConditions - either GeneralBoundaryConditions, which executes something like what we had in place already, or e.g. PhysicalFluxBoundaryConditions, which employ zero flux at the bottom of the domain and could account for precipitation and runoff in the flux for water at the top of the domain. For now, this 2nd option only exists for the SoilWaterModel. Adding it for the heat model can come later.

All tutorials and integration tests have been updated to reflect the change.

### DIFF
--- a/src/Land/Model/Runoff.jl
+++ b/src/Land/Model/Runoff.jl
@@ -1,0 +1,97 @@
+module Runoff
+
+using ...VariableTemplates
+using DocStringExtensions
+
+export AbstractPrecipModel,
+    DrivenConstantPrecip,
+    AbstractSurfaceRunoffModel,
+    NoRunoff,
+    compute_surface_flux
+
+"""
+    AbstractPrecipModel{FT <: AbstractFloat}
+"""
+abstract type AbstractPrecipModel{FT <: AbstractFloat} end
+
+"""
+    DrivenConstantPrecip{FT, F} <: AbstractPrecipModel{FT}
+
+Instance of a precipitation distribution where the precipication value
+is constant across the domain. However, this value can change in time.
+
+# Fields
+$(DocStringExtensions.FIELDS)
+"""
+struct DrivenConstantPrecip{FT, F} <: AbstractPrecipModel{FT}
+    "Mean precipitation in grid"
+    mp::F
+    function DrivenConstantPrecip{FT}(mp::F) where {FT, F}
+        new{FT, F}(mp)
+    end
+end
+
+function (dcp::DrivenConstantPrecip{FT})(t::Real) where {FT}
+    return FT(dcp.mp(t))
+end
+
+"""
+    AbstractSurfaceRunoffModel
+
+Abstract type for different surface runoff models. Currently, only
+`NoRunoff` is supported.
+"""
+abstract type AbstractSurfaceRunoffModel end
+
+"""
+    NoRunoff <: AbstractSurfaceRunoffModel
+
+Chosen when no runoff is to be modeled.
+"""
+struct NoRunoff <: AbstractSurfaceRunoffModel end
+
+
+"""
+    function compute_surface_runoff(
+        runoff_model::NoRunoff,
+        precip_model::AbstractPrecipModel,
+        state::Vars
+    )
+
+Returns zero for net surface runoff when `NoRunoff`
+is used.
+"""
+function compute_surface_runoff(
+    runoff_model::NoRunoff,
+    precip_model::AbstractPrecipModel{FT},
+    state::Vars,
+) where {FT}
+    return FT(0.0)
+end
+
+
+"""
+    compute_surface_flux(
+        runoff_model::AbstractSurfaceRunoffModel,
+        precip_model::AbstractPrecipModel{FT},
+        state::Vars,
+        t::Real,
+    ) where {FT}
+
+Given a runoff model and a precipitation distribution function, compute 
+the surface water flux. This can be a function of time, and state.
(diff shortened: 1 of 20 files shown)


kmdeck:
bors try

gagnelandmanna:
Hi Katherine!
I reviewed it; it looks good to me. I am just maybe confused on the names we chose : general vs physical as both are physical. But I am not sure what to propose as a better option.

kmdeck:
> Hi Katherine!
> I reviewed it; it looks good to me. I am just maybe confused on the names we chose : general vs physical as both are physical. But I am not sure what to propose as a better option.

thanks, @gagnelandmanna! What about RunoffBoundaryConditions?

gagnelandmanna:
Hmm, I think calling them RunoffBoundaryConditions is a good idea. Then we could have RadiativeEnergyBoundaryConditions when we put those in? Will it be clear to the user (or automatic in the code) that they cannot apply also at the same time the general ones? I.e. we cant have two sets of bcs be on. Also, maybe keeping it an option to have zero flux or non zero flux at the bottom is better instead of enforcing zero flux, just because it could be interesting to test runoff and baseflow all at once, and have water exiting the at the bottom should the layer of soil be thing or something?

kmdeck:
> Hmm, I think calling them RunoffBoundaryConditions is a good idea. Then we could have RadiativeEnergyBoundaryConditions when we put those in? Will it be clear to the user (or automatic in the code) that they cannot apply also at the same time the general ones? I.e. we cant have two sets of bcs be on. Also, maybe keeping it an option to have zero flux or non zero flux at the bottom is better instead of enforcing zero flux, just because it could be interesting to test runoff and baseflow all at once, and have water exiting the at the bottom should the layer of soil be thin or something?

all great points! 
we discussed offline, for now:

SurfaceDrivenWaterBoundaryConditions and SurfaceDrivenEnergyBoundaryConditions. We'll have this with zeroflux at the bottom, but if the need arises in the future we will adapt and make the bottom BC something the user can specify.

The user cannot turn on both types of BC, but we will need to make a tutorial explaining the different options for them.

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

src/Land/Model/Runoff.jl
```diff
@@ -0,0 +1,91 @@
...
+$(DocStringExtensions.FIELDS)
+"""
+struct ConstantPrecip{FT, F} <: AbstractPrecipDistribution{FT}
+    "Mean precipitation in grid"
+    mp::F
+    function ConstantPrecip{FT}(; mp::Function = (t) -> FT(0)) where {FT}
+        f = (t) -> FT(mp(t))
+        new{FT, typeof(f)}(f)
+    end
+end
```
jakebolewski, line 30: it might be a bit cleaner to define this as a callable object (so that you can enforce some properties, like the constant precip is actually a constant) and provide some encapsulation.

``` 
struct ConstantPrecip{FT} <:  AbstractPrecipDistribution{FT}
    v::FT
end

function (p::ConstantPrecip{FT})(time::FT) where {FT}
    return p.v * time
end
```

src/Land/Model/Runoff.jl
```diff
@@ -0,0 +1,91 @@
...
+Given a runoff model and a precipitation distribution function, compute 
+the surface water flux. This can be a function of time, and state.
+"""
+function compute_surface_flux(
+    runoff_model::AbstractSurfaceRunoffModel,
+    precip_dist::AbstractPrecipDistribution{FT},
+    state::Vars,
+    t::Real,
+) where {FT}
+    mean_p = precip_dist.mp(t)
```
jakebolewski, line 84: with the above suggested changes this would be `precip_dist(t)`

jakebolewski (approved):

src/Land/Model/Runoff.jl
kmdeck: ok! im doing something very similar now:
```
struct DrivenConstantPrecip{FT, F} <: AbstractPrecipModel{FT}
    "Mean precipitation in grid"
    mp::F
    function DrivenConstantPrecip{FT}(mp::F) where {FT,F}
        new{FT,F}(mp)
    end
end

function (dcp::DrivenConstantPrecip{FT})(t::Real) where {FT}
    return FT(dcp.mp(t))
end
```

kmdeck: Then it is called as precip_model(t), instead of precip_model.mp(t).

thanks!

kmdeck:
bors try

PR #1794, opened 2020-11-27
Revamp bomex EDMF regression test

charleskawczynski:
The BOMEX experiment test, using edmf, is very brittle and compares the computed solution with a previous solution, which is restricting development of the EDMF model.

This PR:
 - Replaces the existing regression test with a newly revamped one that compares against horizontally-averaged PyCLES output at the last matching time with ClimateMachine.
 - Adds wraps `contourf` in a `try-catch` to skip purely zero plots (and throws a warning)
 - Updates some parameters in the BOMEX edmf driver (CFL, polynomial order, number of elements, horizontal extent, the solver type, and adds a vertical velocity filter, which was observed to be needed for the dycore to reach a CFL of 25).

A data-scale normalized Mean-Squared Error (MSE) is computed for ClimateMachine against PyCLES for each variable, and each of these numbers are tabulated and reported against the "best" mse we've achieved. Here's what the table looks like:

Screen Shot 2020-11-27 at 8 36 51 AM

This type of regression test has the advantage of still being brittle in the sense that the regression tests will fail if our MSE increases by more than machine precision-- but it will pass if our MSE improves marginally. These numbers are relatively easy to update as we improve the model and extend the time that we're running the simulation for. In addition, there is an option in the regression test to plot a comparison of ClimateMachine against PyCLES for each field, so that we can set up a nightly QA run.

Right now, the regression test occurs at `t = 400`, which is very early in the simulation. Unfortunately, we're kind of in a deadlock-- increasing the CFL crashes the simulation because the model isn't working yet, but we can't update the model without a more robust regression test. I'm hoping that, as we improve the model, we'll be able to push the CFL limit and run this test for a longer time.

### Description


test/Atmos/EDMF/compute_mse.jl
```diff
@@ -0,0 +1,221 @@
...
+    z_cm = get_z(grid; rm_dupes = true)
+    @test maximum(z_les) + last(diff(z_les)) / 2 ≈ z_cm[end]
+    time_les = ds["t"][:]
+
+    # Find the nearest matching final time:
+    t_min = min(time_cm[end], time_les[end])
+    i_t_cm = argmin(abs.(t_min .- time_cm))
+    i_t_les = argmin(abs.(t_min .- time_les))
+    t_cm = time_cm[i_t_cm]
+    t_les = time_les[i_t_les]
```
ilopezgp, line 69: The only situation I am concerned about is when we might get worse MSE because we change the CFL such that `abs(t_cm-t_les)` is larger than in a previous experiment. Then the MSE will grow but not because the solution is worse, just because we are testing two times that are farther in time.

charleskawczynski: Agreed. I just realized that we're using the 2D spline, so we can interpolate the data to the same time for the comparison. Also, I'll add a test to ensure `t_compare > 400` so that we don't pass just by running a shorter simulation.

charleskawczynski:
Perhaps it was a fluke, but I observed different results when re-running this test, despite using `fix_rng_seed = true` in the driver. I relaxed a few of the mse scores so that it should pass. Fortunately, we can back off the mse if this is somehow a bit flakey.

ilopezgp:
> Perhaps it was a fluke, but I observed different results when re-running this test, despite using `fix_rng_seed = true` in the driver. I relaxed a few of the mse scores so that it should pass. Fortunately, we can back off the mse if this is somehow a bit flakey.

That is odd... But if the MSE score is robust to this random noise, I think the source of that "error" should be orthogonal to the test and we can worry about it later on.

ilopezgp (approved):

yairchn (approved):

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #1796, opened 2020-11-30
Call modifier after turbulence_tensors

charleskawczynski:
### Description

It seems that `turbulence_tensors` is often, but not always, followed by a call to `sponge_viscosity_modifier`--should it be? This PR moves `sponge_viscosity_modifier` inside `turbulence_tensors`. I'm not sure if this is correct. Looping in @akshaysridhar.

This effectively adds calls to `sponge_viscosity_modifier` in a few places:

 - https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/bc_energy.jl#L69 - I don't think this will impact anything, since the sponge is above a certain altitude and the bc is (I assume) at the surface
 - https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/courant.jl#L68 - may improve the correctness of the reported the diffusive courant
 - https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Diagnostics/atmos_les_default.jl#L162 - may improve the correctness of the diagnostics

### DIFF
--- a/src/Common/TurbulenceClosures/TurbulenceClosures.jl
+++ b/src/Common/TurbulenceClosures/TurbulenceClosures.jl
@@ -247,8 +247,26 @@ for the returned quantities.
 - `aux` = Array of auxiliary variables
 - `t` = time
 """
-turbulence_tensors(m::TurbulenceClosureModel, bl::BalanceLaw, args...) =
-    turbulence_tensors(m, bl.orientation, bl.param_set, args...)
+function turbulence_tensors(
+    m::TurbulenceClosureModel,
+    bl::BalanceLaw,
+    state,
+    diffusive,
+    aux,
+    t,
+)
+    ν, D_t, τ = turbulence_tensors(
+        m,
+        bl.orientation,
+        bl.param_set,
+        state,
+        diffusive,
+        aux,
+        t,
+    )
+    ν, D_t, τ = sponge_viscosity_modifier(bl, bl.viscoussponge, ν, D_t, τ, aux)
+    return (ν, D_t, τ)
+end
 
 # We also provide generic math functions for use within the turbulence closures,
 # commonly used quantities such as the [principal tensor invariants](@ref tensor-invariants), handling of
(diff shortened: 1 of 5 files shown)


akshaysridhar:
> ### Description
> 
> It seems that `turbulence_tensors` is often, but not always, followed by a call to `sponge_viscosity_modifier`--should it be? This PR moves `sponge_viscosity_modifier` inside `turbulence_tensors`. I'm not sure if this is correct. Looping in @akshaysridhar.
> 
> This effectively adds calls to `sponge_viscosity_modifier` in a few places:

@charleskawczynski The order of the `turbulence_tensors` -> `sponge_viscosity_modifier` call is right ; this is a reasonable change.

>     * https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/bc_energy.jl#L69 - I don't think this will impact anything, since the sponge is above a certain altitude and the bc is (I assume) at the surface

You are right - this interaction would need to be accounted for when applying a non-zero flux at the top-boundary(via a user-specified bc identifier in the present problem-driver setup.) Defaults are not affected. 

>     * https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Atmos/Model/courant.jl#L68 - may improve the correctness of the reported the diffusive courant
> 
>     * https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Diagnostics/atmos_les_default.jl#L162 - may improve the correctness of the diagnostics

charleskawczynski:
@akshaysridhar should we merge this, then? (still need approval)

akshaysridhar (approved):

src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -247,8 +247,25 @@ for the returned quantities.
...
+    ν, D_t, τ = turbulence_tensors(
+        m,
+        bl.orientation,
+        bl.param_set,
+        state,
+        diffusive,
+        aux,
+        t,
+    )
+    ν, D_t, τ = sponge_viscosity_modifier(bl, bl.viscoussponge, ν, D_t, τ, aux)
```
charleskawczynski, line 267:
```suggestion
    return sponge_viscosity_modifier(bl, bl.viscoussponge, ν, D_t, τ, aux)
```

charleskawczynski, line 267:
```suggestion
    ν, D_t, τ = sponge_viscosity_modifier(bl, bl.viscoussponge, ν, D_t, τ, aux)
    return (ν, D_t, τ)
```

PR #1798, opened 2020-11-30
GCM test case: polynomial order set to 5

jiahe23:
### Description

This PR changes the polynomial order from 3 to 5 in GCM test case drivers:
- solid body rotation
- baroclinic wave

Horizontal and vertical element numbers are changed accordingly so that the default resolution remains roughly the same. 

Tests have been performed to make sure that with the above changes, the solid body rotation test still holds the initial condition; and the baroclinic wave test can survive 30 days and generate reasonable outputs.

### DIFF
--- a/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
+++ b/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
@@ -312,9 +312,9 @@ function main()
 
     # Driver configuration parameters
     FT = Float64                             # floating type precision
-    poly_order = 3                           # discontinuous Galerkin polynomial order
-    n_horz = 12                              # horizontal element number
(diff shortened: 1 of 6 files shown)


szy21:
Look good to me. Do you want to change all the GCM drivers? (e.g. HS)

jiahe23:
> Looks good to me. Do you want to change all the GCM drivers? (e.g. HS)

That's a good point. Let me test HS with poly_order=5 before merging it.

jiahe23:
> Looks good to me. Do you want to change all the GCM drivers? (e.g. HS)

@szy21 

HS survives day 80+ with poly_order 5. It is less stable compared with current poly_order = 3.

I suggest we change it to 5 for now and stabilize it later with the new kernel.

szy21 (approved):

PR #1800, opened 2020-11-30
Move first order rain flux to new tend spec

charleskawczynski:
### Description

Move first order rain fluxes to new tendency specification.

I think that this is looking forward a bit, but, I put `eq_tends` for `Rain` in `precipitation.jl` instead of in `AtmosModel.jl`, since all we really need to do is "forward" `eq_tends` into the submodel components. That way we don't clutter `AtmosModel.jl` with all of the submodel tendencies.

### DIFF
--- a/src/Atmos/Model/precipitation.jl
+++ b/src/Atmos/Model/precipitation.jl
@@ -3,6 +3,8 @@ abstract type PrecipitationModel end
 
 export NoPrecipitation, RainModel#, RainSnow
 
+eq_tends(pv::PV, m::PrecipitationModel, ::Flux{O}) where {PV, O} = ()
+
 using ..Microphysics
 
 vars_state(::PrecipitationModel, ::AbstractStateType, FT) = @vars()
@@ -102,6 +104,33 @@ function compute_gradient_flux!(
     diffusive.precipitation.∇q_rai = ∇transform.precipitation.q_rai
 end
 
+"""
+    RainFlux{PV} <: TendencyDef{Flux{FirstOrder}, PV <: Rain}
+
+Computes the rain flux as a sum of air velocity and rain terminal velocity
+multiplied by the advected variable
+"""
+struct RainFlux{PV <: Rain} <: TendencyDef{Flux{FirstOrder}, PV} end
+
+function flux(::RainFlux{Rain}, m, state, aux, t, ts, direction)
+    FT = eltype(state)
+    u = state.ρu / state.ρ
+    q_rai = state.precipitation.ρq_rai / state.ρ
+
+    v_term::FT = FT(0)
+    if q_rai > FT(0)
+        v_term = terminal_velocity(
+            m.param_set,
+            m.param_set.microphys.rai,
+            state.ρ,
+            q_rai,
+        )
+    end
+
+    k̂ = vertical_unit_vector(m, aux)
+    return state.precipitation.ρq_rai * (u - k̂ * v_term)
+end
+
 function flux_first_order!(
     precip::RainModel,
     atmos::AtmosModel,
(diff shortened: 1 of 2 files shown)


src/Atmos/Model/precipitation.jl
```diff
@@ -102,32 +104,45 @@ function compute_gradient_flux!(
...
-    state::Vars,
-    aux::Vars,
-    t::Real,
-    ts,
-    direction,
-)
+"""
+    TerminalVelocity{PV} <: TendencyDef{Flux{FirstOrder}, PV <: Rain}
+
+TODO: Add docs
```
trontrytel, line 110: There is a leftover TODO here about missing docs

trontrytel (changes requested):
Apart from the missing docstring everything looks good

src/Atmos/Model/precipitation.jl
charleskawczynski: Ah, good catch!

charleskawczynski: Is this a good name for the source? And would you like to push a commit for the doc string?

trontrytel: Not sure if it's a good name. What you compute here is a sum of air velocity with which you advect the precipitation plus its own terminal velocity. So I would say it's just a precipitation flux. 

The docs on how terminal velocity is computed are already available in the microphysics parameterization documentation. I don't think there is a need to repeat it here?

src/Atmos/Model/precipitation.jl
```diff
@@ -102,32 +104,45 @@ function compute_gradient_flux!(
...
-    atmos::AtmosModel,
-    flux::Grad,
-    state::Vars,
-    aux::Vars,
-    t::Real,
-    ts,
-    direction,
-)
+"""
+    TerminalVelocity{PV} <: TendencyDef{Flux{FirstOrder}, PV <: Rain}
```
trontrytel, line 108:
```suggestion
    RainVelocity{PV} <: TendencyDef{Flux{FirstOrder}, PV <: Rain}
```

src/Atmos/Model/precipitation.jl
```diff
@@ -102,32 +104,45 @@ function compute_gradient_flux!(
...
+
+TODO: Add docs
+"""
+struct TerminalVelocity{PV <: Rain} <: TendencyDef{Flux{FirstOrder}, PV} end
```
trontrytel, line 112:
```suggestion
struct RainFlux{PV <: Rain} <: TendencyDef{Flux{FirstOrder}, PV} end
```

src/Atmos/Model/precipitation.jl
trontrytel, line 108:
```suggestion
    RainFlux{PV} <: TendencyDef{Flux{FirstOrder}, PV <: Rain}
```

src/Atmos/Model/precipitation.jl
```diff
@@ -102,32 +104,45 @@ function compute_gradient_flux!(
...
+
+function flux(::TerminalVelocity{Rain}, m, state, aux, t, ts, direction)
```
trontrytel, line 114:
```suggestion
function flux(::RainFlux{Rain}, m, state, aux, t, ts, direction)
```

src/Atmos/Model/precipitation.jl
```diff
@@ -162,3 +177,10 @@ function source!(
     source.precipitation.ρq_rai =
         Σsources(eq_tends(Rain(), atmos, tend), args...)
 end
+
+#####
+##### Tendency specifications
+#####
+
+eq_tends(pv::PV, ::RainModel, ::Flux{FirstOrder}) where {PV <: Rain} =
+    (TerminalVelocity{PV}(),)
```
trontrytel, line 186:
```suggestion
    (RainFlux{PV}(),)
```

src/Atmos/Model/precipitation.jl
trontrytel, line 110:
```suggestion
Computes the rain flux as a sum of air velocity and rain terminal velocity 
multiplied by the advected variable
```

trontrytel (approved):
☔

PR #1801, opened 2020-11-30
Move held suarez sources to new tend spec

charleskawczynski:
### Description

Move sources in Held-Suarez to new tendency specification.

### DIFF
--- a/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
+++ b/experiments/AtmosGCM/GCMDriver/GCMDriver.jl
@@ -23,6 +23,7 @@ using ArgParse
 using LinearAlgebra
 using StaticArrays
 using Test
+using UnPack
 
 using ClimateMachine
 using ClimateMachine.Atmos
--- a/experiments/AtmosGCM/GCMDriver/gcm_sources.jl
+++ b/experiments/AtmosGCM/GCMDriver/gcm_sources.jl
@@ -2,44 +2,43 @@
 # This file contains helpers and lists currently available options
 
 # Current options for GCM-specific sources:
-import ClimateMachine.Atmos: atmos_source!
 
 """
-    HeldSuarezForcing <: AbstractSource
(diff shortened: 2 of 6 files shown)


charleskawczynski:
Since `experiments\AtmosGCM\GCMDriver\GCMDriver.jl` does not seem to be used anywhere in the codebase, it would be helpful if someone can run the touched experiments and make sure that things are still running smoothly. cc @LenkaNovak, @bischtob

LenkaNovak (changes requested):
in `heldsuarez_problem.jl` we need to change `boundarycondition` to `boundaryconditions` but otherwise the GCMDriver runs fine :)

charleskawczynski:
> in `heldsuarez_problem.jl` we need to change `boundarycondition` to `boundaryconditions` but otherwise the GCMDriver runs fine :)

Ah, looks like that's actually broken in master. I'm fine with incorporating the fix here. Do the results look okay with these changes?

charleskawczynski:
Bump. I think these are the last of the sources! Would be nice to remove `atmos_source!` in a subsequent PR.

trontrytel (commented):
LGTM! I have two questions:
- do we have to define `HeldSuarezForcing()` and `HeldSuarezForcingTutorial()` separately?
- is this part of qa or a longer CI test?

charleskawczynski:
> * do we have to define `HeldSuarezForcing()` and `HeldSuarezForcingTutorial()` separately?

I asked about merging `HeldSuarez` and `HeldSuarezTutorial` a while back, and the consensus was to keep them separate.

> * is this part of qa or a longer CI test?

We just run held suarez, as far as I know. We should really add a QA run for HS.

trontrytel (approved):
LGTM. I dont know if you want to wait on people that actually run this case. Would be great to add it to qa at some point.

LenkaNovak (approved):

PR #1802, opened 2020-11-30
Add boundary fields to topologies

simonbyrne:
### Description


src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -748,12 +748,20 @@ fluxes, respectively.
...
         e⁺ = ((id⁺ - 1) ÷ Np) + 1
 
         vid⁻, vid⁺ = ((id⁻ - 1) % Np) + 1, ((id⁺ - 1) % Np) + 1
+        if bctag != 0
+            # TODO: we will use vmap⁺ to store the boundary element info
+            #be = e⁺
+            #bcid = vid⁺
+            e⁺ = e⁻
+            vid⁺ = vid⁻
+        end
```
jkozdon, line 764: So eventually all the loads below will be inside this if statement, e.g., it will become an if-else for loading the "plus" side?

src/Numerics/Mesh/BrickMesh.jl
```diff
@@ -1086,5 +1086,38 @@ function connectmesh(
...
+`elemtobndy`. Builds the `bndytoelem` and `bndytoface` tuples.
+"""
+function enumerateboundaryfaces!(elemtoelem, elemtobndy, periodicity, boundary)
+    nb = 0
+    for i in 1:length(periodicity)
+        if !periodicity[i]
+            nb = max(nb, boundary[i]...)
+        end
+    end
+    @assert nb <= 6
```
jkozdon, line 1102:
```suggestion
    # Limitation of the boundary condition unrolling in the DG kernels
    # (should never be violated unless more general unstructured meshes are used
    # since cube meshes only have 6 faces in 3D, and only 1 bcs is currently allowed
    # per face)
  
    @assert nb <= 6
```

src/Numerics/Mesh/BrickMesh.jl
```diff
@@ -1086,5 +1086,38 @@ function connectmesh(
...
+    bndytoelem = ntuple(b -> Vector{Int64}(), nb)
+    bndytoface = ntuple(b -> Vector{Int64}(), nb)
+
+    nface, nelem = size(elemtoelem)
+
+    N = zeros(Int, nb)
+    for e in 1:nelem
+        for f in 1:nface
+            d = elemtobndy[f, e]
+            if d != 0
```
jkozdon, line 1113: Mesh generation doesn't require this.

```suggestion
            @assert 0 <= d <= nb
            if d != 0
```

test/Numerics/Mesh/mpi_connect.jl
```diff
@@ -68,7 +68,7 @@ function main()
         nrealelem = 5
         globalelems = [1, 2, 3, 4, 5, 6, 7, 8, 14, 15]
         elemtoelem_expect = [
-            1 1 4 4 5 6 7 8 9 10
+            1 1 4 2 3 4 7 8 9 10
```
jkozdon, line 71: I don't understand why these numbers changed

src/Numerics/Mesh/Topologies.jl
```diff
@@ -17,17 +17,17 @@ export grid1d, SingleExponentialStretching, InteriorStretching
 export basic_topology_info
 
 """
-    AbstractTopology{dim}
+    AbstractTopology{dim,T,nb}
 
 Represents the connectivity of individual elements, with local dimension `dim`.
 """
```
jkozdon, line 23:
```suggestion
    AbstractTopology{dim, T, nb}

Represents the connectivity of individual elements, with local dimension `dim` with `nb` boundary conditions types. The element coordinates are of type `T`.
"""
```

jkozdon (approved):
Seems reasonable. Basically you are just building the boundary element to volume element mapping right now, and later you will embed the volume to boundary map in the `vmap` and build arrays for the boundary data.

Is the plan to have a single array for the boundary, and index into the monolithic array or to have one array for each boundary type?

test/Numerics/Mesh/mpi_connect.jl
simonbyrne: `elemtoelem` now stores boundary element number if it faces a boundary (previously it just pointed back to the element itself). In this case, face 1 of elements [1,4,5,6] face boundary 1, and so are numbered `[1,2,3,4]`

src/Numerics/DGMethods/DGModel_kernels.jl
simonbyrne: Possibly, I haven't decided yet.

test/Numerics/Mesh/mpi_connect.jl
jkozdon: I see that now. Thanks. (Some how missed that line 🤦 )

src/Numerics/Mesh/Topologies.jl
```diff
@@ -17,17 +17,17 @@ export grid1d, SingleExponentialStretching, InteriorStretching
 export basic_topology_info
 
 """
-    AbstractTopology{dim}
+    AbstractTopology{dim, T, nb}
 
-Represents the connectivity of individual elements, with local dimension `dim`.
+Represents the connectivity of individual elements, with local dimension `dim` with `nb` boundary conditions types. The element coordinates are of type `T`.
```
jkozdon, line 22: My bad, this line should probably be wrapped
```suggestion
Represents the connectivity of individual elements, with local dimension `dim`
with `nb` boundary conditions types. The element coordinates are of type `T`.
```

jkozdon (approved):

simonbyrne:
I will probably have one array for each boundary point type.

src/Numerics/Mesh/BrickMesh.jl
```diff
@@ -1086,5 +1086,44 @@ function connectmesh(
...
+    end
+    # Limitation of the boundary condition unrolling in the DG kernels
+    # (should never be violated unless more general unstructured meshes are used
+    # since cube meshes only have 6 faces in 3D, and only 1 bcs is currently allowed
+    # per face)
+
+    @assert nb <= 6
+
+    bndytoelem = ntuple(b -> Vector{Int64}(), nb)
+    bndytoface = ntuple(b -> Vector{Int64}(), nb)
```
charleskawczynski, line 1110:
```suggestion
    bndytoelem = ntuple(b -> Vector{Int}(), nb)
    bndytoface = ntuple(b -> Vector{Int}(), nb)
```

simonbyrne: True, but we hardcode everything else as `Int64`, so it should be consistent. I'll leave that for another PR.

PR #1803, opened 2020-12-01
Add missing call to rain source

charleskawczynski:
### Description

This PR
 - Adds a missing call to `source!` for the precipitation model.
 - Re-orders the arguments so that we can pass them through in the same order as `source`

@trontrytel This should fix the dycoms with rain QA run.


trontrytel (approved):
LGTM!

PR #1807, opened 2020-12-01
Show more specific !isfinite error

### LINKED ISSUE #1805: Provide more specific error in `isnan` in `call!(::SummaryLogCallback)`
### Description

The infamous [`isnan(normQ) && error("norm(Q) is NaN")`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Callbacks/Callbacks.jl#L88) should be improved to provide more granular information (at least which field has NaNs). This may really help users with debugging.

charleskawczynski:
### Description

Closes #1805

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -9,7 +9,8 @@ using MPI
 using StaticArrays
 
 using ..TicToc
-using ..VariableTemplates: @vars, varsindex, varsindices, varsize
+using ..VariableTemplates:
+    @vars, varsindex, varsindices, varsize, wrap_val, flattened_tup_chain
 
 using Base.Broadcast: Broadcasted, BroadcastStyle, ArrayStyle
 
@@ -27,7 +28,8 @@ using .CMBuffers
 cpuify(x::AbstractArray) = convert(Array, x)
 cpuify(x::Real) = x
 
-export MPIStateArray, euclidean_distance, weightedsum, array_device, vars
+export MPIStateArray,
+    euclidean_distance, weightedsum, array_device, vars, show_not_finite_fields
 
 """
     MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
@@ -853,6 +855,40 @@ end
         @unroll for s in 1:nvar
             buf[n, s, e] = recvbuf[s, i]
         end
+    end
+end
+
+"""
+    show_not_finite_fields(Q::MPIStateArray)
+
+Prints a warning of which fields are not finite.
+
+!!! warn
+    This is an expensive method, which calls `mapreduce`
+    on `Q`.
+"""
+function show_not_finite_fields(Q::MPIStateArray)
+    vs = vars(Q)
+    not_finite_fields = []
+    not_finite =
+        Array(reshape(mapreduce(x -> !isfinite(x), |, Q; dims = (1, 3)), :))
+    for ftc in flattened_tup_chain(vs)
+        i_vars = varsindex(vs, wrap_val.(ftc)...)
+        if length(i_vars) > 1 # variable is an array
+            for (j, i_var) in enumerate(i_vars)
+                if not_finite[i_var]
+                    push!(not_finite_fields, join(string.(ftc), ".") * "[$j]")
+                end
+            end
+        else # variable is a scalar
+            if not_finite[i_vars[1]]
+                push!(not_finite_fields, join(string.(ftc), "."))
+            end
+        end
+    end
+    if !isempty(not_finite_fields)
+        @warn "Field(s) ($(join(not_finite_fields, ", ", ", and "))) are not finite (has NaNs or Inf)"
+        flush(stdout)
     end
 end
 
--- a/src/Driver/Callbacks/Callbacks.jl
+++ b/src/Driver/Callbacks/Callbacks.jl
@@ -85,7 +85,10 @@ function GenericCallbacks.call!(cb::SummaryLogCallback, solver, Q, param, t)
         estimated_remaining,
         normQ,
     )
-    isnan(normQ) && error("norm(Q) is NaN")
+    if !isfinite(normQ)
+        show_not_finite_fields(Q)
+        error("norm(Q) is not finite")
+    end
     return nothing
 end
 function GenericCallbacks.fini!(cb::SummaryLogCallback, solver, Q, param, t)
--- a/src/Utilities/VariableTemplates/VariableTemplates.jl
+++ b/src/Utilities/VariableTemplates/VariableTemplates.jl
@@ -47,6 +47,20 @@ function varsindex(::Type{S}, insym::Symbol) where {S <: NamedTuple}
     end
     error("symbol '$insym' not found")
 end
+
+# return `Symbol`s unchanged.
+wrap_val(sym) = sym
+# wrap integer in `Val`
+wrap_val(i::Int) = Val(i)
+
+# We enforce that calls to `varsindex` on
+# an `NTuple` must be unrapped in `Val`.
+# This is enforced to synchronize failures
+# on the CPU and GPU, rather than allowing
+# CPU-working and GPU-breaking versions.
+# This means that users _must_ wrap `sym`
+# in `Val`, which can be done with `wrap_val`
+# above.
 unval(::Val{i}) where {i} = i
 Base.@propagate_inbounds function varsindex(
     ::Type{S},
--- a/docs/src/APIs/Arrays/Arrays.md
+++ b/docs/src/APIs/Arrays/Arrays.md
@@ -19,3 +19,9 @@ weightedsum
 ```@docs
 CMBuffers.CMBuffer
 ```
+
+## Helpers
+
+```@docs
+show_not_finite_fields
+```
--- a/src/Utilities/VariableTemplates/flattened_tup_chain.jl
+++ b/src/Utilities/VariableTemplates/flattened_tup_chain.jl
@@ -3,7 +3,7 @@ export flattened_tup_chain
 flattened_tup_chain(::Type{NamedTuple{(), Tuple{}}}; prefix = (Symbol(),)) = ()
 flattened_tup_chain(::Type{T}; prefix = (Symbol(),)) where {T <: Real} =
     (prefix,)
-flattened_tup_chain(::Type{T}; prefix = (Symbol(),)) where {T <: SVector} =
+flattened_tup_chain(::Type{T}; prefix = (Symbol(),)) where {T <: SArray} =
     (prefix,)
 flattened_tup_chain(
     ::Type{T};
--- a/test/Arrays/varsindex.jl
+++ b/test/Arrays/varsindex.jl
@@ -22,6 +22,13 @@ const V = @vars begin
     end
 end
 
+const VNT = @vars begin
+    a::Float32
+    e::Tuple{ntuple(3) do i
+        @vars(b::SVector{3, Float32})
+    end...}
+end
+
 @testset "MPIStateArray varsindex" begin
     # check with invalid vars size
     @test_throws ErrorException MPIStateArray{Float32, V}(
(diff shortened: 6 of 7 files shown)


kpamnany:
How's this look; have you tested it?

charleskawczynski:
Good point. I pushed an update, which looks like
```julia
┌ Info: Update
│     simtime =   164.19 / 21600.00
│     wallclock = 00:01:00
│     efficiency (simtime / wallclock) =   2.7316
│     wallclock end (estimated) = 02:11:47
└     norm(Q) = NaN
ERROR: LoadError: Fields ρ, ρu, ρe, and moisture.ρq_tot have NaNs
Stacktrace:
 [1] error(::String) at /Applications/Julia-1.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [2] call!(::ClimateMachine.Callbacks.SummaryLogCallback, ::AdditiveRungeKutta{Float64,Float64,ClimateMachine.MPIStateArrays.MPIStateArray{Float64,NamedTuple{(:ρ, :ρu, :ρe, :turbulence, :hyperdiffusion, :moisture, :precipitation, :turbconv, :radiation, :tracers),Tuple{Float64,SArray{Tuple{3},Float64,1,3},Float64,NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(:ρq_tot,),Tuple{Float64}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}},NamedTuple{(),T```

**Review by kpamnany — APPROVED** [review]

**Comment by charleskawczynski** [comment]
Going to test it for EDMF and see if I can get it to print for each component of vectors (`ρu`)

**Comment by charleskawczynski** [comment]
Ok it's definitely nice having the element-wise vector case working. Also, I'm glad I stopped to address the EDMF model, since that required some changes. Here's a more complicated error:
```julia
┌ Info: Update
│     simtime =     0.16 /   155.11
│     wallclock = 00:02:03
│     efficiency (simtime / wallclock) =   0.0013
│     wallclock end (estimated) = 10:10:08
└     norm(Q) = NaN
ERROR: LoadError: Fields (ρu[1], moisture.ρq_tot, turbconv.environment.ρatke, and turbconv.updraft.1.ρa) have NaNs
Stacktrace:
 [1] error(::String) at /Applications/Julia-1.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [2] call!(::ClimateMachine.Callbacks.SummaryLogCallback, ::LowStorageRungeKutta2N{Float64,Float64,MPIStateArray{Float64,NamedTuple{(:ρ, :ρu, :ρe, :turbulence, :hyperdiffusion, :moisture, :precipitation, :turbconv, :radiation, :tracers),Tuple{Float64,SArray{Tuple{3},Float64,1,3},Float64,NamedTuple{(),Tuple{}},NamedTuple{()
```

@kpamnany Can you take a second look at the implementation? This is a bit more clunky, but I'm not sure if how it can be simplified.

charleskawczynski:
Also, not sure if my comment is reading clearly

src/Driver/Callbacks/Callbacks.jl
```diff
@@ -85,7 +86,29 @@ function GenericCallbacks.call!(cb::SummaryLogCallback, solver, Q, param, t)
...
-    isnan(normQ) && error("norm(Q) is NaN")
+
+    if isnan(normQ)
+        vs = vars(Q)
+        nan_fields = []
+        for ftc in flattened_tup_chain(vs)
+            i_vars = varsindex(vs, wrap_val.(ftc)...)
+            if length(i_vars) > 1 # array
+                for (j, i_var) in enumerate(i_vars)
+                    if isnan(norm(Q[:, :, i_var]))
```
charleskawczynski, line 97:
```suggestion
                    if isnan(norm(Array(Q[:, :, i_var])))
```
Better to make this GPU friendly

src/Driver/Callbacks/Callbacks.jl
```diff
@@ -85,7 +86,29 @@ function GenericCallbacks.call!(cb::SummaryLogCallback, solver, Q, param, t)
...
+                        push!(nan_fields, join(string.(ftc), ".") * "[$j]")
+                    end
+                end
+            else
+                if isnan(norm(Q[:, :, i_vars]))
```
charleskawczynski, line 102:
```suggestion
                if isnan(norm(Array(Q)[:, :, i_vars]))
```

charleskawczynski:
This might be a generally more useful function that we can put in `src/Arrays/MPIStateArrays.jl`.

src/Arrays/MPIStateArrays.jl
```diff
@@ -856,4 +856,33 @@ end
...
+"""
+function throw_norm_error(Q::MPIStateArray)
+    vs = vars(Q)
+    nan_fields = []
+    Qarr = Array(Q)
+    for ftc in flattened_tup_chain(vs)
+        i_vars = varsindex(vs, wrap_val.(ftc)...)
+        if length(i_vars) > 1 # array
+            for (j, i_var) in enumerate(i_vars)
+                if isnan(norm(Qarr[:, :, i_var]))
```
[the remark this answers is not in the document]
charleskawczynski: Good point. Also, I'm indexing into the wrong rank 🤦🏻 . Fixing a bunch of things and adding tests.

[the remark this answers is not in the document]
charleskawczynski: The problem with using the `MPIStateArray` `norm`, though, is that if the data is on the GPU, I'm not sure how we can `getindex` for the specific field.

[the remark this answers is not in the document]
kpamnany: I think you **don't** want the `MPIStateArray` `norm`.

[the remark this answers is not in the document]
charleskawczynski: Wait, wouldn't we want the MPIStateArray norm so that it works on multi-node?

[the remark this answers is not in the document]
kpamnany: We want to use that one on the outside, in the primary check, yes. Now that I think about it, my concern is misplaced -- I was worried that not all nodes would see the `NaN` from the primary `norm` call so they would enter different code paths (and a subsequent MPI collective could block forever), but that's not correct -- if there's a `NaN`, all nodes will see it.

So you're right, you probably want to use the `MPIStateArray` `norm` on this inner check as well otherwise it will be local only, which means it is possible that only some of the nodes will find the `NaN`.

To do so, I think you have to use the `dims` parameter to `norm` rather than indexing into `Qarr`.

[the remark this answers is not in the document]
charleskawczynski: That means that we'd need to fix/merge #590. That seems to be much more work than I set out for. Going to stick with a single rank error for now so that we get some utility out of this.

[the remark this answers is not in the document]
charleskawczynski: Actually, I think we can just use `mapreduce` with `dims`. Fixing it up now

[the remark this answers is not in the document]
charleskawczynski: Ok, updated things

src/Arrays/MPIStateArrays.jl
```diff
@@ -856,4 +858,44 @@ end
...
 
+"""
+    throw_isnan_error(Q::MPIStateArray)
+
+Call if `isnan(norm(Q)) == true` to provide a
+detailed error of which field(s) have NaNs.
+
+!!! note
+    This method assumes that `isnan(norm(Q)) == true`,
+    and throws an error otherwise.
```
kpamnany, line 869: Why? Call this something like `throw_on_isnan` and let it be general.

src/Arrays/MPIStateArrays.jl
```diff
@@ -856,4 +858,44 @@ end
...
+    end
+    if !isempty(nan_fields)
+        error("Fields ($(join(nan_fields, ", ", ", and "))) have NaNs")
+    else
+        if isnan(norm(Q)) # double check user input
+            error("Q has NaNs")
+        else # user is calling this before checking `isnan(norm(Q))`
+            error("This method was called incorrectly. Intended use: `isnan(norm(Q)) && throw_isnan_error(Q)`")
+        end
+    end
```
kpamnany, line 898:
```suggestion
    !isempty(nan_fields) && error("Fields ($(join(nan_fields, ", ", ", and "))) have NaNs")
```

src/Utilities/VariableTemplates/VariableTemplates.jl
```diff
@@ -47,6 +47,20 @@ function varsindex(::Type{S}, insym::Symbol) where {S <: NamedTuple}
...
     error("symbol '$insym' not found")
 end
+
+# return `Symbol`s unchanged.
+wrap_val(sym) = sym
+# wrap integer in `Val`
+wrap_val(i::Int) = Val(i)
+
+# We enforce that calls to `varsindex` on
+# an `NTuple` must be unrapped in `Val`.
```
kpamnany, line 57: It'd be neater to enforce that everything be wrapped in a `Val` no? But this is not very important.

charleskawczynski: It would, but then some code may work on the CPU but break on the GPU. This synchronizes CPU/GPU failures, which helped avoid a lot of debugging headaches.

src/Arrays/MPIStateArrays.jl
charleskawczynski: `norm` is expensive, and my thought was that we don't need to include unnecessary calls of `norm(Q)` if we assume that things are already determined to be broken.

kpamnany: We can assume that it's been called without forcing it -- just document that it's expensive and this is how it _should_ be called and remove the defensive check below? But I'm not insisting...

kpamnany (approved):

src/Arrays/MPIStateArrays.jl
charleskawczynski: Looped in some others for review.

charleskawczynski: I've thought about this more and I think you're right. It's already an expensive function and we might as well just build in the assumption and document that. Updating...

src/Arrays/MPIStateArrays.jl
```diff
@@ -856,4 +858,44 @@ end
...
+
+!!! note
+    This method assumes that `isnan(norm(Q)) == true`,
+    and throws an error otherwise.
+"""
+function throw_isnan_error(Q::MPIStateArray)
+    vs = vars(Q)
+    nan_fields = []
+    normQ_per_var =
+        Array(reshape(mapreduce(x -> x^2, +, Q; dims = (1, 3)), :) .^ (1 / 2))
```
jkozdon, line 875:
```suggestion
        Array(mapreduce(x -> isfinite(x), |, Q; dims = (1, 3))[:]
```

jkozdon, line 875:
```suggestion
    not_finite =
        Array(mapreduce(x -> !isfinite(x), |, Q; dims = (1, 3))[:]
```

src/Arrays/MPIStateArrays.jl
```diff
@@ -856,4 +858,43 @@ end
...
+
+Call to throw an error if `isfinite(norm(Q)) == false`
+to provide a detailed error of which field(s) are not
+finite.
+
+!!! warn
+    This is an expensive method, which calls `norm(Q)`
+"""
+function throw_not_finite_error(Q::MPIStateArray)
+    normQ = norm(Q)
```
kpamnany, line 872: I'm confused -- why call `norm` again here? We always call it in the callback, since we output it, and we only call this function if `isnan(norm(Q))`. If you want to make that `isfinite` instead of `isnan` in the caller, that's fine, but IMO this function can always assume that the `norm` is not right and will inspect the fields to provide more detail -- the name of the function makes it clear that it will throw an error.

charleskawczynski: If we don't call `norm`, then we don't know for sure that `all(isfinite.(norm(Q)))` and that complicated what we'd write for the error message (what error should we throw if `all(isfinite.(norm(Q)))`?). Maybe we can change the function signature to
```julia
throw_not_finite_error(Q::MPIStateArray, normQ = norm(Q))```
? and make sure that we pass the norm in?

Comment (jkozdon) [reply src/Arrays/MPIStateArrays.jl:872]: What about throwing the error with nothing in it? I agree with @kpamnany though, the call makes it clear that an error should be thrown, and if you do not throw the error that's weird.

**Review by jkozdon — COMMENTED** [review]
I'm fine with whatever you decide is best with regards to my comment.

File: src/Arrays/MPIStateArrays.jl (hunk 8 above)
Comment (kpamnany) [reply src/Arrays/MPIStateArrays.jl:872]: I think it's perfectly fine to assume an entering condition of `!isfinite(norm(Q))` as long as it's documented. Someone is calling a function called `throw_not_finite_error` after all.

If you don't like that then passing in the norm is better than calling `norm` again.

Comment (jkozdon) [reply src/Arrays/MPIStateArrays.jl:872]: Another option is to just print a warning and let the user decide what they want to do afterward.

Comment (jkozdon) [reply src/Arrays/MPIStateArrays.jl:872]: (I mean in all cases, and never throw an error. Make it more diagnostic, not fatal.)

Comment (kpamnany) [reply src/Arrays/MPIStateArrays.jl:872]: > (I mean in all cases, and never throw an error. Make it more diagnostic, not fatal.)

I like this idea. Make this something like `show_not_finite_fields` which shows nothing if they're all finite. Then throw the error from the callback as before.

Comment (charleskawczynski) [reply src/Arrays/MPIStateArrays.jl:872]: Ahh, that's a great idea. Updating...

**Review by jkozdon — APPROVED** [review]

File: src/Driver/Callbacks/Callbacks.jl [hunk 9]
```diff
@@ -85,7 +85,10 @@ function GenericCallbacks.call!(cb::SummaryLogCallback, solver, Q, param, t)
         estimated_remaining,
         normQ,
     )
-    isnan(normQ) && error("norm(Q) is NaN")
+    if isnan(normQ)
+        show_not_finite_fields(Q)
+        error("norm(Q) is NaN")
+    end
```
kpamnany, line 91:
```suggestion
    if !isfinite(normQ)
        show_not_finite_fields(Q)
        error("norm(Q) is not finite")
    end
```

kpamnany (approved):

src/Driver/Callbacks/Callbacks.jl
kpamnany: Not sure if this would be correct.

charleskawczynski: Ah, that was a last minute rush. Definitely want this, good catch!

charleskawczynski:
Thanks for the reviews!

PR #1809, opened 2020-12-02
Add `DGFVModel` interface

jkozdon:
Creates a `DGFVModel` for the using spectral element DG in the horizontal and finite volume in the vertical.

The dimensional splitting used implies that the method will be at most (formally) 2nd order accurate.

The finite volume reconstructions are handled with the newly introduced `FVReconstructions` module. Currently the following reconstruction types are supported:

- Constant reconstructions
- Linear reconstructions with limiters
- Linear reconstructions with limiters with a wider stencil (to test the interface for WENO without actually having weno implemented)

WENO should be supported and work once the WENO types are added to the `FVReconstructions` module.

Some things left to do after this PR:

- Add back in hyperdiffusion on with DG
- Hydrostatic balance
- WENO schemes 

Co-authored-by: Daniel Z. Huang <hzyhzy2014@gmail.com>
Co-authored-by: Thomas Gibson <gibsonthomas1120@hotmail.com>
Co-authored-by: Maciej Waruszewski <mwarusz@igf.fuw.edu.pl>

### DIFF
--- a/src/Numerics/DGMethods/DGFVModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGFVModel_kernels.jl
@@ -62,6 +62,7 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
     α,
     β,
     increment,
+    add_source,
 ) where {info, nvertelem, periodicstack}
     @uniform begin
         dim = info.dim
@@ -152,6 +153,7 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
 
         # Storage for the tendency
         local_tendency = MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_source = MArray{Tuple{num_state_prognostic}, FT}(undef)
 
         # XXX: will revisit this later for FVM
         fill!(local_state_prognostic_bottom1, NaN)
@@ -517,6 +519,24 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
                 t,
             )
 
+            if add_source
+                fill!(local_source, -zero(eltype(local_source)))
+                source!(
+                    balance_law,
+                    Vars{vars_state(balance_law, Prognostic(), FT)}(
+                        local_source,
+                    ),
+                    Vars{vars_state(balance_law, Prognostic(), FT)}(local_state_prognostic[stencil_center - 1],),
+                    Vars{vars_state(balance_law, GradientFlux(), FT)}(local_state_gradient_flux[stencil_center - 1],),
+                    Vars{vars_state(balance_law, Auxiliary(), FT)}(local_state_auxiliary[stencil_center - 1],),
+                    t,
+                    (VerticalDirection(),),
+                )
+                @unroll for s in 1:num_state_prognostic
+                    local_tendency[s] += local_source[s]
+                end
+            end
+
             # Update the bottom element:
             # numerical flux is computed with respect to the top element, so
             # `+=` is used to reverse the flux
@@ -540,6 +560,23 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
 
             # Update top element of the stack
             if eV_up == nvertelem
+                if add_source
+                    fill!(local_source, -zero(eltype(local_source)))
+                    source!(
+                        balance_law,
+                        Vars{vars_state(balance_law, Prognostic(), FT)}(
+                            local_source,
+                        ),
+                        Vars{vars_state(balance_law, Prognostic(), FT)}(local_state_prognostic[stencil_center],),
+                        Vars{vars_state(balance_law, GradientFlux(), FT)}(local_state_gradient_flux[stencil_center],),
+                        Vars{vars_state(balance_law, Auxiliary(), FT)}(local_state_auxiliary[stencil_center],),
+                        t,
+                        (VerticalDirection(),),
+                    )
+                    @unroll for s in 1:num_state_prognostic
+                        local_tendency[s] += local_source[s]
+                    end
+                end
                 # If periodic just add in the tendency that we just computed
                 if periodicstack
                     @unroll for s in 1:num_state_prognostic
--- a/src/Numerics/DGMethods/DGFVModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGFVModel_kernels.jl
@@ -385,6 +385,875 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
     end
 end
 
+@kernel function vert_fvm_interface_tendency!(
+    balance_law::BalanceLaw,
+    ::Val{info},
+    ::Val{nvertelem},
+    ::Val{periodicstack},
+    ::VerticalDirection,
+    reconstruction!::FVLinear,
+    numerical_flux_first_order,
+    numerical_flux_second_order,
+    tendency,
+    state_prognostic,
+    state_gradient_flux,
+    state_auxiliary,
+    vgeo,
+    sgeo,
+    t,
+    elemtobndy,
+    elems,
+    α,
+) where {info, nvertelem, periodicstack}
+    @uniform begin
+        dim = info.dim
+        FT = eltype(state_prognostic)
+        num_state_prognostic = number_states(balance_law, Prognostic())
+        num_state_auxiliary = number_states(balance_law, Auxiliary())
+        num_state_gradient_flux = number_states(balance_law, GradientFlux())
+        num_state_hyperdiffusion = number_states(balance_law, Hyperdiffusive())
+        @assert num_state_hyperdiffusion == 0
+
+        vsp = Vars{vars_state(balance_law, Prognostic(), FT)}
+        vsa = Vars{vars_state(balance_law, Auxiliary(), FT)}
+
+        nface = info.nface
+        Np = info.Np
+        Nqk = info.Nqk # can only be 1 for the FVM method!
+        @assert Nqk == 1
+
+        @assert nvertelem >= 3
+
+        # We only have the vertical faces
+        faces = (nface - 1):nface
+
+        # +/- indicate top/bottom elements
+        # top and bottom indicate face states
+
+        local_flux = MArray{Tuple{num_state_prognostic}, FT}(undef)
+
+        local_state_prognostic = MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_state_prognostic⁺ = MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_state_prognostic⁻ = MArray{Tuple{num_state_prognostic}, FT}(undef)
+
+        local_state_primitive = MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_state_primitive⁺ = MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_state_primitive⁻ = MArray{Tuple{num_state_prognostic}, FT}(undef)
+
+        local_state_auxiliary = MArray{Tuple{num_state_auxiliary}, FT}(undef)
+        local_state_auxiliary⁺ = MArray{Tuple{num_state_auxiliary}, FT}(undef)
+        local_state_auxiliary⁻ = MArray{Tuple{num_state_auxiliary}, FT}(undef)
+
+        local_state_gradient_flux⁻ =
+            MArray{Tuple{num_state_gradient_flux}, FT}(undef)
+        local_state_gradient_flux =
+            MArray{Tuple{num_state_gradient_flux}, FT}(undef)
+        local_state_gradient_flux⁺ =
+            MArray{Tuple{num_state_gradient_flux}, FT}(undef)
+
+        local_state_hyperdiffusion⁻ =
+            MArray{Tuple{num_state_hyperdiffusion}, FT}(undef)
+        local_state_hyperdiffusion =
+            MArray{Tuple{num_state_hyperdiffusion}, FT}(undef)
+        local_state_hyperdiffusion⁺ =
+            MArray{Tuple{num_state_hyperdiffusion}, FT}(undef)
+
+        local_state_primitive_bottom =
+            MArray{Tuple{num_state_prognostic}, FT}(undef)
+        local_state_primitive_top =
+            MArray{Tuple{num_state_prognostic}, FT}(undef)
+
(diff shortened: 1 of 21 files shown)


jkozdon:
bors try

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -1,7 +1,72 @@
...
+
+Must have the following properties:
+
+    - `grid`
+    - `balance_law`
+    - `state_auxiliary`
+"""
+abstract type SpaceDiscretization end
+
+struct DGFVMModel{BL, G, NFND, NFD, GNF, AS, DS, D, DD, MD} <:
```
charleskawczynski, line 17: Not a big deal, but would it make sense to call this `DGFVModel` instead?

jkozdon: Good idea. Thanks.

src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -928,6 +1102,427 @@ fluxes, respectively.
...
+∫ₑ ψ⋅ ∂q/∂t dx - ∫ₑ ∇ψ⋅(Fⁱⁿᵛ + Fᵛⁱˢᶜ) dx + ∮ₑ n̂ ψ⋅(Fⁱⁿᵛ⋆ + Fᵛⁱˢᶜ⋆) dS,
+
+or equivalently in matrix form:
+
+dQ/dt = M⁻¹(MS + DᵀM(Fⁱⁿᵛ + Fᵛⁱˢᶜ) + ∑ᶠ LᵀMf(Fⁱⁿᵛ⋆ + Fᵛⁱˢᶜ⋆)).
+
+This kernel computes the surface terms: M⁻¹ ∑ᶠ LᵀMf(Fⁱⁿᵛ⋆ + Fᵛⁱˢᶜ⋆)), where M
+is the mass matrix, Mf is the face mass matrix, L is an interpolator from
+volume to face, and Fⁱⁿᵛ⋆, Fᵛⁱˢᶜ⋆ are the numerical fluxes for the inviscid
+and viscous fluxes, respectively.
```
charleskawczynski, line 1138: Awesome doc string. This is super helpful, very clear, and helps connect the dots 😃

valeriabarra: I agree it is great! I have one question though: 

In the discrete equation there is the term `M⁻¹MS`, which I believe it represents the source term. If I am not blind at the moment (which I actually am partially ^_^") I do not see it in the continuous equation though

Zhengyu-Huang:
This push 
1) updated something in the FVModel.jl (so far this file is disconnected with the code, each time I use it just add include("FVModel.jl") in the DGModels.jl) 

2) added some test cases FV advection-diffusion with different initial conditions and FV isentropic Euler (the convergence rate: level 1 → 2 = 1.5414479846585512  level 2 → 3 = 1.8772338883737951, levels 3 → 4 = 1.8325715325576029) Will revisit and do a comparison between this and the playground code after Charlie finishes the Prog<->Prim (so far the Prog = Prim)

3)this push should not change results

jkozdon:
@charleskawczynski and @Zhengyu-Huang: I have added a possible interface for FVM reconstructions. Let me know what you think; it's not used yet, but the basic sketch is there: [`src/Numerics/DGMethods/FVReconstructions.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/e38a9bda4a52d3eaac9e8c801909ae101bfd553e/src/Numerics/DGMethods/FVReconstructions.jl)

jkozdon:
bors try

jkozdon:
bors try

Zhengyu-Huang:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
bors try

jkozdon:
I believe 🤞 this is ready for review.

The one thing I don't like is that the numbers changed in 65d1ddfb435ae68477d7e090d4fa1d939d74ea93 (`Adding more generic FV kernel`) going to the more generic stencil for one test on the sphere (this could be difference in which normals and geometry metrics are selected). And was only on a higher resolution run for the sphere. Changes were not large, but were there.

src/Numerics/DGMethods/FVReconstructions.jl
```diff
@@ -0,0 +1,174 @@
...
+       - (::AbstractReconstruction)(state_top, state_bottom, cell_states::SVector,
+                                    cell_weights)
+      compute the reconstruction
+
+(::AbstractReconstruction)(
+        state_top,
+        state_bottom,
+        cell_states,
+        cell_weights,
+    )
```
charleskawczynski, line 23: It looks like some of this section could be wrapped in code blocks

src/Numerics/DGMethods/FVReconstructions.jl
```diff
@@ -0,0 +1,174 @@
...
+end
+"""
+    FVLinear{W = 1} <: AbstractReconstruction
+
+Reconstruction type for limited linear reconstruction finite volume methods.
+
+!!! note
+   The optional type parameter `W` is mainly for debuggin purposes and allows
+   the stencil to be artificially widened to make sure the kernels work with
+   wide stencils.
```
charleskawczynski, line 73:
```suggestion
    The optional type parameter `W` is mainly for debugging purposes and allows
    the stencil to be artificially widened to make sure the kernels work with
    wide stencils.
```

src/Numerics/DGMethods/DGFVModel_kernels.jl
```diff
@@ -0,0 +1,800 @@
...
+
+        # Value in the stencil that corresponds to the top face with respect to
+        # face being updated
+        stencil_center = max(stencil_width, 1) + 1
+
+        # 1 → cell i, face i - 1/2
+        # 2 → cell i, face i + 1/2
+        local_state_face_prognostic = ntuple(Val(stencil_diameter)) do _
+            MArray{Tuple{num_state_prognostic}, FT}(undef)
+        end
```
Zhengyu-Huang, line 99: Is it an array of stencil_diameter MArray{Tuple{num_state_prognostic}, FT}(undef) ? stencil_diameter=2 for FVLinear, do we need 5 for WENO?
Do we need that many face prognostic?
What are these 1, 2 in the comments?

src/Numerics/DGMethods/DGFVModel_kernels.jl
```diff
@@ -0,0 +1,800 @@
...
+                Base.Cartesian.@nif 4 w -> (eV_up == w) w -> begin
+                    rng1, rng2 = stencil_center .+ (1 - w, w - 1)
+                    rng = SUnitRange(rng1, rng2)
+                    reconstruction!(
+                        local_state_face_primitive[1],
+                        local_state_face_primitive[2],
+                        local_state_primitive[rng],
+                        local_cell_weights[rng],
+                    )
+                end w -> throw(BoundsError(local_state_primitive, w))
```
Zhengyu-Huang, line 450: Is it rng = SUnitRange(stencil_center - eV_up+1, stencil_center + eV_up-1) ? 
Why do we need to throw error ?

Zhengyu-Huang (approved):
Thanks for your great effort!!!!

I mainly review the FV kernel, it looks great for me except 2 questions related to WENO reconstruction.

I think this PR is fully tested, (the spherical test has 1 element < stencil width in vertical so the FVLinear reconstruction might give slightly different results)

src/Numerics/DGMethods/FVReconstructions.jl
jkozdon: Good catch. Thanks!

src/Numerics/DGMethods/FVReconstructions.jl
jkozdon: Thanks!

src/Numerics/DGMethods/DGFVModel_kernels.jl
jkozdon: For WENO5 you will set the width to `2` in the `FVReconstructions` module for a  new `FVWENO` type. The diameter will be automatically computed.

We can probably shrink the `local_state_prognostic` to `width + 1` later. @mwarusz told me you still need the full auxiliary though. Once we have a fully working code, we will need to do some optimization passes, but I was going for correctness and not terrible performance to start. (I've included some of these in the comment section entitled `Optimization ideas`.)

The `1` and `2` refer to the bottom and top faces of the element being constructed (this is what I tried to indicate in the comment). I've updated the comment let me know if you have other suggestions.

jkozdon: I see that you were asking about face prognostic, yes you are write. This was a typo on my part. Good catch! Will change to `2`

src/Numerics/DGMethods/DGFVModel_kernels.jl
jkozdon: If this was a CPU only code, we could use `eV_up`, but on the GPU we need to generate a serious of `if...elseif....else` statements for each possible value of `eV_up` that could be encountered. So `w` in this code block takes the place of `eV_up` which will be set directly by the [`Base.Cartesian.@nif'](https://docs.julialang.org/en/v1/devdocs/cartesian/#Base.Cartesian.@nif) macro; the trouble with the GPU is though we have "dynamic syntax" the code still needs to be analyzed statically.

The `error` should never get hit since the `@assert stencil_width < 4` at the top of the code will prevent it, but I was just being defensive here. The error will only getting thrown by the `else` statement of the macro. (What's written should work for upto WENO5, if we go to WENO7 we would need to change the `4` to a `5`.

jkozdon: > I think this PR is fully tested, (the spherical test has 1 element < stencil width in vertical so the FVLinear reconstruction might give slightly different results)

I increased the number of elements to 3 in the spherical test (this was necessary to get your original linear reconstruction kernel to run on the test). The kernel @thomasgibson and I wrote, should work with 1 element in the vertical, and we could decrease it again if you wanted (though the numbers would need to be updated).

jkozdon:
> I think this PR is fully tested, (the spherical test has 1 element < stencil width in vertical so the FVLinear reconstruction might give slightly different results)

I increased the number of elements to 3 in the spherical test (this was necessary to get your original linear reconstruction kernel to run on the test). The kernel @thomasgibson and I wrote, should work with 1 element in the vertical, and we could decrease it again if you wanted (though the numbers would need to be updated).

jkozdon:
bors try

Zhengyu-Huang:
When could we merge this branch to the master? Since EDMF people plan to try the FVM for single column setting.

simonbyrne:
I think @mwarusz is still reviewing it.

src/Numerics/DGMethods/DGFVModel_kernels.jl
```diff
@@ -0,0 +1,800 @@
...
+
+        # XXX: will revisit this later for FVM
+        fill!(local_state_prognostic_bottom1, NaN)
+        fill!(local_state_gradient_flux_bottom1, NaN)
+        fill!(local_state_auxiliary_bottom1, NaN)
+
+        # The remainder model needs to know which direction of face the model is
+        # being evaluated for. In this case we only have `VerticalDirection()`
+        # faces
+        face_direction = (VerticalDirection())
```
charleskawczynski, line 159: Based on downstream calls,
```julia
    ::Dirs,
) where {NumSubFluxes, S, A, Dirs <: NTuple{2, Direction}}
```
It looks like this should be a tuple: `face_direction = (VerticalDirection(), )`.

jkozdon: Good catch, I dropped the comma it looks like.

jkozdon:
> I think @mwarusz is still reviewing it.

There are also `fixup!` commits that will require a rebase.

```sh
git rebase origin/master -i --autosquash
```

valeriabarra:
FYI, I just pushed some minor comment style fixes that I noticed while going through this PR. I hope this is ok.

I know @mwarusz mentioned he was going to make some fixes and push them to this PR. I think after those, this can be ready to merge

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -117,13 +199,219 @@ Computes the tendency terms compatible with `IncrementODEProblem`
...
+        state_prognostic,
+        t,
+        α,
+        β;
+        surface = :exterior,
+        dependencies = (
+            comp_stream,
+            exchange_state_prognostic,
+            exchange_state_gradient_flux,
+            #JK exchange_Qhypervisc_grad,
```
valeriabarra, line 391: Is this what needs to be turned on (i.e. decommented) when adding hyperdiffusion back? Should leave a note/comment?

mwarusz:
bors try

src/Numerics/DGMethods/DGModel.jl
jkozdon: Hopefully, this and a few other places.

We also need to add a test for hyperdiffusion with DG-FV.

I don't think that there is a ton that needs to happen since hyperdiffusion is (I believe), for now just going to be on the DG side.

Hopefully will be an easy follow-up PR!

jkozdon:
@mwarusz and @valeriabarra: Thanks for you feedback. I like the new `local_tendency` good catch!

I've squashed and rebased as needed.

src/Numerics/DGMethods/DGModel.jl
valeriabarra: Okay, I fixed it up!

valeriabarra (approved):

src/Numerics/DGMethods/DGModel.jl
valeriabarra: You can bors r+ at your convenience

jkozdon: Thanks.

mwarusz (approved):
LGTM. Great work Jeremy !

valeriabarra:
> LGTM. Great work !

Absolutely! Good job @jkozdon ! This was a big one! :rocket:

jkozdon:
Thanks for everyone's help!

cc: @Zhengyu-Huang @thomasgibson @mwarusz @valeriabarra @charleskawczynski

PR #1810, opened 2020-12-02
Update Grid to save radial direction mesh size Delta x3  for fvm hydrostatic balance preserving

Zhengyu-Huang:
### Description

Add Δx3 in vgeo, this is the 17 entry of vgeo, which is only used for vertical fvm (for hydrostatic balance reconstruction and initialization)

### DIFF
--- a/src/Numerics/Mesh/Grids.jl
+++ b/src/Numerics/Mesh/Grids.jl
@@ -657,6 +657,31 @@ function computegeometry_fvm(elemtocoord, D, ξ, ω, meshwarp)
 
         # compute MH and JvC
         horizontal_vertical_metrics(vgeo, Nq, ω)
+        # update JcV for the vertical FV scheme, which is 
+        # the radial grid size, 
+        # if vertical finite volume is applied in the last dimension
+        if Nq[dim] == 1
+            avg_den = 2 .^ (sum(Nq[1:dim] .== 1) - 1)
+            avg_ind = ntuple(i -> Colon(), dim - 1)
+            vgeo[:, _JcV, :] =
+                sum(
+                    sqrt.(
+                        (
+                            vgeo_N1_flds[_x1][avg_ind..., 2, :] .-
+                            vgeo_N1_flds[_x1][avg_ind..., 1, :]
+                        ) .^ 2 +
+                        (
+                            vgeo_N1_flds[_x2][avg_ind..., 2, :] .-
+                            vgeo_N1_flds[_x2][avg_ind..., 1, :]
+                        ) .^ 2 +
+                        (
+                            vgeo_N1_flds[_x3][avg_ind..., 2, :] .-
+                            vgeo_N1_flds[_x3][avg_ind..., 1, :]
+                        ) .^ 2,
+                    ),
+                    dims = findall(Nq[1:(dim - 1)] .== 1),
+                )[:] / (2avg_den)
+        end
         num_vgeo_handled += 2
 
         # Make sure we handled all the vgeo terms
@@ -869,36 +894,33 @@ function horizontal_vertical_metrics(vgeo, Nq, ω)
     # Compute |r'(ξ3)| for vertical line integrals
     if dim == 1
         MHJH .= 1
-        JcV .= J
     elseif dim == 2
-        map!(JcV, J, ξ1x1, ξ1x2) do J, ξ1x1, ξ1x2
-            x1ξ1 = J * ξ1x2
-            x2ξ2 = J * ξ1x1
-            hypot(x1ξ1, x2ξ2)
-        end
         map!(MHJH, J, ξ2x1, ξ2x2) do J, ξ2x1, ξ2x2
             hypot(J * ξ2x1, J * ξ2x2)
         end
         MHJH .= MH .* MHJH
 
     elseif dim == 3
-        map!(
-            #! format: off
-            JcV, J,
-            ξ1x1, ξ1x2, ξ1x3, ξ2x1, ξ2x2, ξ2x3,
-            #! format: on
-        ) do J, ξ1x1, ξ1x2, ξ1x3, ξ2x1, ξ2x2, ξ2x3
-            x1ξ3 = J * (ξ1x2 * ξ2x3 - ξ2x2 * ξ1x3)
-            x2ξ3 = J * (ξ1x3 * ξ2x1 - ξ2x3 * ξ1x1)
-            x3ξ3 = J * (ξ1x1 * ξ2x2 - ξ2x1 * ξ1x2)
-            hypot(x1ξ3, x2ξ3, x3ξ3)
-        end
         map!(MHJH, J, ξ3x1, ξ3x2, ξ3x3) do J, ξ3x1, ξ3x2, ξ3x3
             hypot(J * ξ3x1, J * ξ3x2, J * ξ3x3)
         end
         MHJH .= MH .* MHJH
     else
         error("dim $dim not implemented")
+    end
+
+    if Nq[dim] > 1
+        Nq_hori = (dim == 1) ? 1 : prod(Nq[1:(dim - 1)])
+        Δx3 =
+            sqrt.(
+                (x1[(end - Nq_hori + 1):end, :, :] - x1[1:Nq_hori, :, :]) .^ 2 +
+                (x2[(end - Nq_hori + 1):end, :, :] - x2[1:Nq_hori, :, :]) .^ 2 +
+                (x3[(end - Nq_hori + 1):end, :, :] - x3[1:Nq_hori, :, :]) .^ 2,
+            )
+        for i in 1:Nq[dim]
+            JcV[(1 + (i - 1) * Nq_hori):(Nq_hori + (i - 1) * Nq_hori), :, :] .=
+                Δx3 ./ 2
+        end
     end
 end
 # }}}


jkozdon (commented):
This is certainly one option for this. I'd like to understand the FVM hydrostatic balance a little better, since `Δx3` doesn't really enter into the coupled scheme directly I wondering if this is really the right quantity on highly skewed meshes.

src/Numerics/Mesh/Grids.jl
```diff
@@ -655,6 +657,50 @@ function computegeometry_fvm(elemtocoord, D, ξ, ω, meshwarp)
             num_vgeo_handled += 1
         end
 
+        # Compute Δx3, the radial grid size for finite volume method, 
+        # if vertical finite volume is applied in the last dimension
+        if Nq[dim] == 1
```
jkozdon, line 662: I don't think that this is quite right if only one of the horizontal dimensions is `N = 0`. I know we do not currently plan to allow this in the kernels, but the mesh stuff generally supports any dimension being `N = 0`, so I'd like to future proof this. (even if its with filling with `NaN` when it will break.

jkozdon:
I wondering if this would be better to handle once we add the surface `aux`. There we will need the surface points, and these could be derived from those when needed.

I imagine adding a surface mesh which by default is not put on the device but can be loaded to the device in the `DGFVMModel` constructor when needed (and for the right faces)

src/Numerics/Mesh/Grids.jl
Zhengyu-Huang: This is only used for vertical direction N=0 case

Zhengyu-Huang: Screen Shot 2020-12-02 at 8 52 01 AM
Screen Shot 2020-12-02 at 8 52 28 AM

jkozdon: I mean that the code doesn't work if polynomial order is something like `(0, 3, 0)`

Zhengyu-Huang: Yes, the code works only for (n,n,0) I think we need to require first two are the same ?

jkozdon: My point is the mesh generator supports any numbers, so I'd like this to do something reasonable in all cases. (Even if it's filling with `NaN` in cases we do not currently handle.

Zhengyu-Huang: Just update a version to handle (0,3,0)

Zhengyu-Huang:
bors try

jkozdon:
I was thinking about this, and I believe this should be encoding the same data as 'jcv' which is the metric for the vertical line integral (I believe that 'Δx3' should be 'Jcv / 2'). We should update the vertical line integral test for 'N=0', and make sure this is correctly calculated. (Cc @valeriabarra)

valeriabarra:
> I was thinking about this, and I believe this should be encoding the same data as 'jcv' which is the metric for the vertical line integral (I believe that 'Δx3' should be 'Jcv / 2'). We should update the vertical line integral test for 'N=0', and make sure this is correctly calculated. (Cc @valeriabarra)

Sorry, I need to understand! What is `jcv` ? And why for FV, we need `Δx3`? Don't we need the elem size along the z direction?

jkozdon:
Jcv is the vertical Jacobian for line integrals in the vertical. So if we want to integrate a quantity up the stack of degrees of freedom this would be the weighting factor for the line integral (sum discretely). This should be the same as the element distance in the vertical (up to a factor of 2 since the reference element is of length 2).

I can explain more latter if this isn't clear.

valeriabarra:
> Jcv is the vertical Jacobian for line integrals in the vertical. So if we want to integrate a quantity up the stack of degrees of freedom this would be the weighting factor for the line integral (sum discretely). This should be the same as the element distance in the vertical (up to a factor of 2 since the reference element is of length 2).
> 
> I can explain more latter if this isn't clear.

Thanks! Much clearer! Wondering what the `c` in `Jcv` stands for then. 

I know that the length of ref elem is 2. I think I was a bit puzzled by the nomenclature chosen. But if `Δx3` stands for the elem size in the vertical (i.e., third direction) then it is clear

jkozdon:
> Thanks! Much clearer! Wondering what the `c` in `Jcv` stands for then.

I think when I named it I was thinking "Jacobian, curved integral, vertical". Terrible names are one of my specialties! (Feel free to rename it!).

Its all used in [`kernel_indefinite_stack_integral!`](https://github.com/CliMA/ClimateMachine.jl/blob/4fbd57d71275626abe27c7ea31509a4a6815d572/src/Numerics/DGMethods/DGModel_kernels.jl#L1980-L2106) and the two tests we want to reproduce for FVM are [`integral_test.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/4fbd57d71275626abe27c7ea31509a4a6815d572/test/Numerics/DGMethods/integral_test.jl) and [`integral_test_sphere.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/4fbd57d71275626abe27c7ea31509a4a6815d572/test/Numerics/DGMethods/integral_test_sphere.jl)

> I know that the length of ref elem is 2. I think I was a bit puzzled by the nomenclature chosen. But if `Δx3` stands for the elem size in the vertical (i.e., third direction) then it is clear

For curved DG, this value will be different at each degree of freedom in the element, hence the reason it is not strictly `Δx3` for DG. But for FVM, it should be `Δx3 / 2` I believe. Since when you approximate the line integral you will have `JcV * w` where `w` is the quadrature weight, and for FVM the quadrature weight will be `w = 2`.

Zhengyu-Huang:
I have a GCM p0 with Z = 30e3 , and 20 vertical elements, the element size in the vertical direction should be 1500, JcV =  749.5014518561622

jkozdon:
Is it uniformly that or does it vary in the stack / element?

Zhengyu-Huang:
> Is it uniformly that or does it vary in the stack / element?

It is uniform I think there is machine error ?

jkozdon:
Thanks. We can take a look next week. We can always just store your computed value in Jcv since I think we will want this to be correct.

jkozdon:
I think the best way forward would be to force the true value into Jcv, so set Jcv = dx3 / 2.

I think that I know what's going on. We are using the approximate 3d metrics to compute the path integral metrics, but since the metrics are approximated there is error.

I believe that if you refined the mesh the result would get better.

We should probably switch to computing the path integral metrics directly, so that we can ensure that even for dg when you integrate a constant along the radius of the sphere you get the difference between the outer and inner radius of the sphere.

Out of interest how accurate is the N=1 value?

Zhengyu-Huang:
Update the Jcv as Dx3/ω for both FV & DG , will write some gcm tests

jkozdon:
> Update the Jcv as Dx3/2 for both FV & DG , since the vertical direction mesh mapping is always linear, is this true ?
> 
> ∫ f(z) dz = ∫ f(ξ₃) JcV dξ₃ = ∑_i f(ξ₃_i) JcV_i ω₃_i = ∑_i f(ξ₃_i) ω₃_i Dx3 / 2

I don't think that this is true with orography, we have warping inside the element too.

> I have a GCM p0 with Z = 30e3 , and 20 vertical elements, the element size in the vertical direction should be 1500, JcV = 749.5014518561622

Can you post the test that you saw this with? I get with my testing that:

```
extrema(grid.vgeo[:, Grids._JcV, :]) = (749.9999999935759, 750.0000000085739)
```

jkozdon:
> > Update the Jcv as Dx3/2 for both FV & DG , since the vertical direction mesh mapping is always linear, is this true ?
> > ∫ f(z) dz = ∫ f(ξ₃) JcV dξ₃ = ∑_i f(ξ₃_i) JcV_i ω₃_i = ∑_i f(ξ₃_i) ω₃_i Dx3 / 2
> 
> I don't think that this is true with orography, we have warping inside the element too.
> 
> > I have a GCM p0 with Z = 30e3 , and 20 vertical elements, the element size in the vertical direction should be 1500, JcV = 749.5014518561622
> 
> Can you post the test that you saw this with? I get with my testing that:
> 
> ```
> extrema(grid.vgeo[:, Grids._JcV, :]) = (749.9999999935759, 750.0000000085739)
> ```

Nevermind, there is an extra `polynomial_order = 4` set in the [`integral_test_sphere.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/7bd44d9990fc452fff133f0f91692662d3b6d71d/test/Numerics/DGMethods/integral_test_sphere.jl#L171-L178)

Is see `extrema(grid.vgeo[:, Grids._JcV, :]) = (744.981471153572, 746.229210745348)` now.

```diff
❯ git diff
diff --git a/test/Numerics/DGMethods/integral_test_sphere.jl b/test/Numerics/DGMethods/integral_test_sphere.jl
index 39a406040..4d2b6908f 100644
--- a/test/Numerics/DGMethods/integral_test_sphere.jl
+++ b/test/Numerics/DGMethods/integral_test_sphere.jl
@@ -153,6 +153,7 @@ function test_run(mpicomm, topl, ArrayType, N, FT, Rinner, Router)
         CentralNumericalFluxSecondOrder(),
         CentralNumericalFluxGradient(),
     )
+    @show extrema(grid.vgeo[:, Grids._JcV, :])

     Q = init_ode_state(dg, FT(0))
     dQdt = similar(Q)
@@ -168,14 +169,12 @@ let

     mpicomm = MPI.COMM_WORLD

-    polynomialorder = 4
+    polynomialorder = 0

-    base_Nhorz = 4
-    base_Nvert = 2
-    Rinner = 1 // 2
-    Router = 1
-
-    polynomialorder = 4
+    base_Nhorz = 10
+    base_Nvert = 20
+    Rinner = 30e3
+    Router = Rinner + 1500base_Nvert

     expected_result = [
         4.662884229467401e-7,
```

PR #1811, opened 2020-12-02
Expose parameters for SBL calibration.

ilopezgp:
### Description

This PR makes the following changes, necessary for a first proof of concept of CalibrateEmulateSample-CLIMAParameters-ClimateMachine joint use:

- Adds a new ArgParse argument option to `stable_bl_les.jl` that allows ClimateMachine to read a certain set of parameters, identified by the path to the parameter definition file.
- Takes two turbulent parameters, `C_smag` and `C_drag`, from CLIMAParameters. Before, they were hardcoded in the example.

### DIFF
--- a/experiments/AtmosLES/stable_bl_les.jl
+++ b/experiments/AtmosLES/stable_bl_les.jl
@@ -1,4 +1,5 @@
 using Random
+
 include("stable_bl_model.jl")
 
 function add_perturbations!(state, localgeo)
@@ -9,26 +10,16 @@ function add_perturbations!(state, localgeo)
     end
 end
 
-function main()
+function set_clima_parameters(filename)
+    eval(:(include($filename)))
+end
 
-    # TODO: this will move to the future namelist functionality
-    sbl_args = ArgParseSettings(autofix_names = true)
-    add_arg_group!(sbl_args, "StableBoundaryLayer")
-    @add_arg_table! sbl_args begin
-        "--surface-flux"
-        help = "specify surface flux for energy and moisture"
-        metavar = "prescribed|bulk"
-        arg_type = String
-        default = "bulk"
-    end
-
-    cl_args = ClimateMachine.init(parse_clargs = true, custom_clargs = sbl_args)
+function main(cl_args)
 
     surface_flux = cl_args["surface_flux"]
 
     FT = Float64
     config_type = AtmosLESConfigType
-
     # DG polynomial order
     N = 4
     # Domain resolution and size
@@ -51,14 +42,14 @@ function main()
     # Choose default IMEX solver
     ode_solver_type = ClimateMachine.ExplicitSolverType()
 
-    C_smag = FT(0.23)
+    C_smag_ = C_smag(param_set) #FT(0.23)
 
     model = stable_bl_model(
         FT,
         config_type,
         zmax,
         surface_flux;
-        turbulence = SmagorinskyLilly{FT}(C_smag),
+        turbulence = SmagorinskyLilly{FT}(C_smag_),
     )
 
     ics = model.problem.init_state_prognostic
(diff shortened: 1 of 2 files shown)


experiments/AtmosLES/stable_bl_les.jl
```diff
@@ -11,11 +11,19 @@ function main()
...
+        help = "specify CLIMAParameters version"
+        arg_type = Union{String, Nothing}
+        default = nothing
     end
 
     cl_args = ClimateMachine.init(parse_clargs = true, custom_clargs = sbl_args)
 
     surface_flux = cl_args["surface_flux"]
+    if !isnothing(cl_args["CP_Version"])
+        include("clima_param_defs_$(cl_args["CP_Version"]).jl")
```
charleskawczynski, line 25: This include must be located at the top level scope (we’re defining global methods). Alternatively, we might be able to use `eval`. Looping in @jakebolewski

ilopezgp: Got it. For reference, this is what that file might look like:

```
CLIMAParameters.Atmos.SubgridScale.
                C_smag(::EarthParameterSet) =
                0.30016549507297424
CLIMAParameters.Atmos.SubgridScale.
                C_drag(::EarthParameterSet) =
                0.0009143650169447696
```

jakebolewski: shouldn't this take a path to a file (with the method overloads?)  It seems a bit hacky to do it this way but probably will work for now.   `include()` evals in the global toplevel scope of the containing module so it should be fine I would think

ilopezgp: FYI, as you said this does not work. I just checked and the values used for these two parameters in the simulation are the default ones.

jakebolewski: Isn't that because of the way the methods are imported into the module namespace? 
https://github.com/CliMA/ClimateMachine.jl/blob/a5dd39fb7abab30523d33a0f3f3703283abcdad5/experiments/AtmosLES/stable_bl_model.jl#L42

doing it this way they are not available for extension:

https://docs.julialang.org/en/v1/manual/modules/#Summary-of-module-usage

charleskawczynski: > `include() evals in the global toplevel scope of the containing module so it should be fine I would think

That would seem odd. Where/when exactly does the include get executed? This `include` is inside a function. I think we need to `eval` into `Main` scope.

charleskawczynski: My thought/suggestion would be to define
```julia
    function set_clima_parameters(filename)
      eval(:(include($filename)))
    end
```
And call this in the function in `main`

jakebolewski (approved):

experiments/AtmosLES/stable_bl_les.jl
```diff
@@ -11,6 +15,11 @@ function main()
         metavar = "prescribed|bulk"
         arg_type = String
         default = "bulk"
+
+        "--CP-Version"
```
jakebolewski, line 19: I think this should specify the path to the definition file rather than a specific version, the you are not reliant on a specific formatting assumption (name of the file and how you are formatting it) in the CES driver

experiments/AtmosLES/stable_bl_les.jl
```diff
@@ -81,4 +90,19 @@ function main()
...
+@add_arg_table! param_args begin
+    "--CP-Version"
+    help = "specify CLIMAParameters version"
+    arg_type = Union{String, Nothing}
+    default = nothing
+end
+
+parsed_args = parse_args(ARGS, param_args)
+if !isnothing(parsed_args["CP_Version"])
+    filename = "clima_param_defs_$(parsed_args["CP_Version"]).jl"
```
jakebolewski, line 104: So here: filename would be the path to the file.  include doesn't have to be a relative path (it can be an absolute path))

charleskawczynski (approved):

jakebolewski (approved):
LGTM

PR #1813, opened 2020-12-02
add simple rain CI checks

trontrytel:
### Description

This PR adds a simple CI test for rain and moisture variables for LES dycoms case.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1041,7 +1041,7 @@ steps:
   - label: "gpu_dycoms_with_precip"
     key: "gpu_dycoms_with_precip"
     command:
-      - "mpiexec julia --color=yes --project experiments/AtmosLES/dycoms.jl --fix-rng-seed --moisture-model nonequilibrium --precipitation-model rain"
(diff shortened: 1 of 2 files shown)


charleskawczynski (commented):
Looks good. Should we put this in a separate function/file `sanity_test(m::AtmosModel, Q)`, as I could imagine these tests/checks getting large and maybe bloating the driver?. Also, we can check `m isa NonEquilibrium` in `sanity_test` instead of `moisture_model == "nonequilibrium"` so that we don't need to pass tons of arguments through (same with the precip model).

jakebolewski (approved):
:100:

trontrytel:
> Looks good. Should we put this in a separate function/file `sanity_test(m::AtmosModel, Q)`, as I could imagine these tests/checks getting large and maybe bloating the driver?. Also, we can check `m isa NonEquilibrium` in `sanity_test` instead of `moisture_model == "nonequilibrium"` so that we don't need to pass tons of arguments through (same with the precip model).

I'm piggybacking on an existing LES experiment for Dycoms. The default setup there is `noprecip` and `equilibrium` options. Thats why I added the command line arguments for it. 

Not sure what is the right folder structure. Up to you guys, I have no preference either way

PR #1815, opened 2020-12-02
Convert entrainment and detrainment to fluxes which are positive definite to avoid issues with negative upd.w

yairchn:
### Description
This PR switches entrainment and detrainment from the fractional (~1/length) form to their flux (ρ/time) form. These fluxes are positive definite irrespective of the sign of updraft vertical velocity. The relation between the fractional and fluxes are:
ρaw * ε_dyn = E_dyn
ρaw * ε_trb  = E_trb
ρaw * δ_dyn = Δ_dyn
 
In addition I have removed the limiter functions as we are not using them now. If needed in the future they will be re-coded according to the new fluxes of entrainment and detrainment.
A single change to the mse value of ρaθ_liq_cv was needed in order to pass the test.

### DIFF
--- a/test/Atmos/EDMF/edmf_kernels.jl
+++ b/test/Atmos/EDMF/edmf_kernels.jl
@@ -533,7 +533,9 @@ function flux_first_order!(
     # Aliases:
     gm = state
     up = state.turbconv.updraft
+    en = state.turbconv.environment
     up_flx = flux.turbconv.updraft
+    en_flx = flux.turbconv.environment
     N_up = n_updrafts(turbconv)
 
     ρ_inv = 1 / gm.ρ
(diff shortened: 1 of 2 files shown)


test/Atmos/EDMF/closures/entr_detr.jl
```diff
@@ -44,20 +44,18 @@ function entr_detr(
...
     ρ_inv = 1 / gm.ρ
     a_up_i = up[i].ρa * ρ_inv
-
-    lim_ϵ = entr.lim_ϵ
+    lim_E = entr.lim_ϵ
     lim_amp = entr.lim_amp
     w_min = entr.w_min
     # precompute vars
     w_up_i = up[i].ρaw / up[i].ρa
+    temp = max(en.ρatke, 0) * ρ_inv / env.a
```
charleskawczynski, line 52: It doesn't look like this variable (`temp`) is used anywhere

charleskawczynski (commented):
Looks good to me. I think we just need to squash and remove the unused variable `temp`

PR #1817, opened 2020-12-03
Add turbconv_source to source_default and modify CBL_model to follow the structure in SBL_model

yairchn:
### Description
This PR fixes a critical bug in the construction of the sources in the Stable and Convective BL model files. 
I also noticed that the two files have slightly different structure and made sure to adjust the convective_bl_model.jl file to be similar to the stable counterpart. 


experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -276,23 +306,28 @@ function convective_bl_model(
                 )),
                 energy = energy_bc,
                 moisture = moisture_bc,
-                turbconv = turbconv_bcs(turbconv)[1],
```
charleskawczynski, line 279: Deleting this means that `boundary_state(::EDMF)` will not be called. Is that intended?

yairchn: Great catch this mean that until now we did not use the BCs for turbconv in SBL or CBL!
I revised in both

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -206,14 +208,9 @@ function convective_bl_model(
...
+    moisture_model = "dry",
 ) where {FT}
-
-    ics = init_convective_bl!     # Initial conditions
-
-    C_smag = FT(0.23)     # Smagorinsky coefficient
-    C_drag = FT(0.001)    # Momentum exchange coefficient
-    z_sponge = FT(2560)     # Start of sponge layer
-    α_max = FT(0.75)       # Strength of sponge layer (timescale)
+c5)       # Strength of sponge layer (timescale)
```
charleskawczynski, line 213: Looks like a typo or something here

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -279,20 +304,26 @@ function convective_bl_model(
...
     model = AtmosModel{FT}(
-        AtmosLESConfigType,
+        config_type,
         param_set;
         problem = problem,
         turbulence = SmagorinskyLilly{FT}(C_smag),
-        moisture = EquilMoist{FT}(; maxiter = 5, tolerance = FT(0.1)),
-        source = source,
+        moisture = moisture,
+        source = source_default,
```
charleskawczynski, line 323:
```suggestion
        source = source,
```
It looks like we want `source` and not `source_default`

yairchn: Does that me and that in 236 it should be source rather than source_default as well?

charleskawczynski:
bors try

experiments/AtmosLES/convective_bl_model.jl
```diff
@@ -206,14 +208,9 @@ function convective_bl_model(
...
     turbconv = NoTurbConv(),
+    moisture_model = "dry",
 ) where {FT}
 
-    ics = init_convective_bl!     # Initial conditions
-
-    C_smag = FT(0.23)     # Smagorinsky coefficient
-    C_drag = FT(0.001)    # Momentum exchange coefficient
-    z_sponge = FT(2560)     # Start of sponge layer
-    α_max = FT(0.75)       # Strength of sponge layer (timescale)
```
ilopezgp, line 216: Where did these definitions go?

experiments/AtmosLES/stable_bl_model.jl
```diff
@@ -329,7 +332,7 @@ function stable_bl_model(
         problem = problem,
         turbulence = SmagorinskyLilly{FT}(C_smag),
         moisture = moisture,
-        source = source_default,
+        source = source,
```
ilopezgp, line 335: I think this is source_default in the convective_bl case, in case we want to homogenize notation

charleskawczynski: No, this needs to be `source`, otherwise it'll break the `NonEquil` moisture model case (look above).

ilopezgp (approved):

charleskawczynski (approved):
LGTM

PR #1818, opened 2020-12-03
add solid body rotation test with topography (DCMIP 2-0-0)

jiahe23:
### Description

co-authored with @akshaysridhar 

This PR enables enables users to define topography on the cubed sphere. It adds the following functionalities in `Topologies.jl`:
- `compute_lat_long` function to compute lat/long given cubed sphere coordinates and face id;
- `AnalyticalTopography` abstract type to hold topography function;
- `cubedshelltopowarp` function to compute the warp grid given topography function.

A `solid_body_rotation_mountain.jl` test case into the `TestCase` suite, to test the hydrostatic balance with topography. It follows the DCMIP2012 2-0-0 mountain set-up.

CI test on this case only makes sure the simulation can run but it does not check on the relative error. It will be later on added to the weekly test for visualization.

### DIFF
--- a/experiments/TestCase/solid_body_rotation_mountain.jl
+++ b/experiments/TestCase/solid_body_rotation_mountain.jl
@@ -1,0 +1,225 @@
+#!/usr/bin/env julia --project
+using ClimateMachine
+ClimateMachine.init(parse_clargs = true)
+
+using ClimateMachine.Atmos
+using ClimateMachine.Orientations
+using ClimateMachine.ConfigTypes
+using ClimateMachine.NumericalFluxes
+using ClimateMachine.Diagnostics
+using ClimateMachine.GenericCallbacks
+using ClimateMachine.ODESolvers
+using ClimateMachine.TurbulenceClosures
+using ClimateMachine.SystemSolvers: ManyColumnLU
+using ClimateMachine.Mesh.Filters
+using ClimateMachine.Mesh.Grids
+using ClimateMachine.Mesh.Topologies
+using ClimateMachine.Mesh.Interpolation
+using ClimateMachine.TemperatureProfiles
+using ClimateMachine.VariableTemplates
+using ClimateMachine.Thermodynamics: air_density, total_energy
+
+using LinearAlgebra
+using StaticArrays
+
+using CLIMAParameters
+using CLIMAParameters.Planet: day, planet_radius
+struct EarthParameterSet <: AbstractEarthParameterSet end
+const param_set = EarthParameterSet()
+
+function set_topofun(f, r_inner, r_outer, topography)
+    function wrapper_topo(a, b, c)
+        return f(
+            a,
+            b,
+            c,
+            max(abs(a), abs(b), abs(c));
+            r_inner = r_inner,
+            r_outer = r_outer,
+            topography = topography,
+        )
+    end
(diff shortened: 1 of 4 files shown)


jakebolewski (approved):
The test definitions LGTM

experiments/TestCase/solid_body_rotation_mountain.jl
```diff
@@ -0,0 +1,320 @@
...
+using ClimateMachine.Mesh.Filters
+using ClimateMachine.Mesh.Grids
+using ClimateMachine.Mesh.Interpolation
+using ClimateMachine.TemperatureProfiles
+using ClimateMachine.VariableTemplates
+using ClimateMachine.Thermodynamics: air_density, total_energy
+
+using LinearAlgebra
+using StaticArrays
+using Test
```
jakebolewski, line 23: this is not needed if no tests are defined

experiments/TestCase/solid_body_rotation_mountain.jl
```diff
@@ -0,0 +1,320 @@
...
+function cubedshelltopowarp(
+    a,
+    b,
+    c,
+    R = max(abs(a), abs(b), abs(c));
+    r_inner = _planet_radius,
+    r_outer = _planet_radius + domain_height,
+)
+
+    function f(sR, ξ, η, faceid)
```
tapios, line 53: It seems the cubed sphere setup should be going somewhere where it can be reused more generally, not into this test setup.

experiments/TestCase/solid_body_rotation_mountain.jl
```diff
@@ -0,0 +1,320 @@
...
+            φ = atan(Y * cos(atan(X)))
+        elseif faceid == 5
+            λ = atan(X, -Y) + π
+            φ = atan(1 / sqrt(δ - 1))
+        elseif faceid == 6
+            λ = atan(X, Y)
+            φ = -atan(1 / sqrt(δ - 1))
+        end
+
+        r_m = acos(sin(φ_m) * sin(φ) + cos(φ_m) * cos(φ) * cos(λ - λ_m))
```
akshaysridhar, line 89: @jiahe23 We could update this to allow a user to simply specify the coordinate dependent radial warps; the face-identification can happen in the inner meshwarp function, in general I imagine we don't need access to that component at the experiment/driver level (?) [i.e. collapse lines 89-106 in some sensible way]

experiments/TestCase/solid_body_rotation_mountain.jl
jiahe23: Done!

experiments/TestCase/solid_body_rotation_mountain.jl
jiahe23: The following functionalities are moved to `Topologies.jl`:
- cubed sphere set up and topowarp function
- helper functions that computes lat/long
- function that generate the dcmip 2-0-0 mountain (since it is an archived dcmip case)

In the experiment driver, users can specify the topography (either choosing from what we have in `Topologies.jl` or define their own following the `AnalyticalTopography` type) meshwarp argument.

jiahe23:
bors try

jiahe23:
bors try-

PR #1819, opened 2020-12-03
Move tracers first order flux to new tendency specification

charleskawczynski:
### Description

This PR moves tracers first order flux to the new tendency specification.

### DIFF
--- a/src/Atmos/Model/declare_prognostic_vars.jl
+++ b/src/Atmos/Model/declare_prognostic_vars.jl
@@ -18,4 +18,4 @@ abstract type Precipitation <: PrognosticVariable end
 struct Rain <: Precipitation end
 struct Snow <: Precipitation end
 
-struct Tracers <: PrognosticVariable end
+struct Tracers{N} <: PrognosticVariable end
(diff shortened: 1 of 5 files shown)


src/Atmos/Model/declare_prognostic_vars.jl
```diff
@@ -18,4 +18,4 @@ abstract type Precipitation <: PrognosticVariable end
 struct Rain <: Precipitation end
 struct Snow <: Precipitation end
 
-struct Tracers <: PrognosticVariable end
+struct Tracers{N} <: PrognosticVariable end
```
kpamnany, line 21: Presumably `N` is the number of tracers? I'm not seeing it used though?

charleskawczynski: Yes, should we add a doc string? It's not really being used, but it shows up in the tendency table.

kpamnany: It can be added when it's used.

kpamnany (approved):

PR #1830, opened 2020-12-07
Add prim<->prog conversion

charleskawczynski:
### Description

This adds `primitive_to_prognostic!` and `prognostic_to_primitive!` methods to `BalanceLaws`, as well as an `AtmosModel` implementation for all moisture models.

### DIFF
--- a/src/Atmos/Model/prog_prim_conversion.jl
+++ b/src/Atmos/Model/prog_prim_conversion.jl
@@ -1,0 +1,174 @@
+####
+#### prognostic_to_primitive! and primitive_to_prognostic!
+####
+
+####
+#### Wrappers (entry point)
+####
+
+"""
+    prognostic_to_primitive!(atmos::AtmosModel, prim::Vars, prog::Vars, aux::Vars)
+
+Convert prognostic variables `prog` to primitive
+variables `prim` for the atmos model `atmos`.
+
+!!! note
+    The only field in `aux` required for this
+    method is the geo-potential.
+"""
+prognostic_to_primitive!(atmos::AtmosModel, prim::Vars, prog::Vars, aux) =
+    prognostic_to_primitive!(
+        atmos,
+        atmos.moisture,
+        prim,
+        prog,
+        Thermodynamics.internal_energy(
+            prog.ρ,
+            prog.ρe,
+            prog.ρu,
+            gravitational_potential(atmos.orientation, aux),
+        ),
+    )
+
+"""
+    primitive_to_prognostic!(atmos::AtmosModel, prog::Vars, prim::Vars, aux::Vars)
+
+Convert primitive variables `prim` to prognostic
+variables `prog` for the atmos model `atmos`.
+
+!!! note
+    The only field in `aux` required for this
+    method is the geo-potential.
+"""
+primitive_to_prognostic!(atmos::AtmosModel, prog::Vars, prim::Vars, aux) =
+    primitive_to_prognostic!(
+        atmos,
+        atmos.moisture,
+        prog,
+        prim,
+        gravitational_potential(atmos.orientation, aux),
+    )
+
+####
+#### prognostic to primitive
+####
+
+function prognostic_to_primitive!(
+    atmos,
+    moist::DryModel,
+    prim::Vars,
+    prog::Vars,
+    e_int::AbstractFloat,
+)
+    ts = PhaseDry(atmos.param_set, e_int, prog.ρ)
+    prim.ρ = prog.ρ
+    prim.u = prog.ρu ./ prog.ρ
+    prim.p = air_pressure(ts)
+end
+
+function prognostic_to_primitive!(
+    atmos,
+    moist::EquilMoist,
+    prim::Vars,
+    prog::Vars,
+    e_int::AbstractFloat,
+)
+    FT = eltype(prim)
+    ts = PhaseEquil(
+        atmos.param_set,
+        e_int,
+        prog.ρ,
+        prog.moisture.ρq_tot / prog.ρ,
+        # 20,       # can improve test error with better convergence
+        # FT(1e-3), # can improve test error with better convergence
+    )
+    prim.ρ = prog.ρ
+    prim.u = prog.ρu ./ prog.ρ
+    prim.p = air_pressure(ts)
+    prim.moisture.q_tot = PhasePartition(ts).tot
+end
+
+function prognostic_to_primitive!(
+    atmos,
+    moist::NonEquilMoist,
+    prim::Vars,
+    prog::Vars,
+    e_int::AbstractFloat,
+)
+    q_pt = PhasePartition(
+        prog.moisture.ρq_tot / prog.ρ,
+        prog.moisture.ρq_liq / prog.ρ,
+        prog.moisture.ρq_ice / prog.ρ,
+    )
+    ts = PhaseNonEquil(atmos.param_set, e_int, prog.ρ, q_pt)
+    prim.ρ = prog.ρ
+    prim.u = prog.ρu ./ prog.ρ
+    prim.p = air_pressure(ts)
+    prim.moisture.q_tot = PhasePartition(ts).tot
+    prim.moisture.q_liq = PhasePartition(ts).liq
+    prim.moisture.q_ice = PhasePartition(ts).ice
+end
+
+####
+#### primitive to prognostic
+####
+
+function primitive_to_prognostic!(
+    atmos,
+    moist::DryModel,
+    prog::Vars,
+    prim::Vars,
+    e_pot::AbstractFloat,
+)
+    ts = PhaseDry_ρp(atmos.param_set, prim.ρ, prim.p)
+    e_kin = prim.u' * prim.u / 2
+
+    prog.ρ = prim.ρ
+    prog.ρu = prim.ρ .* prim.u
+    prog.ρe = prim.ρ * total_energy(e_kin, e_pot, ts)
+end
+
+function primitive_to_prognostic!(
+    atmos,
+    moist::EquilMoist,
+    prog::Vars,
+    prim::Vars,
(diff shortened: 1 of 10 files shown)


tapios:
To avoid confusion with common usage of the terms, can we change `prognostic` here throughout to `conservative`? (The issue is that we call all explicitly time dependent variables prognostic, and in that sense, the primitive variables are also prognostic.)

charleskawczynski:
> can we change `prognostic` here throughout to `conservative`?

This would probably be better as it's own PR because `prognostic` is used _very_ extensively throughout the code.

tapios:
> > can we change `prognostic` here throughout to `conservative`?
> 
> This would probably be better as it's own PR because `prognostic` is used _very_ extensively throughout the code.

Ok. But would be good to change globally. This will cause confusion for new folks otherwise.

Zhengyu-Huang:
Hi Charlie, how's this PR going ?

charleskawczynski:
Good, working on the tests now. Should be ready in a day or so 🤞🏻. This is dependent on #1852. Can you review #1852? How are the other pieces coming?

charleskawczynski:
I've added the tests today. We're getting pretty close to finishing this, there are 4 remaining failing identities (out of 38). I'll open this PR for review once I figure out what's going on.

charleskawczynski:
Alright, squashed the bug. Turned out to be a missing density factor. This is nearly ready, the only "loose end" to tie up is in testing that the translation is machine-precision accurate. Right now, we have a few approximations that, IIUC, we'll likely use in production:

 - Not performing saturation adjustment in `primitive_to_prognostic!` (for EquilMoist case only) because the resulting error is very small
 - Not reaching machine epsilon in `prognostic_to_primitive!` (for EquilMoist case only) for performance reasons

With the default settings, the _largest_ error in total energy (O(1e4)) are:
```julia
err_max_fwd = 6.446552080888068
err_max_bwd = 0.6120574663073057
```

I've confirmed that our approximations are the source of the error by adjusting the number of iterations and tolerance to (20, 1e-3) on the `PhaseEquil` constructor, which then yields:
```julia
err_max_fwd = 0.08718594611855224
err_max_bwd = 0.0076692650764016435
```

Improving on this error is a separate (thermo) issue. So, I've used the thermo absolute tolerance on total energy to allow the tests to pass and we can improve things later if needed.

charleskawczynski:
Oof, I just realized that I was iterating in `primitive_to_prognostic!` with the listed numbers above. Without iterating, the largest total energy errors are:
```julia
err_max_fwd = 6113.237107359106
err_max_bwd = 0.6120574663073057
```
Hmm. Might need advisement on what we should do here if this is not okay, or what the default behavior should be. @tapios?

My thought would be to test/leave things with high accuracy (with iterations) for now, and then make approximations later if performance is important.

Zhengyu-Huang:
Out of curiosity, is saturation adjustment done by Newton iteration to find the scalar root? Is there discontinuity in the equation that prevents quadratic convergence?

charleskawczynski:
> Out of curiosity, is saturation adjustment done by Newton iteration to find the scalar root? Is there discontinuity in the equation that prevents quadratic convergence?

Yep, we're using Newton iteration. There are discontinuities, and I'm not sure exactly what convergence rate we're achieving. There's quite a bit of discussion on this in #263

tapios:
> Out of curiosity, is saturation adjustment done by Newton iteration to find the scalar root? Is there discontinuity in the equation that prevents quadratic convergence?

Yes, there's a discontinuity at the phase transition (freezing point).

tapios:
> Oof, I just realized that I was iterating in `primitive_to_prognostic!` with the listed numbers above. Without iterating, the largest total energy errors are:
> 
> ```julia
> err_max_fwd = 6113.237107359106
> err_max_bwd = 0.6120574663073057
> ```
> 
> Hmm. Might need advisement on what we should do here if this is not okay, or what the default behavior should be. @tapios?
> 
> My thought would be to test/leave things with high accuracy (with iterations) for now, and then make approximations later if performance is important.

Can you still make the error arbitrarily small by iterating more/lowering tolerances? I wouldn't worry about large errors without saturation adjustments in your tests--you are testing large variations. The point of not using SA in the PDE code is that the reconstructions only involve nearby points, generally with small temperature differences.

charleskawczynski:
> Can you still make the error arbitrarily small by iterating more/lowering tolerances?

It seems like it. As I reported above, I can reach

```julia
err_max_fwd = 0.08718594611855224
err_max_bwd = 0.0076692650764016435
```
with 20 iterations and `1e-3` temperature tolerance. I tried going lower with the temperature tolerance (e.g., `1e-5`), but then we can't converge in ~100 iterations. These errors are really quite small though considering the magnitude of `ρe`.

tapios:
> > Can you still make the error arbitrarily small by iterating more/lowering tolerances?
> 
> It seems like it. As I reported above, I can reach
> 
> ```julia
> err_max_fwd = 0.08718594611855224
> err_max_bwd = 0.0076692650764016435
> ```
> 
> with 20 iterations and `1e-3` temperature tolerance. I tried going lower with the temperature tolerance (e.g., `1e-5`), but then we can't converge in ~100 iterations. These errors are really quite small though considering the magnitude of `ρe`.

It's fine then. This is not expected to work without SA unless the initial values are close. 

Did you test that conservative -> prognostic -> conservative is the identity?

charleskawczynski:
> It's fine then. This is not expected to work without SA unless the initial values are close.

Sounds good

> Did you test that conservative -> primitive -> conservative is the identity?

Yep, that's basically what we're testing in `prog_prim_conversion.jl`. In the default `PhaseEquil` behavior we have:

```julia
err_max_fwd = 6.446552080888068
err_max_bwd = 0.6120574663073057
```
Which passes the `atol_energy = cv_d(param_set) * atol_temperature` tolerance in the thermo test suite. All other fields are returning an exact identity within machine accuracy.

src/BalanceLaws/prog_prim_conversion.jl
```diff
@@ -0,0 +1,43 @@
...
+    aux::AbstractArray,
+)
+    FT = eltype(prog)
+    primitive_to_prognostic!(
+        Vars{vars_state(bl, Prognostic(), FT)}(prog),
+        Vars{vars_state(bl, Primitive(), FT)}(prim),
+        bl,
+        Vars{vars_state(bl, Auxiliary(), FT)}(aux),
+    )
+end
```
jkozdon, line 31: I like this! Thanks for adding it

src/BalanceLaws/prog_prim_conversion.jl
```diff
@@ -0,0 +1,43 @@
...
+
+# BalanceLaw default:
```
jkozdon, line 33:
```suggestion
# By default the primitive is the prognostic
vars_state(bl::BalanceLaw, ::Primitive, FT) = vars_state(bl, Prognostic(), FT)

```

jkozdon (commented):
Should the file `src/BalanceLaws/prog_prim_conversion.jl` just be merged with `src/BalanceLaws/interface.jl`?

charleskawczynski:
> Should the file `src/BalanceLaws/prog_prim_conversion.jl` just be merged with `src/BalanceLaws/interface.jl`?

Yeah, we can. I imagined that this file might get a bit large if we (eventually) added `Vars` wrappers as we once discussed. So that's why I preemptively split it off.

src/BalanceLaws/prog_prim_conversion.jl
charleskawczynski: I haven't yet added a test that calls the array interface. I'll try to add that when I do last-minute cleaning on this PR.

src/BalanceLaws/prog_prim_conversion.jl
```diff
@@ -0,0 +1,45 @@
...
+    aux::AbstractArray,
+)
+    FT = eltype(prog)
+    primitive_to_prognostic!(
+        Vars{vars_state(bl, Prognostic(), FT)}(prog),
+        Vars{vars_state(bl, Primitive(), FT)}(prim),
+        bl,
+        Vars{vars_state(bl, Auxiliary(), FT)}(aux),
+    )
+end
```
jkozdon, line 31: Usually we put balance law first as an argument

```suggestion
function prognostic_to_primitive!(
    bl::BalanceLaw,
    prim::AbstractArray,
    prog::AbstractArray,
    aux::AbstractArray,
)
    FT = eltype(prim)
    prognostic_to_primitive!(
        bl,
        Vars{vars_state(bl, Primitive(), FT)}(prim),
        Vars{vars_state(bl, Prognostic(), FT)}(prog),
        Vars{vars_state(bl, Auxiliary(), FT)}(aux),
    )
end
function primitive_to_prognostic!(
    bl::BalanceLaw,
    prog::AbstractArray,
    prim::AbstractArray,
    aux::AbstractArray,
)
    FT = eltype(prog)
    primitive_to_prognostic!(
        bl,
        Vars{vars_state(bl, Prognostic(), FT)}(prog),
        Vars{vars_state(bl, Primitive(), FT)}(prim),
        Vars{vars_state(bl, Auxiliary(), FT)}(aux),
    )
end
```

charleskawczynski:
Going to move the atmos prog-prim conversion test to `test/Atmos/prog_prim_conversion.jl`, since this is an atmos-specific test..

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -0,0 +1,172 @@
+####
```
jkozdon, line 1: Should this interface be updated for balance law first argument?

charleskawczynski: Yes, will do

jkozdon (approved):
This looks good to me. Thanks!

PR #1833, opened 2020-12-08
Update line integrals

valeriabarra:
### Description

Opening this PR to get people's feedback for now. 

~~At the moment, the `integral_test.jl` does not work with `p=0` polynomial, for instance with the tuple (4,0).~~ 
Update: Both the `integral_test.jl`  and `integral_test_sphere.jl` tests now pass with different polynomial orders, including 0. 

Edit: ~~When discussing this with @jkozdon we said to apply the same changes as in #1775, i.e., where we renamed the single integer `Nq` as a tuple of integers `Nq1, Nq2, Nq3` in the different directions, with the hope that this would help us sort out the issue for the case with zero degree polynomial. I might have misunderstood the usage of a couple of instances of these terms in the loops though, because they were a bit different from the ones in the Filters I had seen before.~~ This has been resolved now.


src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -2001,28 +2001,31 @@ See [`BalanceLaw`](@ref) for usage.
...
         local_state_prognostic = MArray{Tuple{num_state_prognostic}, FT}(undef)
         local_state_auxiliary = MArray{Tuple{num_state_auxiliary}, FT}(undef)
-        local_kernel = MArray{Tuple{nout, Nq}, FT}(undef)
+        local_kernel = MArray{Tuple{nout, Nq3}, FT}(undef)
     end
 
-    local_integral = @private FT (nout, Nq)
-    s_I = @localmem FT (Nq, Nq)
+    local_integral = @private FT (nout, Nq1)
+    s_I = @localmem FT (Nq1, Nq2)
```
jkozdon, line 2016:
```suggestion
    s_I = @localmem FT (Nq3, Nq3)
```

src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -2001,28 +2001,31 @@ See [`BalanceLaw`](@ref) for usage.
...
 
     _eh = @index(Group, Linear)
     i, j = @index(Local, NTuple)
 
     @inbounds begin
-        @unroll for n in 1:Nq
+        @unroll for n in 1:Nq3
             s_I[i, n] = Imat[i, n]
         end
```
jkozdon, line 2024:
```suggestion
@unroll for iv in i:Nq1:Nq3
        @unroll for jv in j:Nq2:Nq3
            s_I[iv, jv] = Imat[iv, jv]
        end
end
```

src/Numerics/DGMethods/DGModel_kernels.jl
```diff
@@ -2070,19 +2073,19 @@ See [`BalanceLaw`](@ref) for usage.
 
             # Evaluate the integral up the element
             @unroll for s in 1:nout
-                @unroll for k in 1:Nq
-                    @unroll for n in 1:Nq
+                @unroll for k in 1:Nq3
+                    @unroll for n in 1:Nq1
```
jkozdon, line 2077:
```suggestion
                @unroll for k in 1:Nq3
                    @unroll for n in 1:Nq3
```

valeriabarra:
@jkozdon just to give you an update, I think I correctly rebased on your branch and applied the suggestions. I do not see the loop @unroll issue, but the test is still not giving the desired result.

jkozdon:
> @jkozdon just to give you an update, I think I correctly rebased on your branch and applied the suggestions. I do not see the loop @unroll issue, but the test is still not giving the desired result.

Great.

So I think there are probably three issues in the test right now

1. The test function be integrated cannot be exactly integrated in the vertical
2. The values we get back from the integration are faces values, whereas the exact solution is being computed at the cell center.
3. The reverse integral (which compute the top value minus each nodal value should probably all be shifted by 1)

Maybe let's chat about this?

valeriabarra:
> 
> Maybe let's chat about this?

Sure thing!

valeriabarra:
Update: 

@jkozdon helped me tremendously in this PR. 

At the moment, this PR contains a couple of bug fixes in the `kernel_indefinite_stack_integral!` and `kernel_reverse_indefinite_stack_integral!` that were not exposed earlier because not exercised with different polynomial orders in the horizontal/vertical directions and in 2D/3D. 

For the N=0 case (FVM) the `integral_test.jl` at the moment still needs some thoughts and fixing. We can either merge this now and fix the FVM case in a subsequent PR or try to tackle that case in here.

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

test/Numerics/DGMethods/integral_test_sphere.jl
```diff
@@ -193,7 +193,7 @@ function test_run(mpicomm, topl, ArrayType, N, FT, Rinner, Router)
         @test all(Router - Rinner .≈ aux[:, :, end, int_r_ind, end, :])
         # FIXME: Reverse integral is currently off by one cell width
         Δ = (Router - Rinner) / nvertelem
-        @test all(Router - Rinner - Δ .≈ aux[:, :, 1, rev_int_r_ind, 1, :])
+        @test all(Δ .≈ aux[:, :, 1, rev_int_r_ind, 1, :])
```
valeriabarra, line 196: @jkozdon I changed the `Router - Rinner - Δ` to `Δ` only, which turns out to be the same thing. 

This reverse integral is still off by one cell width. A hack would be 
```suggestion
        @test all(Δ .≈ aux[:, :, 1, rev_int_r_ind, 1, :]/2)
```
:joy: 
but I was trying to understand why the value of this integral was off by one cell width in the first place

jkozdon:
```suggestion
         @test all(Router - Rinner .≈ aux[:, :, 1, rev_int_r_ind, 1, :]/2)
```
 Didn't work?

jkozdon:
```julia
        #TODO: Update to test all faces (similar to `integral_test.jl`
        # Only the last value of the forward integral is currently correct
        @test all(Router - Rinner .≈ aux[:, :, end, int_r_ind, end, :])
        # Currently only the first value of the reverse integral is correct
        @test all(Router - Rinner .≈ aux[:, :, 1, rev_int_r_ind, 1, :])
        # All the `JcV` (line integral metrics) values should be `Δ / 2`
        Δ = (Router - Rinner) / nvertelem
        @test all(Δ .≈ 2grid.vgeo[:, Grids._JcV, :])
```

valeriabarra:
bors try

valeriabarra:
The windstress test fails in this PR, but I rebased onto master :/

test/Numerics/DGMethods/integral_test_sphere.jl
```diff
@@ -193,7 +193,7 @@ function test_run(mpicomm, topl, ArrayType, N, FT, Rinner, Router)
         @test all(Router - Rinner .≈ aux[:, :, end, int_r_ind, end, :])
         # FIXME: Reverse integral is currently off by one cell width
```
jkozdon, line 194:
```suggestion
```

jkozdon (commented):
We could add a refinement test on the sphere like we do for the `N > 0` case. I would do this like we did for the`integral_test` averaging to the cell centers of the computed values to compare with the numerical values.

valeriabarra:
@jkozdon I tried to implement the refinement for the case `N=0` in the `integral_test_sphere.jl`, based on what was done in the `integral_test.jl` although the data layouts were different in the two tests. I need to figure out why we are not converging now. It might be just a matter of indices that I might have mingled in the attempt of translating the reshaped variables. Any help would be much appreciated. Thanks.

valeriabarra:
Thank you so much for your latest changes in the grid @jkozdon ! Now it passes also the sphere test for all polynomial orders! 
This looks good to me.

valeriabarra:
bors try

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -655,7 +654,7 @@ function reverse_indefinite_stack_integral!(
     dim = dimensionality(grid)
     N = polynomialorders(grid)
     Nq = N .+ 1
-    Nqk = dim == 2 ? 1 : Nq[dim]
+    Nqk = dim == 2 ? 1 : Nq[2]
```
mwarusz, line 657: Could we change this name to `Nqj` or `Nq2` ? `Nqk` is defined differently in other places.

valeriabarra: Done in [4ad0db5](https://github.com/CliMA/ClimateMachine.jl/pull/1833/commits/4ad0db58b11701d82f4ed967f2dbda429ab4c9c6)

valeriabarra:
bors try

Zhengyu-Huang (approved):
Thank you for the great effort, I think you have already fully tested and verified all changes!

valeriabarra:
Thanks for the approval @Zhengyu-Huang .
If this is ok, I will reduce the number of commits (but keep authorships) and bors r+

valeriabarra:
bors try

PR #1834, opened 2020-12-08
Move second order moisture flux to new tend spec

charleskawczynski:
### Description

This PR
 - moves second order moisture fluxes (`MoistureDiffusion`) to the new tendency specification.
 - Changes the argument order in `flux_second_order!(::MoistureModel, args...)` so that arguments can be easily passed through

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -498,13 +498,13 @@ equations.
     t::Real,
     direction,
 )
-    ρu_pad = SVector(1, 1, 1)
+    flux_pad = SVector(1, 1, 1)
     ts = recover_thermo_state(m, state, aux)
     tend = Flux{FirstOrder}()
     args = (m, state, aux, t, ts, direction)
-    flux.ρ = Σfluxes(eq_tends(Mass(), m, tend), args...)
-    flux.ρu = Σfluxes(eq_tends(Momentum(), m, tend), args...) .* ρu_pad
-    flux.ρe = Σfluxes(eq_tends(Energy(), m, tend), args...)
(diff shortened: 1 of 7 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -615,17 +615,18 @@ function. Contributions from subcomponents are then assembled (pointwise).
     aux::Vars,
     t::Real,
 )
+    ρ_pad = SVector(1, 1, 1)
     ρu_pad = SVector(1, 1, 1)
     ts = recover_thermo_state(atmos, state, aux)
     tend = Flux{SecondOrder}()
     args = (atmos, state, aux, t, ts, diffusive, hyperdiffusive)
-    flux.ρ = Σfluxes(eq_tends(Mass(), atmos, tend), args...) .* ρu_pad
+    flux.ρ = Σfluxes(eq_tends(Mass(), atmos, tend), args...) .* ρ_pad
```
yairchn, line 623: I am not sure that I follow, but was the change from 'ρu_pad' to 'ρ_pad' accompanied by a change in the definition of 'Σfluxes' ?

charleskawczynski: I just changed this because the previous name was bad. Actually, ideally we shouldn't have this pad at all for `ρ_pad`. Going to see if we can remove it.

charleskawczynski: Going to try just removing this for now.

charleskawczynski: It turns out that we can't remove the padding because, for example, the MMS model doesn't have any second-order fluxes. So we need to ensure that the LHS and RHS ranks match.

trontrytel (approved):

PR #1727, opened 2020-11-09
specify boundary conditions as a tuple

simonbyrne:
### Description

The main changes are:
1. A `BalanceLaw` defines an additional function `boundary_conditions(::BL)` which should return a tuple of boundary conditions (currently these can be anything, but we may change that in future).
2. Instead of passing the integer `bctype` to the boundary condition functions, we now look up the corresponding element of the tuple and pass that (this matches how Atmos and Ocean were doing it internally, so we can remove that code).
3. I've moved the boundary condition argument to be the 2nd one, to make it easier to do dispatch.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #1838, opened 2020-12-08
added sketch of 3d config and use

kmdeck:
### Description

We've added in a MultiColumnLandModel configuration using the StackedBrickTopology. This takes user supplied boundary conditions at the top and bottom of the domain, but applies zero flux at the horizontal boundaries.

If we want to make the horizontal boundary condition something else, I'm wondering how to make this more general without having a bunch of if/else statements on bc_type...and want to make sure this current iteration doesn't box us out of things later/require refactoring later.

Also wondering if using if/else statements on bc_type is still the favored approach.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -516,6 +516,15 @@ steps:
       queue: central
       slurm_ntasks: 1
 
+  - label: "cpu_soil_test_bc_3d"
+    key: "cpu_soil_test_bc_3d"
+    command:
+      - "mpiexec julia --color=yes --project test/Land/Model/test_bc_3d.jl "
+    agents:
+      config: cpu
+      queue: central
+      slurm_ntasks: 1
+
   - label: "gpu_varsindex"
     key: "gpu_varsindex"
     command:
@@ -1330,6 +1339,16 @@ steps:
     key: "gpu_soil_test_bc"
     command:
       - "mpiexec julia --color=yes --project test/Land/Model/test_bc.jl "
+    agents:
+      config: gpu
+      queue: central
+      slurm_ntasks: 1
+      slurm_gres: "gpu:1"
+
+  - label: "gpu_soil_test_bc_3d"
+    key: "gpu_soil_test_bc_3d"
+    command:
+      - "mpiexec julia --color=yes --project test/Land/Model/test_bc_3d.jl "
     agents:
       config: gpu
       queue: central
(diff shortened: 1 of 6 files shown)


kmdeck:
@simonbyrne, @blallen pointed out your recent PR https://github.com/CliMA/ClimateMachine.jl/pull/1727. I saw you added in this line:
https://github.com/CliMA/ClimateMachine.jl/blob/5aa7a3198763ae864a835a0995fafe9320421af4/src/Land/Model/land_bc.jl#L3
I just added in a third integer here, to reflect lateral boundaries.

Wondering your thoughts..

to make Land look more like Atmos and Ocean, we'd need to, among other things, specify the BC as a tuple  (bottom, top, sides) for the bc_type  integers (1,2,3). We already have the objects Dirchlet() and Neumann(), where the user can specify functions to use for those - but they would need to be modified
https://github.com/CliMA/ClimateMachine.jl/blob/585a602516343cdcc9321db1c50c6c6cb9ef64c3/src/Land/Model/soil_boundary_types.jl#L15
to only hold one condition (right now, if the user only wants Dirichlet on bottom, they pass `nothing` in the entry where they dont want to apply - e.g. `Dirichlet(nothing, bottom_state)`, where `bottom_state` is a function.

It doesnt seem too hard, just a lot of change. We'd also need to specify the tuple in a hierarchical way, since the user can choose to run e.g. without a water model, and only with a heat model, and in that case we dont need water BC. The `boundary_state` functions in this case dispatch on the water model type...

This approach does seem more flexible, and a little nicer in that when specifying e.g. Dirichlet BC as in the above example, they just give the condition they want, and dont need to also say "and no dirichlet BC at the top".

Anyways, happy to make these changes, but I might need some guidance.

simonbyrne:
@kmdeck Yes: it was my intention to do the same thing for Land as we did for Atmos and Ocean, but I didn't have time to figure out exactly how to make it work. Defining it as `(1,2)` was just a cheap trick to make the existing behaviour work.

I'll have a stab at updating `Land` along these lines: would it be better to work off master or this PR?

kmdeck:
> @kmdeck Yes: it was my intention to do the same thing for Land as we did for Atmos and Ocean, but I didn't have time to figure out exactly how to make it work. Defining it as `(1,2)` was just a cheap trick to make the existing behaviour work.
> 
> I'll have a stab at updating `Land` along these lines: would it be better to work off master or this PR?

It might be simpler to go from master - then the only surfaces we apply BC to are the top and bottom (2 and 1, I think). I can add in 3D (another bctype integer) after that. 

I think I could do a lot of the refactoring, with input from you. Maybe I can think about what needs to be done, and then we can meet to discuss before I do it? We could also pair program. Whatever is easiest!

kmdeck:
superseded by https://github.com/CliMA/ClimateMachine.jl/pull/1873, closing.

PR #1839, opened 2020-12-08
GCM Power Spectrum Modification

LenkaNovak:
### Description

Modifies the GCM power spectrum diagnostics to be computed using the full KE, making them directly comparable to the spectra used in the Jablonowski paper on the baroclinic wave test, which CI uses as a benchmark test. 


.gitignore
```diff
@@ -35,3 +35,6 @@ src/**/Manifest.toml
 
 # Julia System Images
 *.so
+
+*.sh
+*.out
```
kpamnany, line 40:
```suggestion
```
I think `.gitignore` should contain only generated files.

kpamnany (approved):
Changes look fine to me apart from my single comment. I think the GCMDriver uses this diagnostics group, I assume you've verified the output. :)

.gitignore
LenkaNovak: Sounds good! I guess it makes life easier for people who use bash scripts to run the CM, but it's not a biggie either way

PR #1841, opened 2020-12-08
Move second order tracer flux to new tend spec

charleskawczynski:
### Description

This PR
 - Moves the second order tracer fluxes to the new tendency specification
 - Adds `tendencies_tracer.jl` to `src/Atmos/Model/`
 - Moves the first order flux definition to `tendencies_tracer.jl` (I think I missed following this pattern in the first order tracer flux PR)

I'm not 100% sure about the name `Diffusion` here. cc @akshaysridhar?

### DIFF
--- a/src/Atmos/Model/tendencies_tracers.jl
+++ b/src/Atmos/Model/tendencies_tracers.jl
@@ -1,0 +1,29 @@
+##### Moisture tendencies
+
+#####
+##### First order fluxes
+#####
+
+function flux(::Advect{Tracers{N}}, m, state, aux, t, ts, direction) where {N}
+    u = state.ρu / state.ρ
+    return (state.tracers.ρχ .* u')'
+end
+
+#####
+##### Second order fluxes
+#####
+
+function flux(
+    ::Diffusion{Tracers{N}},
+    m,
+    state,
+    aux,
+    t,
+    ts,
+    diffusive,
+    hyperdiff,
+) where {N}
+    ν, D_t, τ = turbulence_tensors(m, state, diffusive, aux, t)
+    d_χ = (-D_t) * aux.tracers.δ_χ' .* diffusive.tracers.∇χ
+    return d_χ * state.ρ
+end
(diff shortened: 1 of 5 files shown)


src/Atmos/Model/tendencies_tracers.jl
```diff
@@ -0,0 +1,29 @@
+##### Moisture tendencies
+
+#####
+##### First order fluxes
+#####
+
+function flux(::Advect{Tracers{N}}, m, state, aux, t, ts, direction) where {N}
+    u = state.ρu / state.ρ
+    return (state.tracers.ρχ .* u')'
```
trontrytel, line 9: Can you teach me what are the `'` doing here?

trontrytel (commented):
LGTM. My only question is do we have any simulation that actually uses tracers? And if yes, is it used as a test in CI or as a QA simulation?

src/Atmos/Model/tendencies_tracers.jl
charleskawczynski: `'` is the adjoint (IIUC, here, this is effectively a transpose). Size-wise, I think we have
```
(state.tracers.ρχ .* u')' = ((N x 1) .* (3 x 1)')' = ((N x 1) .* (1 x 3))' = (N x 3)' = (3 x N) # (3 for 3 component flux, N for N tracers)
```

charleskawczynski:
> My only question is do we have any simulation that actually uses tracers? And if yes, is it used as a test in CI or as a QA simulation?

Yes, we do. `schar_scalar_advection.jl` use the tracer model (in CI). Also, several of the DGMethod tests:
```
test/Numerics/DGMethods/Euler/acousticwave_1d_imex.jl
test/Numerics/DGMethods/Euler/acousticwave_mrigark.jl
test/Numerics/DGMethods/Euler/acousticwave_variable_degree.jl
```
And more :)

trontrytel (approved):

Issue #1641, opened 2020-10-19
Command line options


valeriabarra:
### Description

As I am trying to get familiar with the code by playing around with examples/experiments, I was wondering how to quickly  change element number (i.e., resolution) and/or polynomial order. I think these two are the most important CL options to be able to run systematic convergence studies (via h-refinement, p-refinement, or hp-refinement), and the FEM or DG discretizations are particularly favorable to those, so it is common in many Finite Elements codes to have these command line options.

I read the discussion in #947 and I think one concern was to have too many possible CL options. I think when you group them in a config/input/namelist file, their number shouldn't be worrysome and, as pointed out, it is easier to keep the input file with the output files for easy bookkeeping of your numerical experiements (maybe I'm biased by my experience as a PETSc user, but it sounds like in scientific computing packages there is never too many CL options XD).

If we are not ready to have a full Namelist yet at the top level user interface, I think at least having these two CL options would help also in the investigations that were mentioned lately that are necessary for refactoring the code to allow different order in the horizontal/vertical, and, as over integartion is added, also being able to set the number of extra quadrature points at CL becomes important. 

Other than a simpler user experience there are also additional benefits: in fact, having these CL options allows user to run jobs in batch either locally or send them in batch when on the cluster, to loop over polynomial order for strong/weak scaling studies, and to not have Git complain about your changes when you are just trying to run several different numerical experiments and observe their different outcomes.

valeriabarra:
Closed via PR #1845 and #1946

PR #1845, opened 2020-12-09
Add polynomial degrees CL option

valeriabarra:
### Description

A first attempt at adding the polynomial degrees as a CL option. 
This partially addresses #1641. 

Still to do:
- [x] Need to figure out how to pass this option to the `DriverConfiguration` 
- [x] Test

cc: @kpamnany 

### DIFF
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -296,7 +296,7 @@ function parse_commandline(
         default = get_setting(:fixed_number_of_steps, defaults, global_defaults)
         "--degree"
         help = "tuple of horizontal and vertical polynomial degrees for spatial discretization order (no space before/after comma)"
-        metavar = "horizontal,vertical"
+        metavar = "<horizontal>,<vertical>"
         arg_type = NTuple{2, Int}
         default = get_setting(:degree, defaults, global_defaults)
     end
--- a/src/Driver/Driver.jl
+++ b/src/Driver/Driver.jl
@@ -75,10 +75,10 @@ Base.@kwdef mutable struct ClimateMachine_Settings
     array_type::Type = Array
     sim_time::Float64 = NaN
     fixed_number_of_steps::Int = -1
+    degree::NTuple{2, Int} = (-1, -1)
 end
 
 const Settings = ClimateMachine_Settings()
-
 
 """
     ClimateMachine.array_type()
@@ -89,7 +89,6 @@ line, from an environment variable, or from experiment code) after
 `ClimateMachine.init()` is called.
 """
 array_type() = Settings.array_type
-
 
 """
     get_setting(setting_name::Symbol, settings, defaults)
@@ -110,7 +109,13 @@ function get_setting(setting_name::Symbol, settings, defaults)
         return convert(setting_type, settings[setting_name])
     elseif haskey(ENV, setting_env)
         env_val = ENV[setting_env]
-        v = setting_type == String ? env_val : tryparse(setting_type, env_val)
+        if setting_type == String
+            v = env_val
+        elseif setting_type == NTuple{2, Int}
+            v = ArgParse.parse_item(setting_type, env_val)
+        else
+            v = tryparse(setting_type, env_val)
+        end
         if isnothing(v)
             error("cannot parse ENV $setting_env value $env_val, to setting type $setting_type")
         end
@@ -133,7 +138,6 @@ function get_gpu_setting(setting_name::Symbol, settings, defaults)
     # fallback behavior
     return get_setting(setting_name, settings, defaults)
 end
-
 
 """
     parse_commandline(
@@ -290,6 +294,11 @@ function parse_commandline(
         metavar = "<number>"
         arg_type = Int
         default = get_setting(:fixed_number_of_steps, defaults, global_defaults)
+        "--degree"
+        help = "tuple of horizontal and vertical polynomial degrees for spatial discretization order (no space before/after comma)"
+        metavar = "horizontal,vertical"
+        arg_type = NTuple{2, Int}
+        default = get_setting(:degree, defaults, global_defaults)
     end
     # add custom cli argparse settings if provided
     if !isnothing(custom_clargs)
@@ -297,7 +306,6 @@ function parse_commandline(
     end
     return parse_args(s)
 end
-
 
 """
     ClimateMachine.init(;
@@ -366,6 +374,8 @@ Recognized keyword arguments are:
         run for the specified time (in simulation seconds)
 - `fixed_number_of_steps::Int = -1`:
         if `≥0` perform specified number of steps
+- `degree::NTuple{2, Int} = (-1, -1)`:
+        tuple of horizontal and vertical polynomial degrees for spatial discretization order
 
 Returns `nothing`, or if `parse_clargs = true`, returns parsed command line
 arguments.
@@ -460,7 +470,6 @@ function init(;
     return cl_args
 end
 
-
 """
     init_runtime(settings::ClimateMachine_Settings)
 
@@ -521,11 +530,9 @@ function init_runtime(settings::ClimateMachine_Settings)
     return
 end
 
-
 include("driver_configs.jl")
 include("solver_configs.jl")
 include("diagnostics_configs.jl")
-
 
 """
     ClimateMachine.ConservationCheck
--- a/src/Driver/driver_configs.jl
+++ b/src/Driver/driver_configs.jl
@@ -25,6 +25,40 @@ struct SingleStackSpecificInfo <: ConfigSpecificInfo end
 include("SolverTypes/SolverTypes.jl")
 
 """
+    ArgParse.parse_item
+
+Parses custom command line option for tuples of two integers.
+"""
+function ArgParse.parse_item(::Type{NTuple{2, Int}}, s::AbstractString)
+
+    str_array = split(s, ",")
+    horizontal = parse(Int, str_array[1])
+    vertical = parse(Int, str_array[2])
+
+    return (horizontal, vertical)
+end
+
+"""
+    get_polyorders
+
+Utility functions that gets the polynomial orders for the given configuration
+either passed from command line or as default values
+"""
+function get_polyorders(N)
+
+    (polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
+
+    # Check if polynomial degree was passed as a CL option
+    if ClimateMachine.Settings.degree != (-1, -1)
+        ClimateMachine.Settings.degree
+    elseif N isa Int
+        (N, N)
+    else
+        N
+    end
+end
+
+"""
     ClimateMachine.DriverConfiguration
 
 Collects all parameters necessary to set up a ClimateMachine simulation.
@@ -135,7 +169,7 @@ function AtmosLESConfiguration(
     numerical_flux_gradient = CentralNumericalFluxGradient(),
 ) where {FT <: AbstractFloat}
 
-    (polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
+    (polyorder_horz, polyorder_vert) = get_polyorders(N)
 
     print_model_info(model)
 
@@ -162,14 +196,14 @@ function AtmosLESConfiguration(
     @info @sprintf(
         """
 Establishing Atmos LES configuration for %s
-    precision              = %s
-    horiz polynomial order = %d
-    vert polynomial order  = %d
-    domain                 = %.2f m x%.2f m x%.2f m
-    resolution             = %dx%dx%d
-    MPI ranks              = %d
-    min(Δ_horz)            = %.2f m
-    min(Δ_vert)            = %.2f m""",
+    precision               = %s
+    horiz polynomial order  = %d
+    vert polynomial order   = %d
+    domain                  = %.2f m x%.2f m x%.2f m
+    resolution              = %dx%dx%d
+    MPI ranks               = %d
+    min(Δ_horz)             = %.2f m
+    min(Δ_vert)             = %.2f m""",
         name,
         FT,
         polyorder_horz,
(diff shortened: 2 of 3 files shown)


kpamnany (commented):
You'll have to overload `Base.parse` or [`ArgParse.parse_item`](https://carlobaldassi.github.io/ArgParse.jl/stable/custom/#Parsing-to-custom-types-1) to get a tuple from the command line. It'd be simpler to add two command line options instead -- `hdegree` and `vdegree` maybe.

Then, you could add some code [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Driver.jl#L442) to convert the arguments into a tuple to put into `Settings`, but that might get tricky. Or you could have `hdegree` and `vdegree` (or whatever you name them) as separate `Int`s in `Settings` as well.

To actually use these though will require changes [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/driver_configs.jl#L144) and in the other configuration functions in that file. Presumably, we want these command line options, if provided, to override the experiment-specified polynomial orders. To do that, the defaults for the command line options should be `-1`, so that you can easily determine whether they were provided or not. Then you could change this:
```julia
    (polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
```
To something like:
```julia
    (polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
    if ClimateMachine.Settings.hdegree != -1
        polyorder_horz = ClimateMachine.Settings.hdegree
    end
    if ClimateMachine.Settings.vdegree != -1
        polyorder_vert = ClimateMachine.Settings.vdegree
    end
```
Or pull this code into a little function which you can call from each of the configuration functions.

Ping me to talk about this if you like.

src/Driver/solver_configs.jl
```diff
@@ -79,6 +79,8 @@ the ODE solver, and return a `SolverConfiguration` to be used with
 # - `direction=EveryDirection()`: restrict diffusivity direction.
 # - `timeend_dt_adjust=true`: should `dt` be adjusted to hit `timeend` exactly
 # - `CFL_direction=EveryDirection()`: direction for `calculate_dt`
+# - `sim_time`: run for the specified time (in simulation seconds).
+# - `fixed_number_of_steps`: if `≥0` perform specified number of steps.
```
kpamnany, line 83: Good catch, thanks!

kpamnany:
Oh, please also update [the documentation](https://github.com/CliMA/ClimateMachine.jl/blob/master/docs/src/GettingStarted/RunningClimateMachine.md#command-line-argumentsid-climatemachine-clargs) on the command line arguments with the PR.

valeriabarra:
bors try

valeriabarra:
@kpamnany I added the custom parser for the CL option for the tuple of integers. I tested this with the following example
```
julia --project experiments/AtmosLES/bomex_les.jl --degree 4,2
```
and it works just fine! 

I was also trying to automatically generate the environment variable for this CL option. However, I tried to override the `Base.tryparse` function so that it could work with a Tuple of integers, but I encountered an issue with type piracy. I guess one option could be defining a custom type wrapper for this function to overcome the type piracy warning, but I am not this proficient in Julia yet. 

Thanks for any help you might give!

valeriabarra:
bors try

src/Driver/driver_configs.jl
```diff
@@ -17,13 +17,27 @@ struct AtmosLESSpecificInfo <: ConfigSpecificInfo end
 struct AtmosGCMSpecificInfo{FT} <: ConfigSpecificInfo
     domain_height::FT
     nelem_vert::Int
-    nelem_horz::Int
+    nelem_horiz::Int
```
charleskawczynski, line 20: ❤️. The previous name bothered me quite a bit

valeriabarra: haha I am glad I was not the only one ;)

valeriabarra:
I'm going to fix the docs build in the morning.

src/Driver/Driver.jl
```diff
@@ -366,6 +368,8 @@ Recognized keyword arguments are:
         run for the specified time (in simulation seconds)
 - `fixed_number_of_steps::Int = -1`:
         if `≥0` perform specified number of steps
+- `degree::NTuple{2, Int} = -1,-1`:
```
kpamnany, line 371:
```suggestion
- `degree::NTuple{2, Int} = (-1, -1)`:
```

src/Driver/Driver.jl
```diff
@@ -366,6 +368,8 @@ Recognized keyword arguments are:
...
+        tuple of horizontal and vertical polynomial degrees for spatial discretization order (no space before/after comma)
```
kpamnany, line 372:
```suggestion
        tuple of horizontal and vertical polynomial degrees for spatial discretization order
```
As this is Julia code, a tuple and its syntax are well-defined.

src/Driver/driver_configs.jl
```diff
@@ -135,13 +149,17 @@ function AtmosLESConfiguration(
     numerical_flux_gradient = CentralNumericalFluxGradient(),
 ) where {FT <: AbstractFloat}
 
-    (polyorder_horz, polyorder_vert) = isa(N, Int) ? (N, N) : N
+    (polyorder_horiz, polyorder_vert) = isa(N, Int) ? (N, N) : N
+    # Check if polynomial degree was passed as a CL option
+    if ClimateMachine.Settings.degree != (-1, -1)
+        (polyorder_horiz, polyorder_vert) = ClimateMachine.Settings.degree
+    end
```
kpamnany, line 156:
```suggestion
    (polyorder_horiz, polyorder_vert) = get_polyorders(N)
```
Where:
```julia
function get_polyorders(N)
    if ClimateMachine.Settings.degree != (-1, -1)
        ClimateMachine.Settings.degree
    elseif N isa Int
        (N, N)
    else
        N
    end
end
```
And avoid the repetitions below. But it's not a biggie.

kpamnany (commented):
Nice!

I'd like to see the environment variable also working before we merge this. Let's talk about the type piracy error you're seeing -- if we can fix that, you can get rid of the `ArgParse.parse_item` definition (as ArgParse simply uses `Base.parse`).

But it might also make sense to avoid adding a method for `Base.parse` and `Base.tryparse` because [invalidations](https://julialang.org/blog/2020/08/invalidations/). And we can avoid that by changing [this code](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Driver.jl#L113-L116) to call a parsing function we define.

Any comments @jakebolewski?

kpamnany:
> I'm going to fix the docs build in the morning.

It's coming from [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/diagnostics_configs.jl#L88) or [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/diagnostics_configs.jl#L133).

You'll find quite a number of `horz`s in the code.
```bash
~/ClimateMachine.jl$ grep -R -L horz * | wc -l
519
```
Not sure it's worth the effort to change them all, and inconsistent naming is not a good thing either.

jkozdon:
> You'll find quite a number of `horz`s in the code.

Am I alone in finding `vert` and `horz` appealing that they have the same number of characters?

I find something satisfying about the alignment of variables names:
This
```julia
    nelem_vert
    nelem_horz
```
as opposed to
```julia
    nelem_vert
    nelem_horiz
```

That said, its total bikeshedding :bike::house: so I don't really care that much.

jakebolewski:
For the CLI options it seems simpler to just have two flags (one for vert and horiz), that doesn't require any special additional parsing and not subject to julia's syntax (behaves just like any other cli flag)

valeriabarra:
> For the CLI options it seems simpler to just have two flags (one for vert and horiz), that doesn't require any special additional parsing and not subject to julia's syntax (behaves just like any other cli flag)

Thanks @jakebolewski this was deliberately not chosen. I personally didn't like this route because it would have "authorized" the user to see the two options as separate conceptually, and potentially just specified one of the two. Whereas we see the tuple conceptually as it is, a tuple, and we would like the user to think of these degrees as to be expressed together

src/Driver/driver_configs.jl
```diff
@@ -17,13 +17,27 @@ struct AtmosLESSpecificInfo <: ConfigSpecificInfo end
...
 include("SolverTypes/SolverTypes.jl")
 
+"""
+    ArgParse.parse_item
+
+Parses custom command line option for tuples of two integers.
+"""
+function ArgParse.parse_item(::Type{NTuple{2, Int}}, s::AbstractString)
+
+    str_array = split(s, ",")
```
jakebolewski, line 34: you can lstrip and rstrip the resulting split string to get rid of the whitespace issue

kpamnany: I think a space after the comma will make `ArgParse` treat the second part of the tuple as a separate argument?

jakebolewski: I see, this is why having non-magical cli argument syntax is nice!

kpamnany: I don't disagree. 😜 

Might not be the best supporting argument in favor, but gcc has command line arguments like this, in which commas cannot be followed by spaces (e.g. `-Wa,`).

valeriabarra:
> > You'll find quite a number of `horz`s in the code.
> 
> Am I alone in finding `vert` and `horz` appealing that they have the same number of characters?
> 
> I find something satisfying about the alignment of variables names:
> This
> 
> ```julia
>     nelem_vert
>     nelem_horz
> ```
> 
> as opposed to
> 
> ```julia
>     nelem_vert
>     nelem_horiz
> ```
> 
> That said, its total bikeshedding  so I don't really care that much.

I understand that visually the same number of characters look nice, but what was bothering me was that the string printed in the `Info` instead used `nelem_horiz`. So I wanted to make that consistent. If this is not desired, of course I can revert my change. 
There should be a consensus before I change all the 519 instances :sweat_smile: 

@charleskawczynski  and I seem to prefer `nelem_horiz`. 
@jkozdon @mwarusz @simonbyrne  seem to prefer the other way.
Any other suggestions?

kpamnany:
IMO, making so many changes for something cosmetic that isn't even uniformly agreed upon is not a good idea.

valeriabarra:
I will put the `nelem_horz` back. Not a problem!

src/Driver/Driver.jl
valeriabarra: Sorry! From our last conversation I understood that we wanted to make sure to document the no comma requirement (especially for `bash`), but yeah here it was not needed. I went with auto-pilot mode, adding it to too many places!

src/Driver/driver_configs.jl
valeriabarra: Thanks for the suggestion! I put it in [5194752](https://github.com/CliMA/ClimateMachine.jl/pull/1845/commits/519475263f3b5476e35fe225bbcd1c9aa853fbf0)

src/Driver/Driver.jl
valeriabarra: Thanks! Also done in [5194752](https://github.com/CliMA/ClimateMachine.jl/pull/1845/commits/519475263f3b5476e35fe225bbcd1c9aa853fbf0)

valeriabarra:
bors try

valeriabarra:
> Nice!
> 
> I'd like to see the environment variable also working before we merge this. Let's talk about the type piracy error you're seeing -- if we can fix that, you can get rid of the `ArgParse.parse_item` definition (as ArgParse simply uses `Base.parse`).
> 
> But it might also make sense to avoid adding a method for `Base.parse` and `Base.tryparse` because [invalidations](https://julialang.org/blog/2020/08/invalidations/). And we can avoid that by changing [this code](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Driver/Driver.jl#L113-L116) to call a parsing function we define.
> 
I think I got it in [3f721ae](https://github.com/CliMA/ClimateMachine.jl/pull/1845/commits/3f721aea24a7241208cc99d25dc9c27102a1207c)! I tested it with

```
$  CLIMATEMACHINE_SETTINGS_DEGREE=4,3  julia --project experiments/AtmosLES/bomex_les.jl
```
And it worked just fine! 

Of course, if you have further suggestions for improvement , please let me know :)

valeriabarra:
Not sure I understand the CI failure in the Land Model test, which is unrelated to this PR. @jakebolewski any clue?

jakebolewski:
There is a race condition in the way we use Artifacts with the ArtifactsHelper in the experiments.   I would just try again as it's fairly rare to hit (two jobs that download the same artifact have to basically be installing it at a similar time)

valeriabarra:
bors try

src/Driver/Driver.jl
```diff
@@ -290,6 +294,11 @@ function parse_commandline(
         metavar = "<number>"
         arg_type = Int
         default = get_setting(:fixed_number_of_steps, defaults, global_defaults)
+        "--degree"
+        help = "tuple of horizontal and vertical polynomial degrees for spatial discretization order (no space before/after comma)"
+        metavar = "horizontal,vertical"
```
kpamnany, line 299:
```suggestion
        metavar = "<horizontal>,<vertical>"
```
For consistency with our other arguments' metavars.

kpamnany (approved):
For completeness, please also test that `ClimateMachine.init(degree = (h, v))` works as expected (I see no reason it shouldn't).

kpamnany:
> ## try
> Build failed:
> 
> * [buildkite/climatemachine-ci](https://buildkite.com/clima/climatemachine-ci/builds/926)

🤔 I thought you'd fixed this @charleskawczynski, but I see that the fix was for squall line. Looks like that fix (or an enhanced one, which adds code to ensure that this is done on MPI rank 0 only, with an MPI barrier after) needs to be applied in other places as well...

src/Driver/Driver.jl
valeriabarra: Thanks for catching it. Updated it

valeriabarra:
> ClimateMachine.init(degree = (h, v))

Yup. Tested it in the same bomex experiment and it works just fine.

valeriabarra:
bors try

valeriabarra:
Squashed and I can bors r+ once CI finishes (just to make sure)

Issue #1832, opened 2020-12-07
Patterns for dispatch on subcomponents


simonbyrne:
We often use a pattern such as:
```julia
f(atmos::AtmosModel, args...) = f(atmos.component, atmos, args...)
f(component::ComponentTypeA, atmos, args...) = # do something
f(component::ComponentTypeB, atmos, args...) = # do something else
```
The original reason for this is that it avoided the explosion of type parameters when trying to do it directly:
```julia
f(atmos::AtmosModel{C1, C2, ..., ComponentTypeA}, args...) where {C1, C2, ...} = # do something
f(atmos::AtmosModel{C1, C2, ..., ComponentTypeB}, args...) where {C1, C2, ...} = # do something else
```
(and remembering which type parameter goes where).

One problem @kpamnany pointed out was that this only lets you dispatch on a single item.

An alternative pattern would be to define the signature via a function
```julia
f(atmos::p(AtmosModel, C=ComponentTypeA), args...) = # do something
f(atmos::p(AtmosModel, C=ComponentTypeB), args...) = # do something else
```
where `p(AtmosModel, C=XXX)` returns an appropriately parametrized type, e.g. something like
```julia
function p(T::DataType; params...)
    if !isempty(params)
        error("Invalid parameters $params")
    end
    return T
end

function p(T::UnionAll; params...)
    if haskey(params, T.var.name)
        rparams = Base.structdiff(values(params), NamedTuple{(T.var.name,)})
        return p(T{params[T.var.name]}; rparams...)
    else
        return (p(T{P}; params...) where {P})
    end
end
```
which gives
```julia
julia> struct Foo{A,B}
           a::A
           b::B
       end

julia> p(Foo, A=Float32)
Foo{Float32,P} where P

julia> p(Foo, B=Float32)
Foo{P,Float32} where P

julia> p(Foo)
Foo{P,P1} where P1 where P

julia> p(Foo, B=Float32,A=Int)
Foo{Int64,Float32}
```

An alternative would be to match via subtypes instead of name, e.g.
```julia
function psub(T::DataType, ts...)
    if !isempty(ts)
        error("Invalid parameters $params")
    end
    return T
end

function psub(T::UnionAll, ts...)
    for (i,t) in enumerate(ts)
        if t <: T.var.ub
            rts = (ts[1:i-1]..., ts[i+1:end]...)
            return psub(T{t}, rts...)
        end
    end
    return (psub(T{P}, ts...) where {P})
end
```
[...]

PR #1850, opened 2020-12-10
Reformulate source arguments, add precomputed NamedTuple

charleskawczynski:
### Description

This is a tracer bullet for my [proposed solution](https://github.com/CliMA/ClimateMachine.jl/issues/1832#issuecomment-742168958) to #1832 specifically for `source` in the AtmosModel--but we can extend this to `flux` as well. The basic idea is that the argument list is:

```julia
source(::S, ::BalanceLaw, args::NamedTuple, precomputed::NamedTuple) where {S <: TendencyDef{Source, PrognosticVariable}}
```

`precomputed` is computed in `source!` as
```julia
precompute(atmos::AtmosModel, args, ::Source) =
    (ts = recover_thermo_state(atmos, args.state, args.aux),)

function source!(m::AtmosModel, ...)
    tend = Source()
    args = (
        state = state,
        aux = aux,
        t = t,
        direction = direction,
        diffusive = diffusive,
    )
    precomputed = precompute(m, args, tend)

    source.ρ = Σsources(eq_tends(Mass(), m, tend), m, args, precomputed)
...
end
```

We can compute `precomputed` in a nested, model-dependent, way so that we only precompute things when needed:

```julia
precompute(atmos::AtmosModel, args, tt::Flux{SecondOrder}) =
    (
     ts = recover_thermo_state(atmos, args.state, args.aux),
     turbulence = precompute(atmos.turbulence, atmos, args, tt),
     )
function precompute(m::TurbulenceClosureModel, atmos::AtmosModel, args, ::Flux{SecondOrder})
     ν, D_t, τ = turbulence_tensors(atmos, args.state, args.diffusive, args.aux, args.t)
    return (ν = ν, D_t = D_t, τ = τ)
end
```

That way, in the `flux` kernels, we'll have access to, for example, `precomputed.turbulence.D_t`. This should be easy to extend and inject nearly anything we want from the top-level `source!` into any nested submodel. Finally, `args` can be treated as effectively the same args that are available at the top-level `source!`. Thoughts?

Some solution, that allows us to optionally precompute things, is really needed to convert the remaining (EDMF) sources to the new tendency specification. Once we've converted all of the sources, we'll be able to cut out a bunch of unnecessary boilerplate (e.g., removing `atmos_source!`).

## Update

I've incorporated @simonbyrne's suggestion and used `merge` so that `precomputed` is stored in `args` as `args.precomputed`, so that we only need a single argument 🎉.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -800,6 +800,9 @@ function init_state_auxiliary!(
     )
 end
 
+precompute(atmos::AtmosModel, args, ::Source) =
+    (ts = recover_thermo_state(atmos, args.state, args.aux),)
+
 """
     source!(
         m::AtmosModel,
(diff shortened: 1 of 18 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -821,13 +824,22 @@ function source!(
...
-    source!(m.precipitation, source, args...)
+
+    args = (
+        state = state,
+        aux = aux,
+        t = t,
+        direction = direction,
+        diffusive = diffusive,
+    )
+    precomputed = precompute(m, args, tend)
```
simonbyrne, line 835: we could do
```suggestion
    _args = (
        state = state,
        aux = aux,
        t = t,
        direction = direction,
        diffusive = diffusive,
    )
    args = merge(_args, (precomputed = precompute(m, args, tend),))
```
then not need to pass `precomputed` seperately.

simonbyrne (approved):
👍 for the line diff!

charleskawczynski:
bors try

PR #1852, opened 2020-12-10
Add thermo ProfileSet Iterator

charleskawczynski:
### Description

Working on #1830, I realized that it would be really helpful to iterate, pointwise, through the set of thermo profiles. This way we can test the primitive-prognostic conversion across the same set of profiles that thermodynamics is tested. This PR adds and tests this iterator.

### DIFF
--- a/test/Common/Thermodynamics/profiles.jl
+++ b/test/Common/Thermodynamics/profiles.jl
@@ -34,6 +34,33 @@ struct ProfileSet{AFT, QPT, PT}
     phase_type::PT  # Phase type (e.g., `PhaseDry`, `PhaseEquil`)
 end
 
+function Base.iterate(ps::ProfileSet, state = 1)
+    state > length(ps.z) && return nothing
+    nt = (
+        z = ps.z[state],
+        T = ps.T[state],
+        p = ps.p[state],
+        RS = ps.RS[state],
+        e_int = ps.e_int[state],
+        ρ = ps.ρ[state],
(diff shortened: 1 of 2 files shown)


jkozdon (commented):
Not sure I have much to say about this...

test/Common/Thermodynamics/profiles.jl
```diff
@@ -34,6 +34,33 @@ struct ProfileSet{AFT, QPT, PT}
...
+        w = ps.w[state],
+        e_kin = ps.e_kin[state],
+        phase_type = ps.phase_type,
+    )
+    return (nt, state + 1)
+end
+Base.IteratorSize(::ProfileSet) = Base.HasLength()
+Base.length(ps::ProfileSet) = length(ps.z)
+
 """
```
Zhengyu-Huang, line 64: what is the state here ?  is it the nodal point id for looping all nodes?

Zhengyu-Huang (approved):
It looks good for me! Very fancy! Thank you!

tapios (approved):
LGTM

test/Common/Thermodynamics/profiles.jl
charleskawczynski: `state` is the iteration state. I followed the terminology in the julia [interfaces docs](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration) (the "Squares" example).

PR #1859, opened 2020-12-14
Jmc/split explicit cleaning

jm-c:
### Description
Split-Explicit Ocean model:
 - simplify and clean BC ; change sign of second-order fluxes ;
 - simplify checkpoints ; add (back) bits for checking State-Check in test

### DIFF
--- a/src/Ocean/SplitExplicit01/BarotropicModel.jl
+++ b/src/Ocean/SplitExplicit01/BarotropicModel.jl
@@ -85,7 +85,7 @@ end
     t,
 )
     ν = viscosity_tensor(m)
-    D.ν∇U = ν * G.U
+    D.ν∇U = -ν * G.U
 
     return nothing
 end
@@ -132,7 +132,7 @@ end
     t::Real,
 )
     # numerical diffusivity for stability
-    F.U -= D.ν∇U
+    F.U += D.ν∇U
 
     return nothing
 end
--- a/test/Ocean/SplitExplicit/simple_box_ivd.jl
+++ b/test/Ocean/SplitExplicit/simple_box_ivd.jl
@@ -16,6 +16,7 @@ using ClimateMachine.GenericCallbacks
 using ClimateMachine.VTK
 using ClimateMachine.Checkpoint
 
+using Test
 using MPI
 using LinearAlgebra
 using StaticArrays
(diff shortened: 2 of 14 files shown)


jm-c:
bors try

src/Ocean/SplitExplicit01/BarotropicModel.jl
```diff
@@ -85,7 +85,7 @@ end
     t,
 )
     ν = viscosity_tensor(m)
-    D.ν∇U = ν * G.U
+    D.ν∇U = -ν * G.U
```
glwagner, line 88: 🙏

src/Ocean/SplitExplicit01/BarotropicModel.jl
```diff
@@ -132,7 +132,7 @@ end
     t::Real,
 )
     # numerical diffusivity for stability
-    F.U -= D.ν∇U
+    F.U += D.ν∇U
```
glwagner, line 135: 🙏

test/Ocean/SplitExplicit/simple_box_ivd.jl
```diff
@@ -209,30 +210,24 @@ function main(; restart = 0)
 
     if restart > 0
         direction = EveryDirection()
-        Q_3D, A_3D, t0 =
+        Q_3D, _, t0 =
             read_checkpoint(vtkpath, "baroclinic", ArrayType, mpicomm, restart)
-        Q_2D, A_2D, _ =
+        Q_2D, _, _ =
             read_checkpoint(vtkpath, "barotropic", ArrayType, mpicomm, restart)
```
glwagner, line 216: Eventually, the function `read_checkpoint` probably shouldn't return auxiliary fields at all, since they aren't needed, right?

glwagner (approved):
Looks good. @jm-c can you provide a comment for future generations about why the reference values are changed here? Does it have to do with flipping the sign in the viscous flux?

test/Ocean/SplitExplicit/simple_box_ivd.jl
jm-c: Yes. I could have tried to modify read_ & write_ checkpoint to allow this, but wanted to keep this set of changes internal to split-explicit. Wll be addressed in the (near) future.

jm-c:
> Looks good. @jm-c can you provide a comment for future generations about why the reference values are changed here? Does it have to do with flipping the sign in the viscous flux?

This PR does not change results at all (flipping the sign give exactly the same results). But changes made before (such as vertical line integral) had a small impact on results (at machine truncation level). Since we are currently not testing anymore the State-Check values against these refvals, it does pass. However, since I put back the test for these State-Check values, I wanted these
set-up to pass so I lower the requirement for this 1 intermediate/diagnostic variable Delta.Eta.

glwagner:
> This PR does not change results at all (flipping the sign give exactly the same results). But changes made before (such as vertical line integral) had a small impact on results (at machine truncation level). Since we are currently not testing anymore the State-Check values against these refvals, it does pass. However, since I put back the test for these State-Check values, I wanted these
set-up to pass so I lower the requirement for this 1 intermediate/diagnostic variable Delta.Eta.

Ok, thanks for the explanation!

jm-c:
bors try

christophernhill (approved):

jm-c:
bors try

Issue #1678, opened 2020-10-29
ArtifactWrappers is not multi-process safe


kpamnany:
### Description

See [this failure](https://buildkite.com/clima/climatemachine-ci/builds/572#d76710eb-c466-4c30-ad3d-4636e5b33bd6/208-210).

There was a similar problem with the compile cache in Julia, see [this](https://github.com/JuliaLang/julia/pull/36638) for how it was solved (and the problem with the solution).

kpamnany:
Cc: @charleskawczynski

charleskawczynski:
Looks like #1684 didn't actually fix this issue. Seeing
```
ERROR: LoadError: ArgumentError: '/central/scratchio/esm/climaci/climatemachine-ci/717/depot/gpu/artifacts/ff73fa6a0b6a807e71a6921f7ef7d0befe776edd' exists. `force=true` is required to remove '/central/scratchio/esm/climaci/climatemachine-ci/717/depot/gpu/artifacts/ff73fa6a0b6a807e71a6921f7ef7d0befe776edd' before copying.
  | Stacktrace:
  | [1] checkfor_mv_cp_cptree(::String, ::String, ::String; force::Bool) at ./file.jl:308
  | [2] cp(::String, ::String; force::Bool, follow_symlinks::Bool) at ./file.jl:345
  | [3] rename(::String, ::String; force::Bool) at ./file.jl:897
  | [4] rename at ./file.jl:894 [inlined]
  | [5] #mv#13 at ./file.jl:393 [inlined]
  | [6] mv at ./file.jl:392 [inlined]
  | [7] create_artifact(::ClimateMachine.ArtifactWrappers.var"#2#5"{ArtifactWrapper}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Artifacts.jl:229
  | [8] get_data_folder(::ArtifactWrapper) at /central/scratchio/esm/climaci/climatemachine-ci/717/climatemachine-ci/src/InputOutput/ArtifactWrappers/ArtifactWrappers.jl:79
  | [9] top-level scope at /central/scratchio/esm/climaci/climatemachine-ci/717/climatemachine-ci/test/Land/Model/haverkamp_implicit_test.jl:42
  | [10] include(::Function, ::Module, ::String) at ./Base.jl:380
  | [11] include(::Module, ::String) at ./Base.jl:368
[...]

PR #1860, opened 2020-12-14
Fix race condition in Squal line artifact

charleskawczynski:
### Description

This PR should hopefully fix the race condition we're seeing with the squal line experiment in #1678. I'm not sure if this is the best way to grab the job ID.

### DIFF
--- a/experiments/AtmosLES/squall_line.jl
+++ b/experiments/AtmosLES/squall_line.jl
@@ -122,8 +122,17 @@ end
   Read the original squall sounding
 """
 function read_sounding()
+
+    # Artifact creation is not thread-safe
+    #      https://github.com/JuliaLang/Pkg.jl/issues/1219
+    # To avoid race conditions from multiple jobs running this
+    # driver at the same time, we must store artifacts in a
+    # separate folder.
+
+    artifact_folder = mktempdir(@__DIR__; prefix = "artifacts_")
+
     soundings_dataset = ArtifactWrapper(
-        joinpath(@__DIR__, "Artifacts.toml"),
+        joinpath(artifact_folder, "Artifacts.toml"),
         "soundings",
         ArtifactFile[ArtifactFile(
             url = "https://caltech.box.com/shared/static/rjnvt2dlw7etm1c7mmdfrkw5gnfds5lx.nc",


experiments/AtmosLES/squall_line.jl
```diff
@@ -122,8 +122,19 @@ end
...
 """
 function read_sounding()
+
+    # Artifact creation is not thread-safe
+    #      https://github.com/JuliaLang/Pkg.jl/issues/1219
+    # To avoid race conditions from multiple jobs running this
+    # driver at the same time, we must store artifacts in a
+    # separate, job-unique, folder.
+
+    job_id = ENV["BUILDKITE_AGENT_META_DATA_JOBID"]
```
kpamnany, line 132: 🤔 What if this isn't running under Buildkite?

Maybe just use [`mktempdir`](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktempdir-Tuple{AbstractString})?

charleskawczynski: That's exactly what I was thinking. I forgot about `mktempdir`, good idea!

kpamnany (approved):

experiments/AtmosLES/squall_line.jl
```diff
@@ -122,8 +122,17 @@ end
...
 """
 function read_sounding()
+
+    # Artifact creation is not thread-safe
+    #      https://github.com/JuliaLang/Pkg.jl/issues/1219
+    # To avoid race conditions from multiple jobs running this
+    # driver at the same time, we must store artifacts in a
+    # separate folder.
+
+    artifact_folder = mktempdir(@__DIR__; prefix = "artifacts_")
```
kpamnany, line 132: Should note that if this is run with more than one MPI rank, each will download their own copy.

charleskawczynski: Ah, right, makes sense. I'll update the first PR comment.

simonbyrne: Another solution is to
```julia
if MPI.Comm_rank(comm) == 0
    # download file
end
MPI.Barrier(comm)
```

charleskawczynski: These are single-rank MPI jobs, so, IIUC, `mpirank = 1` for each separate job.

simonbyrne: ranks are 0-based (so there will always be a rank 0)

simonbyrne: Ah, but i see that the issues is across jobs, not within MPI jobs

PR #1864, opened 2020-12-15
Move precompute method to balance law

charleskawczynski:
### Description

This should have been incorporated in #1850, we want `precompute` to be a `BalanceLaw` method, so that we can inherit/modularize its use across different models.

### DIFF
--- a/src/BalanceLaws/tendency_types.jl
+++ b/src/BalanceLaws/tendency_types.jl
@@ -17,7 +17,7 @@
 
 export PrognosticVariable
 export FirstOrder, SecondOrder
-export Flux, Source
+export AbstractTendencyType, Flux, Source
 export TendencyDef
 
 """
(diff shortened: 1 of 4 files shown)


src/BalanceLaws/tendency_types.jl
```diff
@@ -121,3 +121,14 @@ function fluxes(bl::BalanceLaw, order::O) where {O <: AbstractOrder}
     tend = filter(x -> x ≠ nothing, tend)
     return Tuple(Iterators.flatten(tend))
 end
+
+"""
+    precompute(bl, args, ::AbstractTendencyType)
+
+A nested NamedTuple of precomputed (cached) values
+and or objects. This is useful for "expsensive"
```
kpamnany, line 129:
```suggestion
and or objects. This is useful for "expensive"
```

kpamnany (approved):

PR #1865, opened 2020-12-15
Reformulate first order flux arguments, add precomputed NamedTuple

charleskawczynski:
### Description

This is the first-order flux counter-part to #1850.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -479,6 +479,9 @@ include("prog_prim_conversion.jl")   # prognostic<->primitive conversion
 include("atmos_tendencies.jl")        # specify atmos tendencies
 include("get_prognostic_vars.jl")     # get tuple of prognostic variables
 
+precompute(atmos::AtmosModel, args, tt::Flux{FirstOrder}) =
+    (ts = recover_thermo_state(atmos, args.state, args.aux),)
+
 """
     flux_first_order!(
         m::AtmosModel,
@@ -492,25 +495,28 @@ Computes and assembles non-diffusive fluxes in the model
 equations.
 """
 @inline function flux_first_order!(
-    m::AtmosModel,
+    atmos::AtmosModel,
     flux::Grad,
     state::Vars,
     aux::Vars,
     t::Real,
     direction,
 )
-    flux_pad = SVector(1, 1, 1)
-    ts = recover_thermo_state(m, state, aux)
-    tend = Flux{FirstOrder}()
-    args = (m, state, aux, t, ts, direction)
-    flux.ρ = Σfluxes(eq_tends(Mass(), m, tend), args...) .* flux_pad
-    flux.ρu = Σfluxes(eq_tends(Momentum(), m, tend), args...) .* flux_pad
-    flux.ρe = Σfluxes(eq_tends(Energy(), m, tend), args...) .* flux_pad
 
-    flux_first_order!(m.moisture, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.precipitation, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.tracers, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.turbconv, m, flux, state, aux, t, ts, direction)
+    flux_pad = SVector(1, 1, 1)
+    tend = Flux{FirstOrder}()
+    _args = (state = state, aux = aux, t = t, direction = direction)
+    args = merge(_args, (precomputed = precompute(atmos, _args, tend),))
+    flux.ρ = Σfluxes(eq_tends(Mass(), atmos, tend), atmos, args) .* flux_pad
+    flux.ρu =
+        Σfluxes(eq_tends(Momentum(), atmos, tend), atmos, args) .* flux_pad
+    flux.ρe = Σfluxes(eq_tends(Energy(), atmos, tend), atmos, args) .* flux_pad
+
+    flux_first_order!(atmos.moisture, atmos, flux, args)
+    flux_first_order!(atmos.precipitation, atmos, flux, args)
+    flux_first_order!(atmos.tracers, atmos, flux, args)
+    flux_first_order!(atmos.turbconv, atmos, flux, args)
+
 end
 
 function compute_gradient_argument!(
@@ -812,7 +818,7 @@ function init_state_auxiliary!(
     )
 end
 
-precompute(atmos::AtmosModel, args, ::Source) =
+precompute(atmos::AtmosModel, args, tt::Source) =
     (ts = recover_thermo_state(atmos, args.state, args.aux),)
 
 """
--- a/src/Atmos/Model/moisture.jl
+++ b/src/Atmos/Model/moisture.jl
@@ -12,16 +12,7 @@ function atmos_nodal_update_auxiliary_state!(
     aux::Vars,
     t::Real,
 ) end
-function flux_first_order!(
-    ::MoistureModel,
-    ::AtmosModel,
-    flux::Grad,
-    state::Vars,
-    aux::Vars,
-    t::Real,
-    ts,
-    direction,
-) end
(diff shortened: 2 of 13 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

kpamnany:
It looks like this adds overhead:
```julia
julia> function f(a, b, c)
           return a + b + c
       end
f (generic function with 1 method)

julia> @code_warntype f(1, 2, 3)
Variables
  #self#::Core.Compiler.Const(f, false)
  a::Int64
  b::Int64
  c::Int64

Body::Int64
1 ─ %1 = (a + b + c)::Int64
└──      return %1
```
Vs:
```julia
julia> function g(comp)
           @unpack a, b, c = comp
           a + b + c
       end
g (generic function with 1 method)

julia> @code_warntype g((a = 1, b = 2, c = 3))
Variables
  #self#::Core.Compiler.Const(g, false)
  comp::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
  253::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
  a::Int64
  b::Int64
  c::Int64

Body::Int64
1 ─       (253 = comp)
│   %2  = Base.getproperty(UnPack, :unpack)::Core.Compiler.Const(UnPack.unpack, false)
│   %3  = 253::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
│   %4  = Core.apply_type(Main.Val, :a)::Core.Compiler.Const(Val{:a}, false)
│   %5  = (%4)()::Core.Compiler.Const(Val{:a}(), false)
│         (a = (%2)(%3, %5))
│   %7  = Base.getproperty(UnPack, :unpack)::Core.Compiler.Const(UnPack.unpack, false)
│   %8  = 253::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
│   %9  = Core.apply_type(Main.Val, :b)::Core.Compiler.Const(Val{:b}, false)
│   %10 = (%9)()::Core.Compiler.Const(Val{:b}(), false)
│         (b = (%7)(%8, %10))
│   %12 = Base.getproperty(UnPack, :unpack)::Core.Compiler.Const(UnPack.unpack, false)
│   %13 = 253::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
│   %14 = Core.apply_type(Main.Val, :c)::Core.Compiler.Const(Val{:c}, false)
│   %15 = (%14)()::Core.Compiler.Const(Val{:c}(), false)
│         (c = (%12)(%13, %15))
│         253
│   %18 = (a + b + c)::Int64
└──       return %18
```
Cc: @vchuravy

charleskawczynski:
The previously used `@unpack_fields`:

```julia
julia> macro unpack_fields(stuff, syms...)
         thunk = Expr(:block)
         for sym in syms
           push!(thunk.args, :($(esc(sym)) = $(esc(stuff))[$(QuoteNode(sym))]))
         end
         push!(thunk.args, nothing)
         return thunk
       end
@unpack_fields (macro with 1 method)

julia> function h(comp)
          @unpack_fields comp a b c
          return a + b + c
       end
h (generic function with 1 method)

julia> @code_warntype h((a=1,b=2,c=3))
Variables
  #self#::Core.Compiler.Const(h, false)
  comp::NamedTuple{(:a, :b, :c),Tuple{Int64,Int64,Int64}}
  a::Int64
  b::Int64
  c::Int64

Body::Int64
1 ─      (a = Base.getindex(comp, :a))
│        (b = Base.getindex(comp, :b))
│        (c = Base.getindex(comp, :c))
│   %4 = (a + b + c)::Int64
└──      return %4
```

kpamnany:
Looking deeper, the `@code_llvm`s for functions that take composite type parameters vs functions that take a `NamedTuple` of composite types as a parameter looks very similar to me.

@vchuravy: any insight into whether using this pattern really costs us anything?

vchuravy:
`code_warntype` is the wrong thing to look at, since it tries to make it easier to understand, but hides the actual code.
`code_typed` should be representative.

charleskawczynski:
If that's the case, then the only difference I see is a call to `getproperty` to unpack things, which seems innocuous to me.

charleskawczynski:
bump

src/Atmos/Model/AtmosModel.jl
```diff
@@ -499,18 +502,20 @@ equations.
...
-    flux.ρ = Σfluxes(eq_tends(Mass(), m, tend), args...) .* flux_pad
-    flux.ρu = Σfluxes(eq_tends(Momentum(), m, tend), args...) .* flux_pad
-    flux.ρe = Σfluxes(eq_tends(Energy(), m, tend), args...) .* flux_pad
-
-    flux_first_order!(m.moisture, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.precipitation, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.tracers, m, flux, state, aux, t, ts, direction)
-    flux_first_order!(m.turbconv, m, flux, state, aux, t, ts, direction)
+    _args = (state = state, aux = aux, t = t, direction = direction)
+    args = merge(_args, (precomputed = precompute(m, _args, tend),))
```
kpamnany, line 509: Any reason for `precomputed` to be specifically labelled? We could simply merge whatever `precompute` returns into `args` directly?

kpamnany: Also, since you're touching this code, let's please use `atmos::AtmosModel` consistently -- it is `m::AtmosModel` in some places and that's very annoying. 😱

src/Atmos/Model/moisture.jl
```diff
@@ -154,16 +145,11 @@ function flux_first_order!(
...
-    t::Real,
-    ts,
-    direction,
+    args,
 )
     tend = Flux{FirstOrder}()
-    args = (atmos, state, aux, t, ts, direction)
     flux.moisture.ρq_tot =
-        Σfluxes(eq_tends(TotalMoisture(), atmos, tend), args...)
+        Σfluxes(eq_tends(TotalMoisture(), atmos, tend), atmos, args)
```
kpamnany, line 152: I see `atmos` is added to the arguments for `Σfluxes` here, but goes into `args` in sub-components... I sort of see why, but more consistency might be better?

kpamnany (approved):
LGTM and there shouldn't be a negative effect from the `@unpack`ing, but please compare the CI run times from a try here and the last PR merge?

Also, someone better versed in the physics should also review...

src/Atmos/Model/AtmosModel.jl
charleskawczynski: > Any reason for precomputed to be specifically labelled? We could simply merge whatever precompute returns into args directly?

I kind of like the idea of encapsulating the pre-computed things separately from the top-level arguments (which is basically what `_args` is), so that we know what is cached vs what is a top-level argument/positional arg when we're deep in the tendency kernels.

> Also, since you're touching this code, let's please use atmos::AtmosModel consistently -- it is m::AtmosModel in some places and that's very annoying. 😱

Yes, I can update that. It's been bothering me also.

src/Atmos/Model/moisture.jl
charleskawczynski: It should be consistent (in `precipitation.jl`), thanks!

charleskawczynski:
bors try

charleskawczynski:
@trontrytel @ilopezgp @yairchn can one of you take a look?

charleskawczynski:
bors try

trontrytel (approved):
LGTM. I only see changes to the interface, no changes to the physics

charleskawczynski:
Hope we can automate something like this soon (runtime in seconds):

PR

charleskawczynski:
The variation doesn't seem alarming to me. bors r+

PR #1867, opened 2020-12-16
Reformulate second order flux arguments, add precomputed NamedTuple

charleskawczynski:
### Description

This is the second-order flux counter-part to #1850.

I think we'll be able to remove the call to `thermo_state` in second-order fluxes, but it will be more clear that we can once we've moved all second-order fluxes to the new specification.

I also removed a few unused pieces in `TurbulenceClosures.jl`

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -619,6 +619,9 @@ function transform_post_gradient_laplacian!(
     )
 end
 
+precompute(atmos::AtmosModel, args, ::Flux{SecondOrder}) =
+    (ts = recover_thermo_state(atmos, args.state, args.aux),)
+
 """
     flux_second_order!(
         atmos::AtmosModel,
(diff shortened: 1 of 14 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -619,6 +619,9 @@ function transform_post_gradient_laplacian!(
     )
 end
 
+precompute(atmos::AtmosModel, args, ::Flux{SecondOrder}) =
+    (ts = recover_thermo_state(atmos, args.state, args.aux),)
```
kpamnany, line 623: Wondering if you want to add: `ν, D_t, τ = turbulence_tensors(atmos, state, diffusive, aux, t)` to this.

kpamnany (approved):
LGTM.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I thought the same thing, and I did at first, but then I thought that it might be PMO because `turbulence_tensors` is cheap AFAICT. It's an easy thing to change. I'll incorporate it in a follow-up PR, so we can compare the buildkite runtime to see if there's a noticeable gain.

Issue #1831, opened 2020-12-07
Compare NewtonsMethod with NewtonsMethodAD


charleskawczynski:
### Description

Per [Tapio's suggestion](https://github.com/CliMA/ClimateMachine.jl/pull/1763#discussion_r537788543), we should compare `NewtonsMethod` with `NewtonsMethodAD` w.r.t. performance/accuracy, and explore:

> If AD is fast enough, it would be nice because it would allow us to use arbitrary (smooth) phase partitioning functions.

PR #1869, opened 2020-12-16
Make swapping sat adjust numerical methods easier, add GPU tests

charleskawczynski:
### Description

This PR is a step towards fixing #1831 by using the numerical method type, rather than a function, as an input to `saturation_adjustment`. This should help us more rigorously test/compare the performance and robustness of numerical methods in saturation adjustment for `ρ, e_int, q_tot` and `ρ, p, q_tot` thermo variables, which are the only two cases that are potentially used beyond initialization.

I've also battled the formatter a bit in the `@print` statements, and added a test section for performance with BenchmarkTools.

This also:
 - adds a convenience method, `PhaseEquil_dev_only`, _for development only_ so that we can more easily toggle one option at a time while maintaining synchronization with all other arguments.
 - Adds GPU tests

### DIFF
--- a/src/Common/Thermodynamics/config_numerical_method.jl
+++ b/src/Common/Thermodynamics/config_numerical_method.jl
@@ -1,0 +1,119 @@
+# These functions (variants of sa_numerical_method)
+# return an instance of a numerical method to solve
+# saturation adjustment, for different combinations
+# of thermodynamic variable inputs.
+
+# @print only accepts literal strings, so we must
+# branch to print which method is being used.
+function print_numerical_method(
+    ::Type{sat_adjust_method},
+) where {sat_adjust_method}
+    if sat_adjust_method <: NewtonsMethod
+        @print("    Method=NewtonsMethod")
+    elseif sat_adjust_method <: NewtonsMethodAD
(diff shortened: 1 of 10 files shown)


tapios (approved):
Looks nice! I'm looking forward to seeing the performance results.

src/Common/Thermodynamics/config_numerical_method.jl
```diff
@@ -0,0 +1,119 @@
...
+    elseif sat_adjust_method <: NewtonsMethodAD
+        @print("    Method=NewtonsMethodAD")
+    elseif sat_adjust_method <: SecantMethod
+        @print("    Method=SecantMethod")
+    elseif sat_adjust_method <: RegulaFalsiMethod
+        @print("    Method=RegulaFalsiMethod")
+    else
+        error("Unsupported numerical method")
+    end
+end
```
charleskawczynski, line 22: @vchuravy is there a cleaner way?

vchuravy: Yeah we should implement `@printf` proper https://github.com/JuliaGPU/KernelAbstractions.jl/issues/73
cc: @leios

PR #1873, opened 2020-12-17
Refactor land BCs

kmdeck:
### Description
Refactoring land BC to be more in line with the way atmos and the ocean do it, motivated by PR:
https://github.com/CliMA/ClimateMachine.jl/pull/1727
and discussion here:
https://github.com/CliMA/ClimateMachine.jl/pull/1838

This also adds in the 3D config for the land model, to make sure this works for that as well (not just single stack).

The default for all boundary conditions is to apply no boundary condition, which is helpful because (a) not all components of the models are turned on at the same time, depending on what the user chooses), and (b) we dont always need to supply a BC for the lateral sides - only when running in 3d.

To run a soil water model in 1D (vertical), you could do something like:
```
    bc = LandBCStruct(
        bottom_bc = LandBC(
            soil_water = Neumann(bottom_flux)
        ),
        surface_bc = LandBC(
            soil_water = Dirichlet(surface_state)
        )
    )
```
and then land.boundary_conditions would be set to this. Internally, this is computed:
```
function boundary_conditions(land::LandModel)
    bc = land.boundary_conditions # of type LandBCStruct
    mytuple = (bc.bottom_bc, bc.surface_bc, bc.lateral_bc)
    #faces integer 1,2,3 are bottom, top, lateral sides.
    return mytuple
end
```

To run both heat and water in 3d, you'd need something like:
```
    bc = LandBCStruct(
        bottom_bc = LandBC(
            soil_water = Neumann(bottom_flux),
            soil_heat = Neumann(bottom_heat_flux)
        ),
        surface_bc = LandBC(
            soil_water = Dirichlet(surface_state),
            soil_heat = Neumann(top_heat_flux)
        ),
     lateral_bc = LandBC(soil_water = whatever, soil_heat = whatever)
    )
```


kmdeck:
@simonbyrne I started this branch to refactor the land BC....

The test script is ./test/Land/Model/heat_analytic_unit_test.jl.

I've made it so the boundary conditions for the land model are specified as part of the LandModel object, in the form: `(LandBC(soil_water = WaterSurfaceBC, soil_heat = HeatSurfaceBC), LandBC(soil_water = WaterBottomBC, soil_heat = HeatBottomBC))`
where `boundary_conditions(land::LandModel) = land.boundary_conditions
`

I think the code in the background will iterate over this tuple, using element 1 for the face denoted with 1 in the driver config, using element 2 for the face denoted with 2 in the driver config...

It isn't quite working yet. (The issue is somewhere in land_bc.jl - it's trying to call the soil_boundary_state!() function with the arguments needed for a flux BC, but with Dirichlet() as the bctype..and I didnt define this method because it isnt required. Not sure how to fix this yet).

I have a couple other questions as well...

kmdeck:
> @simonbyrne I started this branch to refactor the land BC....
> 
> The test script is ./test/Land/Model/heat_analytic_unit_test.jl.
> 
> I've made it so the boundary conditions for the land model are specified as part of the LandModel object, in the form: `(LandBC(soil_water = WaterSurfaceBC, soil_heat = HeatSurfaceBC), LandBC(soil_water = WaterBottomBC, soil_heat = HeatBottomBC))`
> where `boundary_conditions(land::LandModel) = land.boundary_conditions `
> 
> I think the code in the background will iterate over this tuple, using element 1 for the face denoted with 1 in the driver config, using element 2 for the face denoted with 2 in the driver config...
> 
> It isn't quite working yet. (The issue is somewhere in land_bc.jl - it's trying to call the soil_boundary_state!() function with the arguments needed for a flux BC, but with Dirichlet() as the bctype..and I didnt define this method because it isnt required. Not sure how to fix this yet).
> 
> I have a couple other questions as well...

got it to work!

kmdeck:
I think this is working now. I changed it so the land.boundary_conditions object has fields labeled bottom_bc, surface_bc, and lateral_bc. Then the boundary_functions(land::LandModel) extracts those and stores them in a tuple in the correct order. this seemed better to be then expecting the user to remember what order the faces are in (i also kept mixing up top and bottom).

the defaults seem to be working. Next up would be testing this in a 3D config using the lateral_bc as well.

LandBCStruct isnt a great name, also.

kmdeck:
Marking as ready for review...though @simonbyrne pointed out this wont work as we intended when we have warped surfaces. Right now we assume that the normal vector is aligned with the cartesian vectors (or anti-aligned). Some more thought is needed to determine the correct way to specify the BC in this case, for land - using the normal vector or the projected normal vector along the vertical direction.

kmdeck:
bors try

src/Land/Model/LandModel.jl
```diff
@@ -211,15 +217,15 @@ function init_state_prognostic!(
...
 include("land_bc.jl")
 include("SoilWaterParameterizations.jl")
 using .SoilWaterParameterizations
 include("SoilHeatParameterizations.jl")
 using .SoilHeatParameterizations
-include("Runoff.jl")
-using .Runoff
 include("soil_model.jl")
-include("soil_boundary_types.jl")
+#include("soil_boundary_types.jl")
```
simonbyrne, line 228:
```suggestion
```

simonbyrne (approved):

src/Land/Model/LandModel.jl
kmdeck: will change!

src/Driver/driver_configs.jl
```diff
@@ -477,6 +477,99 @@ Establishing single stack configuration for %s
     )
 end
 
+function MultiColumnLandModel(
```
valeriabarra, line 480: Sorry I'm just noticing this, but this function `MultiColumnLandModel` needs to be added to the summary of all configs at the top of the file. I also missed it in my recent PR in which I touched this file :/

kmdeck: do you mean the comments at the top? If it's ok, I can make the change in the next PR we have for land, even if it is unrelated to driver configs.

valeriabarra: Yup, that summary/comment at the top of this file. Thanks!

valeriabarra: Hey @kmdeck , I actually took care of it in #1960 , don't worry about it

kmdeck: ok! thank you! :)

PR #1876, opened 2020-12-18
Add TKE budget variables to GradientFlux().

ilopezgp:
### Description

This PR adds TKE budget variables to the GradientFlux() variables of the EDMF scheme. Adding these variables to a state is necessary if we want to output them with the current version of the model.

### DIFF
--- a/test/Atmos/EDMF/edmf_kernels.jl
+++ b/test/Atmos/EDMF/edmf_kernels.jl
@@ -143,6 +143,11 @@ function vars_state(::Environment, ::GradientFlux, FT)
         ∇θ_liq_q_tot_cv::SVector{3, FT},
         ∇θv::SVector{3, FT},
         ∇e::SVector{3, FT},
+        K_m::FT,
+        l_mix::FT,
+        shear_prod::FT,
+        buoy_prod::FT,
+        tke_diss::FT
     )
 
 end
@@ -370,6 +375,7 @@ function compute_gradient_flux!(
     up_∇tf = ∇transform.turbconv.updraft
     en_dif = diffusive.turbconv.environment
     en_∇tf = ∇transform.turbconv.environment
+    en = state.turbconv.environment
 
     @unroll_map(N_up) do i
         up_dif[i].∇w = up_∇tf[i].w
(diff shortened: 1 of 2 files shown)


test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -151,7 +151,12 @@ function vars_state(m::EDMF, st::GradientFlux, FT)
     @vars(
         S²::FT, # should be conditionally grabbed from atmos.turbulence
         environment::vars_state(m.environment, st, FT),
-        updraft::vars_state(m.updraft, st, FT)
+        updraft::vars_state(m.updraft, st, FT),
```
yairchn, line 154: is there a reason not to add these variables in the environment in 

https://github.com/CliMA/ClimateMachine.jl/blob/8e2ef08353183dbf6daaf3e5e12d3f5b5a1db0f9/test/Atmos/EDMF/edmf_kernels.jl#L135

test/Atmos/EDMF/stable_bl_edmf.jl
```diff
@@ -198,7 +200,7 @@ function main(::Type{FT}) where {FT}
     grid = driver_config.grid
 
     # state_types = (Prognostic(), Auxiliary(), GradientFlux())
-    state_types = (Prognostic(), Auxiliary())
+    state_types = (Prognostic(), Auxiliary(), GradientFlux())
```
yairchn, line 203: if we are exporting GradientFlux and entrainment and detrainment are in the output could we compute all the needed values offline ?

test/Atmos/EDMF/edmf_kernels.jl
ilopezgp: Good point, I will add them to the environment

test/Atmos/EDMF/stable_bl_edmf.jl
ilopezgp: Well yes but I would have to rewrite the EDMF offline, I don't see the point if I can just get the output here. The same is applicable to entrainment and detrainment.

ilopezgp: The ideal think would be if I could get the `atmos_source!` variables

charleskawczynski (approved):
I'm fine with these changes assuming that this will help edmf development. Adding fields to `vars_state(::GradientFlux)` allocates more global memory, which we'll need to eventually remove. So, it'll behoove us to avoid depending on these fields inside `flux_first_order!`/`source!` since we'll need to eventually remove them.

ilopezgp:
> I'm fine with these changes assuming that this will help edmf development. Adding fields to `vars_state(::GradientFlux)` allocates more global memory, which we'll need to eventually remove. So, it'll behoove us to avoid depending on these fields inside `flux_first_order!`/`source!` since we'll need to eventually remove them.

I agree! We will have to remove it at some point, this is just temporary.

PR #1877, opened 2020-12-19
Move hyperdiffusion fluxes to new tend spec

charleskawczynski:
### Description

The PR moves hyperdiffusion to the new tendency specification.

### DIFF
--- a/docs/src/APIs/Common/TurbulenceClosures.md
+++ b/docs/src/APIs/Common/TurbulenceClosures.md
@@ -12,6 +12,8 @@ TurbulenceClosures
 
 ```@docs
 TurbulenceClosureModel
+hyperdiff_enthalpy_and_momentum_flux
+hyperdiff_momentum_flux
 WithDivergence
 WithoutDivergence
 ConstantViscosity
--- a/src/Atmos/Model/tendencies_energy.jl
+++ b/src/Atmos/Model/tendencies_energy.jl
@@ -26,6 +26,16 @@ function flux(::ViscousFlux{Energy}, atmos, args)
     return τ * state.ρu
 end
 
+function flux(::HyperdiffViscousFlux{Energy}, atmos, args)
+    @unpack state, hyperdiffusive = args
+    return hyperdiffusive.hyperdiffusion.ν∇³u_h * state.ρu
+end
+
+function flux(::HyperdiffEnthalpyFlux{Energy}, atmos, args)
+    @unpack state, hyperdiffusive = args
+    return hyperdiffusive.hyperdiffusion.ν∇³h_tot * state.ρ
+end
+
 struct DiffEnthalpyFlux{PV <: Energy} <: TendencyDef{Flux{SecondOrder}, PV} end
 function flux(::DiffEnthalpyFlux{Energy}, atmos, args)
     @unpack state, aux, t, diffusive = args
--- a/src/Atmos/Model/tendencies_moisture.jl
+++ b/src/Atmos/Model/tendencies_moisture.jl
@@ -47,6 +47,11 @@ function flux(::MoistureDiffusion{IceMoisture}, atmos, args)
     return d_q_ice * state.ρ
 end
 
+function flux(::HyperdiffViscousFlux{TotalMoisture}, atmos, args)
+    @unpack state, hyperdiffusive = args
+    return hyperdiffusive.hyperdiffusion.ν∇³q_tot * state.ρ
+end
+
 #####
 ##### Sources
 #####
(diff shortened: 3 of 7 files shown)


src/Atmos/Model/tendencies_energy.jl
```diff
@@ -26,6 +26,16 @@ function flux(::ViscousFlux{Energy}, atmos, args)
     return τ * state.ρu
 end
 
+function flux(::HyperdiffViscousFlux{Energy}, atmos, args)
+    @unpack state, hyperdiffusive = args
+    return hyperdiffusive.hyperdiffusion.ν∇³u_h * state.ρu
```
trontrytel, line 31: Maybe some different names could be used here? `hyperdiffusive.hyperdiffusion` seems a little too hyper?

src/Atmos/Model/tendencies_moisture.jl
```diff
@@ -47,6 +47,11 @@ function flux(::MoistureDiffusion{IceMoisture}, atmos, args)
     return d_q_ice * state.ρ
 end
 
+function flux(::HyperdiffViscousFlux{TotalMoisture}, atmos, args)
+    @unpack state, hyperdiffusive = args
+    return hyperdiffusive.hyperdiffusion.ν∇³q_tot * state.ρ
```
trontrytel, line 52: Should this also work with nonequilibrium moisture model? Then we would also need this for cloud liquid and ice variables? How about precipitation variables?

trontrytel (commented):
I left two comments. Looks good otherwise.

src/Atmos/Model/tendencies_energy.jl
charleskawczynski: I agree that it's a bit hyper, but it might be nice to change the name across the board since `hyperdiffusive.hyperdiffusion` exists in a bunch of other places. Maybe in a follow-up PR?

src/Atmos/Model/tendencies_moisture.jl
charleskawczynski: Yep, absolutely. We can add `flux(::HyperdiffViscousFlux{LiquidMoisture}` and `flux(::HyperdiffViscousFlux{IceMoisture}` methods if that's needed in the future.

src/Atmos/Model/tendencies_energy.jl
trontrytel: sounds good

src/Atmos/Model/tendencies_moisture.jl
trontrytel: @jiahe23 - Should we do it now? Or is it obvious to everyone that it should not be used with moisture and precipitation vars right now?

jiahe23: For now, we don't have a hyperdiffussion written for nonequil. We only have these two options
- [DryBiharmonic](https://github.com/CliMA/ClimateMachine.jl/blob/71c045457c177c0c1f2cfbc9008162fa52b87435/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L892) 
- [EquilMoistBiharmonic](https://github.com/CliMA/ClimateMachine.jl/blob/71c045457c177c0c1f2cfbc9008162fa52b87435/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L804)

How about leaving the non-equil related var out for now?

docs/src/APIs/Common/TurbulenceClosures.md
```diff
@@ -12,6 +12,8 @@ TurbulenceClosures
 
 ```@docs
 TurbulenceClosureModel
+hyperdiff_enthalpy_and_visc_flux
```
tapios, line 15: I'd call this `hyperdiff_enthalpy_and_momentum_flux` throughout (clearer than viscous flux, and parallel to enthalpy).

docs/src/APIs/Common/TurbulenceClosures.md
```diff
@@ -12,6 +12,8 @@ TurbulenceClosures
...
+hyperdiff_visc_flux
```
tapios, line 16: Same here and elsewhere: `hyperdiff_momentum_flux`

charleskawczynski:
The one thing that I'm not super pleased about is having to add the extra functions (e.g., `hyperdiff_momentum_flux`). Ideally, this would be called `eq_tends`, but I realized that if we want to share this model among different balance laws, and different prognostic variables (which, I'm not sure how ready this is at the moment), then we have to leave the interface relatively generic and we can't extend `eq_tends` because it requires typing the prognostic variable, which is model-dependent. Do we plan on using hyperdiffusion outside of the atmosphere? Are/can these kernels actually be shared?

tapios:
Yes, hyperdiffusion is also useful for the ocean.

trontrytel (approved):

PR #1879, opened 2020-12-21
Remove `atmos_source!` from the code

charleskawczynski:
### Description

This PR removes `atmos_source!` from the code. Some of the logic has been flipped-- now `filter_source(pv, m, s)` returns `s` (the source definition) by default, instead of `nothing`. So, it would be nice to get a few pairs of eyes on this PR.

@trontrytel pointed out that the model-dependent filter logic is simply no longer needed, which absolutely makes sense, now that it's been inverted.


src/Atmos/Model/AtmosModel.jl
```diff
@@ -844,7 +843,7 @@ Computes (and assembles) source terms `S(Y)` in:
 ```
 """
 function source!(
-    m::AtmosModel,
+    atmos::AtmosModel,
```
kpamnany, line 846: ❤️

kpamnany (approved):
👍

Issue #1469, opened 2020-08-25
Add points in tested profiles to 3D plots


charleskawczynski:
# Description

It would be nice to see where our [tested profiles](https://clima.github.io/ClimateMachine.jl/latest/HowToGuides/Common/Thermodynamics/#Moist-Phase-in-thermodynamic-equilibrium) fall in the full [3D plots](https://clima.github.io/ClimateMachine.jl/latest/HowToGuides/Common/Thermodynamics/#Converged-cases-(3D-view)) that show all permutations of inputs.

# Additional context

Add any other reasons why this feature will help with the goals of CLIMA.

# For CLIMA Developers

- [ ] This feature can be added (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been (or will be) implemented (Please link the PR)

PR #1885, opened 2020-12-21
Move and improve thermo convergence plots

charleskawczynski:
### Description

This PR moves the thermo convergence percentage analysis to the developer docs, and adds plots for more numerical methods. Partially addresses #1469.


docs/src/DevDocs/Common/Thermodynamics/saturation_adjustment.jl
```diff
@@ -0,0 +1,213 @@
...
+dims = (10, 10, 10);
+ρ = range(min(ρ...), stop = max(ρ...), length = dims[1]);
+e_int = range(min(e_int...), stop = max(e_int...), length = dims[2]);
+q_tot = range(min(q_tot...), stop = max(q_tot...), length = dims[3]);
+
+# Suggested extreme ranges (based on discussions and existing tests):
+# 0 < ρ < 1.6
+# 0 < q_tot < 0.16902787156074564
+# -86590.81344095738 < e_int < 449782.0552978233
+# We don't reach quite the same ρ range, but all others are well tested
```
tapios, line 33: I'd remove the above commented lines.

docs/src/DevDocs/Common/Thermodynamics/saturation_adjustment.jl
```diff
@@ -0,0 +1,213 @@
...
+    println("Convergence percentages:")
+    for (k, v) in convergence_percent
+        println("$k = $v")
+    end
+
+    @warn "Note that the temperature axis for the non-converged
+    plot is not necessarily accurate, since the temperatures are
+    the result of a non-converged saturation adjustment. Additionally,
+    non-converged cases (resulting in NaNs) do not appear in the
+    non-converged plots"
```
tapios, line 212: Why don't you plot the true temperature instead? I think this would be more revealing w.r.t. where failure occurs.

src/Common/Thermodynamics/config_numerical_method.jl
```diff
@@ -82,8 +82,10 @@ function sa_numerical_method(
     _T_min::FT = T_min(param_set)
     q_pt = PhasePartition(q_tot, FT(0), q_tot) # Assume all ice
     T_2 = air_temperature(param_set, e_int, q_pt)
```
tapios, line 84: We can give rigorous bounds here:
```suggestion
    T_2 = T_0 +
           ( e_int + q.tot * e_int_i0 )  / cv_d
```
(This is not easily evaluated from the thermodynamics functions because we assume all ice in the numerator and dry air in the denominator, to rigorously bound the max.)

src/Common/Thermodynamics/config_numerical_method.jl
```diff
@@ -82,8 +82,10 @@ function sa_numerical_method(
...
-    T_1 = max(_T_min, air_temperature(param_set, e_int, PhasePartition(q_tot))) # Assume all vapor
-    T_2 = bound_upper_temperature(T_1, T_2)
+    T_1 = air_temperature(param_set, e_int, PhasePartition(q_tot)) # Assume all vapor
```
tapios, line 85: We can also give a rigorous bound here:
```suggestion
    T_1 = T_0 +
           ( e_int - q.tot * e_int_v0 )  / (cv_d + (cv_v - cv_d) * q.tot)  # Assume all vapor for energy, and all liquid for specific heat
```
(This likewise is not easily computed from the thermodynamics functions, because it assumes all vapor for the energy, and all liquid for the specific heat, to get a rigorous lower bound.)

src/Common/Thermodynamics/config_numerical_method.jl
```diff
@@ -82,8 +82,10 @@ function sa_numerical_method(
...
+    # Sometimes T_1 is slightly larger than the root.
+    # So, we subtract a small amount to ensure the root is bounded
+    T_1 = max(_T_min, T_1 - FT(0.1))
```
tapios, line 88: Don't need this with the change in l. 85:
```suggestion
```

docs/src/DevDocs/Common/Thermodynamics/saturation_adjustment.jl
charleskawczynski: We can plot true temperature, I'll adjust the temperature plot to use true temperature, and I'll remove this warning. One reason I wanted to still have internal energy was so that, if we could ensure convergence in a known 3D input space, we could use something like
```julia
L1 = e_int_min < e_int < e_int_max
L2 = ρ_min < ρ < ρ_max
L3 = q_tot_min < q_tot < q_tot_max
if !all((L1,L2,L3))
    error("Attempting to initialize an unphsical
        thermodynamic state with e_int=$e_int, ρ=$ρ, q_tot=$q_tot")
end
```
in `PhaseEquil`. At this point in the code, we don't have temperature.

src/Common/Thermodynamics/config_numerical_method.jl
charleskawczynski: When I tried doing this (and the suggestions below) but more iterations were needed. I forget how many, my thought is that this may require a bit of experimenting/thought.

docs/src/DevDocs/Common/Thermodynamics/saturation_adjustment.jl
tapios: I prefer not to have such hard bounds in the code. They have caused me much grief (and time wasted tracking them down in code) earlier in my career: What's unphysical on Earth is not necessarily unphysical on other planets, and it's nice to be able to simulate other planets.

src/Common/Thermodynamics/config_numerical_method.jl
tapios: Maybe it uses a few more iterations, because it is a rigorous bound (i.e., maximally wide). But if the point of regula falsi is to bracket a guaranteed solution, I think we have to use rigorous brackets. What's currently there does not always bracket the zero.

docs/src/DevDocs/Common/Thermodynamics/saturation_adjustment.jl
charleskawczynski: Ah, ok, that's fair. I'll change over to the true temperature plots entirely.

src/Common/Thermodynamics/config_numerical_method.jl
charleskawczynski: Actually, this fails even on the reference profiles (it's somehow not bracketing).
```julia
ERROR: LoadError: AssertionError: y0 * y1 < 0
Stacktrace:
 [1] find_zero(::ClimateMachine.Thermodynamics.var"#2#3"{EarthParameterSet,Float64,Float64,Float64,UnionAll}, ::RegulaFalsiMethod{Float64}, ::Float64, ::Float64, ::CompactSolution, ::ResidualTolerance{Float64}, ::Int64) at /Users/charliekawczynski/.julia/packages/RootSolvers/3Wfce/src/RootSolvers.jl:338
```

charleskawczynski: I'm fine with skipping the RegulaFalsi changes for now, and just updating the plots to include the reference profiles. Does this sound okay @tapios?

tapios: Why not including the rigorous bounds in RF? I don't want these to get lost (I didn't write down the derivation anywhere else.)

PR #1886, opened 2020-12-21
EKE power spectrum diagnostic for the baroclinic_wave test

LenkaNovak:
Adds the spectrum diagnostics group to the baroclinic wave test case. 

### Description

### DIFF
--- a/experiments/TestCase/baroclinic_wave.jl
+++ b/experiments/TestCase/baroclinic_wave.jl
@@ -21,6 +21,7 @@ using ClimateMachine.Thermodynamics:
     air_density, air_temperature, total_energy, internal_energy, PhasePartition
 using ClimateMachine.TurbulenceClosures
 using ClimateMachine.VariableTemplates
+using ClimateMachine.Spectra: compute_gaussian!
 
 using CLIMAParameters
 using CLIMAParameters.Planet: MSLP, R_d, day, grav, Omega, planet_radius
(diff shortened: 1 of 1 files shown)


experiments/TestCase/baroclinic_wave.jl
```diff
@@ -296,15 +297,36 @@ function config_diagnostics(FT, driver_config)
...
+        step = FT(1000), # in m
+    ))
+
+    #boundaries = [
+    #    FT(-90.0) FT(-180.0) _planet_radius
+    #    FT(90.0) FT(180.0) FT(_planet_radius + info.domain_height)
+    #]
+
+    #resolution = (FT(2), FT(2), FT(1000)) # in (deg, deg, m)
+
```
jiahe23, line 325: L319-324: Could these lines be deleted or do we want to keep it commented out as a record?

LenkaNovak: I kept it in in case someone didn't want to use the Gaussian grid :)

jiahe23: > I kept it in in case someone didn't want to use the Gaussian grid :)

haha that's how I figured~ sounds good

I'm running it now and will get the outputs/plots very soon :P

jakebolewski (approved):

jiahe23:
@LenkaNovak This current setup crashes at Day 17 while we want to plot the spectra at Day 30 (as in the reference). Shall we switch back to p_vert=5 and n_vert=4 ?

jiahe23:
> @LenkaNovak This current setup crashes at Day 17 while we want to plot the spectra at Day 30 (as in the reference). Shall we switch back to p_vert=5 and n_vert=4 ?

Or shall we plot the KE spectra for Day 15 instead (then we will expect signals from small scale wave lengths and this will be quite different from the Jablonowski's reference)?

LenkaNovak:
> > @LenkaNovak This current setup crashes at Day 17 while we want to plot the spectra at Day 30 (as in the reference). Shall we switch back to p_vert=5 and n_vert=4 ?
> 
> Or shall we plot the KE spectra for Day 15 instead (then we will expect signals from small scale wave lengths and this will be quite different from the Jablonowski's reference)?

I think we should just plot the spectra at the same times as the other diagnostics for now (the new QA notebook is setup that way). The numerics team wanted to run this for the split polynomial order, even if it is not stable. We could have a separate QA test that runs to day 30, but that doesn't seem to be required right now. We can chat offline tomorrow! :)

jiahe23:
> I think we should just plot the spectra at the same times as the other diagnostics for now (the new QA notebook is setup that way). 

👍 

> The numerics team wanted to run this for the split polynomial order, even if it is not stable. 

👍 

> We could have a separate QA test that runs to day 30, but that doesn't seem to be required right now. We can chat offline tomorrow! :)

Yes, this could wait.

jiahe23 (approved):

PR #1887, opened 2020-12-22
Attempt to speed up Dycoms: Use `Type{FT}`, lower diagnostics freq, cache O2 fluxes

charleskawczynski:
### Description

This PR is an attempt to speed up Dycoms a bit via:
 - Using `::Type{FT}` constraint in the driver, to ensure we avoid dynamic closures
 - Lower the diagnostics output frequency
 - Cache some of the 2nd order fluxes (both `turbulence_tensors` and some precipitation sources)

### DIFF
--- a/experiments/AtmosLES/dycoms.jl
+++ b/experiments/AtmosLES/dycoms.jl
@@ -247,7 +247,7 @@ function init_dycoms!(problem, bl, state, aux, localgeo, t)
 end
 
 function config_dycoms(
-    FT,
+    ::Type{FT},
     N,
     resolution,
     xmax,
@@ -255,7 +255,7 @@ function config_dycoms(
     zmax,
     moisture_model = "equilibrium",
     precipitation_model = "noprecipitation",
-)
+) where {FT}
     # Reference state
     T_profile = DecayingTemperatureProfile{FT}(param_set)
     ref_state = HydrostaticState(T_profile)
@@ -394,8 +394,9 @@ function config_dycoms(
     return config
 end
 
-function config_diagnostics(driver_config)
-    interval = "1000steps"
+function config_diagnostics(driver_config, timeend)
+    ssecs = cld(timeend, 2) + 10
+    interval = "$(ssecs)ssecs"
     dgngrp = setup_atmos_default_diagnostics(
         AtmosLESConfigType(),
         interval,
@@ -469,7 +470,7 @@ function main()
         init_on_cpu = true,
         Courant_number = Cmax,
     )
-    dgn_config = config_diagnostics(driver_config)
+    dgn_config = config_diagnostics(driver_config, timeend)
 
     if moisture_model == "equilibrium"
         filter_vars = ("moisture.ρq_tot",)
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -622,9 +622,12 @@ function transform_post_gradient_laplacian!(
 end
 
 function precompute(atmos::AtmosModel, args, tt::Flux{SecondOrder})
-    ts = recover_thermo_state(atmos, args.state, args.aux)
+    @unpack state, diffusive, aux, t = args
+    ts = recover_thermo_state(atmos, state, aux)
+    ν, D_t, τ = turbulence_tensors(atmos, state, diffusive, aux, t)
+    turbulence = (ν = ν, D_t = D_t, τ = τ)
     turbconv = precompute(atmos.turbconv, atmos, args, ts, tt)
-    return (; ts, turbconv)
+    return (; ts, turbconv, turbulence)
 end
 
 """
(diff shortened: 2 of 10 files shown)


trontrytel:
@jakebolewski, @charleskawczynski: I timed Dycoms with this branch and it runs at 3 hours! This is great.

trontrytel (approved):

experiments/AtmosLES/dycoms.jl
```diff
@@ -395,7 +395,7 @@ function config_dycoms(
 end
 
 function config_diagnostics(driver_config)
-    interval = "1000steps"
+    interval = "10000steps"
```
kpamnany, line 398: I see this sort of interval specification in quite a few experiments and it is problematic: change the poly order, the time stepper, the CFL, or any number of other things and the number of timesteps changes and you have to remember to change this also (and it's is easy to miss). I suggest passing `timeend` to this function and using something like:
```julia
ssecs = cld(timeend, 2) + 10
interval = "$(ssecs)ssecs"
```
Which should produce diagnostics output at `t = 0`, `t ~= timeend/2` and `t=timeend`.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -619,8 +619,14 @@ function transform_post_gradient_laplacian!(
...
-precompute(atmos::AtmosModel, args, ::Flux{SecondOrder}) =
-    (ts = recover_thermo_state(atmos, args.state, args.aux),)
+function precompute(atmos::AtmosModel, args, ::Flux{SecondOrder})
+    @unpack state, diffusive, aux, t = args
+    ν, D_t, τ = turbulence_tensors(atmos, state, diffusive, aux, t)
+    return (
+        ts = recover_thermo_state(atmos, state, aux),
+        turbulence = (ν = ν, D_t = D_t, τ = τ),
+    )
+end
```
kpamnany, line 628: Is there any experiment out there that _doesn't_ use these for second order flux computation?

kpamnany (approved):
LGTM.

kpamnany:
Looks like this needs a rebase+squash.

experiments/AtmosLES/dycoms.jl
charleskawczynski: Yikes, I did not know this. Will update.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: The `AtmosModel` hard-codes using `ViscousStress{Momentum}()` and `ViscousStress{Energy}()` (which was the case before the tendency specification layer was added). So `turbulence_tensors` is always needed/called. If/when we want to make this more flexible, we can change this to `turbulence = precompute(atmos.turbulence, atmos, args, Flux{SecondOrder})` and compute this, or not, depending on the turbulence model.

charleskawczynski: It looks like we can remove `ts` here for all cases except when the EDMF model is used. So we could use `ts = atmos.turbconv isa EDMF ? recover_thermo_state(atmos, state, aux) : NamedTuple()`.

src/Atmos/Model/precipitation.jl
```diff
@@ -65,6 +68,9 @@ vars_state(::RainModel, ::Prognostic, FT) = @vars(ρq_rai::FT)
 vars_state(::RainModel, ::Gradient, FT) = @vars(q_rai::FT)
 vars_state(::RainModel, ::GradientFlux, FT) = @vars(∇q_rai::SVector{3, FT})
 
+precompute(::RainModel, atmos::AtmosModel, args, ts, ::Source) =
+    (cache = warm_rain_sources(atmos, args, ts),)
```
charleskawczynski, line 72: @trontrytel what if we just did this?
```suggestion
precompute(::RainModel, atmos::AtmosModel, args, ts, ::Source) = warm_rain_sources(atmos, args, ts)
```

Then, below, we'd have something like
```julia
source(::WarmRain_1M{Energy}, m, args) = args.precomputed.precipitation.S_ρ_e
```

experiments/AtmosLES/dycoms.jl
charleskawczynski: Thanks!

trontrytel: Thanks! This is a great tip. I will update my other setups too.

src/Atmos/Model/precipitation.jl
trontrytel: Personally, it is not more readable to me. You have `args` as arguments everywhere and in the way you actually call the precomputed value. I would keep at least that name separate?

src/Atmos/Model/AtmosModel.jl
kpamnany: 😢 If we must.

charleskawczynski: I don't think we need to. I'm fine with keeping it as is

trontrytel:
bump?

PR #1891, opened 2020-12-22
Detect CI environments and download Artifacts to unique path prefixes

jakebolewski:
Prevents race conditions with multiple tests using the same Artifact
name through ArtifactWrappers running in parallel.

### Description


jakebolewski:
bors try

jakebolewski:
bors try

charleskawczynski (approved):
I like this solution! Thanks!

charleskawczynski:
I think this closes #1678, right?

jakebolewski:
Hmm maybe? I don't know if ArtifactWrappers can be (easily) be made process / thread safe.  I think it's easier to workaround this by using the artifacts definitions more in the way they are intended (static file checked in).  This is more of a workaround when there can be concurrency.

PR #1892, opened 2020-12-22
Cache some computations in turbconv

charleskawczynski:
### Description

This PR adds some computations to the cache in `turbconv`, in preparation to adding the tendency specification layer.

We get hit a bit hard with the formatter here, I may try to see if we can alleviate this.

### DIFF
--- a/test/Atmos/EDMF/closures/mixing_length.jl
+++ b/test/Atmos/EDMF/closures/mixing_length.jl
@@ -3,41 +3,35 @@
     mixing_length(
         m::AtmosModel{FT},
         ml::MixingLengthModel,
-        state::Vars,
-        diffusive::Vars,
-        aux::Vars,
-        t::Real,
+        args,
         Δ::Tuple,
         Et::Tuple,
-        ts,
+        ts_gm,
+        ts_en,
         env,
     ) where {FT}
 
 Returns the mixing length used in the diffusive turbulence closure, given:
  - `m`, an `AtmosModel`
  - `ml`, a `MixingLengthModel`
- - `state`, state variables
- - `diffusive`, additional variables
- - `aux`, auxiliary variables
- - `t`, the time
+ - `args`, the top-level arguments
  - `Δ`, the detrainment rate
  - `Et`, the turbulent entrainment rate
- - `ts`, NamedTuple of thermodynamic states
+ - `ts_gm`, grid-mean thermodynamic states
+ - `ts_en`, environment thermodynamic states
  - `env`, NamedTuple of environment variables
 """
 function mixing_length(
     m::AtmosModel{FT},
     ml::MixingLengthModel,
-    state::Vars,
-    diffusive::Vars,
-    aux::Vars,
-    t::Real,
+    args,
     Δ::Tuple,
     Et::Tuple,
-    ts,
+    ts_gm,
+    ts_en,
     env,
 ) where {FT}
-
+    @unpack state, aux, diffusive, t = args
     # TODO: use functions: obukhov_length, ustar, ϕ_m
 
     # Alias convention:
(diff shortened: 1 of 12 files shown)


test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -3,41 +3,35 @@
...
         m::AtmosModel{FT},
         ml::MixingLengthModel,
-        state::Vars,
-        diffusive::Vars,
-        aux::Vars,
-        t::Real,
+        args,
         Δ::Tuple,
         Et::Tuple,
         ts,
```
yairchn, line 9: maybe we should call the grid mean TS as "ts_gm"

test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -3,41 +3,35 @@
...
  - `ml`, a `MixingLengthModel`
- - `state`, state variables
- - `diffusive`, additional variables
- - `aux`, auxiliary variables
- - `t`, the time
+ - `args`, the top-level arguments
  - `Δ`, the detrainment rate
  - `Et`, the turbulent entrainment rate
- - `ts`, NamedTuple of thermodynamic states
+ - `ts`, grid-mean thermodynamic states
```
yairchn, line 20: here as well

test/Atmos/EDMF/closures/turbulence_functions.jl
```diff
@@ -54,19 +50,19 @@ function compute_buoyancy_gradients(
     _R_d::FT = R_d(m.param_set)
     _R_v::FT = R_v(m.param_set)
     ε_v::FT = 1 / molmass_ratio(m.param_set)
-    p = air_pressure(ts.gm)
+    p = air_pressure(ts)
```
yairchn, line 53: here as well

test/Atmos/EDMF/closures/turbulence_functions.jl
```diff
@@ -22,29 +22,25 @@ end
...
+ - `ts_en`, environment thermodynamic state
 """
 function compute_buoyancy_gradients(
     m::AtmosModel{FT},
-    state::Vars,
-    diffusive::Vars,
-    aux::Vars,
-    t::Real,
+    args,
     ts,
```
yairchn, line 40: here as well

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -698,6 +656,87 @@ function flux_first_order!(
...
+        bl,
+        bl.turbconv.mix_len,
+        args,
+        Δ_dyn,
+        E_trb,
+        ts,
+        ts_en,
+        env,
+    )
+    ρa_up = compute_ρa_up(bl, state, aux)
```
ilopezgp, line 685: `ρa_up` and `env` are precomputed for the first order fluxes. Is there a way to unpack their values here instead of computing them again?

test/Atmos/EDMF/closures/mixing_length.jl
charleskawczynski: Ah, good catch. I dropped this earlier because the cache uses NamedTuple and the symbols must match (e.g, `@unpack ts = precomputed`), but we can unpack without that syntax directly into a `ts_gm` alias.

test/Atmos/EDMF/edmf_kernels.jl
charleskawczynski: No, not easily. `flux_first_order!`, `flux_second_order!` and `source!` are separate calls, so they can't easily share a cache. We'd need for `flux_first_order!` (assuming that's the first one called) to return the cache and then pass it into `flux_second_order!` and `source!` (which would change the interface).

ilopezgp (approved):

ilopezgp:
bors try

PR #1881, opened 2020-12-21
Hydrostatic balance for DG-FV model

mwarusz:
hydrostatic balance for DG finite volume model

- Hydrostatic FV reconstruction on top of FV reconstruction for any reconstruction schemes
    - HBFVReconstruction struct
- Discrete hydrostatic balanced reference state 
    ```
     T_virt, p = atmos.ref_state.virtual_temperature_profile(atmos.param_set, z)
    aux.ref_state.p = p
    aux.ref_state.ρ = p / (T_virt * R_d(atmos.param_set))
    ```
    ```
     pᵢ₋₁ - pᵢ =  g ρᵢ Δzᵢ/2 + g ρᵢ₋₁ Δzᵢ₋₁/2
        for i = 2:nvertelem 
              ρᵢ  = (pᵢ₋₁ - pᵢ - g ρᵢ₋₁ Δzᵢ₋₁/2) / (g  Δzᵢ/2)
        end
    ```

  The **failure** is in the T_ref for the hydrostatic balance test.
  The change is in T_virt computation, 
     - in the master, we compute T_virt from the virtual_temperature_profile
     - in this version, we compute T_virt as T_virt  = p / (ρ R_d(atmos.param_set))
   since we have modified ρ in the correction, these two are different, which leads to the 10^-6 error in T_ref.

- Compute ∇Φ for DG-FV

Several tests are in 
- Hydrostatic balance unit tests:  test/Atmos/Model/discrete_hydrostatic_balance.jl 
- Hydrostatic balanced reconstruction unit tests:  test/Numerics/DGMethods/fv_reconstrution_test.jl 
- Hydrostatic balance GCM test: test/Numerics/DGMethods/Euler/fvm_balance_sphere.jl 
- Hydrostatic balance LES test: test/Numerics/DGMethods/Euler/fvm_balance_box.jl 
 

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #1896, opened 2020-12-26
Comput and use correct face auxiliary in FV vert_interface_tendency

Zhengyu-Huang:
This PR updates face auxiliary variables in FV kernel

In the present FV  implementation, we do not have nodal points at the top/bottom cell faces, 
but we need to compute flux at the top/bottom cell faces, therefore, we need to construct the face auxiliary variables.

The geopotential is constructed as 
```
Φᵢ₊₀.₅ =  Φᵢ + gΔzᵢ/2
```
The `aux.moisture.temperature`  is updated in `primitive_to_prognostic!`, which is called at the cell faces

The `prognostic_to_primitive!`, which is only called at the cell centers is updated, which uses the auxiliary variables computed from `update_auxiliary_state!`

This PR also requires using ```Δzᵢ```  as cell weights (same modification in  #1881 )

### Description

### DIFF
--- a/src/Atmos/Model/prog_prim_conversion.jl
+++ b/src/Atmos/Model/prog_prim_conversion.jl
@@ -15,20 +15,12 @@ variables `prim` for the atmos model `atmos`.
 !!! note
     The only field in `aux` required for this
     method is the geo-potential.
+    The other states in `aux` have been 
+    computed in update_auxiliary_state, 
+    which can be used to speed up the compuation
 """
-prognostic_to_primitive!(atmos::AtmosModel, prim::Vars, prog::Vars, aux) =
-    prognostic_to_primitive!(
-        atmos,
-        atmos.moisture,
-        prim,
-        prog,
-        Thermodynamics.internal_energy(
-            prog.ρ,
-            prog.ρe,
-            prog.ρu,
-            gravitational_potential(atmos.orientation, aux),
-        ),
-    )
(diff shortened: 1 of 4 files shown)


src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -93,14 +107,10 @@ function prognostic_to_primitive!(
...
-    e_int::AbstractFloat,
+    aux::Vars,
 )
-    q_pt = PhasePartition(
-        prog.moisture.ρq_tot / prog.ρ,
-        prog.moisture.ρq_liq / prog.ρ,
-        prog.moisture.ρq_ice / prog.ρ,
-    )
-    ts = PhaseNonEquil(atmos.param_set, e_int, prog.ρ, q_pt)
+    ts = recover_thermo_state(atmos, prog, aux)
```
charleskawczynski, line 112: I think this is wrong. `recover_thermo_state` is intended to assume that `aux.T` exists and is correctly populated.

charleskawczynski:
I think it would be simpler / safer to leave the `prim` <-> `prog` interface as is for now, and tackle that in a later PR if it's hurting performance.

Zhengyu-Huang:
> I think it would be simpler / safer to leave the `prim` <-> `prog` interface as is for now, and tackle that in a later PR if it's hurting performance.

That sounds good!

PR #1898, opened 2020-12-27
Fix temp BCs for SBL.

ilopezgp:
### Description

This PR simplifies the `surface_temperature_variation` function in the `stable_bl_model`, making it consistent with the description of these conditions in the original publication (Beare et al, 2006). Although analytically theta_liq, theta and T should be equivalent for a DryModel at the surface, I found that the T resulting from setting the BC for theta_liq is not what it is supposed to be. Its value oscillates around the value set by theta_liq. Since the variable of interest is T, the BC should be in this variable. 

### DIFF
--- a/experiments/AtmosLES/stable_bl_model.jl
+++ b/experiments/AtmosLES/stable_bl_model.jl
@@ -178,15 +178,7 @@ end
 
 function surface_temperature_variation(bl, state, t)
     FT = eltype(state)
-    ρ = state.ρ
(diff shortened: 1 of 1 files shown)


ilopezgp:
bors try

yairchn:
LGTM

yairchn:
LGTM

ilopezgp:
bors try

charleskawczynski (approved):

PR #1906, opened 2021-01-07
Add NodalStack utils and iterator

charleskawczynski:
### Description

This PR adds `NodalStack` with `iterate(::NodalStack)`. `iterate(::NodalStack)` traverses the nodal stack and returns a `NamedTuple` of prognostic, auxiliary, diffusive, and hyperdiffusive `Vars`, so that we can write a simple callback that calls each of the physics kernels (and `precompute` caches). For example:

```julia
    nodal_stack = NodalStack(solver_config; interp = true) # Allows for on-the-fly interpolating
    ρ = [local_states.prog.ρ for local_states in nodal_stack] # `local_states.prog` is a `Vars{vars_state()}()` instance.
    # ρ is a vector of values along the interpolated nodal stack (at `i=1,j=1`, optional kwargs to `NodalStack`)
```

Next, a cache diagnostics (to be included in a later PR) will be:

```julia
function cache_diagnostics(solver_config)
    bl = solver_config.dg.balance_law
    t = gettime(solver_config.solver)
    direction = solver_config.diffdir
    return [
        begin
            @unpack prog, aux, ∇flux, hyperdiff = local_states
            diffusive = ∇flux
            state = prog
            hyperdiffusive = hyperdiff

            _args_fx1 = (;state,aux,t,direction)
            _args_fx2 = (;state,aux,t,diffusive, hyperdiffusive)
            _args_src = (;state,aux,t,direction, diffusive)

            cache_fx1 = precompute(bl, _args_fx1, Flux{FirstOrder}())
            cache_fx2 = precompute(bl, _args_fx2, Flux{SecondOrder}())
            cache_src = precompute(bl, _args_src, Source())

            merge(cache_fx1, cache_fx2, cache_src)
        end for local_states in NodalStack(solver_config)
    ]
end
```

This has some nice long-term benefits:
 - Users (single-stack for now) will be less inclined to dump diagnostics fields into `aux`, which has been very common and will eventually need to be undone because doing so adds global memory reads/writes at every tendency evaluation.
 - The performance/debugging can be easily managed, from the driver level, by reducing the diagnostics frequency.
 - Changing what is incorporated in these diagnostics is even easier than adding new `aux` fields (no need to modify `vars_state(, ::Auxiliary, )`)

Edit: renamed `GridStack` -> `NodalStack`

### DIFF
--- a/src/Utilities/SingleStackUtils/SingleStackUtils.jl
+++ b/src/Utilities/SingleStackUtils/SingleStackUtils.jl
@@ -7,9 +7,11 @@ export get_vars_from_nodal_stack,
     reduce_nodal_stack,
     reduce_element_stack,
     horizontally_average!,
-    dict_of_nodal_states
+    dict_of_nodal_states,
+    NodalStack
 
 using OrderedCollections
+using UnPack
 using StaticArrays
 import KernelAbstractions: CPU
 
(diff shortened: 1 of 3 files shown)


src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -453,4 +454,258 @@ function dict_of_nodal_states(
...
+        j = 1,
+        interp = true,
+    )
+        states = States(prognostic, auxiliary, diffusive, hyperdiffusive)
+        vrange = 1:size(prognostic, 3)
+        N = polynomialorders(grid)
+        dim = dimensionality(grid)
+        Nq = N .+ 1
+        Nqk = dim == 2 ? 1 : Nq[dim]
+        Nk = dim == 2 ? 1 : N[dim]
```
kpamnany, line 542: Use [`basic_grid_info`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/DGMethods/DGModel.jl#L130) (and [`basic_topology_info`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/Mesh/Topologies.jl#L1333) if needed) instead.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -453,4 +454,258 @@ function dict_of_nodal_states(
...
+    N = polynomialorders(grid)
+    dim = dimensionality(grid)
+    Nq = N .+ 1
+    # Code assumes the same polynomial order in all horizontal directions
+    @inbounds begin
+        Nq1 = Nq[1]
+        Nq2 = Nq[2]
+        Nqk = dim == 2 ? 1 : Nq[dim]
+    end
+    Np = dofs_per_element(grid)
```
kpamnany, line 623: Ditto.

Also: is there a plan/expectation/need to support different polynomial orders in horizontal directions? If there is, then we shouldn't be adding code that doesn't support it, and if there isn't, then there's no need for the comment and why  do we have a way to do such a thing?

src/Utilities/SingleStackUtils/SingleStackUtils.jl
```diff
@@ -453,4 +454,258 @@ function dict_of_nodal_states(
...
+
+    n_vars_prog = size(prognostic, 2)
+    n_vars_aux = size(auxiliary, 2)
+    n_vars_diff = size(diffusive, 2)
+    n_vars_hd = size(hyperdiffusive, 2)
+
+    elemtobndy = convert(Array, grid.elemtobndy)
+    vmap⁻ = convert(Array, grid.vmap⁻)
+    vmap⁺ = convert(Array, grid.vmap⁺)
+    vgeo = convert(Array, grid.vgeo)
```
kpamnany, line 643: This makes this code very tightly bound to the internals of grids. It would be better to add interface functions in that module to support the functionality you need below than to do this.

kpamnany (approved):
Much of this functionality looks to me like it should be in Mesh. Most of SingleStackUtils is this way (much more so than Diagnostics at this point) and unless this is required urgently, it would be good to take the time to design some generalized helper functions to add into Mesh.

kpamnany:
@kmdeck: this isn't doing what you want, but it illustrates some of the complexities of grid access.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
charleskawczynski: Will do, thanks!

src/Utilities/SingleStackUtils/SingleStackUtils.jl
charleskawczynski: > This makes this code very tightly bound to the internals of grids.
Meaning the use of `elemtobndy`,`vmap⁻`,`vmap⁺`,`vgeo`? Or the `convert`?

> It would be better to add interface functions in that module to support the functionality you need below than to do this.

Providing an interface to access elements of `elemtobndy`,`vmap⁻`,`vmap⁺`,`vgeo`?

charleskawczynski:
> Much of this functionality looks to me like it should be in Mesh. Most of SingleStackUtils is this way (much more so than Diagnostics at this point) and unless this is required urgently, it would be good to take the time to design some generalized helper functions to add into Mesh.

I completely agree. @yairchn and @ilopezgp would really benefit from the debugging that this can/will provide. My inclination is to add for now and move/refactor later. It at least provides a concrete data point of desired functionality/interface.

src/Utilities/SingleStackUtils/SingleStackUtils.jl
kpamnany: Yes, the use of `elemtobndy` etc. And the interface should be to find out what you want to find out rather than to access the data structures to get the information. But as you say below, this provides a concrete data point for what's desired.

PR #1910, opened 2021-01-08
add diffusion in the updrafts at the lower levels

yairchn:
### Description
This draft PR adds updraft diffusion with an exponentially decaying diffusivity 
K=exp(-αz); the goal is for this diffusivity to transport vertically the updraft buoyancy from the first to the second level while having w=0 at the first model level. Currently it does not work in bomex, as the results without Smagorinsky are very bad. I would ask your opinion on why is it the case 

### DIFF
--- a/test/Atmos/EDMF/edmf_kernels.jl
+++ b/test/Atmos/EDMF/edmf_kernels.jl
@@ -1098,7 +1098,8 @@ function flux(::Diffusion{up_ρa{i}}, atmos, args) where {i}
     up_dif = diffusive.turbconv.updraft
     gm = state
     z = altitute(atmos, aux)
-    α = FT(5e-2)
+    # α = FT(5e-2)
+    α = atmos.turbconv.surface.α_surf
     K = exp(-α*z)
     ẑ = vertical_unit_vector(atmos, aux)
     return -K * up_dif.∇ρa[3] * ẑ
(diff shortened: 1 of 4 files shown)


test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -1074,6 +1090,63 @@ function flux(::Diffusion{en_ρatke}, atmos, args)
...
+function flux(::Diffusion{up_ρaw{i}}, atmos, args) where {i}
+    @unpack state, aux = args
+    @unpack ρa_up = args.precomputed.turbconv
+    FT = eltype(atmos)
+    up = state.turbconv.updraft
+    up_dif = diffusive.turbconv.updraft
+    gm = state
+    z = altitute(atmos, aux)
+    α = atmos.turbconv.surface.α_surf
+    K = exp(-α*z)
```
ilopezgp, line 1117: Can you centralize the definition of `K` in some file?

charleskawczynski: Sounds like a good idea. Alternatively, we could add it to `precompute` and unpack it (for `precompute(, ::Flux{SecondOrder})` only).

yairchn: I used a function 
K = surface_upd_diff(α,z)
which is defined in surface_functions.jl

PR #1911, opened 2021-01-09
fix a bug in FVM kernel

Zhengyu-Huang:
### Description

In the FVM kernel, 
When increment is off, and for periodic top/bottom case: 

the top element tendency is updated in 2 places, in the first place(ie = 1) it is set as = the second place(ie=nelem) should be updated as +=

### DIFF
--- a/src/Numerics/DGMethods/DGFVModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGFVModel_kernels.jl
@@ -590,7 +590,7 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
                             else
                                 T = local_tendency[s]
                             end
-                            tendency[n, s, eH + eV_up] = T
+                            tendency[n, s, eH + eV_up] += T
                         end
                     end
                 else
--- a/src/Numerics/DGMethods/DGFVModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGFVModel_kernels.jl
@@ -583,14 +583,8 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
                         if increment
                             tendency[n, s, eH + eV_up] += local_tendency[s]
                         else
-                            if β != 0
-                                T =
-                                    local_tendency[s] +
-                                    β * tendency[n, s, eH + eV_up]
-                            else
-                                T = local_tendency[s]
-                            end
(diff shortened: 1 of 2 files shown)


Zhengyu-Huang:
bors try

src/Numerics/DGMethods/DGFVModel_kernels.jl
```diff
@@ -590,7 +590,7 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
                             else
                                 T = local_tendency[s]
                             end
-                            tendency[n, s, eH + eV_up] = T
+                            tendency[n, s, eH + eV_up] += T
```
mwarusz, line 593: Good catch. Does this mean that we also shouldn't scale by beta again here ? So the whole thing can be simplified to just `tendency += local_tendency` regardless of `increment` ?

Zhengyu-Huang: What is beta? modify to tendency += local_tendency

jkozdon: `beta` is a scaling factor for the previous tendency. See the description here:
https://github.com/CliMA/ClimateMachine.jl/blob/c693e9e7d85ceff7ba55f71ebdf3ee577d36dc71/src/Numerics/DGMethods/DGModel.jl#L189-L202

jkozdon (commented):
Can we add a test to make sure this gets hit? Must not have before I guess?

src/Numerics/DGMethods/DGFVModel_kernels.jl
```diff
@@ -583,14 +583,8 @@ A finite volume reconstruction is used to construction `Fⁱⁿᵛ⋆`
...
-                            if β != 0
-                                T =
-                                    local_tendency[s] +
-                                    β * tendency[n, s, eH + eV_up]
-                            else
-                                T = local_tendency[s]
-                            end
-                            tendency[n, s, eH + eV_up] = T
+                            # This tendency has already been updated in the first element (eV_up=1)
+                            tendency[n, s, eH + eV_up] += local_tendency[s]
```
jkozdon, line 587:
```julia
                            # This tendency has already been updated in the first element (eV_up=1)
                            tendency[n, s, eH + eV_up] += local_tendency[s]
```

PR #1915, opened 2021-01-11
Support for kernels on subsets of variables

simonbyrne:
### Description

Allows use of MPIStateArrays with different state variables then defined in the kernel. Should make it easier to operate on subsets of state.


simonbyrne:
bors try

src/BalanceLaws/state_types.jl
```diff
@@ -38,6 +40,12 @@ solved for by the ODE solver.
 """
 struct Prognostic <: AbstractStateType end
 
+struct PrognosticIn <: AbstractStateType end
+struct PrognosticOut <: AbstractStateType end
```
charleskawczynski, line 44:
```suggestion
abstract type AbstractStreamType end
struct In <: AbstractStreamType end
struct Out <: AbstractStreamType end
struct InOut <: AbstractStreamType end
struct Prognostic{T<: AbstractStreamType} <: AbstractStateType end
Prognostic() = Prognostic{InOut}()
```
Would something like this be simpler if we plan to extend this to other `AbstractStateType`'s?

simonbyrne: Possibly. There is the chance we might want them to be specific to each kernel as well.

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

PR #1916, opened 2021-01-11
refine hyperdiffusion in turbulence closure

jiahe23:
### Description
Co-authored with @charleskawczynski 

This PR refines the biharmonics formula in `TurbulenceClosures.jl` and removes the duplicates block of codes between `DryBiharmincs` and `EquilMoistBiharmonics`. 


src/Common/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -799,38 +800,63 @@ struct NoHyperDiffusion <: HyperDiffusion end
 hyperviscosity_tensors(m::HyperDiffusion, bl::BalanceLaw, args...) =
     hyperviscosity_tensors(m, bl.orientation, bl.param_set, args...)
 
+
+abstract type HyperdiffMoistureModel <: BalanceLaw end
+struct HyperdiffDryModel <: HyperdiffMoistureModel end
+struct HyperdiffEquilMoist{FT} <: HyperdiffMoistureModel
+    τ_timescale_q_tot::FT
+end
+struct HyperdiffNonEquilMoist <: HyperdiffMoistureModel end
```
charleskawczynski, line 809: Overall the PR looks good to me. The only part that I'm not super happy about is here, which is a bit of a trade-off. We're kind of duplicating types so that we don't have to duplicate kernels. Seems like a fine tradeoff, but I added @kpamnany to review.

kpamnany: I don't quite see why `HyperdiffMoistureModel` is a `BalanceLaw` so I guess I don't understand the problem. Will talk about it.

charleskawczynski: If it's not a `BalanceLaw`, then we need to define `vars_state` for all empty cases (e.g., `vars_state(:: HyperdiffEquilMoist, ::Auxiliary, FT) = @vars()`), this was the only reason.

kpamnany: I think it's better to do that than to make it a `BalanceLaw` for convenience. At some point, we might want a different base type for `Atmos` components?

charleskawczynski: Either way is fine to me

PR #1917, opened 2021-01-12
Add single stack diagnostics, replace dons in some experiments

charleskawczynski:
### Description

This PR adds `single_stack_diagnostics`, which returns an array of entries. Unlike `dict_of_nodal_states`, which returns a dictionary of arrays, `single_stack_diagnostics` returns an array of (basically) dictionaries. This may not be super-performant, but allows us to contain all of the `Vars` instances in each entry, so that we can compute and also include all quantities returned from, for example, `precompute()`--which we hope to be helpful for debugging. The other advantage of adding this is that we'll be able to (in the single stack configurations, at least) stop using aux for diagnostics, which results in global memory reads, and performs unnecessary computations at each RHS evaluation.

A next step might be splitting `precompute` (in `BalanceLaws`) into `precompute` and `diagnostics`. `precompute` can compute all quantities needed to solve the system of equations, and `diagnostics` (which can work in exactly the same way), can compute additional diagnostics quantities.

### DIFF
--- a/src/Utilities/VariableTemplates/flattened_tup_chain.jl
+++ b/src/Utilities/VariableTemplates/flattened_tup_chain.jl
@@ -1,4 +1,4 @@
-export flattened_tup_chain
+export flattened_tup_chain, flattened_named_tuple
 
 flattened_tup_chain(::Type{NamedTuple{(), Tuple{}}}; prefix = (Symbol(),)) = ()
 flattened_tup_chain(::Type{T}; prefix = (Symbol(),)) where {T <: Real} =
@@ -9,6 +9,7 @@ flattened_tup_chain(
     ::Type{T};
     prefix = (Symbol(),),
 ) where {T <: SHermitianCompact} = (prefix,)
+flattened_tup_chain(::Type{T}; prefix = (Symbol(),)) where {T} = (prefix,)
 
 """
     flattened_tup_chain(::Type{T}) where {T <: Union{NamedTuple,NTuple}}
@@ -33,3 +34,53 @@ function flattened_tup_chain(
     Iterators.flatten |>
     collect
 end
+flattened_tup_chain(::AbstractVars{S}) where {S} = flattened_tup_chain(S)
+
+"""
+    flattened_named_tuple(v::AbstractVars)
+
+A flattened NamedTuple, given a `Vars` instance.
+
+# Example:
+
+```julia
+using Test
+using ClimateMachine.VariableTemplates
+nt = (x = 1, a = (y = 2, z = 3, b = ((a = 1,), (a = 2,), (a = 3,))));
+fnt = flattened_named_tuple(nt);
+@test keys(fnt) == (:x, :a_y, :a_z, :a_b_1_a, :a_b_2_a, :a_b_3_a)
+@test length(fnt) == 6
+@test fnt.x == 1
+@test fnt.a_y == 2
+@test fnt.a_z == 3
+@test fnt.a_b_1_a == 1
+@test fnt.a_b_2_a == 2
+@test fnt.a_b_3_a == 3
+```
+"""
+function flattened_named_tuple end
+
+function flattened_named_tuple(v::AbstractVars)
+    ftc = flattened_tup_chain(v)
+    keys_ = Symbol.(join.(ftc, :_))
+    vals = map(x -> getproperty(v, x), ftc)
+    return (; zip(keys_, vals)...)
+end
+flattened_named_tuple(v::Nothing) = NamedTuple()
+
+function flattened_named_tuple(nt::NamedTuple)
+    ftc = flattened_tup_chain(typeof(nt))
+    keys_ = Symbol.(join.(ftc, :_))
+    vals = flattened_nt_vals(nt)
+    return (; zip(keys_, vals)...)
+end
+
+flattened_nt_vals(a::NamedTuple) = flattened_nt_vals(Tuple(a))
+flattened_nt_vals(a::NamedTuple{(), Tuple{}}) = (nothing,)
+flattened_nt_vals(a) = (a,)
+flattened_nt_vals(a::NamedTuple, b...) =
+    tuple(flattened_nt_vals(a)..., flattened_nt_vals(b...)...)
+flattened_nt_vals(a::NamedTuple{(), Tuple{}}, b...) =
+    tuple(nothing, flattened_nt_vals(b...)...)
+flattened_nt_vals(a, b...) = tuple(values(a), flattened_nt_vals(b...)...)
+flattened_nt_vals(x::Tuple) = flattened_nt_vals(x...)
--- a/test/Atmos/EDMF/report_mse_bomex.jl
+++ b/test/Atmos/EDMF/report_mse_bomex.jl
@@ -33,6 +33,24 @@ function test_mse(computed_mse, best_mse, key)
     mse_not_regressed || @show key
 end
 
+#! format: off
+dons(diag_vs_z) = Dict(
+    "z" => [ca.z for ca in diag_vs_z],
+    "ρ" => [ca.prog.ρ for ca in diag_vs_z],
+    "ρu[1]" => [ca.prog.ρu[1] for ca in diag_vs_z],
+    "moisture.ρq_tot" => [ca.prog.moisture.ρq_tot for ca in diag_vs_z],
+    "turbconv.updraft[1].ρa" => [ca.prog.turbconv.updraft[1].ρa for ca in diag_vs_z],
+    "turbconv.updraft[1].ρaw" => [ca.prog.turbconv.updraft[1].ρaw for ca in diag_vs_z],
+    "turbconv.updraft[1].ρaθ_liq" => [ca.prog.turbconv.updraft[1].ρaθ_liq for ca in diag_vs_z],
+    "turbconv.updraft[1].ρaq_tot" => [ca.prog.turbconv.updraft[1].ρaq_tot for ca in diag_vs_z],
+    "turbconv.environment.ρatke" => [ca.prog.turbconv.environment.ρatke for ca in diag_vs_z],
+    "turbconv.environment.ρaθ_liq_cv" => [ca.prog.turbconv.environment.ρaθ_liq_cv for ca in diag_vs_z],
+    "turbconv.environment.ρaq_tot_cv" => [ca.prog.turbconv.environment.ρaq_tot_cv for ca in diag_vs_z],
+)
+#! format: on
+
+dons_arr = [dons(diag_vs_z) for diag_vs_z in diag_arr]
+
 computed_mse = Dict(
     k => compute_mse(
         solver_config.dg.grid,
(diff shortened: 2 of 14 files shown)


src/Utilities/VariableTemplates/flattened_tup_chain.jl
```diff
@@ -33,3 +34,54 @@ function flattened_tup_chain(
     Iterators.flatten |>
     collect
 end
+flattened_tup_chain(::AbstractVars{S}) where {S} = flattened_tup_chain(S)
+
+export flattened_named_tuple
```
kpamnany, line 39: Please move all `export` statements to the top of the module defining file (`VariableTemplates.jl` in this case).

src/Utilities/VariableTemplates/flattened_tup_chain.jl
```diff
@@ -33,3 +34,54 @@ function flattened_tup_chain(
...
+A flattened NamedTuple, given a `Vars` instance.
+
+# Example:
+
+```julia
+using Test
+using ClimateMachine.VariableTemplates
+nt = (x = 1, a = (y = 2, z = 3, b = ((a = 1,), (a = 2,), (a = 3,))));
+fnt = flattened_named_tuple(nt);
+@test keys(fnt) == (:x, :a_y, :a_z, :a_b_1_a, :a_b_2_a, :a_b_3_a)
```
kpamnany, line 52: This is demonstrating support for a regular tuple mixed in with `NamedTuple`s? But `Vars` doesn't have those, does it?

test/Atmos/EDMF/report_mse_bomex.jl
```diff
@@ -33,6 +33,24 @@ function test_mse(computed_mse, best_mse, key)
...
+    "ρ" => [ca.prog.ρ for ca in cache_vs_z],
+    "ρu[1]" => [ca.prog.ρu[1] for ca in cache_vs_z],
+    "moisture.ρq_tot" => [ca.prog.moisture.ρq_tot for ca in cache_vs_z],
+    "turbconv.updraft[1].ρa" => [ca.prog.turbconv.updraft[1].ρa for ca in cache_vs_z],
+    "turbconv.updraft[1].ρaw" => [ca.prog.turbconv.updraft[1].ρaw for ca in cache_vs_z],
+    "turbconv.updraft[1].ρaθ_liq" => [ca.prog.turbconv.updraft[1].ρaθ_liq for ca in cache_vs_z],
+    "turbconv.updraft[1].ρaq_tot" => [ca.prog.turbconv.updraft[1].ρaq_tot for ca in cache_vs_z],
+    "turbconv.environment.ρatke" => [ca.prog.turbconv.environment.ρatke for ca in cache_vs_z],
+    "turbconv.environment.ρaθ_liq_cv" => [ca.prog.turbconv.environment.ρaθ_liq_cv for ca in cache_vs_z],
+    "turbconv.environment.ρaq_tot_cv" => [ca.prog.turbconv.environment.ρaq_tot_cv for ca in cache_vs_z],
```
kpamnany, line 48: This is performing 11 passes through `cache_vs_z` when it could use just 1 with 11 `push!`s...

test/Atmos/EDMF/stable_bl_single_stack_implicit.jl
```diff
@@ -2,6 +2,7 @@ using ClimateMachine
 using ClimateMachine.SystemSolvers
 using ClimateMachine.ODESolvers
 using ClimateMachine.MPIStateArrays
+@show ClimateMachine.array_type()
```
kpamnany, line 5: Maybe output this in the `@info` where the `DriverConfiguration` is created?

src/Utilities/VariableTemplates/flattened_tup_chain.jl
kpamnany: Ah, EDMF does not use names for the different updrafts?

src/Diagnostics/single_stack_cache_diagnostics.jl
```diff
@@ -0,0 +1,74 @@
...
+            _args_fx2 = (; state, aux, t, diffusive, hyperdiffusive)
+            _args_src = (; state, aux, t, direction, diffusive)
+
+            cache_fx1 = precompute(bl, _args_fx1, Flux{FirstOrder}())
+            cache_fx2 = precompute(bl, _args_fx2, Flux{SecondOrder}())
+            cache_src = precompute(bl, _args_src, Source())
+
+            # cache_fx1, cache_fx2, and cache_src have overlapping
+            # data, only need one copy, so merge:
+            cache = merge(cache_fx1, cache_fx2, cache_src)
```
kpamnany, line 42: Maybe better if we introduce a common parent type for `Source` et al. which can be used by a `precompute` implementation that doesn't repeat computations?

kpamnany (commented):
If these are debugging diagnostics that will be used a lot going forward, it would be better to build it in the standard diagnostics group framework rather than hacking it in this way? I see that the `report_mse` code is doing some validation of the output in one case, but it could simply open the NC file to do that. Otherwise, why not leave it in `SingleStackUtils`, which we know needs a close look and refactoring when EDMF moves into `src/`?

This code looks expensive, from `NodalStack` to the output arrays (`dons` is being created too!) and we should be able to leverage (with some tweaking and additions) the `dump_state`/`dump_aux` groups instead. We will be building a tendencies output diagnostics group shortly also, so we should be able to get all this functionality within the existing design.

src/Diagnostics/single_stack_cache_diagnostics.jl
charleskawczynski: This would be tricky/involved because the cache would need to persist through `flux_first_order!`, `flux_second_order!` and `source!`. I think repeating the computations is fine since the frequency of the diagnostic is typically much lower than the frequency of the RHS evaluation.

src/Utilities/VariableTemplates/flattened_tup_chain.jl
charleskawczynski: > But Vars doesn't have those, does it?

It does, updrafts are indexed as tuples.

test/Atmos/EDMF/report_mse_bomex.jl
charleskawczynski: Hmm, it looks like the dynamic allocations from `push!` are more costly:
```julia
julia> # let

       #! format: off
       function dons_single_pass(cache_vs_z)
           dons_sp = Dict()
           FT = typeof(cache_vs_z[1].z)
           dons_sp["z"] = FT[]
           dons_sp["ρ"] = FT[]
           dons_sp["ρu[1]"] = FT[]
           dons_sp["moisture.ρq_tot"] = FT[]
           dons_sp["turbconv.updraft[1].ρa"] = FT[]
           dons_sp["turbconv.updraft[1].ρaw"] = FT[]
           dons_sp["turbconv.updraft[1].ρaθ_liq"] = FT[]
           dons_sp["turbconv.updraft[1].ρaq_tot"] = FT[]
           dons_sp["turbconv.environment.ρatke"] = FT[]
           dons_sp["turbconv.environment.ρaθ_liq_cv"] = FT[]
           dons_sp["turbconv.environment.ρaq_tot_cv"] = FT[]
           for ca in cache_vs_z
               push!(dons_sp["z"], ca.z)
               push!(dons_sp["ρ"], ca.prog.ρ)
               push!(dons_sp["ρu[1]"], ca.prog.ρu[1])
               push!(dons_sp["moisture.ρq_tot"], ca.prog.moisture.ρq_tot)
               push!(dons_sp["turbconv.updraft[1].ρa"], ca.prog.turbconv.updraft[1].ρa)
               push!(dons_sp["turbconv.updraft[1].ρaw"], ca.prog.turbconv.updraft[1].ρaw)
               push!(dons_sp["turbconv.updraft[1].ρaθ_liq"], ca.prog.turbconv.updraft[1].ρaθ_liq)
               push!(dons_sp["turbconv.updraft[1].ρaq_tot"], ca.prog.turbconv.updraft[1].ρaq_tot)
               push!(dons_sp["turbconv.environment.ρatke"], ca.prog.turbconv.environment.ρatke)
               push!(dons_sp["turbconv.environment.ρaθ_liq_cv"], ca.prog.turbconv.environment.ρaθ_liq_cv)
               push!(dons_sp["turbconv.environment.ρaq_tot_cv"], ca.prog.turbconv.environment.ρaq_tot_cv)
           end
           return dons_sp
       end
dons_single_pass (generic function with 1 method)

julia> let
       #! format: on

       @btime dons_arr = [dons(cache_vs_z) for cache_vs_z in cache_arr]
       @btime dons_arr = [dons_single_pass(cache_vs_z) for cache_vs_z in cache_arr]
       nothing
       end
  29.269 μs (200 allocations: 108.81 KiB)
  515.792 μs (10727 allocations: 441.56 KiB)
```
FWIW, this is only done once at the end of the simulation to compare with our ground truth.

charleskawczynski:
Hmm, changing some names...

charleskawczynski:
> If these are debugging diagnostics that will be used a lot going forward, it would be better to build it in the standard diagnostics group framework rather than hacking it in this way?

I didn't want to add diagnostics while you were in the middle of a re-vamp. Also, I'm not sure how to handle the updrafts in the existing diagnostics (the tuples of `Vars`).

> I see that the report_mse code is doing some validation of the output in one case, but it could simply open the NC file to do that.

Yeah, this is kind of multi-purposing things. I wanted to avoid dumping too much stuff to disc since this is utilized in CI. Also, I figure that making some of these data structures available in the REPL is helpful for the edmf team while they're fixing/improving the model.

> Otherwise, why not leave it in SingleStackUtils, which we know needs a close look and refactoring when EDMF moves into src/?

Leave what in `SingleStackUtils`? The `single_stack_diagnostics`? I'm fine moving `single_stack_diagnostics` into `SingleStackUtils` if you think that makes sense.

> This code looks expensive, from NodalStack to the output arrays (dons is being created too!) and we should be able to leverage (with some tweaking and additions) the dump_state/dump_aux groups instead. We will be building a tendencies output diagnostics group shortly also, so we should be able to get all this functionality within the existing design.

I agree that it looks expensive, but the single-stack configuration is a small domain, and `bomex_edmf.jl` runs in about the same time on my laptop. The `dons` is only computed once at the end. Also, from a scalability POV, this allows us to remove auxiliary fields in `turbconv`, which only exist for diagnostics / visualization. Ultimately, it's far easier to drop/replace the diagnostics callback than remove the existing dependence on the aux fields. My suggestion would be to follow through with this until proper diagnostics are ready, it should be easy to replace.

If adding this functionality can be available in the next day or so, I'm fine with waiting, but I know that the edmf team has been waiting, and is ready now, to use this. Their existing workflow is very limited by the amount of available information from a given simulation, and the only existing way to get more information is to dump things into the aux state, which is a step in the wrong direction IMO (since it will eventually need to be undone).

I'm about to meet with some of the edmf team, let's discuss this later today.

kpamnany (approved):

PR #1918, opened 2021-01-12
Move random perturbations to LES experiment files

ilopezgp:
### Description

The random perturbations needed to break symmetry are moved to the LES experiment files, the only case when they are needed. Furthermore, the random perturbations are now centered around zero.

### DIFF
--- a/experiments/AtmosLES/bomex_les.jl
+++ b/experiments/AtmosLES/bomex_les.jl
@@ -1,4 +1,14 @@
+using Random
 include("bomex_model.jl")
+
+function add_perturbations!(state, localgeo)
+    FT = eltype(state)
+    z = localgeo.coord[3]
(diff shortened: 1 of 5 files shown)


yairchn (approved):
great idea

ilopezgp:
bors try

ilopezgp:
Screen Shot 2021-01-12 at 2 48 46 PM

yairchn:
-7% in the TKE use , is that a lot ? do we understand why ?

ilopezgp:
> -7% in the TKE use , is that a lot ? do we understand why ?

I would say the former one was correct for the wrong reasons. There is also a 14% improvement in qt. The issue is that we were adding energy at every level below 400 m.

charleskawczynski (approved):

PR #1921, opened 2021-01-12
Traverse all prog vars in top-level `flux_first_order!`, `flux_second_order!`, and `source!`

charleskawczynski:
### Description

This PR: introduces `get_prog_state` (there's probably a better name out there) to `BalanceLaws`, and is defined in `ClimateMachine.Atmos` for all prognostic variables. This allows us to:
 - Loop over all prognostic variables in AtmosModel's `flux_first_order!`, `flux_second_order!` and `source!` without having to call `flux_first_order!`, `flux_second_order!` and `source!` in AtmosModel's submodels (and therefore we can delete all of the sub-model implementations).
 - Move the pads (`flux_pad` / `vec_pad`) inside `Σfluxes` / `Σsources`, which is done by adding the prognostic variable type as an argument to `Σfluxes` / `Σsources`, and using dispatch based on `PrognosticVariable`, `AbstractMomentum`, and `AbstractTracers`.

One nice aspect of this PR is that, with this PR, the tendency table will be automatically synchronized with what is computed in the model. Previously, we could have accidentally missed (or messed up) calling, for example, `flux_second_order!(atmos.turbconv, ...)`. In this case, the table would have said the turbconv submodel fluxes were included, but in reality, they were not. Now, that cannot happen because `show_tendencies`, `flux_first_order!`, `flux_second_order!` and `source!` all rely on `prognostic_vars` to traverse the prognostic variables.


charleskawczynski:
The second commit works on the CPU, but fails on the GPU --dynamic function invocation (call to overdub). I suspect it's an issue with the padding, but I haven't spent any time looking into it. Would like to get feedback before jumping into the weeds.

charleskawczynski:
Note that the first commit, which has effectively the same complexity as the second commit, works on the GPU and should pass CI.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -511,16 +511,14 @@ equations.
...
-    flux_first_order!(atmos.tracers, atmos, flux, args)
-    flux_first_order!(atmos.turbconv, atmos, flux, args)
 
+    pvs = prognostic_vars(atmos)
+    ntuple(Val(length(pvs))) do i
+        prog = pvs[i]
+        prog_flux = get_prog_state(flux, prog)
+        val = Σfluxes(eq_tends(prog, atmos, tend), atmos, args) .* flux_pad
+        setproperty!(prog_flux[1], prog_flux[2], val)
+    end
```
simonbyrne, line 521: Does this work?
```suggestion
    map(pvs) do prog
        prog_flux = get_prog_state(flux, prog)
        val = Σfluxes(eq_tends(prog, atmos, tend), atmos, args) .* flux_pad
        setproperty!(prog_flux[1], prog_flux[2], val)
    end
```

simonbyrne:
bors try

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Haven't tried, but I can. The very first commit worked (on the GPU), so it might be good to experiment off of that one.

charleskawczynski:
Looks like some of the padding was not done properly (`flux_pad` and `src_pad`), which handle cases where there are/are not fluxes/sources for scalar/vector equations.

charleskawczynski:
That very well may be why the gpu jobs are failing

simonbyrne:
Why do we need `flux_pad`?

charleskawczynski:
We could get rid of it. But we'd need to somehow update `Σfluxes`:
```julia
function Σfluxes(
    fluxes::NTuple{N, TendencyDef{Flux{O}, PV}},
    args...,
) where {N, PV, O}
    return ntuple_sum(ntuple(Val(N)) do i
        flux(fluxes[i], args...)
    end)
end
Σfluxes(fluxes::Tuple{}, args...) = 0            # <------- If we have no fluxes, this needs to return the right size, but we don't know which equation we're summing fluxes for, so it could be a scalar or a vector.
```

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
> We could get rid of it. But we'd need to somehow update `Σfluxes`:
> 
> ```julia
> function Σfluxes(
>     fluxes::NTuple{N, TendencyDef{Flux{O}, PV}},
>     args...,
> ) where {N, PV, O}
>     return ntuple_sum(ntuple(Val(N)) do i
>         flux(fluxes[i], args...)
>     end)
> end
> Σfluxes(fluxes::Tuple{}, args...) = 0            # <------- If we have no fluxes, this needs to return the right size, but we don't know which equation we're summing fluxes for, so it could be a scalar or a vector.
> ```

The simple solution here was to pass the prognostic variable to Σfluxes / Σsources, and dispatch on `BalanceLaws.PrognosticVariable` vs `BalanceLaws.AbstractMomentum`, which is turning out to be really helpful. Thanks @kpamnany!

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
charleskawczynski: `map` seems to be working!

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
It looks like CI is passing for all except [this ocean test](https://buildkite.com/clima/climatemachine-ci/builds/1313#04d2e670-cfb7-4cb8-80d1-c1e5e6676cf5/279-282), but I can't reproduce the error locally (neither on CPU nor GPU). AFAICT, the changes introduced here should not affect the ocean test, I suspect it's unrelated to this PR.

charleskawczynski:
@yairchn, @ilopezgp, this undoes the `if-else` that effectively removes EDMF moisture variable tendencies. So, we'll want to put this logic back in (maybe just via `eq_tends`) in a separate PR (CI is still passing with these changes).

charleskawczynski:
> It looks like CI is passing for all except [this ocean test](https://buildkite.com/clima/climatemachine-ci/builds/1313#04d2e670-cfb7-4cb8-80d1-c1e5e6676cf5/279-282), but I can't reproduce the error locally (neither on CPU nor GPU). AFAICT, the changes introduced here should not affect the ocean test, I suspect it's unrelated to this PR.

Now we're seeing the same error in a [different test](https://buildkite.com/clima/climatemachine-ci/builds/1314#ec276310-9efa-47d2-a046-85efc9e050b1/279-299).

src/Atmos/Model/AtmosModel.jl
```diff
@@ -537,21 +537,17 @@ equations.
...
-    flux_first_order!(atmos.energy, atmos, flux, args)
-    flux_first_order!(atmos.moisture, atmos, flux, args)
-    flux_first_order!(atmos.precipitation, atmos, flux, args)
-    flux_first_order!(atmos.tracers, atmos, flux, args)
-    flux_first_order!(atmos.turbconv, atmos, flux, args)
 
+    pvs = prognostic_vars(atmos)
+    map(pvs) do prog
+        prog_flux = get_prog_state(flux, prog)
+        prop = getproperty(prog_flux[1], prog_flux[2])
```
charleskawczynski, line 547:
```suggestion
```

charleskawczynski: And similar elsewhere

src/Atmos/Model/AtmosModel.jl
```diff
@@ -537,21 +537,15 @@ equations.
...
-    flux.ρu =
-        Σfluxes(eq_tends(Momentum(), atmos, tend), atmos, args) .* tens_pad
-
-    flux_first_order!(atmos.energy, atmos, flux, args)
-    flux_first_order!(atmos.moisture, atmos, flux, args)
-    flux_first_order!(atmos.precipitation, atmos, flux, args)
-    flux_first_order!(atmos.tracers, atmos, flux, args)
-    flux_first_order!(atmos.turbconv, atmos, flux, args)
 
+    map(prognostic_vars(atmos)) do prog
```
simonbyrne, line 544: `foreach` might be a better choice here?

```suggestion
    foreach(prognostic_vars(atmos)) do prog
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -682,23 +676,15 @@ function. Contributions from subcomponents are then assembled (pointwise).
...
-
-    flux_second_order!(atmos.energy, flux, atmos, args)
-    flux_second_order!(atmos.moisture, flux, atmos, args)
-    flux_second_order!(atmos.precipitation, flux, atmos, args)
-    flux_second_order!(atmos.tracers, flux, atmos, args)
-    flux_second_order!(atmos.turbconv, flux, atmos, args)
+    foreach(prognostic_vars(atmos)) do prog
+        prog_flux = get_prog_state(flux, prog)
+        val = Σfluxes(prog, eq_tends(prog, atmos, tend), atmos, args)
+        setproperty!(prog_flux[1], prog_flux[2], val)
```
simonbyrne, line 686:
```suggestion
        var, name = get_prog_state(flux, prog)
        val = Σfluxes(prog, eq_tends(prog, atmos, tend), atmos, args)
        setproperty!(var, name, val)
```

src/Atmos/Model/get_prognostic_vars.jl
```diff
@@ -27,4 +27,15 @@ prognostic_vars(m::AtmosModel) = (
     prognostic_vars(m.turbconv)...,
 )
 
+get_prog_state(state, ::Mass) = (state, :ρ)
```
simonbyrne, line 30: This needs a docstring.

src/Atmos/Model/AtmosModel.jl
simonbyrne: Or even if we could combine it into a single function call?:
```suggestion
        Σfluxes!(flux, prog, eq_tends(prog, atmos, tend), atmos, args)
```
this might make it easier, as `Σfluxes!` can figure out the shape from the variable

simonbyrne:
> > 
> 
> Now we're seeing the same error in a [different test](https://buildkite.com/clima/climatemachine-ci/builds/1314#ec276310-9efa-47d2-a046-85efc9e050b1/279-299).

That was a faulty GPU

simonbyrne (approved):
A very nice cleanup, and simplifies the code considerably!

src/Atmos/Model/AtmosModel.jl
simonbyrne: Given the scope of this PR, we can also change this later

src/Atmos/Model/get_prognostic_vars.jl
charleskawczynski: I added a doc string to `BalanceLaws`

charleskawczynski:
I messed up the doc string

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I missed the second suggestion here. I like the idea. So, `Σfluxes` would basically be defined as the contents of what is removed?

charleskawczynski: I'll open a follow-up PR

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Interesting. Switching from `map` to `foreach` resulted in `getindex` at `var, name = get_prog_state`. Maybe this is a potential improvement to julialang? (assuming it's not already been improved in 1.6)

charleskawczynski:
Going to move forward with `map`.

PR #1930, opened 2021-01-15
Implements SplitExplicitOceanSuperModel

glwagner:
### Description

This PR implements a "Super Model" similar to the `HydrostaticBoussinesqSuperModel` for the "split-explicit" formulation of the hydrostatic Boussinesq equations called `SplitExplicitOceanSuperModel`. The split-explicit formulation of the hydrostatic Boussinesq equations integrates the barotropic (vertically-integrated) component of the solution with a small "fast" time step, while integrating the baroclinic (deviation from the vertical integral) component with the large "slow" time step. The barotropic component is time-averaged and "reconciled" with the baroclinic component prior to a baroclinic time step.

For the purpose of this PR, I have adapted `tutorials/Ocean/internal_wave.jl` to use `SplitExplicitOceanSuperModel` rather than `HydrostaticBoussinesqSuperModel`. In the future, we would like to re-implement the tests for code in `/Ocean/SplitExplicit01` to use `SplitExplicitOceanSuperModel`.

I hope to add a discussion of the keyword arguments for `SplitExplicitOceanSuperModel` to this PR soon.

- [ ] Documentation has been added/updated OR N/A
- [ ] `Base.show` method for `SplitExplicitOceanSuperModel`
- [ ] Internal wave example runs successfully

### DIFF
--- a/tutorials/Ocean/internal_wave.jl
+++ b/tutorials/Ocean/internal_wave.jl
@@ -98,6 +98,21 @@ time_step = 0.005 # close to Δx / c = 0.5 * 1/16, where Δx is nominal resoluti
 
 # and build a model with a smidgeon of viscosity and diffusion,
 
+using ClimateMachine.Ocean.SuperModels: SplitExplicitOceanSuperModel
+
+model = SplitExplicitOceanSuperModel(
+    domain = domain,
+    parameters = NonDimensionalParameters(),
+    barotropic_time_step = time_step,
+    baroclinic_time_step = time_step / 24,
+    initial_conditions = initial_conditions,
+    turbulence_closure = (νʰ = 1e-6, νᶻ = 1e-6, κʰ = 1e-6, κᶻ = 1e-6, κᶜ = 1.0),
+    coriolis = (f₀ = f, β = 0),
+    buoyancy = (αᵀ = αᵀ,),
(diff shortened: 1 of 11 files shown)


tutorials/Ocean/internal_wave.jl
```diff
@@ -98,14 +98,15 @@ time_step = 0.005 # close to Δx / c = 0.5 * 1/16, where Δx is nominal resoluti
...
+using ClimateMachine.Ocean.SuperModels: SplitExplicitOceanSuperModel
 
-model = HydrostaticBoussinesqSuperModel(
+model = SplitExplicitOceanSuperModel(
     domain = domain,
-    time_step = time_step,
-    initial_conditions = initial_conditions,
     parameters = NonDimensionalParameters(),
-    turbulence_closure = (νʰ = 1e-6, νᶻ = 1e-6, κʰ = 1e-6, κᶻ = 1e-6),
+    barotropic_time_step = time_step,
```
jm-c, line 106: Are we sure of these 2 time-step setting ? Would look better to me if it was switched, i.e,
 barotropic_time_step = time_step / 24
 baroclinic_time_step = time_step

glwagner: Agree, they got switched up! Even fixing that I'm having trouble finding stable values for these; but possibly I need to run this problem at higher resolution.

PR #1946, opened 2021-01-24
Add element and domain sizes CL options

valeriabarra:
### Description

This PR adds a few command line options to facilitate HPC runs and users experience:
- `--nelems` and `--domain-height` for GCM and single-stack configurations
- `--resolution` and  `--domain-min`/`--domain-max` for LES and `MultiColumnLandModel` configurations

While at it, when testing these additions, I also renamed `test/Driver/driver_test.jl` -> `test/Driver/les_driver_test.jl` for consistency with the GCM case that is called `test/Driver/gcm_driver_test`.


valeriabarra:
I tested the new options in all possible ways, e.g., 

```
julia --project experiments/AtmosLES/bomex_les.jl --resolution 50,50,50 --domain-min 50,50,50 --domain-max 5000,5000,5000
```

```
CLIMATEMACHINE_SETTINGS_DOMAIN_HEIGHT=200 julia --project test/Driver/gcm_driver_test.jl
```
or passed as arguments to `ClimateMachine.init()`

src/Driver/Driver.jl
```diff
@@ -300,6 +306,31 @@ function parse_commandline(
         metavar = "<horizontal>,<vertical>"
         arg_type = NTuple{2, Int}
         default = get_setting(:degree, defaults, global_defaults)
+        "--nelems"
+        help = "tuple of number of elements in each direction: 2 for GCM or 1 for single-stack (no space before/after comma)"
```
kpamnany, line 310: This help string seems to say that a single element tuple is expected for single-stack, but the code seems to require a two element tuple?

kpamnany (approved):

src/Driver/Driver.jl
valeriabarra: The option can either be a tuple of two elements for GCM or a single value for single stack.

valeriabarra:
bors try

src/Driver/driver_configs.jl
```diff
@@ -32,22 +32,35 @@ Parses custom command line option for tuples of two integers.
 function ArgParse.parse_item(::Type{NTuple{2, Int}}, s::AbstractString)
 
     str_array = split(s, ",")
-    horizontal = parse(Int, str_array[1])
-    vertical = parse(Int, str_array[2])
+    int_1 = length(str_array) > 0 ? parse(Int, str_array[1]) : 0
+    int_2 = length(str_array) > 1 ? parse(Int, str_array[2]) : 0
```
valeriabarra, line 36: @kpamnany here I took care of the case in which a single integer is passed instead of two. In that case, a 0 is placed as the second entry of the tuple. I tested this for single stack configs, e.g., after adding `ClimateMachine.init(parse_clargs = true)` in `test/Utilities/SingleStackUtils/ssu_tests.jl`, 
```
julia --project test/Utilities/SingleStackUtils/ssu_tests.jl --domain-height 2 --nelems 10
```

and the relevant info printed when the test is run is:

```
Info: Establishing single stack configuration for SingleStackUtilsTest
│     precision               = Float64
│     horiz polynomial order  = 5
│     vert polynomial order   = 5
│     domain_min              = 0.00 m, 0.00 m, 0.00 m
│     domain_max              = 1.00 m, 1.00 m, 2.00 m
│     # vert elems            = 10
│     MPI ranks               = 1
│     min(Δ_horz)             = 0.12 m
└     min(Δ_vert)             = 0.02 m
```

valeriabarra:
This closed issue #1641 .

valeriabarra:
Note that this PR didn't touch the `OceanBoxGCMConfiguration` case because I was confused on how to use the number of elements in that case, since it is formulated differently from the other cases. 

See the calculation of [brickrange](https://github.com/CliMA/ClimateMachine.jl/blob/fe19efd685881ead178842fb58363d818582bc7f/src/Driver/driver_configs.jl#L423) compared to the other [cases](https://github.com/CliMA/ClimateMachine.jl/blob/fe19efd685881ead178842fb58363d818582bc7f/src/Driver/driver_configs.jl#L214). 
If anyone wants to add it, if they have a better idea of the Ocean config

valeriabarra:
@blallen 
if you think it could be quickly added by someone, please let me know. Or if you could please clarify to me how the brickrange is constructed differently in the ocean case I can try to do it. Thanks

blallen:
> @blallen
> if you think it could be quickly added by someone, please let me know. Or if you could please clarify to me how the brickrange is constructed differently in the ocean case I can try to do it. Thanks

I think you can treat it like the LES case. Both are rectangular boxes, the only difference is the ocean height goes down, e.g. from -H to 0 instead of from 0 to H.

valeriabarra:
Thanks. But why it uses the `length = Nˣ + 1`, which if I interpreted correctly the variable's meaning, it represents the number of elements plus 1, and the other configuration uses `elemsize`?

blallen:
> Thanks. But why it uses the `length = Nˣ + 1`, which if I interpreted correctly the variable's meaning, it represents the number of elements plus 1, and the other configuration uses `elemsize`?

Ah because we are used to specifying things in terms of number of elements like 40x40x10 instead of by resolution which would be like 100kmx100kmx100m. There's no substantial difference between the two

valeriabarra:
Great, thanks. Please see #1960

Issue #500, opened 2019-10-23
Allocations on the CPU


mwarusz:
Today I saw that we do a lot of allocations inside our main computational kernels on the CPU,
see `TimerOutputs` results for the DYCOMS example here

https://gist.github.com/mwarusz/786e4790f4d82c47d26bd1f442625623

The biggest culprit seems to be creation of `Vars` and `Grad` wrappers inside the kernels.
Manually hoisting them outside inner loops when possible gets rid of most of the allocations
but is a bit painful, I only tried it for one kernel.

This does not affect the GPU performance at all so, inspired by this, I modified `GPUifyLoops` to run the pass that forces inlining also on the CPU. This got rid of most of the allocations and the resulting speedup is not insignificant (~1.8x).

Not sure what would be a good solution for this problem.

@simonbyrne @kpamnany @vchuravy

kpamnany:
Nice! Can you link the branch the manually hoisted version is in please?

mwarusz:
https://github.com/climate-machine/CLIMA/tree/mw/lessalloc

Here I did it for the `volumerhs!` kernel which is a trivial case. Other kernels are more tricky.

jkozdon:
Glad you were able to track this down!

> This does not affect the GPU performance at all so, inspired by this, I modified `GPUifyLoops` to run the pass that forces inlining also on the CPU. This got rid of most of the allocations and the resulting speedup is not insignificant (~1.8x).

Do you know if the speedup came from the inlining and unrolling on the CPU vs the removing of allocations?

Is there a reason not to just roll with the inlining and unrolling on the CPU in the way we do on the GPU?

mwarusz:
>Do you know if the speedup came from the inlining and unrolling on the CPU vs the removing of allocations?

I don't now exactly. The speedup I got from inlining for the volume kernel seems to be similar to the one I got from removing allocations manually but its hard to say if this would be the same for the other kernels.

> Is there a reason not to just roll with the inlining and unrolling on the CPU in the way we do on the GPU?

I don't see any reason since we can still disable inlining if it have to via `@noinline` but I would like to hear input from others.

smarras79:
A merge of the latest `upstream/master` breaks the code with an array allocation error (see stacktrace below) when using `device=CPU()` in `Q = init_ode_state(dg, FT(0); device=CPU())` 

```
ERROR: LoadError: Type Array does not have a definite size.
Stacktrace:
[...]

PR #1947, opened 2021-01-25
Add some Vars wrappers

charleskawczynski:
### Description

This PR revives some old work with adding `Vars` wrappers to `BalanceLaws`, so that the DG kernels don't have to wrap arrays with every call. The previous branch I was working on got a bit out of control, so I'm leaning towards doing this more incrementally.

This PR only adds vars wrappers for:
 - `source`
 - `flux_first_order!`
 - `flux_second_order!`

When trying to rely on dispatch between `AbstractArray` and `Vars` methods, I hit method ambiguities (due to, e.g., `source!(::AtmosAcousticLinearModel, _...) = nothing`), so my thought was that appending `_arr` to the array-based methods would be the simplest solution. I'm definitely open to other ideas.

Pros/cons:
 - [con] Performance? (based on observations in #500)
 - [pro] Thins the layer in which `Vars` is used, which makes VariableTemplates closer to being replaceable
 - [pro] Improves readability of DG kernels

I also flattened out a few doc strings to make them a bit more readable.


jkozdon (commented):
This is roughly what I was proposing.

Why do we need the suffix `_arr!`?

I started this for the [numerical fluxes too](https://github.com/CliMA/ClimateMachine.jl/blob/fe19efd685881ead178842fb58363d818582bc7f/src/Numerics/DGMethods/NumericalFluxes.jl#L703-L913) when I did the FVM kernels. But I didn't fully fix the DG kernels; instance [here](https://github.com/CliMA/ClimateMachine.jl/blob/fe19efd685881ead178842fb58363d818582bc7f/src/Numerics/DGMethods/DGModel_kernels.jl#L785-L837) and elsewhere

jkozdon:
I'd be super happy is 'Vars' and 'Grad' were totally gone from the `DGModel_kernels.jl` and `DGFVModel_kernels.jl` files. I think it will make them more formatter friendly and easier to read in general (we shouldn't need to know about vars at this level).

jkozdon:
Also, instead of putting them in a separate file, they could just go in the [`interface.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/fe19efd685881ead178842fb58363d818582bc7f/src/BalanceLaws/interface.jl) file.

charleskawczynski:
> This is roughly what I was proposing.

👍🏻

> Why do we need the suffix _arr!?

At first I didn't have the `_arr` suffix, but then it lead to method ambiguities. We could fix the method ambiguities by changing, for example:
```julia
source!(::AtmosAcousticLinearModel, _...) = nothing
```
to
```julia
source!(
    ::AtmosAcousticLinearModel,
    source::Vars,
    state::Vars,
    diffusive::Vars,
    aux::Vars,
    t::Real,
    direction,
) = nothing
```
but I thought there might be a lot of these (across `source!`, `flux_first_order!`, `flux_second_order!`, `boundary_state!` etc.). So, I figured the suffix might be simpler. I'm open to trying it though. Maybe we'd be able to get away with, e.g., `source!(::AtmosAcousticLinearModel, ::Vars, _...) = nothing`. I'm not sure what the best solution is.

> I started this for the numerical fluxes too when I did the FVM kernels. But I didn't fully fix the DG kernels; instance here and elsewhere

I started working on the numerical fluxes, but things started ballooning, so I decided it might be easier to do this incrementally.

jkozdon:
> > Why do we need the suffix _arr!?
> 
> At first I didn't have the `_arr` suffix, but then it lead to method ambiguities. We could fix the method ambiguities by changing, for example:
> 
> ```julia
> source!(::AtmosAcousticLinearModel, _...) = nothing
> ```
> 
> to
> 
> ```julia
> source!(
>     ::AtmosAcousticLinearModel,
>     source::Vars,
>     state::Vars,
>     diffusive::Vars,
>     aux::Vars,
>     t::Real,
>     direction,
> ) = nothing
> ```
> 
> but I thought there might be a lot of these (across `source!`, `flux_first_order!`, `flux_second_order!`, `boundary_state!` etc.). So, I figured the suffix might be simpler. I'm open to trying it though. Maybe we'd be able to get away with, e.g., `source!(::AtmosAcousticLinearModel, ::Vars, _...) = nothing`. I'm not sure what the best solution is.

This makes sense. `_arr` is fine. I cannot think of anything worse, so stick with it.

Should we be consistent and do the same for the numerical flux wrappers?

 
> > I started this for the numerical fluxes too when I did the FVM kernels. But I didn't fully fix the DG kernels; instance here and elsewhere
> 
> I started working on the numerical fluxes, but things started ballooning, so I decided it might be easier to do this incrementally.

This totally make sense!

jkozdon (approved):
This is a great start. Thanks!

PR #1953, opened 2021-01-26
add hooks for gradient and tendency filters

simonbyrne:
### Description

This should make it easier to do tendency-level filtering to emulate overintegration.

cc: @jkozdon @Zhengyu-Huang 


Zhengyu-Huang (approved):
That looks great to me!  Thank you for the effort. 
I will try this!

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -294,6 +306,17 @@ function (dgfvm::DGFVModel)(tendency, state_prognostic, _, t, α, β)
...
         )
 
+        if dgfvm.gradient_filter !== nothing
+            wait(device, comp_stream)
+            Filters.apply!(
+                dgfvm.state_gradient_flux,
+                1:num_state_gradient_flux,
+                dgfvm.grid,
+                dgfvm.gradient_filter,
+            )
```
jkozdon, line 316: It might be good to modify `Filters.apply!` take an optional event so that this doesn't slow everything down.

The logic gets a tad complicated since if the user doesn't pass an `event` you would wait in the `apply!` otherwise you would return the new event.

simonbyrne:
TODO:
- add `--cutoff-degree` option to the driver
- make filter asynchronous

src/Numerics/DGMethods/DGModel.jl
simonbyrne: Probably better to have a separate `async_apply!` function

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
bors try

simonbyrne:
Okay, this is ready for review. It does need a test: I don't know what would be best to try out though.

simonbyrne:
bors try

simonbyrne:
@thomasgibson did you want to modify the isothermal zonal flow test?

simonbyrne:
bors try

src/Driver/driver_configs.jl
```diff
@@ -252,12 +259,24 @@ function AtmosLESConfiguration(
         meshwarp = meshwarp,
     )
 
+    if cutofforder_horz < polyorder_horz || cutofforder_vert < polyorder_vert
```
valeriabarra, line 262: Do we want a logical OR here or a logical AND? I thought both conditions were needed. Can we have a case where only one of the two is satisfied and we still can apply the filter?

src/Driver/driver_configs.jl
```diff
@@ -252,12 +259,24 @@ function AtmosLESConfiguration(
...
+
+
     @info @sprintf(
         """
 Establishing Atmos LES configuration for %s
     precision               = %s
     horiz polynomial order  = %d
     vert polynomial order   = %d
+    cutofforder_horz        = %d
+    cutofforder_vert        = %d
```
valeriabarra, line 279: for consistency with the other prints, I'd use something like:
```suggestion
    horiz cutoff order        = %d
    vert cutoff order          = %d
```

src/Driver/driver_configs.jl
valeriabarra: The same question applies to the following configurations

simonbyrne: we only need to apply the filter if the cutoff order is < polyorder

valeriabarra: Yes, but for both its components, horizontal and vertical, no?

valeriabarra:
Note:

While testing this PR, I realized that in PR #1946 I did not check for backward compatibility, and unfortunately missed testing the `--degree` option that was already there. I'm pushing a fix to your branch right now.

src/Driver/driver_configs.jl
```diff
@@ -196,9 +200,12 @@ function AtmosLESConfiguration(
     numerical_flux_gradient = CentralNumericalFluxGradient(),
     fv_reconstruction = nothing,
     grid_stretching = (nothing, nothing, nothing),
+    Ncutoff = N,
```
valeriabarra, line 203: I'm testing this by running a different experiment than the `experiments/TestCase/isothermal_zonal_flow.jl` and this would break, if the default test arguments were used. I think we should give `Ncutoff` another default value, not the same as `N`.

src/Driver/driver_configs.jl
```diff
@@ -196,9 +214,12 @@ function AtmosLESConfiguration(
...
```
valeriabarra, line 217: I'm testing this by running a different experiment than the `experiments/TestCase/isothermal_zonal_flow.jl` and this would break, if the default test arguments were used. I think we should give `Ncutoff` another default value, not the same as `N`.

valeriabarra (changes requested):

src/Driver/driver_configs.jl
```diff
@@ -252,12 +273,24 @@ function AtmosLESConfiguration(
         meshwarp = meshwarp,
     )
 
+    if cutofforder_horz < polyorder_horz || cutofforder_vert < polyorder_vert
+        filter = CutoffFilter(
+            grid,
+            (cutofforder_horz + 1, cutofforder_horz + 1, cutofforder_vert + 1),
+        )
+    else
+        filter = nothing
```
valeriabarra, line 282: If we fall in this `else` condition, we should change the values of `cutofforder_horz` and `cutofforder_vert` to `nothing`, so that in the print that follows, the user indeed sees that in that case the filter was not applied. Otherwise, the way it is right now, the user would see the attempted cutoff filter degrees, not the actual (nothing) one and it would be misleading.

valeriabarra (dismissed):

src/Driver/driver_configs.jl
valeriabarra: I changed it in [c575dc3](https://github.com/CliMA/ClimateMachine.jl/pull/1953/commits/c575dc3358e3e202eea2e057078fda5b0550c311)

src/Driver/driver_configs.jl
simonbyrne: we could make it `get_polyorder(N)`?

src/Driver/driver_configs.jl
simonbyrne: Thanks!

src/Driver/driver_configs.jl
simonbyrne: if we apply a cutoff filter that is the same degree or higher, it won't do anything (so is safe to apply the filter in the other direction).

valeriabarra: I don't think so. Since currently there is an `assert` in the `Filter.jl` module. I tested this and got the following stack:

```
ERROR: LoadError: AssertionError: all(0 .<= Nc .<= N)
Stacktrace:
 [1] CutoffFilter(::DiscontinuousSpectralElementGrid{Float64,3,(5, 5, 5),216,Array,Tuple{Array{Float64,1},Array{Float64,1},Array{Float64,1}},Tuple{Array{Float64,2},Array{Float64,2},Array{Float64,2}},Array{Float64,3},Array{Float64,4},Array{Int64,1},Array{Int64,2},Array{Int64,3},ClimateMachine.Mesh.Topologies.StackedCubedSphereTopology{Float64,2},Tuple{Array{Float64,2},Array{Float64,2},Array{Float64,2}}}, ::Tuple{Int64,Int64,Int64}) at /home/valeria/CliMA/ClimateMachine.jl/src/Numerics/Mesh/Filters.jl:264
```

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -512,6 +545,16 @@ function (dg::DGModel)(tendency, state_prognostic, _, t, α, β)
...
         )
 
+        if dg.gradient_filter !== nothing
+            wait(device, comp_stream)
+            Filters.apply!(
+                dg.state_gradient_flux,
+                1:num_state_gradient_flux,
+                dg.grid,
+                dg.gradient_filter,
+            )
```
jkozdon, line 555:
```suggestion
            comp_stream = Filters.apply_async!(
                dg.state_gradient_flux,
                1:num_state_gradient_flux,
                dg.grid,
                dg.gradient_filter;
                dependencies = comp_stream
            )
```

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -720,6 +763,14 @@ function (dg::DGModel)(tendency, state_prognostic, _, t, α, β)
...
     # this function is finished.
     wait(device, comp_stream)
+    if dg.tendency_filter !== nothing
+        Filters.apply!(
+            tendency,
+            1:num_state_tendency,
+            dg.grid,
+            dg.tendency_filter,
+        )
+    end
```
jkozdon, line 773:
```suggestion
    if dg.tendency_filter !== nothing
        comp_stream = Filters.apply!(
            tendency,
            1:num_state_tendency,
            dg.grid,
            dg.tendency_filter;
            dependencies = comp_stream
        )
    end
    wait(device, comp_stream)
```

jkozdon (commented):
Two questions:

- Hyperdiffusion should probably be filtered as well
- Shouldn't the filter be applied in 'init_ode_state' as well?

jkozdon:
Looking good though!

valeriabarra:
bors try

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -719,6 +762,15 @@ function (dg::DGModel)(tendency, state_prognostic, _, t, α, β)
...
     # this function is finished.
+    if dg.tendency_filter !== nothing
+        comp_stream = Filters.apply!(
+            tendency,
+            1:num_state_tendency,
+            dg.grid,
+            dg.tendency_filter;
+            dependencies = comp_stream
+        )
+    end
```
valeriabarra, line 773: I'm trying to debug this. 
I tested this with `experiments/TestCase/isothermal_zonal_flow.jl` and here `comp_stream = nothing` and the code breaks at the following line, with error:
```
ERROR: LoadError: MethodError: no method matching wait(::KernelAbstractions.CPU, ::Nothing)
```
I'm trying to understand why in this case the `apply_async!` returns nothing to the `Filters.apply!`. @simonbyrne , @jkozdon any clue?

valeriabarra: Ah! I rolled back to commit [ac05980](https://github.com/CliMA/ClimateMachine.jl/pull/1953/commits/ac059805ed895efb7e08bf2e7e453a6f1c1094d1), that was before applying @jkozdon 's suggestion and it works. Hence, something needs to be improved, building on that suggestion. Thanks :)

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -398,6 +421,16 @@ function (dgfvm::DGFVModel)(tendency, state_prognostic, _, t, α, β)
         ),
     )
 
+    if dgfvm.tendency_filter !== nothing
+        Filters.apply_async!(
```
mwarusz, line 425:
```suggestion
        comp_stream = Filters.apply_async!(
```

src/Numerics/DGMethods/DGModel.jl
```diff
@@ -719,6 +762,15 @@ function (dg::DGModel)(tendency, state_prognostic, _, t, α, β)
     # The synchronization here through a device event prevents CuArray based and
     # other default stream kernels from launching before the work scheduled in
     # this function is finished.
+    if dg.tendency_filter !== nothing
+        comp_stream = Filters.apply!(
```
mwarusz, line 766:
```suggestion
        comp_stream = Filters.apply_async!(
```

valeriabarra: Thanks @mwarusz but that still doesn't work :/

src/Driver/driver_configs.jl
valeriabarra: @simonbyrne  I fixed this in my branch `valeria/fix-cutoff-degree-cl-option`. You can merge it here when the issue regarding the `apply_async!` is sorted out.

src/Numerics/DGMethods/DGModel.jl
mwarusz: Weird, works for me. Just to make sure: I had two separate suggestions, your latest has the first one but not this one. Is this what you wanted ?

valeriabarra: Oh, yes! Sorry! I had accidentally missed one of the two. Thanks! Just testing it now.

valeriabarra:
bors try

src/Numerics/DGMethods/DGModel.jl
valeriabarra: Update: This has been resolved by @mwarusz 's fixes! Thanks @jkozdon  and @mwarusz !!!

src/Driver/driver_configs.jl
valeriabarra: Update: I pushed it to this branch. So it is fixed now and I dismissed my request for changes in this PR

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra (approved):

valeriabarra:
@simonbyrne I think this PR is ready to go. I will reduce the number of commits to a small number of significant commits to preserve authorship of all contributors to this PR, and then `r+` if that's ok. Thanks

jkozdon:
Should we allow in some way for different filters in the `vertical` and `horiztonal`? Not sure if mixing polynomial orders is still in the cards, but if you wanted overintegration with this we might need two different filters

valeriabarra:
That's what I thought too, when testing this and was wondering if we wanted to allow the filtering to happen only in one direction and for instance `nothing` in the other. We would need to erase all the `@assert (0 .<= Nc .<= N)` in the grid, if we want to go that way.

simonbyrne:
Thanks everyone for the contributions, I really appreciate the help.

> Should we allow in some way for different filters in the `vertical` and `horiztonal`? Not sure if mixing polynomial orders is still in the cards, but if you wanted overintegration with this we might need two different filters

Can't you specify different orders for the cutoff filter? Or you mean apply different filters?

> That's what I thought too, when testing this and was wondering if we wanted to allow the filtering to happen only in one direction and for instance `nothing` in the other. We would need to erase all the `@assert (0 .<= Nc .<= N)` in the grid, if we want to go that way.

If `Nc == N`, won't that do nothing (since you're filtering to the same degree)?

jkozdon:
> Can't you specify different orders for the cutoff filter? Or you mean apply different filters?

You're right, the cutoff filter takes a tuple of polynomial orders. I forgot!

> If `Nc == N`, won't that do nothing (since you're filtering to the same degree)?

'Nc' is the first cutoff mode. So you would need 'Nc = N + 1' would be a non-op (though the assert doesn't allow this). Would should probably allow for it in someway, but maybe that's a separate pr?

valeriabarra:
> 'Nc' is the first cutoff mode. So you would need 'Nc = N + 1' would be a non-op (though the assert doesn't allow this). Would should probably allow for it in someway, but maybe that's a separate pr?

I would leave that to a separate PR

simonbyrne:
I updated the merge. Any final comments @valeriabarra?

valeriabarra:
Thanks Simon. No more comments from me. If we want to separate the directions and remove all the asserts `@assert (0 .<= Nc .<= N)`, I think it could belong to a separate PR.

PR #1847, opened 2020-12-09
Transfer over atmos linear model tendencies to new specification

thomasgibson:
### Description

WIP. Transfers over atmos linear model tendencies to the new type specification.

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #1954, opened 2021-01-26
Add linearized_pressure wrapper, more compact namedtuples

charleskawczynski:
### Description

Peeled off from #1847. This PR adds a wrapper for `linearized_pressure` and uses Julia's `(;)` NamedTuple constructor syntax.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -506,7 +506,7 @@ equations.
 
     flux_pad = SVector(1, 1, 1)
     tend = Flux{FirstOrder}()
-    _args = (state = state, aux = aux, t = t, direction = direction)
+    _args = (; state, aux, t, direction)
     args = merge(_args, (precomputed = precompute(atmos, _args, tend),))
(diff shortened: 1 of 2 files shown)


src/Atmos/Model/AtmosModel.jl
```diff
@@ -506,7 +506,7 @@ equations.
 
     flux_pad = SVector(1, 1, 1)
     tend = Flux{FirstOrder}()
-    _args = (state = state, aux = aux, t = t, direction = direction)
+    _args = (; state, aux, t, direction)
```
mwarusz, line 509: I think this syntax was added 1.5. Our compat bound is still 1.4. Should we update it ?

mwarusz (approved):

src/Atmos/Model/AtmosModel.jl
charleskawczynski: That's a good point. I think we're using this syntax elsewhere, but the compat probably should match.

charleskawczynski: Opened #1955 to do this separately.

Issue #1952, opened 2021-01-26
Re-enable Sysimg CI build when Julia v1.5.4 is released.


jakebolewski:
### Description

Running the tests with a custom system image has been disabled due to Julia runtime bugs in 1.5.2 and 1.5.3.  Re-enable testing with custom system image when Julia 1.5.4 is released https://github.com/JuliaLang/julia/pull/39351

jakebolewski:
In addition we need to undo: https://github.com/CliMA/ClimateMachine.jl/pull/1958/commits/5fa0b830fe47cebb5cfc40e0efd82e70fa1c1d94

PR #1971, opened 2021-01-29
Move long tests to buildkite

charleskawczynski:
### Description

This PR moves some of the longer tests over to buildkite, to help with CI. To note, some tests are already being run in buildkite, so they were just removed from the GHA CI.

Maybe we can revert this once #1952 is fixed.


charleskawczynski:
It looks like this will bring our CI times back down to < 1hr.

simonbyrne:
Thanks @charleskawczynski. Fine by me, any thoughts @jakebolewski ?

kpamnany (commented):
GHA gives us cross-platform testing. Not sure if the tests moved will weaken our assurance of platform portability; I would guess not, but if we're not sure, perhaps it would be better to keep a representative test from each group on GHA rather than move entire groups?

jakebolewski:
This LGTM, I agree that we loose x-platform testing but I don't know if longer term that will honestly be sustainable for expensive tests as the number of tests just continues to increase.

jakebolewski (approved):

charleskawczynski:
Had to fix a few labels (they were not unique on the first few r+'s)

charleskawczynski:
Ok, maybe not quite < 1 hr, but we're close

Issue #1441, opened 2020-08-14
Dump non-interpolated data


LenkaNovak:
- several groups have asked for an option to dump non-interpolated data (on nodal points) into NetCDF files.

blallen:
Cc; @christophernhill  @ali-ramadhan

kpamnany:
This is essentially the inverse of #1092 and the solution is in progress. Essentially, we need the ability to _optionally_ specify an interpolation for any diagnostics group. Currently, some diagnostics groups require an interpolation configuration and some cannot use one.

kpamnany:
I need additional information on this one. I don't see a way to output non-interpolated data into NetCDF-4 as the format does not seem to support what's needed (see [this comment](https://github.com/CliMA/ClimateMachine.jl/issues/1092#issuecomment-627497058) which seems to need something like [this](https://publicwiki.deltares.nl/display/NETCDF/Finite+Element+based+CF+proposal+for+Unstructured+Grid+data+model) and this has not been included into the NetCDF spec yet AFAICT).

I can treat nodal points and elements as "dimensions" but that will require special treatment in post-processing, i.e. you won't be able to load these NC files into Paraview for instance.

Or, I can try to generalize the VTK module to allow output of arbitrary fields (not just prognostic and auxiliary variables) and use that instead.

Another complication is supporting multi-node output. We can handle this to a certain extent for interpolated data now and have some ideas on how to scale up, but it isn't clear how to manage this for non-interpolated data.

Thoughts?

christophernhill:
A suggestion

1.  just dumping to index based grid , along with “world coords“ as a
variable would be super useful. It’s easier to slurp netCDF offline into
python, matlab, Julia etc than reading VTK.

2. It would also be useful to have a world topology to locate cells one
day. Even on sphere we have a semi structured mesh. Generating that as
topology meta data would be useful.

3. Once we have some basic thing for 1. exporting meta data on basis
functions is possible

1. alone would be great to have - rest can come as and when.

Chris

kpamnany:
> 1. just dumping to index based grid , along with “world coords“ as a variable would be super useful.

I'm sorry I don't understand this @christophernhill -- do you mean simply the DG grid, which is indexed on nodal points and elements? And what would be in the world coords?

christophernhill:
Yes - out with dog can you zoom around 10am?

kpamnany:
[...]

Issue #114, opened 2019-03-19
What do you guys do for NetCDF output?


ali-ramadhan:
For Oceananigans.jl we settled on supporting NetCDF output early as it's the most ubiquitous and familiar data format (see https://github.com/climate-machine/Oceananigans.jl/issues/31).

We've been using [NetCDF.jl](https://github.com/JuliaGeo/NetCDF.jl) but it seems like an inactive package so we're thinking of switching to the more active and thoroughly tested [NCDatasets.jl](https://github.com/Alexander-Barth/NCDatasets.jl). Output writing and compression was so slow compared to time stepping on the GPU that we ended up introducing asynchronous output writing (https://github.com/climate-machine/Oceananigans.jl/pull/137).

Anyways, just curious if NetCDF output is something you guys have looked at yet. If not, might be worth it to work on a common solution?

ali-ramadhan:
> Another option, not sure if it is good as I have never used it, is a meta package like <https://www.olcf.ornl.gov/center-projects/adios/> which supports parallel io and is somewhat file format agnostic.
<sub>[View in Slack](https://climate-machine.slack.com/archives/CAV9VKEQJ/p1553011741102200)</sub>

ali-ramadhan:
> <@UCSM0C54P> re: your NetCDF issue, why not HDF5?
<sub>[View in Slack](https://climate-machine.slack.com/archives/CAV9VKEQJ/p1553011287098400)</sub>

ali-ramadhan:
> <@UCSM0C54P> re: your NetCDF issue, why not HDF5?
<sub>[View in Slack](https://climate-machine.slack.com/archives/CAV9VKEQJ/p1553011287098400)</sub>

ali-ramadhan:
> if you do need NetCDF compatibility (e.g. for exporting to another program), it might be possible to enforce the necessary restrictions at the Julia level
<sub>[View in Slack](https://climate-machine.slack.com/archives/CAV9VKEQJ/p1553012268107000)</sub>

ali-ramadhan:
> My short exposure to HDF5 has all been positive
<sub>[View in Slack](https://climate-machine.slack.com/archives/CAV9VKEQJ/p1553013251110100)</sub>

ali-ramadhan:
> The main complaints I have heard about HDF5 have been when people try to update or mutate data, and corrupt their dataset
[...]

PR #1977, opened 2021-01-29
Add `VTKFieldWriter`

kpamnany:
### Description

Allows writing of computed fields into VTK.

For example:
```julia
using ClimateMachine.VTK
# ...

    function pres_fun(atmos::AtmosModel, prognostic::Vars, auxiliary::Vars)
        ts = recover_thermo_state(atmos, prognostic, auxiliary)
        air_pressure(ts)
    end
    fwriter = VTKFieldWriter(solver_config.name, [("pressure", pres_fun)])
    cbfw = GenericCallbacks.EveryXSimulationTime(60) do
        fwriter(solver_config.dg, solver_config.Q)
    end
```
Then pass `cbfw` into `invoke!` (or `solve!`) as a callback. You can use a similar pattern to add other computed fields; just add tuples to the vector passed in to the `VTKFieldWriter` constructor.

### DIFF
--- a/src/InputOutput/VTK/fieldwriter.jl
+++ b/src/InputOutput/VTK/fieldwriter.jl
@@ -1,0 +1,126 @@
+import KernelAbstractions: CPU
+
+using MPI
+using Printf
+
+using ..BalanceLaws
+using ..MPIStateArrays
+using ..DGMethods: SpaceDiscretization
+using ..VariableTemplates
+
+"""
+    VTKFieldWriter(
+        name::String,
+        FT::DataType,
+        fields::Vector{<:Tuple{String, <:Function}};
+        path_prefix = ".",
+        number_sample_points = 0,
+    )
+
+Construct a callable type that computes the specified `fields` and
+writes them to a VTK file in `path_prefix`/. `number_sample_points`
+are passed to [`writevtk`](@ref).
+
+Intended for use in a callback passed to a running simulation; files
+are suffixed with a running number beginning with 0.
+
+# Example
+```julia
+    function pres_fun(atmos::AtmosModel, prognostic::Vars, auxiliary::Vars)
+        ts = recover_thermo_state(atmos, prognostic, auxiliary)
+        air_pressure(ts)
+    end
+    fwriter = VTKFieldWriter(solver_config.name, [("pressure", pres_fun)])
+    cbfw = GenericCallbacks.EveryXSimulationTime(60) do
+        fwriter(solver_config.dg, solver_config.Q)
+    end
+```
+"""
+mutable struct VTKFieldWriter
+    path_prefix::String
+    name::String
+    number_sample_points::Int
+    nfields::Int
+    field_names::Vector{String}
+    field_funs::Vector{<:Function}
+    vars_type::DataType
+    num::Int
+
+    function VTKFieldWriter(
+        name::String,
+        FT::DataType,
+        fields::Vector{<:Tuple{String, <:Function}};
+        path_prefix = ".",
+        number_sample_points = 0,
+    )
+        nfields = length(fields)
+        field_names = [name for (name, _) in fields]
+        field_funs = [fun for (_, fun) in fields]
+        vars_type = NamedTuple{
+            tuple(Symbol.(field_names)...),
+            Tuple{[FT for _ in 1:length(fields)]...},
+        }
+        new(
+            path_prefix,
+            name * "_fields",
+            number_sample_points,
+            nfields,
+            field_names,
+            field_funs,
+            vars_type,
+            0,
+        )
+    end
+end
+function (vfw::VTKFieldWriter)(dg::SpaceDiscretization, Q::MPIStateArray)
+    bl = dg.balance_law
+    fQ = similar(Q, Array; vars = vfw.vars_type, nstate = vfw.nfields)
+
+    if array_device(Q) isa CPU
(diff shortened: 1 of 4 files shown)


kpamnany:
@smarras79 and @OsKnoth: here's how you can get pressure (or other such computed fields) output to VTK. Please pick one of our experiments and I'll add code that uses this into it for you to use as a pattern.

kpamnany:
@jkozdon: could you take a look at this please and see if it looks okay to you?

@smarras79: does this work for you?

src/InputOutput/VTK/fieldwriter.jl
```diff
@@ -0,0 +1,126 @@
...
+    field_funs::Vector{<:Function}
+    vars_type::DataType
+    num::Int
+
+    function VTKFieldWriter(
+        name::String,
+        FT::DataType,
+        fields::Vector{<:Tuple{String, <:Function}};
+        path_prefix = ".",
+        number_sample_points = 0,
```
jkozdon, line 54: Minor, but other places this is a keyword argument.

jkozdon (approved):
Looks good. Can we add a test? There are some other vtk tests, mainly to make sure the interfaces continue to work.

We could possibly move to only using this and defaulting to a version that dumps all the data in aux and state? Or alternatively should this be merged with the other vtk data dump functionality so you dump state, aux, and computed fields.

I'm also fine with leaving as is though, as far as I can tell it's correct.

src/InputOutput/VTK/fieldwriter.jl
kpamnany: It's a keyword argument here also, as is `path_prefix`. :)

smarras79:
This solution is not trivial. A user should not be worrying about writing code lines to simply have pressure (a fundamental quantity in fluid flow solver) written to VTK.

src/InputOutput/VTK/fieldwriter.jl
jkozdon: Oops. Missed the semicolon, thought it was a default value!

simonbyrne:
@smarras79 Unfortunately the current design of the VTK output is really only intended for prognostic quantities; derived quantities are designed to go through the diagnostics system. Changing how this works will require significant changes to the codebase, which we don't have the resources to do at the moment.

smarras79:
> @smarras79 Unfortunately the current design of the VTK output is really only intended for prognostic quantities; derived quantities are designed to go through the diagnostics system. Changing how this works will require significant changes to the codebase, which we don't have the resources to do at the moment.

smarras79:
[closed by mistake] reopened

smarras79:
The VTK is there to help debug the code because, as of now, it is the only format in climatemachine that writes data on the native numerical (DG) grid. This file contains quantities that are totally unimportant from the physical point of view of what problems we are solving (total energy, density, momentum, reference density). 
If we have useless quantities given to a developer who needs to debug the code, effective and efficient debugging is close to impossible (as demonstrated so far).

At the level of writing the VTK file, it is 1 call to the thermodynamic function that computes pressure from the equation of state that needs to be called, compute pressure, and write it to file. If this is not easy to do, then there is clearly something very wrong in the code structure and needs to be addressed as soon as possible. 
cc.: @bischtob @tapios @fxgiraldo

kpamnany:
All efforts on output thus far have been focused on the Diagnostics module which currently only writes NetCDF. Some discussion of output formats is in #114.

There are three tasks on the TODO list relating to diagnostics output:
1. #2007 to support parallel NetCDF.
2. #1441 to allow writing data on the native grid into NetCDF.
3. #2006 to allow the Diagnostics module to write to VTK.

These are listed in order of priority. 

We do not currently have any plans to enhance the VTK module beyond the capability introduced in this PR.

kpamnany:
> Looks good. Can we add a test? There are some other vtk tests, mainly to make sure the interfaces continue to work.

@jkozdon: I added a test, but it feels a little silly; can you take a look please?

smarras79:
> All efforts on output thus far have been focused on the Diagnostics module which currently only writes NetCDF. Some discussion of output formats is in #114.
> 
> There are three tasks on the TODO list relating to diagnostics output:
> 
> 1. #2007 to support parallel NetCDF.
> 2. #1441 to allow writing data on the native grid into NetCDF.
> 3. #2006 to allow the Diagnostics module to write to VTK.
> 
> These are listed in order of priority.
> 
> We do not currently have any plans to enhance the VTK module beyond the capability introduced in this PR.

@kpamnany #2006 may not be necessary if you are already doing #1441

jkozdon:
> @jkozdon: I added a test, but it feels a little silly; can you take a look please?

Yeah. Our VTK tests are a bit silly, but we did run into a problem with interface changes in the past, so just wasn't sure if we wanted something that exercises the interface.

I'm fine removing if you think it's not needed.

jkozdon (approved):

kpamnany:
> #2006 may not be necessary if you are already doing #1441

I think we don't yet know if any of the standard NetCDF visualization tools people are using (ParaView, etc.) understand the UGRID conventions. If they do, then you're right. If they don't, then #1441 will only be useful for people writing their own visualization scripts and #2006 might be higher priority.

kpamnany:
> I'm fine removing if you think it's not needed.

I don't think the test adds enough value to make up for the overheads.

PR #1980, opened 2021-02-01
Tendency diagnostics

### LINKED ISSUE #1036: Extract fluxes / sources for diagnostics and budget evaluations

kpamnany:
### Description

Adds a diagnostics group that dumps all currently available tendencies (sources and first and second order flux terms) to a NetCDF file. Requires interpolation.

A more flexible way to specify and get tendencies will be added to DiagnosticsMachine.

This group has been added to `test/Atmos/EDMF/bomex_edmf.jl`.

Closes #1036.


kpamnany:
What else is needed to close #1527?

charleskawczynski:
Closing #1527 is a bit more difficult, because while we're dumping `F_1(Y)`, `F_2(Y, G)` and `S(Y,G)` in `∂_t Y_i + (∇•F_1(Y))_i + (∇•F_2(Y,G)))_i = (S(Y,G))_i`, but we need `∇•F_1(Y)` and `∇•F_2(Y,G)` to report the budget. But this is a good first step.

src/Diagnostics/dump_tendencies.jl
```diff
@@ -0,0 +1,251 @@
...
+                progname = prog_name(pv)
+                for tend in eq_tends(pv, bl, tend_type)
+                    flat_vals = flattened_tuple(
+                        FlattenArr(),
+                        tend_fun(tend, bl, tend_args),
+                    )
+                    for i in 1:length(flat_vals)
+                        varname = "$(progname)_$(tend_name(tend))_$(i)"
+                        vars[varname] =
+                            (tuple(collect(keys(dims))...), FT, Dict())
```
charleskawczynski, line 126: Should `tuple(collect(keys(dims))...)` be precomputed? Or will Julia optimize this away?

kpamnany: Probably can hoist it; doubt Julia will. It's pretty minor though.

szy21:
I'm not familiar with what the tendencies should look like in EDMF. Maybe @yairchn or @ilopezgp could take a look?

test/Atmos/EDMF/bomex_edmf.jl
```diff
@@ -267,13 +267,19 @@ function config_diagnostics(driver_config, timeend)
...
     )
-    dgngrp = setup_dump_state_diagnostics(
+    ds_dgngrp = setup_dump_state_diagnostics(
         SingleStackConfigType(),
         interval,
         driver_config.name,
         interpol = interpol,
     )
-    return ClimateMachine.DiagnosticsConfiguration([dgngrp])
+    dt_dgngrp = setup_dump_tendencies_diagnostics(
```
charleskawczynski, line 276: How is this getting exercised?

kpamnany: It's only added here for evaluation (could put it somewhere else) and I wasn't intending to put it somewhere it would be exercised on CI -- we don't exercise every diagnostic group (e.g. `Dump*`) because they dump quite a bit of output and there's not a lot of logic to test. But I'm happy to do so if you feel it's necessary.

src/Diagnostics/dump_tendencies.jl
```diff
@@ -0,0 +1,251 @@
...
+                le,
+                :,
+            ))
+            hyperdiffusivei = Vars{vars_state(bl, Hyperdiffusive(), FT)}(view(
+                all_hd_data,
+                lo,
+                la,
+                le,
+                :,
+            ))
```
charleskawczynski, line 206:
```suggestion
            vars_view(st, data) = Vars{vars_state(bl, st, FT)}(view(
                data,
                lo,
                la,
                le,
                :,
            ))
            statei = vars_view(Prognostic(), all_state_data)
            auxi = vars_view(Auxiliary(), all_aux_data)
            diffusivei = vars_view(GradientFlux(), all_gf_data)
            hyperdiffusivei = vars_view(Hyperdiffusive(), all_hd_data)
```

kpamnany:
> I'm not familiar with what the tendencies should look like in EDMF. Maybe @yairchn or @ilopezgp could take a look?

Easy enough to add it to any experiment @szy21; can either copy what's in `config_diagnostics` in `bomex_edmf.jl` to another experiment, or I can add it somewhere for you. I tagged you because you opened #1527, so I thought you'd like to see this output. :-)

src/Diagnostics/dump_tendencies.jl
```diff
@@ -0,0 +1,251 @@
...
+        ::ClimateMachineConfigType,
+        interval::String,
+        out_prefix::String;
+        writer = NetCDFWriter(),
+        interpol = nothing,
+    )
+
+Create the "DumpTendencies" `DiagnosticsGroup` which contains all the
+tendencies for the simulation's `BalanceLaw`. These are collected on
+an interpolated grid (`interpol` _must_ be specified).
```
charleskawczynski, line 12: Should we include a warning that this is extremely expensive?

szy21:
> Easy enough to add it to any experiment @szy21; can either copy what's in `config_diagnostics` in `bomex_edmf.jl` to another experiment, or I can add it somewhere for you. I tagged you because you opened #1527, so I thought you'd like to see this output. :-)

Oh! I didn't even remember... then yes, could please add it to bomex_les.jl? Thanks!

src/Diagnostics/dump_tendencies.jl
charleskawczynski: We should also note that this is more of a physics-debugging tool rather than numerics debugging, since these tendency evaluations will not match up exactly with what the RHS evaluation that the solver sees.

kpamnany: Please post a suggestion for the text and I'll add it as is?

src/Diagnostics/dump_tendencies.jl
```diff
@@ -0,0 +1,233 @@
...
+
+!!! warning This diagnostics group can produce a lot of output and is
+expensive.
```
charleskawczynski, line 15:
```suggestion
!!! warn
    This diagnostics group can produce a lot of output and is
    very expensive.

!!! warn
    These diagnostics are intended for physics-debugging, and not numerics debugging, because
    `flux` and `source` are called on interpolated data, which may not exactly match the
    non-interpolated evaluation of fluxes and sources that the solver uses.
```

kpamnany:
> Oh! I didn't even remember... then yes, could please add it to bomex_les.jl? Thanks!

@szy21: I've added it to `bomex_les.jl` and run it. You can either pull this branch and run it yourself, or you can take a look at the output NC file from my run: `~kpamnany/ClimateMachine.jl/output/BOMEX_DumpTendencies-2021-02-01T10.18.14.367.nc`.

szy21:
What are those numbers (1,2,3,...9) for? Different directions?

kpamnany:
> What are those numbers (1,2,3,...9) for? Different directions?

When there are 3, they are `x`, `y`, and `z`. I'm not sure about 9 -- @charleskawczynski can explain better.

szy21:
Thanks! And would it be easy to calculate horizontal mean for these tendencies? @akshaysridhar for LES, do you think 3d output or 1d output of tendencies is more helpful?

charleskawczynski:
For momentum, fluxes have 9-components (3 spatial components and 3 momentum components), all other equations are scalar equations, so fluxes only have 3 components

szy21:
> For momentum, fluxes have 9-components (3 spatial components and 3 momentum components), all other equations are scalar equations, so fluxes only have 3 components

Thanks! And are these fluxes or tendencies (divergence of fluxes)?

kpamnany:
> would it be easy to calculate horizontal mean for these tendencies?

Yes, but I'd prefer to do that in `DiagnosticsMachine`, in a separate PR. This is a simple full dump; something like what we're planning to do in case of a crash (kernel sees a NaN -> dump state and all tendencies -> exit).

charleskawczynski:
> And are these fluxes or tendencies (divergence of fluxes)?

These are fluxes and sources. Getting the divergence of fluxes is a bit more tricky, and we'll need to do that in a later PR.

charleskawczynski:
There are actually a bunch of tendency diagnostics-related PRs, I think we can probably close #1036 with this PR, since it mostly refers to fluxes/sources, whereas #1527 emphasizes the budget.

szy21:
Here are x-z slices of TotalMoisture_BomexTendencies_1. The top and bottom figures are the first (t=0) and the second (t=600) timestep in the output.
From the driver, the total moisture tendency is ρ∂qt∂t - state.ρ * w_s * dot(k̂, diffusive.moisture.∇q_tot). I wonder why the two timesteps are very different. It seems that the first figure doesn't have the subsidence tendency while the second figure has it. @akshaysridhar Any thoughts?
image
image

charleskawczynski:
> I wonder why the two timesteps are very different.

The magnitude appears very small

charleskawczynski:
Also, we could further break up those tendencies and re-run to further dissect 😄

szy21:
> The magnitude appears very small

They should be small (~1e-8). I was more curious why it's zero above 500m at t=0. ρ∂qt∂t is zero above 500m, and w_s is zero above 2100m. That's where my guess is from:)

charleskawczynski:
We'll need names for each one! 🎉

szy21:
> We'll need names for each one! 🎉

We could call them AdvectiveTendency and SubsidenceTendency, respectively. It's a bit tricky for Energy though, for which the "AdvectiveTendency" includes a fake radiative tendency...

charleskawczynski:
Does `AdvectiveTendency` makes sense in `source`? Also, maybe we can reuse the AtmosModel's `Subsidence`

szy21:
It's like large-scale advective tendency, so maybe `large-scale tendency`? I think we cannot use `Subsidence` before because w is a function of z, but I'm not sure whether it's still a problem.

src/Diagnostics/dump_tendencies.jl
```diff
@@ -0,0 +1,239 @@
...
+    t,
+    direction,
+)
+    _args_fx1 = (; state, aux, t, direction)
+    _args_fx2 = (; state, aux, t, diffusive, hyperdiffusive)
+    _args_src = (; state, aux, t, direction, diffusive)
+
+    cache_fx1 = precompute(bl, _args_fx1, Flux{FirstOrder}())
+    cache_fx2 = precompute(bl, _args_fx2, Flux{SecondOrder}())
+    cache_src = precompute(bl, _args_src, Source())
```
charleskawczynski, line 68: Another thing that would be super useful would be to return/flatten this named tuple and dump into the .nc file. I think we can tackle that in a separate PR, and we'll need to filter out some keys (e.g., `ts::ThermodynamicState`).

ilopezgp (approved):

kpamnany:
@szy21: should I merge this?

szy21:
> @szy21: should I merge this?

Yes, let's merge it. I don't think the problem (if any) is from this PR.

PR #1981, opened 2021-02-01
Land/soil/runoffv1

kmdeck:
### Description
This is a first pass at computing surface runoff, inspired by the switching boundary condition approach.

 If the precipitation rate exceeds the infiltration flux into the soil, a ponding condition (Dirichlet, state = porosity, i.e. soil is saturated) is appropriate; if the precipitation rate is smaller than the flux into the soil, we would want a flux condition (flux = precipitation). 

However, in both cases, we apply a Neumann BC. In the ponding case, we compute the flux that would result if we had instead set the correct dirichlet BC. This flux depends on the grid resolution - we approximate it using a simple g'(z) = [g(z+dz)-g(z)]/dz = [g_bc -g-]/dz. We thought this wouldnt work, but offline tests show that it converges to the right thing as dz-> 0. This is somewhat surprising to me, because g(z+dz) = g_bc in this case is not a computed value that is guaranteed to be continuous with the rest of the profile. The plot shows this - the bonan solution is what we expect when a dirichlet condition is set. the other curves show what we get when we set a neumann BC, and change the resolution. The flux at the surface does not diverge. this happens with FV and DG (npoly = 1). 
neumann_bc_converge

There are three tests for runoff in `test_physical_bc.jl`. The first shows that when we want to no runoff to occur, and all precipitation to enter the soil (flux_bc =P), that that is what we recover. The second shows that our approximation of the flux that would result if the dirichlet BC was set matches what occurs when we set the dirichlet BC directly. The last shows a case where ponding should not occur (we set P = 0). We also set the profile to be in hydrostatic equilibrium. we recover that nothing occurs as expected (the right bc - flux_bc = P =0  - occurs). So, we test that the two branches of the if/else statement do what we want. We do not yet test whether or not the threshold used in the if/else statement acts as we want, but this is an approach taken in other land models.


test/Land/Model/test_physical_bc.jl
```diff
@@ -0,0 +1,420 @@
...
+        S = FT((FT(1) + (α * (z - zm))^n)^(-m))
+        return FT(S * porosity)
+    end
+    N_poly = (1, 2)
+    nelem_vert = 50
+
+    # Specify the domain boundaries
+    zmax = FT(0)
+    zmin = FT(-0.35)
+    Δz = FT(0.35 / 50 / 2)
```
kmdeck, line 174: what should this be in DG? This seems to work? In FV the first point is half a layer down, so this defn makes sense for that. Can we avoid making the grid first and just use a representative value in DG as well? (note: land model likely will use FV in vertical, but whatever we do should work for both).

kmdeck:
keep seeing this error: https://github.com/CliMA/ClimateMachine.jl/runs/1912152211?check_suite_focus=true#step:7:382 not sure why?

test/Land/Model/test_physical_bc.jl
kmdeck: @Zhengyu-Huang do you know what value we would use in DG for the "delta z" in the denominator of the approximated flux (using the value we would use as a dirichlet condition at the surface to estimate the flux that would result)? In DG it's more of an approx because we are using the surface state (state-), so dz is ~0. But using a value typical for the spacing of points (Length of domain/(n_elem*N_poly)) seems to work ok?

kmdeck:
> keep seeing this error: https://github.com/CliMA/ClimateMachine.jl/runs/1912152211?check_suite_focus=true#step:7:382 not sure why?
> 
> error doesnt occur on my computer, only in CI

@jakebolewski noted that this happens when you run the tests with
`julia --project -e 'using Pkg; Pkg.test(test_args=["Land"])'` but not when you include the file in the julia REPL.

I rebased to make sure there was the latest version of the Project.toml...still occurs

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors cancel

kmdeck:
bors try

jakebolewski (approved):
It might be good to add a comment as to why we are skipping over the first timestep for posterity.

kmdeck:
> It might be good to add a comment as to why we are skipping over the first timestep for posterity.

Gradients are not defined at t=0, so if you retrieve the gradient flux variables before calling invoke!, you will be getting an un-initialized array of values back. On a Mac computer, these may get set to zero, but on other operating systems they may not be. This led to an issue for us where a test was passing on some systems but not on others. A recommendation to help avoid this might be to use `ClimateMachine.init(debug_init=true)` which will initialize any balance law state arrays to NaNs.

kmdeck:
> > keep seeing this error: https://github.com/CliMA/ClimateMachine.jl/runs/1912152211?check_suite_focus=true#step:7:382 not sure why?


comment regarding how to avoid error is in the conversation.

test/Land/Model/test_physical_bc.jl
```diff
@@ -0,0 +1,420 @@
...
+    ClimateMachine.invoke!(solver_config; user_callbacks = (callback,))
+
+    dons = dict_of_nodal_states(solver_config, state_types; interp = true)
+    push!(dons_arr, dons)
+    push!(time_data, gettime(solver_config.solver))
+
+    # Get z-coordinate
+    z = get_z(solver_config.dg.grid; rm_dupes = true)
+    N = length(dons_arr)
+    computed_surface_flux = [dons_arr[k]["soil.water.K∇h[3]"][end] for k in 2:N]
```
jakebolewski, line 134: add comment?

```# we take time 2:N here as gradients are not defined at t == 0```

File: src/Land/Model/Runoff.jl [hunk 3]
```diff
@@ -52,46 +57,114 @@ struct NoRunoff <: AbstractSurfaceRunoffModel end
...
+
+Chosen when no subgrid effects are to be modeled.
 """
-function compute_surface_runoff(
-    runoff_model::NoRunoff,
-    precip_model::AbstractPrecipModel{FT},
-    state::Vars,
-) where {FT}
-    return FT(0.0)
+struct CoarseGridRunoff{FT, F} <: AbstractSurfaceRunoffModel
```
kmdeck, line 64: Changing this to:
```
struct CoarseGridRunoff{FT} <: AbstractSurfaceRunoffModel
    "Mean vertical resolution at the surface"
    Δz::FT
end
```

src/Land/Model/Runoff.jl
```diff
@@ -52,46 +57,114 @@ struct NoRunoff <: AbstractSurfaceRunoffModel end
...
+    precip_model::AbstractPrecipModel,
+    n̂,
+    state⁻::Vars,
+    diff⁻::Vars,
+    aux⁻::Vars,
+    t::Real,
+)
+    FT = eltype(state⁻)
+    incident_water_flux = precip_model(t)
+    K∇h⁺ = n̂ * (-FT(2) * incident_water_flux) - diff⁻.soil.water.K∇h
```
kmdeck, line 166: using inflow flux here, I think this is OK

src/Land/Model/soil_bc.jl
```diff
@@ -255,12 +254,52 @@ function soil_boundary_flux!(
...
+        state⁻::Vars,
+        diff⁻::Vars,
+        aux⁻::Vars,
+        t,
+    )
+
+The Surface Driven BC method for `soil_boundary_flux!` for the
+`SoilWaterModel`.
+"""
+function soil_boundary_state!(
```
kmdeck, line 291: we dont need this again, because it will go to the default method

kmdeck (commented):
left comments

test/Land/Model/test_physical_bc.jl
kmdeck: will do!

test/Land/Model/test_physical_bc.jl
kmdeck: @Zhengyu-Huang any thoughts on this?

Zhengyu-Huang: For Dg, we should use the first nodal distance, it is ugly, but the vertical nodal distance is
```
using ClimateMachine.Grids
mygrid = solver_config.dg.grid
mygrid.vgeo[1, Grids._x3, 2] - mygrid.vgeo[1, Grids._x3, 1]
```
The first index is the element id, the last index is the quadrature id, the second index is the index for x3 direction

Zhengyu-Huang (approved):
May need to add a function to compute \Delta z , others look great for me, may l

kmdeck:
> May need to add a function to compute \Delta z , others look great for me, may l

is this (for a regular cartesian grid) the same as (zmin-zmax)/nelem_vert? i.e. the polynomial order doesnt affect it?

test/Land/Model/test_physical_bc.jl
kmdeck: Npoly doesnt affect this?

charleskawczynski: We should absolutely write a separate function, in `ClimateMachine.Grids` that computes this. Maybe `lower_Δz(::Grid)` and `upper_Δz(::Grid)`? There is no concern about laterally changing `Δz` here?

kmdeck: i dont think we'll change resolution near the surface laterally.. when I tested what Daniel sent, though, it looked like it was just (zmin-zmax)/nelem_vert. If that is the case (at least for cartesian coords) I almost want to punt on writing a function for it because then we end up having to make a grid with an empty model, then create the model, then make the actual grid using the same params...especially if it as simple as the size of a vertical element.

kmdeck: in FV the interpretation is more clear (dz = distance from boundary to first point). for DG the first point is at the surface. But perhaps a way to think about it is that in either case, we want to apply the same boundary condition (function of dz) between the two models if they have the same resolution, so we use the same value of dz, even if it is not the distance to the first point anymore?

kmdeck: dg_poly2_convergence


this is testing the above hypothesis - looks fine? it is using dz = 1/2(zmin-zmax)/nelem_vert as we'd do for FV.

kmdeck: Given this, my preference would be to use: `    Δz = abs(FT((zmin-zmax) / nelem_vert / 2))` and to open an issue reminding me that in the future we want a function which calculates this. if i understand correctly, it will only matter if we dont have cartesian/rectangular coordinates, or if we stretch the points. Getting it from the grid is great, but we dont have access to the grid at this point, before we make the model, and I think the solution to that particular issue is pending since it has come up for a number of people.

kmdeck: ok having spoken with Daniel we agree this makes sense for DG, but we also want to confirm it. Please see https://github.com/CliMA/ClimateMachine.jl/issues/2058.

kmdeck:
all land tests pass using `julia --project -e 'using Pkg; Pkg.test(test_args=["Land"])'` trying again...

PR #1988, opened 2021-02-02
Implement SurfaceFluxes in EDMF

ilopezgp:
### Description

This PR makes use of the SurfaceFluxes module for EDMF single stack simulations. The necessary changes with respect to the previous implementation are:

- Diagnostic of friction velocity and Obuhkov length, instead of hardcoding their values.
- Diagnostic of heat, momentum and moisture surface fluxes following Monin Obukhov Similarity Theory and Nishizawa et al (2018). These fluxes have to be passed to the grid mean and to the surface functions in the EDMF scheme.

It would probably be useful to cache the results of the iterative solver somewhere so that they do not have to be recomputed in different places. They are needed right now in the model files (e.g. `stable_bl_model.jl`), `surface_functions.jl` and the obukhov length should be accessible to `mixing_length.jl`.

There is also the need to pass the boundary state and the first interior point state to SurfaceFluxes. This is easy for the pass in `surface_functions.jl`, but not for `stable_bl_model.jl`.

### DIFF
--- a/experiments/AtmosLES/stable_bl_model.jl
+++ b/experiments/AtmosLES/stable_bl_model.jl
@@ -35,6 +35,7 @@ using ClimateMachine.TurbulenceClosures
 using ClimateMachine.TurbulenceConvection
 using ClimateMachine.VariableTemplates
 using ClimateMachine.BalanceLaws
+using ClimateMachine.SurfaceFluxes
 import ClimateMachine.BalanceLaws: source
 
 using CLIMAParameters
@@ -259,6 +260,9 @@ function stable_bl_model(
             (state, aux, t, normPu_int) -> C_drag,
             (state, aux, t) -> q_sfc,
         )
+        # elseif surface_flux == "Nishizawa2018"
+        #     energy_bc = PrescribedEnergyFlux((state, aux, t) -> LHF + SHF)
+        #     moisture_bc = PrescribedMoistureFlux((state, aux, t) -> moisture_flux)
     else
         @warn @sprintf(
             """
--- a/src/Common/SurfaceFluxes/SurfaceFluxes.jl
+++ b/src/Common/SurfaceFluxes/SurfaceFluxes.jl
@@ -114,10 +114,8 @@ Surface conditions given
  - `x_ave` volume-averaged value for variable `x`
  - `x_s` surface value for variable `x`
  - `z_0` roughness length for variable `x`
- - `F_exchange` flux at the top for variable `x`
  - `VDSE_scale` virtual dry static energy (i.e., basic potential temperature)
- - `Δz` layer thickness (not spatial discretization)
- - `z` coordinate axis
+ - `Δz` layer thickness
  - `VDSE_flux_star` potential temperature flux (optional)
 
 If `VDSE_flux_star` is not given, then it is computed by iteration
@@ -129,11 +127,8 @@ function surface_conditions(
     x_ave::AbstractVector,
     x_s::AbstractVector,
     z_0::AbstractVector,
-    F_exchange::AbstractVector,
     VDSE_scale::FT,
-    qt_bar::FT,
     Δz::FT,
-    z::FT,
     scheme,
     VDSE_flux_star::Union{Nothing, FT} = nothing,
 ) where {FT <: AbstractFloat, AbstractEarthParameterSet}
@@ -143,7 +138,6 @@ function surface_conditions(
     @assert length(x_ave) == n_vars
     @assert length(x_s) == n_vars
     @assert length(z_0) == n_vars
-    @assert length(F_exchange) == n_vars
     local sol
 
     args = (;
@@ -181,17 +175,8 @@ function surface_conditions(
     VDSE_flux_star = -u_star^3 * VDSE_scale / (_von_karman_const * _grav * L_MO)
     flux = -u_star * x_star
 
-    K_exchange = exchange_coefficients(
-        param_set,
-        z,
-        F_exchange,
-        x_star,
-        VDSE_scale,
-        L_MO,
-    )
-
     C_exchange =
-        get_flux_coefficients(param_set, z, x_star, VDSE_scale, L_MO, z_0)
+        get_flux_coefficients(param_set, Δz, x_star, VDSE_scale, L_MO, z_0)
 
     VFT = typeof(flux)
     return SurfaceFluxConditions{FT, VFT}(
--- a/test/Atmos/EDMF/closures/mixing_length.jl
+++ b/test/Atmos/EDMF/closures/mixing_length.jl
@@ -49,10 +49,6 @@ function mixing_length(
     tke_en = max(en.ρatke, 0) * ρinv / env.a
 
     # buoyancy related functions
-    # compute obukhov_length and ustar from SurfaceFlux.jl here
-    ustar = m.turbconv.surface.ustar
-    obukhov_length = m.turbconv.surface.obukhov_length
-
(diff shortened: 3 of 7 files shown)


test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -63,14 +59,8 @@ function mixing_length(
...
 
     # compute L2 - law of the wall
-    surf_vals = subdomain_surface_values(
-        m.turbconv.surface,
-        m.turbconv,
-        m,
-        gm,
-        gm_aux,
-        m.turbconv.surface.zLL,
-    )
```
ilopezgp, line 73: I removed this because I don't think we can call `subdomain_surface_values` here: The states passed to that function are assumed to be at the surface, not at any height. (This function was not used though, it was a placeholder for later use).

experiments/AtmosLES/stable_bl_model.jl
```diff
@@ -259,6 +260,9 @@ function stable_bl_model(
             (state, aux, t, normPu_int) -> C_drag,
             (state, aux, t) -> q_sfc,
         )
+        # elseif surface_flux == "Nishizawa2018"
+        #     energy_bc = PrescribedEnergyFlux((state, aux, t) -> LHF + SHF)
+        #     moisture_bc = PrescribedMoistureFlux((state, aux, t) -> moisture_flux)
```
ilopezgp, line 265: Any ideas on how to pass the boundary and first interior point states to the call that we would need here, @charleskawczynski ?

test/Atmos/EDMF/closures/surface_functions.jl
```diff
@@ -1,6 +1,8 @@
 #### Surface model kernels
 
 using Statistics
+using ClimateMachine.SurfaceFluxes
+const ArrayType = ClimateMachine.array_type()
```
charleskawczynski, line 5:
```suggestion
```

test/Atmos/EDMF/closures/surface_functions.jl
```diff
@@ -30,26 +36,71 @@ function subdomain_surface_values(
...
-    # these value should be given from the SurfaceFluxes.jl once it is merged
-    oblength = turbconv.surface.obukhov_length
-    ustar = turbconv.surface.ustar
+    # Retrieve surface fluxes based on MO similarity
+    ## Initial guesses for MO parameters
+    LMO_init = eps(FT)
+    u_star_init = FT(0.1)
+    th_star_init = eps(FT)
+    qt_star_init = eps(FT)
+    x_init = ArrayType(FT[LMO_init, u_star_init, th_star_init, qt_star_init])
```
charleskawczynski, line 66:
```suggestion
    x_init = MArray{Tuple{4}, FT}(LMO_init, u_star_init, th_star_init, qt_star_init)
```
We might be able to get away with using tuples for inputs: `x_init = (LMO_init, u_star_init, th_star_init, qt_star_init)` because `surface_conditions` is allocating. But `MArray`'s should work here

experiments/AtmosLES/stable_bl_model.jl
```diff
@@ -35,6 +35,7 @@ using ClimateMachine.TurbulenceClosures
 using ClimateMachine.TurbulenceConvection
 using ClimateMachine.VariableTemplates
 using ClimateMachine.BalanceLaws
+using ClimateMachine.SurfaceFluxes
```
charleskawczynski, line 38:
```suggestion
using ClimateMachine.SurfaceFluxes
using ClimateMachine.SurfaceFluxes: DGScheme
```

test/Atmos/EDMF/closures/surface_functions.jl
```diff
@@ -30,26 +36,71 @@ function subdomain_surface_values(
...
+
+    surf_flux_cond = surface_conditions(
+        atmos.param_set,
+        x_init,
+        x_ave,
+        x_s,
+        z_rough,
+        VDSE_scale,
+        zLL,
+        ClimateMachine.SurfaceFluxes.DGScheme(),
```
charleskawczynski, line 94:
```suggestion
        DGScheme(),
```

test/Atmos/EDMF/closures/surface_functions.jl
charleskawczynski: Will need to apply this elsewhere

src/Common/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -114,10 +114,8 @@ Surface conditions given
  - `x_ave` volume-averaged value for variable `x`
  - `x_s` surface value for variable `x`
  - `z_0` roughness length for variable `x`
- - `F_exchange` flux at the top for variable `x`
  - `VDSE_scale` virtual dry static energy (i.e., basic potential temperature)
- - `Δz` layer thickness (not spatial discretization)
- - `z` coordinate axis
+ - `Δz` layer thickness
```
charleskawczynski, line 118: We might want to change the surface fluxes pieces in a separate PR. I can take this on if you'd like

charleskawczynski: Actually, I don't think we'll need a ton of changes

experiments/AtmosLES/stable_bl_model.jl
charleskawczynski: I think we have two options:
 - add properties (`state_int`, `aux_int`) to `PrescribedEnergyFlux`, which we can then use in, for example, `src/Atmos/Model/bc_energy.jl`.
 - Update the anonymous function to also use `state_int`, `aux_int`. I.e., change `energy_bc = PrescribedEnergyFlux((state, aux, t) -> LHF + SHF)` to `energy_bc = PrescribedEnergyFlux((state, aux, t, state_int, aux_int) -> LHF + SHF)`
I think both will require changing `src/Atmos/Model/bc_energy.jl` and `src/Atmos/Model/bc_moisture.jl` etc.

charleskawczynski: I think I'd probably lean towards the latter option

test/Atmos/EDMF/closures/mixing_length.jl
```diff
@@ -49,10 +49,6 @@ function mixing_length(
     tke_en = max(en.ρatke, 0) * ρinv / env.a
 
     # buoyancy related functions
-    # compute obukhov_length and ustar from SurfaceFlux.jl here
-    ustar = m.turbconv.surface.ustar
```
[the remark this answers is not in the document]
ilopezgp: Good point. Yes, we will need an option to hardcode the friction velocity for some idealized cases.

ilopezgp:
This PR depends on #2038.

ilopezgp:
This PR depends on #2072. With the added changes from that PR, we would like to

- Construct the `SurfaceFluxConditions` structure within a common `bc_precompute`, and distribute it to the `bc_val` of each BC type.
- Add a dependency of `bc_precompute` on the energy BC type. If the energy fluxes are prescribed, `SurfaceFluxConditions` is only used to obtain the Monin-Obukhov parameters. Otherwise, the flux is constructed from the output of `SurfaceFluxConditions`, either by using a bulk formula with `C_exchange`, or using the returned flux directly.

Apart from this, the Monin Obukhov length is not readily available to the `mixing_length` function at each nodal point due to the fact that this call would be nonlocal, as noted by @charleskawczynski. Changes in the structure of the code would be required to allow this call, but this is **not** an urgent task so it can be left for future work.

PR #1997, opened 2021-02-03
Change the length scale in hyperdiffusion to horizontal nodal distance

szy21:
### Description

Co-authored-by: @akshaysridhar 

This PR :

- Adds a function to calculate the horizontal effective resolution in Geometry.jl
- Changes the length scale in hyperdiffusion(`DryBiharmonic` and `EquilMoistBiharmonic`) to the horizontal effective resolution. 

The parameters in the hyperdiffusion are e.g. `(ν₄, aux.hyperdiffusion.Δ, τ_timescale) = (3.359104202786882e15, 235880.1990528821, 28800.0)`, which are reasonable.

Thank @jkozdon for his help on the calculation of horizontal nodal distance!

### DIFF
--- a/src/Numerics/Mesh/Geometry.jl
+++ b/src/Numerics/Mesh/Geometry.jl
@@ -19,7 +19,7 @@ using ..Grids:
     _x3,
     _JcV
 
-export LocalGeometry, lengthscale, resolutionmetric
+export LocalGeometry, lengthscale, resolutionmetric, lengthscale_horizontal
 
 """
     LocalGeometry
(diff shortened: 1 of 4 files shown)


jiahe23:
LGTM

jiahe23:
Do we want to change the `lengthscale` to be `lengthscale_volume` (or some better name)?

szy21:
> Do we want to change the `lengthscale` to be `lengthscale_volume` (or some better name)?

@akshaysridhar What do you think?

szy21:
@jkozdon I would like to add a unit test for `lengthscale_horizontal` in `test/Numerics/Mesh/Geometry.jl`. Any suggestion?

jiahe23 (approved):

src/Numerics/Mesh/Geometry.jl
```diff
@@ -121,4 +121,33 @@ The effective grid resolution at the point.
...
+    #    horizontal direction 2: √( ∑_j ∂x_2 / ∂ξ_j) * 2
+    #
+    # To get these we need to have the Jacobian matrix:
+    #
+    #    J[i,j] =  ∂x_i / ∂ξ_j
+    #
+    # To get this we can solve the system with the unit basis vectors
+    # (division by `N` gives the local average node distance)
+    Δ1 = norm(invJ \ SVector(1, 0, 0)) * 2 / g.polyorder[1]
+    Δ2 = norm(invJ \ SVector(0, 1, 0)) * 2 / g.polyorder[2]
```
akshaysridhar, line 147: \xi_1, \xi_2 represent the 2 tangential directions, and the Jacobian inverse is available in stored local geometry properties; this is a reasonable estimate for the (local) horizontal lengthscale.

test/Numerics/Mesh/Geometry.jl
```diff
@@ -45,15 +45,16 @@ MPI.Initialized() || MPI.Init()
...
-    N = polynomialorder
-    Np = (N + 1)^3
+    N = (polynomialorder, polynomialorder)
+    Np = (polynomialorder + 1)^3
     for e in 1:size(grid.vgeo, 3)
         for n in 1:size(grid.vgeo, 1)
             g = LocalGeometry{Np, N}(grid.vgeo, n, e)
             @test lengthscale(g) ≈ Savg
-            @test Geometry.resolutionmetric(g) ≈ M
+            @test lengthscale_horizontal(g) ≈ Shoriavg
```
akshaysridhar, line 57: Why is the geometric average (via resolutionmetric)check being removed from the test here?

akshaysridhar (commented):
> Do we want to change the `lengthscale` to be `lengthscale_volume` (or some better name)?

This is still a measure of the (local) geometric mean resolution, so some name that highlights this could be useful. Perhaps updating the docstring from `The effective grid resolution at the point.` to `The effective (geometric mean) grid resolution at the point.` is useful; this should clarify the returned value.

test/Numerics/Mesh/Geometry.jl
szy21: Because it's not updated for the horizontal and vertical splitting. I checked with @jkozdon but currently we don't know how to change it. On the other hand, `resolutionmetric` is not used anywhere in the code.

Issue #1765, opened 2020-11-18
Better naming scheme for output?


szy21:
### Description

Follow-up on the discussion with @jiahe23 @kpamnany @jakebolewski

Currently the output is named with an appended timestamp, which is inconvenient for scripted runs. What would be a better naming scheme?

glwagner:
What's the motivation for having a naming scheme?

Here's my alternative viewpoint: I think it's best to refrain from imposing any naming scheme at all except when necessary. The main exception is when output must be split into multiple files, either because:

* computations are performed on distributed architectures (so we have `_rankXXXX`, etc)
* restrictions are imposed on the file size, either due to user dictate or for the purpose of having a practical default (eg, checkpointing)

This is the approach in Oceananigans. For checkpointing, we append user-provided prefixes with `_iterationXXXXXX` where `XXXXXX` is the model iteration the checkpoint pertains to. If users specify a `max_filesize`, we append user-provided prefixes with `_partX`, where the `X`s are ordered chronologically starting at 1.

Otherwise, user input is either respected or appended with the appropriate file extension.

jiahe23:
> What's the motivation for having a naming scheme?
> 
> Here's my alternative viewpoint: I think it's best to refrain from imposing any naming scheme at all except when necessary. The main exception is when output must be split into multiple files, either because:
> 
> * computations are performed on distributed architectures (so we have `_rankXXXX`, etc)
> * restrictions are imposed on the file size, either due to user dictate or for the purpose of having a practical default (eg, checkpointing)
> 
> This is the approach in Oceananigans. For checkpointing, we append user-provided prefixes with `_iterationXXXXXX` where `XXXXXX` is the model iteration the checkpoint pertains to. If users specify a `max_filesize`, we append user-provided prefixes with `_partX`, where the `X`s are ordered chronologically starting at 1.
> 
> Otherwise, user input is either respected or appended with the appropriate file extension.

The current netcdf output is named as `ExperimentName_DiagnosticsType_TimestampOfSim.nc`. The timestamp is to differentiate new sim outputs (e.g., restart run) from existing data file but it is quite inconvenient for setting up auto postprocessing scriprs/notebooks. It also makes a lot more work for the weekly test and visualization. We might want to remove the timestamp but think about other ways of differentiate between simulations.

valeriabarra:
Could another option be `ExperimentName_DiagnosticsType_NumerOfTimeStep.nc` ? 

[...]

PR #1998, opened 2021-02-03
Enable custom suffix for LES default diagnostics

ilopezgp:
### Description

This PR enables appending a suffix to the name of the LES default diagnostics netcdf file. The suffix is passed as an optional argument to the setup of the diagnostics, and is contained in the `DiagnosticsParams` structure. If `!nothing`, the suffix substitutes the default timestamp suffix.

This feature is very useful for handling batches of simulations that are performed in parallel, where the timestamp-based suffix is not informative.

### DIFF
--- a/experiments/AtmosLES/stable_bl_les.jl
+++ b/experiments/AtmosLES/stable_bl_les.jl
@@ -75,7 +75,8 @@ function main(cl_args)
         init_on_cpu = true,
         Courant_number = CFLmax,
     )
-    dgn_config = config_diagnostics(driver_config)
+    dgn_config =
+        config_diagnostics(driver_config, out_suffix = cl_args["cparam_file"])
 
     check_cons = (
         ClimateMachine.ConservationCheck("ρ", "1mins", FT(0.0001)),
--- a/experiments/AtmosLES/stable_bl_model.jl
+++ b/experiments/AtmosLES/stable_bl_model.jl
@@ -318,11 +318,12 @@ function stable_bl_model(
     return model
 end
 
-function config_diagnostics(driver_config)
+function config_diagnostics(driver_config; out_suffix = nothing)
     default_dgngrp = setup_atmos_default_diagnostics(
         AtmosLESConfigType(),
         "2500steps",
-        driver_config.name,
+        driver_config.name;
+        out_suffix = out_suffix,
     )
     core_dgngrp = setup_atmos_core_diagnostics(
         AtmosLESConfigType(),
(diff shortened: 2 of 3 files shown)


charleskawczynski (commented):
I'm fine with these changes, but I'll leave this up to @kpamnany, since this touches diagnostics.

ilopezgp:
bors try

kpamnany (commented):
This is a nicely done solution, but it has a couple of problems:
1. It pushes the naming problem to the user who has to specify a command line parameter.
2. It's a singular solution, which only addresses the issue for `AtmosLESDefault` and this particular experiment.

As discussed in #1765, the plan was to address this by removing the timestamp suffix altogether, for all the diagnostics groups. This would cause existing output files to be overwritten, but that problem can be managed by changing `output_dir`.

If this would be an acceptable solution for you, please let me know and I can open that PR today. It would also move [this](https://github.com/CliMA/ClimateMachine.jl/blob/4c22bfefdfc92c821453b72e8e204cea92d002e6/src/Driver/Driver.jl#L546) to `invoke!`, so that apart from being able to specify `output_dir` on the command line, you could also add:
```julia
ClimateMachine.Settings.output_dir = cl_args["cparam_file"] * ".output"
```
Over [here](https://github.com/CliMA/ClimateMachine.jl/blob/4c22bfefdfc92c821453b72e8e204cea92d002e6/src/Driver/Driver.jl#L546).

Thoughts?

ilopezgp:
> This is a nicely done solution, but it has a couple of problems:
> 
> 1. It pushes the naming problem to the user who has to specify a command line parameter.
> 2. It's a singular solution, which only addresses the issue for `AtmosLESDefault` and this particular experiment.
> 
> As discussed in #1765, the plan was to address this by removing the timestamp suffix altogether, for all the diagnostics groups. This would cause existing output files to be overwritten, but that problem can be managed by changing `output_dir`.
> 
> If this would be an acceptable solution for you, please let me know and I can open that PR today. It would also move [this](https://github.com/CliMA/ClimateMachine.jl/blob/4c22bfefdfc92c821453b72e8e204cea92d002e6/src/Driver/Driver.jl#L546) to `invoke!`, so that apart from being able to specify `output_dir` on the command line, you could also add:
> 
> ```julia
> ClimateMachine.Settings.output_dir = cl_args["cparam_file"] * ".output"
> ```
> 
> Over [here](https://github.com/CliMA/ClimateMachine.jl/blob/4c22bfefdfc92c821453b72e8e204cea92d002e6/src/Driver/Driver.jl#L546).
> 
> Thoughts?

The solution you propose (changing the output dir name) would be perfect for me. I agree that it is much cleaner as it can be done for all diagnostics at once. Thanks!

ilopezgp:
@kpamnany you can close this PR if you want.

kpamnany:
Superseded by #2005.

Issue #1995, opened 2021-02-03
Pass grid information to TurbulenceClosures.jl


szy21:
### Description

We would like to access grid information, at least min_node_distance in TurbulenceClosures.jl. For example, when calculating the [lengthscale](https://github.com/CliMA/ClimateMachine.jl/blob/5d3658ff57b95622f2c07d1d46bad718408db770/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L912) for hyperdiffusion we use the geometric mean, but what we need is the min_nod_distance in the horizontal direction.

@charleskawczynski @simonbyrne What would be the best way to access grid information? Thanks!

charleskawczynski:
Just spoke with @akshaysridhar about this. The basic idea of my suggestion is to call `min_node_distance` inside the `DiscontinuousSpectralElementGrid` constructor and store it as a property. That way we'll have access inside `update_auxiliary_state!`, and we'll just need to update the interface to include an extra argument.

PR #2001, opened 2021-02-04
Helper functions for nodal distances via empty-model

akshaysridhar:
### Description

Includes two functions to allow access to grid prior to generating AtmosModel instance in GCM / LES configurations so minimal nodal distances can be accessed by model subcomponents without being stored in aux variables. Includes additional driver-keyword to provide an option to suppress model information (default setting will print info to screen). 

Closes issue https://github.com/CliMA/ClimateMachine.jl/issues/1995 
 
- [N/A] Code is exercised in an integration test OR N/A.


src/Driver/driver_configs.jl
```diff
@@ -404,6 +406,60 @@ Establishing Atmos GCM configuration for %s
     )
 end
 
+export EmptyParamSet
+export init_empty!, get_grid_spacing_les, get_grid_spacing_gcm
+struct EmptyParamSet <: AbstractEarthParameterSet end
```
charleskawczynski, line 411: We'll need to remove this-- we're restricting these definitions to top-level files, outside of `src/`.

src/Driver/driver_configs.jl
```diff
@@ -404,6 +406,60 @@ Establishing Atmos GCM configuration for %s
     )
 end
 
+export EmptyParamSet
+export init_empty!, get_grid_spacing_les, get_grid_spacing_gcm
```
charleskawczynski, line 410: Let's move these to the top of file.

src/Driver/driver_configs.jl
```diff
@@ -404,6 +406,60 @@ Establishing Atmos GCM configuration for %s
...
+        "EmptyModel",
+        N_poly,
+        (nelem_horz, nelem_vert),
+        domain_height,
+        EmptyParamSet(),
+        init_empty;
+        suppress_model_info = true,
+    )
+    Δh = min_node_distance(empty_config.grid, HorizontalDirection())
+    Δv = min_node_distance(empty_config.grid, VerticalDirection())
```
valeriabarra, line 455: I'm a bit confused. Is this calculating the grid size before declaring any discretization (i.e., either SEM or FV, or any other type)? If so, how does this differ from the [resolution](https://github.com/CliMA/ClimateMachine.jl/blob/6026b0a2ea86e6fdf0a3fc764635a4b8a81000ca/src/Driver/driver_configs.jl#L205) used to compute the [elemsize](https://github.com/CliMA/ClimateMachine.jl/blob/6026b0a2ea86e6fdf0a3fc764635a4b8a81000ca/src/Driver/driver_configs.jl#L225)?

test/Atmos/Model/driver_spacing_test.jl
```diff
@@ -0,0 +1,40 @@
...
+using ClimateMachine.ConfigTypes
+using ClimateMachine.GenericCallbacks
+using ClimateMachine.DGMethods
+using ClimateMachine.Mesh.Grids
+using ClimateMachine.Mesh.Geometry
+using ClimateMachine.Orientations
+
+using Test
+
+@testset "EmptyModel Get Nodal Distance GCM" begin
```
valeriabarra, line 16: Based on the same question I had above, should this really be printing `"EmptyModel Get Element Size GCM"`or `"EmptyModel Get Grid Size GCM"` (if I am not misunderstanding the code)? Thanks

test/Atmos/Model/driver_spacing_test.jl
```diff
@@ -0,0 +1,40 @@
...
+    FT = Float64
+    Npoly = (5, 5)
+    Nelem = (8, 5)
+    Δh, Δv = get_grid_spacing_gcm(Npoly, Nelem, FT(30e3))
+    # Values established from "complete" driver run 
+    # min_node_distance is independently tested.
+    @test Δh ≈ 103918.2452822826
+    @test Δv ≈ 704.834028207901
+end
+@testset "EmptyModel Get Nodal Distance LES" begin
```
valeriabarra, line 26: Same question as  on line 16 above. Should this be printing `"EmptyModel Get Element Size LES"` or `"EmptyModel Get Grid Size LES"`? (again, unless I am completely misunderstanding)

src/Driver/driver_configs.jl
```diff
@@ -215,7 +216,7 @@ function AtmosLESConfiguration(
         (xmax, ymax, zmax) = ClimateMachine.Settings.domain_max
     end
 
-    print_model_info(model, mpicomm)
+    !suppress_model_info ? print_model_info(model, mpicomm) : nothing
```
kpamnany, line 219:
```suggestion
    suppress_model_info || print_model_info(model, mpicomm)
```

src/Driver/driver_configs.jl
```diff
@@ -404,6 +406,60 @@ Establishing Atmos GCM configuration for %s
...
 end
 
+export EmptyParamSet
+export init_empty!, get_grid_spacing_les, get_grid_spacing_gcm
+struct EmptyParamSet <: AbstractEarthParameterSet end
+function init_empty!(problem, bl, state, aux, localgeo, t)
+    nothing
+end
+function get_grid_spacing_les(
+    N_poly::Union{Int, NTuple{2, Int}},
```
kpamnany, line 416:
```suggestion
function get_grid_spacing(
    ::AtmosLESConfigType,
    N_poly::Union{Int, NTuple{2, Int}},
```

src/Driver/driver_configs.jl
```diff
@@ -404,6 +406,60 @@ Establishing Atmos GCM configuration for %s
...
+        init_empty;
+        suppress_model_info = true,
+    )
+    Δh = min_node_distance(empty_config.grid, HorizontalDirection())
+    Δv = min_node_distance(empty_config.grid, VerticalDirection())
+    return (Δh, Δv)
+end
+
+function get_grid_spacing_gcm(
+    N_poly::Union{Int, NTuple{2, Int}},
```
kpamnany, line 440:
```suggestion
function get_grid_spacing(
    ::AtmosGCMConfigType,
    N_poly::Union{Int, NTuple{2, Int}},
```

kpamnany (approved):
The new `get_grid_spacing` functions should probably be in a new file: `src/Driver/utilities.jl` or something?

charleskawczynski (commented):
I think I have some ideas on how to improve this

src/Driver/driver_configs.jl
kmdeck: maybe along those lines - the name could be `get_min_node_distance` instead of `get_grid_spacing` to be clearer?

kmdeck:
This is a general comment, but it feels odd that we need to create a model, create a param set, create even an `init_les!` function...just to get a grid resolution. It has to be this way now because we create the grid when the driver config is created. maybe that can be split into two pieces - a function creating the grid, which the user could call, and then they could define the model, and then the model and grid would be passed to the Driver config creator function.

This would be a bigger refactor, though.

kmdeck (commented):
This is a general comment, but it feels odd that we need to create a model, create a param set, create even an init_les! function...just to get a grid resolution. It has to be this way now because we create the grid when the driver config is created. maybe that can be split into two pieces - a function creating the grid, which the user could call, and then they could define the model, and then the model and grid would be passed to the Driver config creator function.

This would be a bigger refactor, though.

kpamnany:
> This is a general comment, 

Completely agree. Until now, it hasn't been a problem, and these driver configuration functions exist solely to reduce boilerplate code in the experiments. Splitting them would raise the question of what purpose they serve.

This needs a rethink.

kpamnany:
Until a more general refactoring of the driver configurations can be done (replacing them with [`Domains`](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Ocean/Domains/rectangular_domain.jl) from the Ocean model), if the need is to have the grid created before the model, will the following work in the experiment code (from [`surfacebubble.jl`](https://github.com/CliMA/ClimateMachine.jl/blob/master/experiments/AtmosLES/surfacebubble.jl#L114) as an example)?
```julia
    config = ClimateMachine.AtmosLESConfiguration(
        "SurfaceDrivenBubble",
        N,
        resolution,
        xmax,
        ymax,
        zmax,
        param_set,
        init_surfacebubble!,
        solver_type = ode_solver,
        model = nothing,
    )
    # Use config.grid to get minimum nodal distance or whatever is needed
    config.bl = AtmosModel{FT}(
        AtmosLESConfigType,
        param_set;
        problem = problem,
        moisture = EquilMoist{FT}(),
        source = (Gravity(),),
    )
```
You will still need the `suppress_model_info` parameter added in this PR, _or_ to add `isnothing(model) && return` [here](https://github.com/CliMA/ClimateMachine.jl/blob/as/driver-nodal-dist/src/Driver/driver_configs.jl#L151).

src/Driver/driver_configs.jl
akshaysridhar: Updated names for clarity

akshaysridhar: > I'm a bit confused. Is this calculating the grid size before declaring any discretization (i.e., either SEM or FV, or any other type)? If so, how does this differ from the [resolution](https://github.com/CliMA/ClimateMachine.jl/blob/6026b0a2ea86e6fdf0a3fc764635a4b8a81000ca/src/Driver/driver_configs.jl#L205) used to compute the [elemsize](https://github.com/CliMA/ClimateMachine.jl/blob/6026b0a2ea86e6fdf0a3fc764635a4b8a81000ca/src/Driver/driver_configs.jl#L225)?

In the current driver setup it uses the empty model to get the discretization type; then allows this to be passed to `<: BalanceLaw`

valeriabarra: Thanks @akshaysridhar Shouldn't also the name of the return variables, e.g. `Δh` and `Δv`, should reflect this? Say, call them `minΔh`, `minΔv`? I am just a bit concerned because earlier on  in this same file, we use `(Δx, Δy, Δz) = ClimateMachine.Settings.resolution` in `AtmosLESConfiguration` and that is for the average node distance, not the min one

PR #2002, opened 2021-02-04
update resolutionmetric for the horizontal and vertical splitting

szy21:
### Description

Fix a bug in `resolutionmetric` which results in an error for the horizontal and vertical splitting.

### DIFF
--- a/src/Numerics/Mesh/Geometry.jl
+++ b/src/Numerics/Mesh/Geometry.jl
@@ -109,7 +109,7 @@ Cartesian coordinates and `M = resolutionmetric(g)`, `sqrt(u'*M*u)` is the
 degree-of-freedom density in the direction of `u`.
 """
 function resolutionmetric(g::LocalGeometry)
-    S = g.polyorder * g.invJ / 2
+    S = SDiagonal(g.polyorder) * g.invJ / 2
     S' * S # TODO: return an eigendecomposition / symmetric object?
 end
 


jkozdon:
I agree that what is there right now is not quite right (assumes an isotopic grid). I don't think this change is quite right either, since it's saying that mesh skew ness is defined by the polynomial order, but that's sort of unrelated. What we should be doing is having something that depends not on the Jacobian determinant, but the Jacobian matrix (or something related to this). I'd need to understand / look at how and where this is used to make a specific recommendation.

szy21:
Thanks Jeremy. It seems the only place that uses `resolutionmetric` is in the [test](https://github.com/CliMA/ClimateMachine.jl/blob/159d9aae514395bd391d18e5a0dae17604acc43d/test/Numerics/Mesh/Geometry.jl#L56), which assumes the same horizontal and vertical polynomial orders.

jkozdon:
It is a bit odd that this is just in a test:
```
❯ rg resolutionmetric
src/Common/TurbulenceClosures/TurbulenceClosures.jl
47:import ..Mesh.Geometry: LocalGeometry, resolutionmetric, lengthscale

src/Numerics/Mesh/Geometry.jl
22:export LocalGeometry, lengthscale, resolutionmetric
105:    resolutionmetric(g::LocalGeometry)
108:Cartesian coordinates and `M = resolutionmetric(g)`, `sqrt(u'*M*u)` is the
111:function resolutionmetric(g::LocalGeometry)

src/Atmos/Model/AtmosModel.jl
77:    LocalGeometry, lengthscale, resolutionmetric, DGModel

docs/src/APIs/Numerics/Meshes/Mesh.md
52:Geometry.resolutionmetric

test/Numerics/Mesh/Geometry.jl
56:            @test Geometry.resolutionmetric(g) ≈ M
```

Are you trying to use it for something? Or should the function just be deleted?

szy21:
This is related to #2003. So in the hyperdiffusion in TurbulenceClosures.jl the [length scale](https://github.com/CliMA/ClimateMachine.jl/blob/0d2162739dc3b9da844d2623350640832db253f3/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L912) uses the geometric mean. But what should be used is the horizontal (min) nodal distance. [Here](https://github.com/CliMA/ClimateMachine.jl/blob/4002748977ec28ea94250a58b62556f677846e3d/src/Common/TurbulenceClosures/TurbulenceClosures.jl#L915) is how we try to get the horizontal nodal distance using `resolutionmetric`. (Note this is not correct because we haven't figured out how to get the unit vector in the latitude or longitude direction.) Is there a good way to get the horizontal nodal distance?

jkozdon:
By using minimum node distance are you really trying to get at element size in the horizontal? Min node distance seems a bit odd since it's dependent on the choice of basis functions, which doesn't make sense to me really.

szy21:
> By using minimum node distance are you really trying to get at element size in the horizontal? Min node distance seems a bit odd since it's dependent on the choice of basis functions, which doesn't make sense to me really.

I'm trying to follow the design docs to get the minimal horizontal node distance. But if it is strange the horizontal element size (node distance?) is fine too. I think the more important thing is only the horizontal length scale matters for the hyperdiffusion.

@tapios Could you clarify what we should use here?

tapios:
> > By using minimum node distance are you really trying to get at element size in the horizontal? Min node distance seems a bit odd since it's dependent on the choice of basis functions, which doesn't make sense to me really.
> 
> I'm trying to follow the design docs to get the minimal horizontal node distance. But if it is strange the horizontal element size (average node distance?) is fine too. I think the more important thing is only the horizontal length scale matters for the hyperdiffusion. What is used in NUMA?
> 
> @tapios Could you clarify what we should use here?

Yes, either minimum nodal distance or average (e.g., geometric or arithmetic mean) nodal distance is fine, but as you say, it's important it is only in the horizontal.

jkozdon:
What about avoiding the resolution metric all together?
```julia
function init_aux_hyperdiffusion!(
    ::DryBiharmonic,
    bl::BalanceLaw,
    aux::Vars,
    geom::LocalGeometry,
)
    # inverse Jacobian matrix:
    #    invJ[i,j] =  ∂ξ_i / ∂x_j
    invJ = geom.invJ

    # The local horizontal grid horizontal are:
    #
    #    horizontal direction 1: √( ∑_j ∂x_1 / ∂ξ_j) / 2
    #    horizontal direction 2: √( ∑_j ∂x_2 / ∂ξ_j) / 2
    #
    # To get these we need to have the Jacobian matrix:
    #
    #    J[i,j] =  ∂x_i / ∂ξ_j
    #
    # To get this we can solve the system with the unit basis vectors
    # (division by `N` gives the local average node distance)

    Δ1 = norm(invJ \ SVector(1, 0, 0)) / 2geo.polyorder[1]
    Δ2 = norm(invJ \ SVector(0, 1, 0)) / 2geo.polyorder[2]

    # Set hyperdiffusion to the max of the two calculated values
    aux.hyperdiffusion.Δ = max(Δ1, Δ2)
    # or hypot?
    # aux.hyperdiffusion.Δ = hypot(Δ1, Δ2)
end
```

I think that this might be closer to what you actually want.

szy21:
Thanks Jeremy! I will give it a try.

szy21:
@jkozdon I wonder if you mean `Δ1 = norm(invJ \ SVector(1, 0, 0)) / 2 / geom.polyorder[1]`? With your code the `aux.hyperdiffusion.Δ` I get is about 60km, while the min horizontal nodal distance is ~100km and I expect the average horizontal nodal distance to be ~200km.

jkozdon:
These should be equivalent `2geom.polyorder[1]` should be the same as `2 * geom.polyorder[1]`.

Two things to try:

   - On a brick mesh does it give you back the right thing?
   - If you refine does it reduce as you expect? (maybe we don't need the divide by 2?)

> Maybe hypot makes more sense? How should I interpret Δ1 and Δ2?

I expected `Δ1` and `Δ2` to approximately the two horizontal element sizes.

szy21:
> * On a brick mesh does it give you back the right thing?

On a brick mesh, when I set the average node distance to 100m and the polynomial order to 4, I get Δ1 = Δ2 = 25m.

> * If you refine does it reduce as you expect? (maybe we don't need the divide by 2?)

For the sphere, If I increase the horizontal number of elements by a factor of 2 (and keep the same polynomial order), Δ1 and Δ2 do decrease by a factor of 2.

jkozdon:
Let me write things out again on paper, but you are probably right that the divide by two should probably be a multiply.

jkozdon:
OK its multiply by `2`, sorry about that. Here is the reason

---
Assume we have the simple transformation:
```julia
x₁ = α₁ ξ₁ / 2
x₂ = α₂ ξ₂ / 2
x₃ = α₃ ξ₃ / 2
```
with `-1 ≤ ξ₁, ξ₂, ξ₃ ≤ 1`. The lengthscale of the three directions are then
```julia
α₁
α₂
α₃
```
The Jacobian is
```julia
    [∂x₁/∂ξ₁ ∂x₁/∂ξ₂ ∂x₁/∂ξ₃]   [α₁ / 2    0      0  ]
J = [∂x₂/∂ξ₁ ∂x₂/∂ξ₂ ∂x₂/∂ξ₃] = [   0   α₂ / 2    0  ]
    [∂x₃/∂ξ₁ ∂x₃/∂ξ₂ ∂x₃/∂ξ₃]   [   0      0   α₃ / 2]
```
So we have that
```julia
√((∂x₁/∂ξ₁)² + (∂x₂/∂ξ₁)² + (∂x₃/∂ξ₁)²) = α₁ / 2
√((∂x₁/∂ξ₂)² + (∂x₂/∂ξ₂)² + (∂x₃/∂ξ₂)²) = α₂ / 2
√((∂x₁/∂ξ₃)² + (∂x₂/∂ξ₃)² + (∂x₃/∂ξ₃)²) = α₃ / 2
```
and solving for the lengthscale scales gives
```julia
α₁ = 2√((∂x₁/∂ξ₁)² + (∂x₂/∂ξ₁)² + (∂x₃/∂ξ₁)²)
α₂ = 2√((∂x₁/∂ξ₂)² + (∂x₂/∂ξ₂)² + (∂x₃/∂ξ₂)²)
α₃ = 2√((∂x₁/∂ξ₃)² + (∂x₂/∂ξ₃)² + (∂x₃/∂ξ₃)²)
```

szy21:
Thank you Jeremy! I'm going to close this PR since we are not sure how to modify `resolutionmetric`. The modification to the hyperdiffusion is in another PR (#1997).

PR #2003, opened 2021-02-04
Add min nodal distance to DG grid

charleskawczynski:
### Description

This PR adds minimum nodal distance to the grids, while keeping `min_node_distance` defined. So, calling `min_node_distance` now has significantly reduced overhead. I'm not sure if/what tests to add, but I'm pretty sure that everything should be fully exercised.

### DIFF
--- a/src/Numerics/Mesh/Grids.jl
+++ b/src/Numerics/Mesh/Grids.jl
@@ -18,6 +18,19 @@ struct HorizontalDirection <: Direction end
 struct VerticalDirection <: Direction end
 Base.in(::T, ::S) where {T <: Direction, S <: Direction} = T == S
 
+"""
+    MinNodalDistance{FT}
+
+A struct containing the minimum nodal distance
+along horizontal and vertical directions.
+"""
+struct MinNodalDistance{FT}
+    "horizontal"
+    h::FT
+    "vertical"
+    v::FT
+end
+
 abstract type AbstractGrid{
     FloatType,
     dim,
@@ -167,6 +180,7 @@ struct DiscontinuousSpectralElementGrid{
     DAI3,
     TOP,
     TVTK,
+    MINΔ,
 } <: AbstractGrid{T, dim, N, Np, DA}
     "mesh topology"
     topology::TOP
@@ -221,6 +235,11 @@ struct DiscontinuousSpectralElementGrid{
     other cases these match `vgeo` values
     """
     x_vtk::TVTK
+
+    """
+    Minimum nodal distances for horizontal and vertical directions
+    """
+    minΔ::MINΔ
 
     # Constructor for a tuple of polynomial orders
     function DiscontinuousSpectralElementGrid(
@@ -312,6 +331,13 @@ struct DiscontinuousSpectralElementGrid{
         TOP = typeof(topology)
         TVTK = typeof(x_vtk)
 
+        FT = FloatType
+        minΔ = MinNodalDistance(
+            min_node_distance(vgeo, topology, N, FT, HorizontalDirection()),
+            min_node_distance(vgeo, topology, N, FT, VerticalDirection()),
+        )
+        MINΔ = typeof(minΔ)
+
         new{
             FloatType,
             dim,
@@ -327,6 +353,7 @@ struct DiscontinuousSpectralElementGrid{
             DAI3,
             TOP,
             TVTK,
+            MINΔ,
         }(
             topology,
             vgeo,
@@ -345,6 +372,7 @@ struct DiscontinuousSpectralElementGrid{
             D,
             Imat,
             x_vtk,
+            minΔ,
         )
     end
 end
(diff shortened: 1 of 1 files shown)


akshaysridhar (commented):
Thanks @charleskawczynski  - is this then accessible to `model` components in the way driver-configuration is currently set up ? In 2001 thats what the empty model enables, (in a similar fashion to a solution used in the LandModel setup @kmdeck)

kmdeck:
> Thanks @charleskawczynski - is this then accessible to `model` components in the way driver-configuration is currently set up ? In #2001 thats what the empty model enables, (in a similar fashion to a solution used in the LandModel setup @kmdeck)

I'm not sure this solves the background issue - you still need a model to create the grid (because the grid is made within the DriverConfig, which also needs the model)...but maybe this then makes it more efficient to get the min nodal distance, once the grid is made? I think a different solution to the background problem might be to have a function that creates the grid separately from what makes the driver.

kpamnany (approved):

kpamnany:
> I'm not sure this solves the background issue 

True, but I think it's a worthwhile improvement.

charleskawczynski:
> I'm not sure this solves the background issue - you still need a model to create the grid (because the grid is made within the DriverConfig, which also needs the model)...but maybe this then makes it more efficient to get the min nodal distance, once the grid is made? I think a different solution to the background problem might be to have a function that creates the grid separately from what makes the driver.

It's a bit tricky because, for example, the ocean stores parameters in their model that inform the grid, so we'd need a way to resolve the reverse construction on their side. That said, I have a few ideas, I'm looking into a simpler way that we can incorporate min node distance into the model.

charleskawczynski:
Ok, I think these changes should work. We'll need similar changes for the land (the land model will need to add a `minΔ` property), overall I think this approach is a bit cleaner.

charleskawczynski:
There's a question of whether we should use `Setfield.jl` or `Accessors.jl`. I'm not sure. This could probably be split into two PRs. I think the first commit (with pieces from the 2nd) is a nice improvement aside from adding `minΔ` to the models. @kmdeck how does this look? @kpamnany mind re-reviewing?

kpamnany (commented):
I'm not sure I like this change; it seems to be blurring the separation between the spatial discretization and the equations. If the equations _must_ know grid details, I'd prefer to abstract that into a more general `GridInfo` type, rather than specifics like nodal distances.

If the end goal here is to get this information to the turbulence closure, then can't it be passed to the turbulence closure constructor by the experiment code? This can be achieved as described in my comment [here](https://github.com/CliMA/ClimateMachine.jl/pull/2001#issuecomment-773610113) for now until we can redesign the driver configuration.

Cc: @simonbyrne

jkozdon:
> I'm not sure I like this change; it seems to be blurring the separation between the spatial discretization and the equations. If the equations _must_ know grid details, I'd prefer to abstract that into a more general `GridInfo` type, rather than specifics like nodal distances.

I agree. Another really weird thing about using min node distance is that it makes the physics dependent on your choice of basis functions and all elements are equally effected by mesh skewness. I'm not sure where this is used, but that seems odd to me.

charleskawczynski:
> I agree. Another really weird thing about using min node distance is that it makes the physics dependent on your choice of basis functions and all elements are equally effected by mesh skewness. I'm not sure where this is used, but that seems odd to me.

I 100% agree with you @jkozdon and @kpamnany. I do not like the idea of giving grid information to the physics models-- it can completely ruin spatial convergence, and will likely lead to interpretation problems.

I think part of the issue is that we're still trying to compare with other models, which do use grid information--regardless of whether it's a good design choice or not. In doing so, it seems that several model developers are stuck.

I think that we can replace these quantities in the long run, but it probably involves answering some difficult research questions--and we're not there yet. I have some ideas, but I'll need to talk with the numericists to see what they think.

> If the equations must know grid details, I'd prefer to abstract that into a more general GridInfo type, rather than specifics like nodal distances.

I like this idea. I'll split this PR up.

> If the end goal here is to get this information to the turbulence closure, then can't it be passed to the turbulence closure constructor by the experiment code? This can be achieved as described in my comment here for now until we can redesign the driver configuration.

They could, but they would need to bypass the `AtmosLESConfiguration`. which return immutable `DriverConfiguration`'s, so we cannot redefine the model once it's been added to the driver config.

I don't think that'll work because `DriverConfiguration` is immutable. They could still pass grid info to the model, but they'd need to bypass AtmosLESConfiguration, AtmosGCMConfiguration etc.

kpamnany:
> I don't think that'll work because `DriverConfiguration` is immutable.

Ah yes. But we could make it `mutable` easily enough. It's a quick solution (if it works) that doesn't open a can of worms as I fear this one does. In general, I'm far happier to hack and slash at, or even completely rewrite the driver configuration stuff than to change `AtmosModel` because of current driver configuration limitations. Also, I don't like Accessors/Setfield.

kmdeck:
> Ah yes. But we could make it `mutable` easily enough. It's a quick solution (if it works) that doesn't open a can of worms as I fear this one does. In general, I'm far happier to hack and slash at, or even completely rewrite the driver configuration stuff than to change `AtmosModel` because of current driver configuration limitations. Also, I don't like Accessors/Setfield.

from the soil pov, @kpamnany's suggestion on PR 2001 should work. The above approach that Charlie had would also work. We had been storing the nodal distance in a Source type structure, because it is used in a single source term, rather than the model itself, but the source terms are also part of the model, so it amounts to something v. similar). What we had been doing is something v. similar to what @akshaysridhar had: https://github.com/CliMA/ClimateMachine.jl/blob/0d2162739dc3b9da844d2623350640832db253f3/test/Land/Model/freeze_thaw_alone.jl#L34

szy21:
For the hyperdiffusion Jeremy suggests a way to get the horizontal nodal distance in #2002. @akshaysridhar I wonder if that's all we need, or do we still need min_nodal_distance somewhere?

charleskawczynski:
I've stripped out the controversial changes and left the changes from the first commit. We can follow up on incorporating it into the model in a subsequent PR.

charleskawczynski:
Not sure, but it looks like this may need reapproval @kpamnany

kpamnany (approved):

PR #2005, opened 2021-02-04
Remove timestamps from diagnostics output filenames

### LINKED ISSUE #1765: Better naming scheme for output?

kpamnany:
### Description

Now, any previously existing output files will be overwritten. Use `--output-dir` (or the environment variable or the keyword argument to `init`) to change the output directory if you don't want files overwritten.

Closes #1765.

### DIFF
--- a/src/InputOutput/Writers/netcdf_writer.jl
+++ b/src/InputOutput/Writers/netcdf_writer.jl
@@ -18,8 +18,16 @@ function full_name(writer::NetCDFWriter, filename = nothing)
     end
 end
 
-function init_data(nc::NetCDFWriter, filename, dims, vars)
-    Dataset(full_name(nc, filename), "c") do ds
+function init_data(nc::NetCDFWriter, filename, no_overwrite, dims, vars)
+    nm = full_name(nc, filename)
+    if ispath(nm)
+        if no_overwrite
+            error("$(nm) exists and overwriting is forbidden.")
+        else
+            @warn "$(nm) exists and will be overwritten."
+        end
+    end
+    Dataset(nm, "c") do ds
         # define spatial and time dimensions
         for (dn, (dv, da)) in dims
             defDim(ds, dn, length(dv))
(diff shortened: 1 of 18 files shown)


kpamnany:
bors try

ilopezgp:
LGTM

ilopezgp (approved):

jakebolewski:
LGTM.  Somewhat related to this PR it might be use to have an `--overwite=true/false` flag as well which would through an error, that would be useful for deployments.

jakebolewski:
Looks like one unit test expected the previous behavior: https://github.com/CliMA/ClimateMachine.jl/runs/1834568819#step:9:668

jakebolewski (approved):

src/InputOutput/Writers/netcdf_writer.jl
```diff
@@ -18,8 +18,12 @@ function full_name(writer::NetCDFWriter, filename = nothing)
     end
 end
 
-function init_data(nc::NetCDFWriter, filename, dims, vars)
-    Dataset(full_name(nc, filename), "c") do ds
+function init_data(nc::NetCDFWriter, filename, no_overwrite, dims, vars)
+    nm = full_name(nc, filename)
+    if no_overwrite && ispath(nm)
+        error("$(nm) exists and overwriting is forbidden.")
```
jakebolewski, line 24: maybe there should be a warning thrown when overwriting an existing file as well

jakebolewski (approved):

src/InputOutput/Writers/netcdf_writer.jl
jakebolewski: just following the principal of "least surprise" when debugging  :)

PR #2011, opened 2021-02-05
Fix flux pads so that empty fluxes work

charleskawczynski:
### Description

I wasn't really thorough with the pads the first time around. This PR cleans up the flux pads so that we can use empty `eq_tends` for (all 🤞🏻 ) fluxes. We could alternatively handle this inside `Σfluxes`/`Σsources`, but we'd need to pass the prognostic variable to it, and I'd rather make that transition in #1921 if that materializes.

Thanks @dennisYatunin for pointing this out.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -535,13 +535,14 @@ equations.
     direction,
 )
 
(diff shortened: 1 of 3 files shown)


dennisYatunin (approved):
Thank you! This looks good; it should help with the radiative equilibrium testing.

PR #1940, opened 2021-01-21
TwoDimensionalCompressibleNavierStokesEquations for testing Bickley Jet Setups

blallen:
### Description

Implemented the CNSE2D to compare ClimateMachine against Exasim for Bickley Jet setups. Currently produces comparable results at twice the resolution.

https://user-images.githubusercontent.com/7977055/105417975-68a4ef00-5c0a-11eb-8610-fd29a9244cfd.mp4

https://user-images.githubusercontent.com/7977055/105418071-87a38100-5c0a-11eb-927c-0ceb3c95b25d.mp4

Things to do still:
- [ ] Extract this fully from the Ocean Setup and make it a standalone balance law
- [ ] Rename from BickleyJetModel and BJModel to TwoDimensionalCompressibleNavierStokesEquations and CNSE2D
- [ ] Test out Roe fluxes for this model (can be a separate PR)
- [ ] Lots of different resolution tests
- [ ] Output to JLD2 for Makie viz
- [ ] More stuff that I am forgetting

PR #2013, opened 2021-02-06
Move around events in `Filters.apply!`

jkozdon:
Fix race conditions due to event dependency  in the `Filter.apply!` when `EveryDirection` is used.

### DIFF
--- a/src/Numerics/Mesh/Filters.jl
+++ b/src/Numerics/Mesh/Filters.jl
@@ -338,10 +338,10 @@ function apply!(
 
     nrealelem = length(topology.realelems)
 
+    event = Event(device)
     if direction isa EveryDirection || direction isa HorizontalDirection
         @assert dim == 2 || Nq1 == Nq2
         filtermatrix = filter.filter_matrices[1]
-        event = Event(device)
         event = kernel_apply_filter!(device, (Nq1, Nq2, Nq3))(
             Val(dim),
             Val(N),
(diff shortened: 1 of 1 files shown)


jkozdon:
@mwarusz and @vchuravy: @sandreza and I were poking around, and found what we think is a possible race condition. @sandreza was seeing some race like conditions with multi-threaded ClimateMachine with the filters and this change seems to have fixed the issue.

Not sure I understand what was wrong before with the dependency, can you offer any insight? (Or is it not really a race conditions and a coincidence.)

jkozdon:
If this is a race conditions, not sure if it would apply to #1953 as well (the events are moved around a bit there)

jkozdon:
@blallen: This might fix #1940

blallen:
> @blallen: This might fix #1940

looks like it did! https://buildkite.com/clima/climatemachine-ci/builds/1243#a3d28eb1-fa28-4062-a8b7-2c9683482c22

just need to lower the precision requirements on something

vchuravy:
Right the `event = Event(Device()` shadows the actual incoming event.

mwarusz:
This looks good to me, but is the pattern
```
event  = if ...
    ...
    event
else
    event
end
```
necessary ?  How is it different from
```
if ...
    event = 
end
```

mwarusz (approved):

jkozdon:
Thanks @mwarusz

Yeah I originally had a bug in some cases, and didn't think about it when I put in a fix.

PR #2017, opened 2021-02-08
Add tend spec layer docs

charleskawczynski:
### Description

This PR:
 - Adds a BalanceLaw tendency specification tutorial
 - Adds some options to `show_tendencies` (`include_module ` and `table_complete`)
 - Removes the warning in `show_tendencies` for the AtmosModel 🎉 

### DIFF
--- a/src/BalanceLaws/show_tendencies.jl
+++ b/src/BalanceLaws/show_tendencies.jl
@@ -15,21 +15,35 @@ format_tends(tend_types, include_params) =
     "(" * join(format_tend.(tend_types, include_params), ", ") * ")"
 
 """
-    show_tendencies(bl; include_params = false)
+    show_tendencies(
+        bl;
+        include_params = false,
+        include_module = false,
+        table_complete = false,
+    )
 
 Show a table of the tendencies for each
 prognostic variable for the given balance law.
 
-Using `include_params = true` will include the type
-parameters for each of the tendency fluxes
-and sources.
+## Arguments
+ - `include_params[ = true]` will include the type parameters for each of
+   the tendency fluxes and sources.
+ - `include_module[ = false]` will print not remove the module where each
+   prognostic variable is defined (e.g., `Atmos.Mass`).
+ - `table_complete[ = false]` will print a warning (if false) that the
+   tendency table is incomplete.
 
 Requires definitions for
  - [`prognostic_vars`](@ref)
  - [`eq_tends`](@ref)
 for the balance law.
 """
-function show_tendencies(bl; include_params = false)
+function show_tendencies(
+    bl;
+    include_params = false,
+    include_module = false,
+    table_complete = false,
+)
     ip = include_params
     prog_vars = prognostic_vars(bl)
     if !isempty(prog_vars)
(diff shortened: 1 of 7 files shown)


src/BalanceLaws/show_tendencies.jl
```diff
@@ -15,37 +15,55 @@ format_tends(tend_types, include_params) =
...
+        end
         fmt_tends(tt) = map(prog_vars) do pv
             format_tends(eq_tends(pv, bl, tt), ip)
         end |> collect
         F1 = fmt_tends(Flux{FirstOrder}())
         F2 = fmt_tends(Flux{SecondOrder}())
         S = fmt_tends(Source())
         data = hcat(eqs, F1, F2, S)
-        @warn "This table is temporarily incomplete"
+        table_complete || @warn "This table is temporarily incomplete"
```
charleskawczynski, line 66: We could alternatively change this to
```suggestion
        if table_complete
            @warn "This table is temporarily incomplete"
        else
            @warn "This table is complete, if a prognostic variable or tendency is not showing up in the table, then either 1) the configuration is incorrect, or 2) open an issue on GitHub."
        end
```

tutorials/BalanceLaws/tendency_specification_layer.jl
```diff
@@ -0,0 +1,129 @@
...
+
+# We can also permit tendency definitions to be defined for
+# only the relevant prognostic variables: here, we restrict
+# the second order flux FY2 to be defined only for Y.
+struct FY2{PV <: Y} <: TendencyDef{Flux{SecondOrder}, PV} end
+
+struct S1{PV} <: TendencyDef{Source, PV} end
+struct S2{PV} <: TendencyDef{Source, PV} end
+
+# Define `eq_tends` given
```
kpamnany, line 62: Expand a little on what `eq_tends` does? "Define tendencies for the balance law..." or something?

kpamnany (commented):
Need to add `ref`s to the API documentation for all the functions.

This seems nice and simple, but I wonder if a real world example (some very simple but real balance law?) would be more understandable for physicists?

src/BalanceLaws/show_tendencies.jl
kpamnany: Too noisy, prefer it how you have it.

charleskawczynski:
> Need to add refs to the API documentation for all the functions.

Good idea

> This seems nice and simple, but I wonder if a real world example (some very simple but real balance law?) would be more understandable for physicists?

That was my original thought / plan (I was going to create an AtmosModel), but then I thought this might be simpler. I agree that it looks funny though. Do you think if better names for the flux / source definitions would help make it look more real-world? @jiahe23 any thoughts?

jiahe23:
> > This seems nice and simple, but I wonder if a real world example (some very simple but real balance law?) would be more understandable for physicists?
> 
> That was my original thought / plan (I was going to create an AtmosModel), but then I thought this might be simpler. I agree that it looks funny though. Do you think if better names for the flux / source definitions would help make it look more real-world? @jiahe23 any thoughts?

I agree! It is not clear to me how to enable the tendencies computations (or include/exclude a specific tendency contribution of interest) in setting up an experiment driver. A real example will be very helpful.

tutorials/BalanceLaws/tendency_specification_layer.jl
charleskawczynski: I added more description of what this does..

charleskawczynski:
Ok, @jiahe23 and I went over this in more detail and made some things more concrete/realistic. @kpamnany, if you're okay with these changes, it'd be helpful to have this merged before Thursday 👨🏻‍🏫. Going to test CI, since it's not been tested with cross refs yet.

charleskawczynski:
bors try

kpamnany (approved):

charleskawczynski:
I only `ref`'d the first instances of each method, because adding the `ref` clogs the markdown a bit. I also added some more doc strings. Maybe one last quick lookover @kpamnany?

jiahe23 (approved):

charleskawczynski:
Actually, merging now since it may help make reviewing the presentation a bit easier

PR #2021, opened 2021-02-09
Add incompressible single stack model and driver configuration

charleskawczynski:
### Description

This PR:
 - Adds a `compressibility` submodel to the AtmosModel
 - Adds some thermo wrappers which use the (anelastic) reference state for density
 - Modify equation tendencies for the `Anelastic1D` submodel
 - Adds a `projection` hook to the balance law

It seems like we can take timesteps of at least 15 seconds 🎉 

It would be nice to more elegantly (i.e., by default) handle `g = compressibility isa Compressible ? (Gravity(),) : ()`. I think that using `g = compressibility isa Compressible ? (Gravity(),) : ()` is the simplest existing solution based on how sources are currently passed to the atmos model. We can decide later if the sources should be handled differently, but this would touch many experiments and should be addressed in a separate PR.

### DIFF
--- a/docs/src/APIs/Atmos/AtmosModel.md
+++ b/docs/src/APIs/Atmos/AtmosModel.md
@@ -26,6 +26,13 @@ ClimateMachine.BalanceLaws.source!(m::AtmosModel, source::Vars, state::Vars, dif
 ClimateMachine.BalanceLaws.init_state_prognostic!(m::AtmosModel, state::Vars, aux::Vars, localgeo, t, args...)
 ```
 
+## Compressibility
+
+```@docs
+ClimateMachine.Atmos.Compressible
+ClimateMachine.Atmos.Anelastic1D
+```
+
 ## Reference states
 
 ```@docs
@@ -40,6 +47,8 @@ ClimateMachine.Atmos.NoReferenceState
 ```@docs
 ClimateMachine.Atmos.recover_thermo_state
 ClimateMachine.Atmos.new_thermo_state
+ClimateMachine.Atmos.recover_thermo_state_anelastic
+ClimateMachine.Atmos.new_thermo_state_anelastic
 ```
 
 ## Moisture and Precipitation
--- a/experiments/AtmosLES/stable_bl_model.jl
+++ b/experiments/AtmosLES/stable_bl_model.jl
@@ -45,7 +45,7 @@ struct EarthParameterSet <: AbstractEarthParameterSet end
 const param_set = EarthParameterSet()
 import CLIMAParameters
 
-using ClimateMachine.Atmos: altitude, recover_thermo_state
+using ClimateMachine.Atmos: altitude, recover_thermo_state, density
 
 """
   StableBL Geostrophic Forcing (Source)
@@ -149,14 +149,15 @@ function init_problem!(problem, bl, state, aux, localgeo, t)
         θ_liq = FT(265) + FT(0.01) * (z - z1)
     end
     θ = θ_liq
-    π_exner = FT(1) - _grav / (c_p * θ) * z # exner pressure
-    ρ = p0 / (R_gas * θ) * (π_exner)^(c_v / R_gas) # density
+    p = aux.ref_state.p
     # Establish thermodynamic state and moist phase partitioning
     if bl.moisture isa DryModel
-        TS = PhaseDry_ρθ(bl.param_set, ρ, θ_liq)
+        TS = PhaseDry_pθ(bl.param_set, p, θ)
     else
-        TS = PhaseEquil_ρθq(bl.param_set, ρ, θ_liq, q_tot)
+        TS = PhaseEquil_pθq(bl.param_set, p, θ_liq, q_tot)
     end
+
+    ρ = bl.compressibility isa Compressible ? air_density(TS) : aux.ref_state.ρ
     # Compute momentum contributions
     ρu = ρ * u
     ρv = ρ * v
@@ -190,6 +191,7 @@ function stable_bl_model(
     surface_flux;
     turbulence = ConstantKinematicViscosity(FT(0)),
     turbconv = NoTurbConv(),
+    compressibility = Compressible(),
     moisture_model = "dry",
     ref_state = HydrostaticState(DecayingTemperatureProfile{FT}(param_set),),
 ) where {FT}
@@ -210,9 +212,11 @@ function stable_bl_model(
 
     q_sfc = FT(0)
 
+    g = compressibility isa Compressible ? (Gravity(),) : ()
+
     # Assemble source components
     source_default = (
-        Gravity(),
+        g...,
         StableBLSponge(
             FT,
             zmax,
@@ -324,6 +328,7 @@ function stable_bl_model(
         moisture = moisture,
         source = source,
         turbconv = turbconv,
+        compressibility = compressibility,
     )
 
     return model
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -6,7 +6,9 @@ export AtmosModel,
     HLLCNumericalFlux,
     RoeNumericalFlux,
     RoeNumericalFluxMoist,
-    LMARSNumericalFlux
+    LMARSNumericalFlux,
+    Compressible,
+    Anelastic1D
 
 using UnPack
 using CLIMAParameters
@@ -34,7 +36,7 @@ using ..TurbulenceClosures
 import ..TurbulenceClosures: turbulence_tensors
 using ..TurbulenceConvection
 
-import ..Thermodynamics: internal_energy
+import ..Thermodynamics: internal_energy, soundspeed_air
 using ..MPIStateArrays: MPIStateArray
 using ..Mesh.Grids:
     VerticalDirection,
@@ -48,6 +50,7 @@ using ClimateMachine.Problems
 
 import ..BalanceLaws:
     vars_state,
+    projection,
     flux_first_order!,
     flux_second_order!,
     source!,
@@ -119,6 +122,7 @@ default values for each field.
         radiation,
         source,
         tracers,
+        compressibility,
         data_config,
     )
 
@@ -142,6 +146,7 @@ struct AtmosModel{
     S,
     TR,
     LF,
+    C,
     DC,
 } <: BalanceLaw
     "Parameter Set (type to dispatch on, e.g., planet parameters. See CLIMAParameters.jl package)"
@@ -174,9 +179,32 @@ struct AtmosModel{
     tracers::TR
     "Large-scale forcing (Forcing information from GCMs, reanalyses, or observations)"
     lsforcing::LF
+    "Compressibility switch"
+    compressibility::C
     "Data Configuration (Helper field for experiment configuration)"
     data_config::DC
 end
+
+abstract type Compressibilty end
+
+"""
+    Compressible <: Compressibilty
+
+Dispatch on compressible model (default)
+
+ - Density is prognostic
+"""
+struct Compressible <: Compressibilty end
+
+"""
+    Anelastic1D <: Compressibilty
+
+Dispatch on Anelastic1D model
+
+ - Density is constant in time
+ - Remove momentum z-component tendencies
+"""
+struct Anelastic1D <: Compressibilty end
 
 """
     AtmosModel{FT}()
@@ -207,6 +235,7 @@ function AtmosModel{FT}(
     ),
     tracers::TR = NoTracers(),
     lsforcing::LF = NoLSForcing(),
+    compressibility::C = Compressible(),
     data_config::DC = nothing,
 ) where {
     FT <: AbstractFloat,
@@ -225,6 +254,7 @@ function AtmosModel{FT}(
     S,
     TR,
     LF,
+    C,
     DC,
 }
     @assert !any(isa.(source, Tuple))
@@ -245,6 +275,7 @@ function AtmosModel{FT}(
         source,
         tracers,
         lsforcing,
+        compressibility,
         data_config,
     )
 
@@ -275,6 +306,7 @@ function AtmosModel{FT}(
     source::S = (Gravity(), Coriolis(), turbconv_sources(turbconv)...),
     tracers::TR = NoTracers(),
     lsforcing::LF = NoLSForcing(),
+    compressibility::C = Compressible(),
     data_config::DC = nothing,
 ) where {
     FT <: AbstractFloat,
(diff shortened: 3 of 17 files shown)


yairchn (approved):
At the end we have no changes to the 'stable_bl_model.jl' file, that is great!

docs/src/APIs/Atmos/AtmosModel.md
```diff
@@ -26,6 +26,13 @@ ClimateMachine.BalanceLaws.source!(m::AtmosModel, source::Vars, state::Vars, dif
 ClimateMachine.BalanceLaws.init_state_prognostic!(m::AtmosModel, state::Vars, aux::Vars, localgeo, t, args...)
 ```
 
+## Compressibility
+
+```@docs
+ClimateMachine.Atmos.Compressible
+ClimateMachine.Atmos.Incompressible
```
tapios, line 33:
```suggestion
ClimateMachine.Atmos.Anelastic
```
(It's not really incompressible. But even anelastic may be going too far, because we only implement a 1d version (we do not want 3d elliptic solves...)

yairchn: Thtat is true in scampy (and I believe most single column models that make an anelastic assumption)

src/Atmos/Model/AtmosModel.jl
```diff
@@ -173,10 +178,33 @@ struct AtmosModel{
...
+    Compressible <: Compressibilty
+
+Dispatch on compressible model (default)
+
+ - Density is prognostic
+"""
+struct Compressible <: Compressibilty end
+
+"""
+    Incompressible <: Compressibilty
```
tapios, line 199:
```suggestion
   Anelastic <: Compressibilty
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -173,10 +178,33 @@ struct AtmosModel{
...
+
+Dispatch on incompressible model
```
tapios, line 201:
```suggestion
Dispatch on inelastic model
```

src/Atmos/Model/AtmosModel.jl
```diff
@@ -173,10 +178,33 @@ struct AtmosModel{
...
+
+ - Density is constant in time
+ - Remove momentum z-component tendencies
+"""
+struct Incompressible <: Compressibilty end
```
tapios, line 206:
```suggestion
struct Anelastic <: Compressibilty end
```

src/Atmos/Model/thermo_states.jl
```diff
@@ -2,6 +2,21 @@
 
 export new_thermo_state, recover_thermo_state
 
+"""
+    density(atmos::AtmosModel, state::Vars, aux::Vars)
+
+In the incompressible state, `state.ρ` that is used to
+extract intrinsic values (i.e. `e=state.energy.ρe/state.ρ`)
+is time invariant (by eliminating the tendencies) while
```
tapios, line 10: Why would specific energy be constant in time? And is intrinsic "specific"? -- I do not understand this comment.

src/Atmos/Model/thermo_states.jl
```diff
@@ -2,6 +2,21 @@
...
+`ref_state.ρ` is used to construct a thermodynamic state.
+`density` will get the `ref_state.ρ` for thermodynamic state
+when the model is incompressible.
```
tapios, line 13: Ok. I cannot parse this comment at all (notwithstanding that I should know where this is going). Please rewrite.

tapios (commented):
@charleskawczynski Thanks for putting this together. I'm glad it's working.

But I think the approach here is at the same time trying to be too general, and is too specific in that it does not easily seem to allow the flexibility we need. What I had in mind is simply specifying an equation set for a single stack that 
1) Eliminates hydrostatic balance from the momentum equation (i.e., no dp/dz + g \rho on the RHS of the vertical momentum equation)
2) Uses a fixed density rho_0(z) in the explicit time derivatives and fluxes in the conservative (flux form) equations
3) Uses a fixed pressure p_0 (in hydrostatic balance with rho_0, but this does not need to be discrete) in thermodynamic calculations where needed.

I do not think this requires new thermodynamic states, just passing rho_0 in place of rho as a state variable in thermodynamics. It should eventually also allow users to specify whatever rho_0 and p_0 they want (not necessarily something constructed to be in discrete HS balance; this isn't needed when the hydrostatic terms are explicitly dropped from the momentum balance).

So: can we do this without new thermodynamic states and simply by specifying a different equation set to be used?

@ilopezgp may be able to help if the concepts are unclear.

src/Atmos/Model/thermo_states.jl
ilopezgp: Yes, I think here @yairchn meant specific when he wrote intrinsic. Also, I think the parenthesis just shows an example of the way we compute a specific variable, but "is time invariant" refers to the state.rho. We will rewrite this for clarity.

ilopezgp:
> @charleskawczynski Thanks for putting this together. I'm glad it's working.
> 
> But I think the approach here is at the same time trying to be too general, and is too specific in that it does not easily seem to allow the flexibility we need. What I had in mind is simply specifying an equation set for a single stack that
> 
> 1. Eliminates hydrostatic balance from the momentum equation (i.e., no dp/dz + g \rho on the RHS of the vertical momentum equation)
> 2. Uses a fixed density rho_0(z) in the explicit time derivatives and fluxes in the conservative (flux form) equations
> 3. Uses a fixed pressure p_0 (in hydrostatic balance with rho_0, but this does not need to be discrete) in thermodynamic calculations where needed.
> 
> I do not think this requires new thermodynamic states, just passing rho_0 in place of rho as a state variable in thermodynamics. It should eventually also allow users to specify whatever rho_0 and p_0 they want (not necessarily something constructed to be in discrete HS balance; this isn't needed when the hydrostatic terms are explicitly dropped from the momentum balance).
> 
> So: can we do this without new thermodynamic states and simply by specifying a different equation set to be used?
> 
> @ilopezgp may be able to help if the concepts are unclear.

Thank you, Tapio. We will work on implementing this.

src/Atmos/Model/AtmosModel.jl
```diff
@@ -173,10 +178,33 @@ struct AtmosModel{
     tracers::TR
     "Large-scale forcing (Forcing information from GCMs, reanalyses, or observations)"
     lsforcing::LF
+    "Compressibility switch"
+    compressibility::C
     "Data Configuration (Helper field for experiment configuration)"
     data_config::DC
 end
 
+abstract type Compressibilty end
```
charleskawczynski, line 187: @tapios if we change `Incompressible` to `Anelastic` / `SingleStack`, should `Compressibility` change?

src/Atmos/Model/AtmosModel.jl
```diff
@@ -173,10 +178,33 @@ struct AtmosModel{
     tracers::TR
     "Large-scale forcing (Forcing information from GCMs, reanalyses, or observations)"
     lsforcing::LF
+    "Compressibility switch"
+    compressibility::C
```
charleskawczynski, line 182: @tapios If we change `Incompressible` to `SingleStack`, then what should `compressibility` be? My impression was that everything inside `AtmosModel` should be physics-centric, and agnostic to the configuration (GCM/LES/SingleStack). We're okay with making `config::SingleStack` a property of `AtmosModel`?

src/BalanceLaws/tendency_types.jl
```diff
@@ -94,6 +94,14 @@ corresponding to the column-vector `Yᵢ` in:
 """
 prognostic_vars(::BalanceLaw) = ()
 
+"""
+    projection(bl::BalanceLaw, ::TendencyDef, x)
+
+Provide a hook to project individual tendencies.
+Return identity by defualt
+"""
+projection(bl::BalanceLaw, ::TendencyDef, x) = x
```
charleskawczynski, line 103: @kpamnany can you review this projection hook? It's mainly used in `src/BalanceLaws/sum_tendencies.jl`. I think that this is something that @LenkaNovak and @mwarusz may also need. To make this work on the sphere, we might want to add `args` as an argument, so that we have access to `args.aux`, which is needed for projections on the sphere.

docs/src/APIs/Atmos/AtmosModel.md
ilopezgp: @tapios how about `Anelastic1D` ? I think we already have many SingleStacks (experiments, model, ...) and this would lead to confusion.

test/Atmos/EDMF/stable_bl_edmf_incompressible.jl
```diff
@@ -0,0 +1,256 @@
...
+
+projection(
+    ::AtmosModel,
+    ::TendencyDef{Flux{O}, PV},
+    x,
+) where {O, PV <: Momentum} =
+    x .* SArray{Tuple{3, 3}}(1, 1, 1, 1, 1, 1, 0, 0, 0)
+
+projection(::AtmosModel, ::TendencyDef{Source, PV}, x) where {PV <: Momentum} =
+    x .* SVector(1, 1, 0)
```
charleskawczynski, line 28: @kpamnany, @mwarusz, @LenkaNovak Here's how we're doing the projections for a single experiment. I'm not sure the best way to incorporate this into the model itself. We could add a dispatch mechanism inside the AtmosModel so that we have default projection methods for each model, but I'm not sure that putting a dispatch mechanism inside the AtmosModel is the appropriate place.

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Spoke with @ilopezgp about this. Thanks!

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Spoke with @ilopezgp about this. Thanks!

docs/src/APIs/Atmos/AtmosModel.md
tapios: Sounds good.

src/BalanceLaws/tendency_types.jl
kpamnany: It looks nice, but we should probably discuss.

charleskawczynski:
bors try

test/Atmos/EDMF/stable_bl_edmf_incompressible.jl
```diff
@@ -0,0 +1,260 @@
+using JLD2, FileIO
```
charleskawczynski, line 1: We need to add this to buildkite

experiments/AtmosLES/stable_bl_model.jl
```diff
@@ -148,14 +148,15 @@ function init_problem!(problem, bl, state, aux, localgeo, t)
         θ_liq = FT(265) + FT(0.01) * (z - z1)
     end
     θ = θ_liq
-    π_exner = FT(1) - _grav / (c_p * θ) * z # exner pressure
-    ρ = p0 / (R_gas * θ) * (π_exner)^(c_v / R_gas) # density
+    p = aux.ref_state.p
```
charleskawczynski, line 151: @ilopezgp the other SBL cases are breaking, is it because our initialization is different?

test/Atmos/EDMF/stable_bl_edmf_incompressible.jl
ilopezgp: Do we need it now that we are using the netcdf diagnostics?

charleskawczynski: I just mean the driver file (`test/Atmos/EDMF/stable_bl_edmf_incompressible.jl`)

experiments/AtmosLES/stable_bl_model.jl
ilopezgp: Yes, the error comes from using `density` here, which default the density to state.rho, which is not yet defined. I will push a fix.

ilopezgp (approved):

ilopezgp:
bors try

test/Atmos/EDMF/stable_bl_anelastic1d.jl
```diff
@@ -0,0 +1,260 @@
...
+        zmax,
+        surface_flux;
+        turbulence = SmagorinskyLilly{FT}(0.21),
+        turbconv = turbconv,
+        compressibility = compressibility,
+    )
+
+    # Assemble configuration
+    driver_config = ClimateMachine.SingleStackConfiguration(
+        "SBL_EDMF_INCOMPRESS",
```
yairchn, line 159: should we change this to "SBL_EDMF_ANELASTIC"?

ilopezgp:
bors try

yairchn (approved):

src/Atmos/Model/AtmosModel.jl
```diff
@@ -174,10 +179,33 @@ struct AtmosModel{
...
+"""
+struct Compressible <: Compressibilty end
+
+"""
+    Anelastic1D <: Compressibilty
+
+Dispatch on Anelastic1D model
+
+ - Density is constant in time
+ - Remove momentum z-component tendencies
```
dennisYatunin, line 205: @charleskawczynski Are the z-components actually removed by default? The changes to eq_tends eliminate mass advection and the momentum pressure gradient, but the z-components are only removed in stable_bl_anelastic1d.jl. Either this comment is misleading, or that z-component removal should be done by default.

charleskawczynski: We should remove it by default. I think we can do this with the projection hook. We'll probably want to change the `Σfluxes`/`Σsources` argument list (and not slurp everything).

src/BalanceLaws/tendency_types.jl
charleskawczynski: @kpamnany, @simonbyrne maybe we can discuss this today sometime?

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Ok, now it should be hooked in by default.

charleskawczynski:
bors try

test/Atmos/EDMF/stable_bl_edmf_incompressible.jl
charleskawczynski: This has moved to `src/Atmos/Model/projections.jl`

charleskawczynski:
bors try

charleskawczynski:
@simonbyrne, @kpamnany I think this is ready to go.

PR #2022, opened 2021-02-09
Stability tests using 3D Compressible NSE Model

blallen:
### Description

Extension of #1940 to 3D. Have confirmed that 2.5 bickley jet (w/ and w/o 3D perturbations) and taylor green vortices work. This part is ready to go, just needs some good tests.  

Also working on extending to a sphere, might split into a separate PR. Basic isotropic sphere tests are working, likely requires more development of the balance law for an anisotropic sphere.

Also have a commit with a bunch of @sandreza's visualization stuff. will probably drop that before merging.


blallen:
bors try

blallen:
bors try

bischtob (approved):
LGTM!

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try

blallen:
bors try

test/Numerics/DGMethods/compressible_navier_stokes_equations/three_dimensional/refvals_bickley_jet.jl
```diff
@@ -0,0 +1,43 @@
...
+ [ "state", "ρu[2]",  -2.53378560432881150266e-01,  3.01289778488659565348e-01,  2.01684398256973883116e-04,  6.56919957209414762112e-02 ],
+ [ "state", "ρu[3]",  -2.50004518195408198533e-01,  2.78664912597020086871e-01, -1.46870695308643243668e-04,  4.39167259046288754876e-02 ],
+ [ "state",    "ρθ",  -1.01085368604178116314e+00,  8.88723796619741435165e-01, -2.72714617833441435538e-04,  5.56687696608747506488e-02 ],
+],
+[
+ ["state",     "ρ", 9, 8, 11, 8],
+ ["state", "ρu[1]", 6, 6,  9, 7],
+ ["state", "ρu[2]", 6, 6,  0, 8],
+ ["state", "ρu[3]", 6, 6,  0, 9],
+ ["state",    "ρθ", 5, 6,  0, 7],
```
blallen, line 37: @christophernhill @jkozdon any idea why I have to set the precision so much lower for 4th order than for 1st order? The results seem to be deterministic per setup, but this a much larger difference between tartarus and caltech central than we have observed before

christophernhill: Hmmmm - not sure.  

Not sure if this is a clue but the means look a bit odd to me. In one case it is practically zero, but in the other it seems to have a non-zero value. However, the mean in state check is currently really the sum of the nodal points divided by the nodal point count. So its OK as a checksum, but maybe not as good a diagnostic as the true mean. 

Really we should upgrade state check to use https://github.com/CliMA/ClimateMachine.jl/blob/0d5b384f6b3ec479121e1ddd7fb952d335e1b93a/sandbox/volume_average_test.jl#L19 , but that is a little out of scope here. I think if we were doing true mean and there was still a big difference (especially if one case is almost zero as in here) then that would be a clue to possibly something unexpected. 

Not sure that is a helpful answer to your question though!!

blallen: Ah, I was actually asking more about the max, min, and stdev. I've seen these fluctuations about zero for true zero means before and believe it's caused by exactly what you mentioned. What I don't understand is what could drop the precision for the other three (especially min and max).

jkozdon: Seems odd for sure. Not sure why the high-order would be so much worse than the low order, except you have more floating point operations to be different.

Not sure if the optimizations the code experiences on the V100 (MIT) should be that much different than the P100 (Caltech).

blallen: Any reason to avoid merging with these lowered precisions?

also @sandreza maybe we should setup some crosscheck on satori and cyclops

christophernhill: @blallen that sounds fine. We may discover later that there was a reason, but that is OK.

blallen:
bors try

bischtob (approved):
LGTM

sandreza (approved):
lookin good to merge! Just making a note here that we'll want to change things up for the sphere (Coriolis for example)

blallen:
turns out we really need to actually use the weights when taking the mean because wow I cannot test momentum at all for the zero stays zero test

PR #2024, opened 2021-02-09
add standard sbatch submission script for Caltech Central cluster

simonbyrne:
### Description

This adds a standard job submission script for the Caltech Central HPC cluster

I'm not 100% convinced this is a good idea: we've so far tried to avoid keeping site-specific job submission scripts in the repo, instead keeping the info in the wiki. The downside of this is that the information can easily get out-of-date. Thoughts or alternative suggestions are appreciated.

I've tried to condense the logic into one script. This uses GPUs by default, but also supports CPU-only operation. 

I know that some people use more complicated scripts, e.g. https://github.com/CliMA/ClimateMachine.jl/wiki/Bash-Run-Scripts, so would appreciate some input to see if we can accommodate those workflows.

### DIFF
--- a/contrib/caltech.sh
+++ b/contrib/caltech.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+#SBATCH --ntasks=1
+#SBATCH --gres=gpu:1
+#SBATCH --time=01:00:00
+
+if [[ -z "${SLURM_JOB_ID}" ]]; then
+    cat << EOF
+This script should be used as an sbatch submission script. Usage:
+
+  sbatch [options...] contrib/caltech.sh script [arguments...]
+
+where:
+  [options...]: sbatch options. Default are "--ntasks=1 --gres=gpu:1 --time=01:00:00"
+  script: is the julia script to run
+  [arguments...]: options to be passed to the script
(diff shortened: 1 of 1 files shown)


valeriabarra:
How would this avoid getting the Wiki or documentation info out of sync? Is this script going to generate those?

simonbyrne:
The idea is that we would remove them from the wiki, and encourage users to use this script directly. Of course that means that the script should be sufficiently flexible to handle most users workflows.

valeriabarra:
Oh! I personally like and have followed this [page](https://github.com/CliMA/ClimateMachine.jl/wiki/Caltech-Central-Cluster) quite a lot. I think it's nice because it goes through anything a user would need to know, from how to set up from VPN, to SSH, 2FA, and how to request an interactive node, and so on... I don't think this script only could cover all of that useful information.

simonbyrne:
We can keep the page, just remove the scripts.

kpamnany:
My $0.02 is to put such things in a separate `contrib` repo, as with the visualization scripts.

simonbyrne:
Okay, i'll close for now.

PR #2030, opened 2021-02-11
Alternative balanced ref state initialization for FVM

### LINKED ISSUE #2018: ref state T and density oscillation when using grid stretching vertical FV
### Description

All data shown in the plots are nodal values (No interpolation from diagnostics module).

With `grid_stretching = SingleExponentialStretching(FT(4.5))` in the `solid_body_rotation_fvm.jl`, `ref_state` temperature and density profile has strong oscillation in the low levels. 

While in the vertical DG set-up, no oscillation is observed. However, the very top levels of temperature jumps to a very large value. 

ref_state

mwarusz:
This PR change the way a balanced reference state is computed for FVM in the vertical. Currently we compute
density from the discrete balance relation assuming a pressure profile. In this PR the approach is to integrate
the discrete balance relation for pressure and density assuming a virtual temperature profile. Seems to behave better
for stretched grids.

@jiahe23 Could you test ?

Closes #2018

### DIFF
--- a/src/Atmos/Model/ref_state.jl
+++ b/src/Atmos/Model/ref_state.jl
@@ -260,15 +260,21 @@ end
 
 function fvm_balance_init!(
     m::AtmosModel,
-    aux::Vars,
     aux_top::Vars,
+    aux::Vars,
     Δz::MArray{Tuple{2}, FT},
 ) where {FT}
-    # ρᵢ₋₁  = (pᵢ₋₁ - pᵢ - g ρᵢ Δzᵢ/2) / (g  Δzᵢ₋₁/2)
     _grav::FT = grav(m.param_set)
-    aux.ref_state.ρ =
-        (
-            aux.ref_state.p - aux_top.ref_state.p -
-            _grav * aux_top.ref_state.ρ * Δz[2] / 2
-        ) / (_grav * Δz[1] / 2)
+    _R_d::FT = R_d(m.param_set)
+
+    ref = aux.ref_state
+    topref = aux_top.ref_state
+
+    T_virt = ref.p / (_R_d * ref.ρ)
+    top_T_virt = topref.p / (_R_d * topref.ρ)
+
+    topref.ρ =
+        ref.ρ * (_R_d * T_virt - _grav * Δz[1] / 2) /
+        (_R_d * top_T_virt + _grav * Δz[2] / 2)
+    topref.p = _R_d * topref.ρ * top_T_virt
 end
--- a/src/Numerics/DGMethods/DGFVModel_kernels.jl
+++ b/src/Numerics/DGMethods/DGFVModel_kernels.jl
@@ -1000,15 +1000,10 @@ end
 
 """
     kernel_fvm_balance!(f!, balance_law::BalanceLaw, ::Val{nvertelem}, state_auxiliary, vgeo, elems)
-    
-    To avoid the oscillation in the reference temperature
-    ρ[nvertelem] = ρ[nvertelem] * (1 - ϵ)
-    to ensure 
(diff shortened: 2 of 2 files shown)


ilopezgp:
@mwarusz @jiahe23 An addition that would be extremely useful to the Atmos model would be if we can construct a reference state given a virtual or potential temperature profile. That is, given T (for example, could be another temperature), construct p from \int{dp/p} = -int{g dz / RT(z)}. Then, get rho from p and T.

Zhengyu-Huang (commented):
Looks great! Thank you! Here you fix the T_virt and update  both rho  and p. How does this affect discrete hydrostatic balance? From the unit tests, it seems the hydrostatic balance tests pass

mwarusz:
>Looks great! Thank you! Here you fix the T_virt and update both rho and p. How does this affect discrete hydrostatic balance? >From the unit tests, it seems the hydrostatic balance tests pass

Yes, the balance should be maintained.

mwarusz:
> An addition that would be extremely useful to the Atmos model would be if we can construct a reference state given a virtual or potential temperature profile. That is, given T (for example, could be another temperature), construct p from \int{dp/p} = -int{g dz / >RT(z)}. Then, get rho from p and T.

This is basically what this PR does for FV. Are asking to do this also for DG, or for a better interface to prescribe the reference temperature profile ?

mwarusz:
bors try

ilopezgp:
> > An addition that would be extremely useful to the Atmos model would be if we can construct a reference state given a virtual or potential temperature profile. That is, given T (for example, could be another temperature), construct p from \int{dp/p} = -int{g dz / >RT(z)}. Then, get rho from p and T.
> 
> This is basically what this PR does for FV. Are asking to do this also for DG, or for a better interface to prescribe the reference temperature profile ?

Maybe I didn't understand the code correctly. My understanding is that the temperature is still obtained from the reference pressure and density as
`T_virt = ref.p / (_R_d * ref.ρ)`
instead of accepting as an input a pointwise definition of T (or theta) to construct the reference. Is my interpretation wrong? 

Regarding your second question, yes, it would be great to have this for DG as well.

mwarusz:
> Maybe I didn't understand the code correctly. My understanding is that the temperature is still obtained from the reference pressure and density as
T_virt = ref.p / (_R_d * ref.ρ)
instead of accepting as an input a pointwise definition of T (or theta) to construct the reference. Is my interpretation wrong?

Ah, sorry, the code is a bit confusing because it handles both FV and DG initializations. In the FV case this computation should just give the prescribed virtual temperature profile without modification (because rho and p were computed assuming this profile).

jiahe23:
Yes, the oscillation in the fvm with grid stretching has been resolved. 

One remaining issue is the relatively large T values in the top level in DG here. 

@mwarusz Do you have an idea what might cause it?

ref_state (1)

mwarusz:
The problem with DG is that with this amount of stretching you end up with one element that covers the top 20 km of the atmosphere. A polynomial approximation to the pressure gradient is simply too oscillatory on such large element. I currently don't know of a way to make it work for this setup. You should see that this effect goes away if you increase the number of elements or the polynomial order.

jiahe23:
> The problem with DG is that with this amount of stretching you end up with one element that covers the top 20 km of the atmosphere. A polynomial approximation to the pressure gradient is simply too oscillatory on such large element. I currently don't know of a way to make it work for this setup. You should see that this effect goes away if you increase the number of elements or the vertical polynomial order.

Thanks! This seems to be a different issue irrelevant to this PR. Let's merge this and I'll open up a new issue on the stretched grid with DG.

jiahe23 (approved):

PR #2031, opened 2021-02-11
Add WrapVars type, and use to remove arr suffix

charleskawczynski:
### Description

I started working on the Vars wrapper in numerical fluxes, but then realized that we have multiple methods that would accept arrays (e.g., `numerical_boundary_flux_first_order!`) and the wrapper would have the same interface as the caller, which makes the `_arr` suffix much less appealing/clear. This PR adds a type `WrapVars` to the balance law, and uses this as a first argument whenever we want to wrap the arrays in `Vars`. This should also avoid the method ambiguity problem (since `WrapVars` is an extra argument).

### DIFF
--- a/src/Numerics/DGMethods/NumericalFluxes.jl
+++ b/src/Numerics/DGMethods/NumericalFluxes.jl
@@ -790,7 +790,7 @@ function numerical_flux_first_order!(
     )
 end
 
-function numerical_boundary_flux_first_order!(
+function numerical_boundary_flux_first_order_loop!(
     numerical_flux_first_order,
     bctag::Int,
     balance_law,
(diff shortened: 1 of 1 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
Hmm, now I'm hitting stack overflow (recursive calls). I think the remaining part (numerical fluxes) kind of needs to happen in one go. I think this new type makes the intent of the call more clear, and then also removes the proliferation of function names. Going to try to tackle the whole thing in one go, here.

charleskawczynski:
bors try

charleskawczynski:
Closing in favor of #2039

PR #2038, opened 2021-02-13
Enable call to SurfaceFluxes from boundary conditions, clean up SurfaceFluxes

ilopezgp:
### Description

This PR makes two main changes:

- Cleans up SurfaceFluxes and makes use of more standard notation for some of its arguments.
- Corrects the description of the height at which the SurfaceFluxes functions are evaluated, which is different for DG and FV simulations.
- Enables the use of the output of `surface_conditions` (`SurfaceFluxes` module) in a new type of boundary condition set that can be specified in an experiment setup script. 

The boundary condition specification takes as input functional forms for the temperature and moisture at the ground (similar to `BulkFormulaEnergy`) and for the roughness lengths (Float or Array of floats). The output is the energy flux (for this EnergyBC).

**Disclaimer:** This PR does not add BC based on SurfaceFluxes for momentum and moisture yet. This is because we want to cache the results from the SurfaceFluxes call and use them for all BCs in the future. 

### DIFF
--- a/src/Atmos/Model/bc_energy.jl
+++ b/src/Atmos/Model/bc_energy.jl
@@ -1,10 +1,11 @@
 abstract type EnergyBC end
 
 using ..TurbulenceClosures
-
+using ClimateMachine.SurfaceFluxes:
+    get_energy_flux, surface_conditions, DGScheme
+using ClimateMachine.Atmos: altitude
 """
     Insulating() :: EnergyBC
-
 No energy flux across the boundary.
 """
 struct Insulating <: EnergyBC end
@@ -19,7 +20,6 @@ function atmos_energy_normal_boundary_flux_second_order!(
 
 """
     PrescribedTemperature(fn) :: EnergyBC
-
 Prescribe the temperature at the boundary by `fn`, a function with signature
 `fn(state, aux, t)` returning the temperature (in K).
 """
@@ -76,7 +76,6 @@ end
 
 """
     PrescribedEnergyFlux(fn) :: EnergyBC
-
 Prescribe the net inward energy flux across the boundary by `fn`, a function
 with signature `fn(state, aux, t)`, returning the flux (in W/m^2).
 """
@@ -114,7 +113,6 @@ end
 
 """
     Adiabaticθ(fn) :: EnergyBC
-
 Prescribe the net inward potential temperature flux
 across the boundary by `fn`, a function with signature
 `fn(state, aux, t)`, returning the flux (in kgK/m^2).
@@ -147,11 +145,9 @@ end
 
 """
     BulkFormulaEnergy(fn) :: EnergyBC
-
 Calculate the net inward energy flux across the boundary. The drag
 coefficient is `C_h = fn_C_h(atmos, state, aux, t, normu_int_tan)`. The surface
 temperature and q_tot are `T, q_tot = fn_T_and_q_tot(atmos, state, aux, t)`.
-
 Return the flux (in W m^-2).
 """
 struct BulkFormulaEnergy{FNX, FNTM} <: EnergyBC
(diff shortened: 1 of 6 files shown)


src/Atmos/Model/bc_energy.jl
```diff
@@ -203,3 +198,88 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+        z,
+        DGScheme(),
+    )
+
+    # recover thermo state
+    ts_surf = PhaseEquil_ρTq(atmos.param_set, state⁻.ρ, T, q_tot)
+    # TODO: use the correct density at the surface
+    ρ_avg = average_density(state⁻.ρ, state_int⁻.ρ)
+    # Add sensible heat flux
+    fluxᵀn.energy.ρe -= ρ_avg * θ_flux * cp_m(ts_surf)
```
szy21, line 282: I'm not sure whether `cp_m(ts_surf)` or `cp_m(ts_int)`, or the average of these two is better here. But the effect should be tiny.

ilopezgp: I agree. This is one of the things that I want to discuss... I think we might need to choose different values for FV or DG/CG

src/Atmos/Model/bc_energy.jl
```diff
@@ -203,3 +199,95 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    qt_star_init = FT(1e-5)
+    MO_param_guess = MArray{Tuple{4}, FT}(FT[
+        LMO_init,
+        u_star_init,
+        th_star_init,
+        qt_star_init,
+    ])
+
+    # Roughness and interior lengths
+    z_0 = FT(0.1)
```
szy21, line 270: Should z_0 be an input of the function?

ilopezgp: This is just a simple test to showcase that surface fluxes work, these things will be added later on.

ilopezgp: I ended up adding it :)

src/Atmos/Model/bc_energy.jl
ilopezgp: I ended up using the `surf` values for all the terms in the flux. I think the flux is applied at `z=0`, so this is what makes sense to me.

ilopezgp: Following the discussion about where the momentum flux is defined as well, I think we need to have the guidelines for how to do this written somewhere.

ilopezgp:
bors try

src/Atmos/Model/bc_energy.jl
```diff
@@ -203,3 +199,90 @@ function atmos_energy_normal_boundary_flux_second_order!(
...
+    qt_star_init = q_tot
+    MO_param_guess = MArray{Tuple{4}, FT}(FT[
+        LMO_init,
+        u_star_init,
+        th_star_init,
+        qt_star_init,
+    ])
+
+    # Roughness and interior heights
+    z_0 = bc_energy.fn_z0(atmos, state⁻, aux⁻, t, normu_int⁻_tan)
```
szy21, line 267: A minor point. We may not need `normu_int⁻_tan` in the input?

szy21:
This looks good to me. Thanks Ignacio!

szy21 (approved):

src/Atmos/Model/bc_energy.jl
ilopezgp: I left the same signature as the `C_drag` coefficients for now, in case we want to use some formulation of the roughness height that depends on the tangential velocity.

charleskawczynski:
Is `NishizawaEnergyFlux` exercised anywhere?

ilopezgp:
> Is `NishizawaEnergyFlux` exercised anywhere?

Not by default in the test, only if the argument is parsed. Do we have to add an experiment that uses it as the default?

charleskawczynski:
> Not by default in the test, only if the argument is parsed. Do we have to add an experiment that uses it as the default?

No, but it would be nice if it were exercised, otherwise it will likely get stale pretty quickly. It would be ideal to write a unit test.

ilopezgp:
> > Not by default in the test, only if the argument is parsed. Do we have to add an experiment that uses it as the default?
> 
> No, but it would be nice if it were exercised, otherwise it will likely get stale pretty quickly. It would be ideal to write a unit test.

I don't think we have unit tests for the other `energy_bc`. I am not sure how to create them. We could have an integration test that loops over all boundary conditions, running each one for a couple of seconds... Thoughts on this?

szy21:
> I don't think we have unit tests for the other `energy_bc`. I am not sure how to create them. We could have an integration test that loops over all boundary conditions, running each one for a couple of seconds... Thoughts on this?

When we wrote the other boundary conditions it was hard to add a unit test. Is it still the case now? We ended up adding an integration test as shown [here](https://github.com/CliMA/ClimateMachine.jl/blob/711360d0386d2edbb0dc78d110c0c9a654f9e818/.buildkite/pipeline.yml#L1681).

src/Atmos/Model/bc_energy.jl
```diff
@@ -1,10 +1,11 @@
 abstract type EnergyBC end
 
 using ..TurbulenceClosures
-
+using ClimateMachine.SurfaceFluxes:
+    get_energy_flux, surface_conditions, DGScheme
+using ClimateMachine.Atmos: altitude
```
charleskawczynski, line 6:
```suggestion
using ..SurfaceFluxes:
    get_energy_flux, surface_conditions, DGScheme
```
This file is included in AtmosModel.jl, so we don't need `using ClimateMachine.Atmos: altitude`

charleskawczynski:
> When we wrote the other boundary conditions it was hard to add a unit test.

If we were to write one (without the required boilerplate), what might it look like? This interface:
```julia
atmos_energy_normal_boundary_flux_second_order!(
    nf,
    bc_energy::NishizawaEnergyFlux,
    atmos,
    fluxᵀn,
    n⁻,
    state⁻,
    diffusive⁻,
    hyperdiffusive⁻,
    aux⁻,
    state⁺,
    diffusive⁺,
    hyperdiffusive⁺,
    aux⁺,
    t,
    state_int⁻,
    diffusive_int⁻,
    aux_int⁻,
)
```
is generic, so we can pass `NamedTuple`s through in a unit test. When we type by `::Vars` it requires a bit more leg work, but I think that we can relatively easily construct a unit test for this. I kind of wish we started by writing more unit tests for SurfaceFluxes itself before making this feature available for use.

szy21:
> If we were to write one (without the required boilerplate), what might it look like? This interface:


I was thinking about something like checking whether the output flux is correct given state⁻, aux⁻, state_int⁻, aux_int⁻, etc. But there are many other inputs of the function that I'm not sure how to specify.

ilopezgp:
I will add some more unit tests to SurfaceFluxes here. However, since no BCs have unit tests, I think a unit test of the NishizawaEnergyFlux is beyond the scope of this PR. It should be left for the BC revamp.

PR #2042, opened 2021-02-17
Enable the removal of condensate in the moist BW driver

jiahe23:
### Description

This PR adds the `RemovePrecipitation(true)` to the source terms in the moist baroclinic wave driver, and enables the removal of condensates thru the 0M microphysics scheme. 

It has been tested that the two experiment drivers can run successfully with the changes, and the `RemovePrecipitation` keyword is added to the Mass, Energy, and TotalMoisture in the tendency table.


trontrytel (approved):
LGTM.

The only question would be why we define the same case twice - in AtmosGCM and in TestCase?

jiahe23:
> LGTM.
> 
> The only question would be why we define the same case twice - in AtmosGCM and in TestCase?

The one in the TestCase is the same as described in the DCMIP, and we keeps it there as a standard test. The one in the AtmosGCM is set up with the BW initial condition and run with prescribed SST. It is one step further towards aqua-planet simulations.

PR #2052, opened 2021-02-21
Unit tests and corrections, SurfaceFluxes module

ilopezgp:
### Description

This PR adds _recovery_ unit tests to the `SurfaceFluxes` module. With this motivation:

- I add a `recover_profile` function that generates u(z), θ(z) following Monin-Obukhov Similarity Theory.
- I generate profiles u(z), θ(z) from an initial set of Monin-Obukhov parameters, and test their recovery through `surface_conditions`.

In the development of the tests,

- I found and corrected a problem in the `monin_obukhov_length` function, which allowed division by zero in a physically relevant scenario (no surface fluxes).
- I found the SurfaceFluxes module to **lack robustness in conditions of surface heating** (θ*<0). There exist logarithms of functions that can take negative values in this regime (This can easily be reproduced through the unit tests).

### DIFF
--- a/src/Common/SurfaceFluxes/SurfaceFluxes.jl
+++ b/src/Common/SurfaceFluxes/SurfaceFluxes.jl
@@ -34,7 +34,8 @@ abstract type SurfaceFluxesModel end
 struct FVScheme end
 struct DGScheme end
 
-export surface_conditions, exchange_coefficients
+export surface_conditions,
+    exchange_coefficients, recover_profile, monin_obukhov_length
 
 """
     SurfaceFluxConditions{FT}
@@ -220,6 +221,10 @@ end
 """
     compute_physical_scale(uf, z_in, z_0, x_in, x_s, transport, ::FVScheme)
 
+Returns u_star, θ_star, ... given equations (17), (18) for FV and (8), (9) for DG
+from Nishizawa & Kitamura (2018).
+
+## Arguments
  - `uf` universal function family
  - `z_in` Input height for the similarity functions. It is Δz for FV and the height
     of the first nodal point for DG
@@ -229,7 +234,6 @@ end
  - `x_s` surface values for state variable array `x`
  - `transport` Type of transport (momentum or heat)
 
-    Equations (17), (18) for FV and (8), (9) for DG from Nishizawa & Kitamura (2018)
 """
 function compute_physical_scale(
     uf::AbstractUniversalFunction{FT},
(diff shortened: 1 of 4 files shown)


ilopezgp:
bors try

ilopezgp:
@charleskawczynski is there a GPU-compatible version of` Random.shuffle` ?

charleskawczynski:
> @charleskawczynski is there a GPU-compatible version of` Random.shuffle` ?

I don't think we do. Maybe we can use shuffle if on CPU?

src/Common/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -271,6 +272,40 @@ function compute_physical_scale(
     return _π_group⁻¹ * _von_karman_const / Σterms * (x_in - x_s)
 end
 
+"""
+    recover_profile(z, x_star, x_s, z_0, L_MO, transport, ::DGScheme, uf)
```
charleskawczynski, line 276: Can we improve the name here? I'm not sure what profile this refers to

ilopezgp:
> > @charleskawczynski is there a GPU-compatible version of` Random.shuffle` ?
> 
> I don't think we do. Maybe we can use shuffle if on CPU?

This is an OK solution for me provided that we are also testing in CPU in CI. Is this the case?

src/Common/SurfaceFluxes/SurfaceFluxes.jl
ilopezgp: I am referring to u(z), θ(z), ... inside the computational domain, so the vertical profiles of the variables. Would `recover_vert_profiles` work better?

charleskawczynski: Ok, maybe just clarify what's being returned in the doc string

charleskawczynski:
> Can I add a `cpu_surface_fluxes` test in CI?

Yes

ilopezgp:
bors try

ilopezgp:
bors try

charleskawczynski:
@ilopezgp I can take a look at what's going on. It might be easier to diagnose locally

ilopezgp:
> @ilopezgp I can take a look at what's going on. It might be easier to diagnose locally

I think I found the bug in build kite

ilopezgp:
bors try

src/Common/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -271,6 +272,64 @@ function compute_physical_scale(
...
+ - `z` Input height for evaluation
+ - `x_star` Surface layer scales for state variable array `x`.
+ - `x_s` surface values for state variable array `x`
+ - `z_0` roughness lengths for state variable array `x`
+ - `L_MO` Obukhov length
+ - `transport` Type of transport (momentum or heat)
+ - `uf` universal function family
+
+    Recover vertical profiles u(z), θ(z), ... using equations (4) and (5)
+    for DG and (12), (13) for FV from Nishizawa & Kitamura (2018).
```
charleskawczynski, line 287:
```suggestion
Recover vertical profiles u(z), θ(z), ... using equations (4) and (5)
for DG and (12), (13) for FV from Nishizawa & Kitamura (2018).

## Arguments
  - `z` Input height for evaluation
 - `x_star` Surface layer scales for state variable array `x`.
 - `x_s` surface values for state variable array `x`
 - `z_0` roughness lengths for state variable array `x`
 - `L_MO` Obukhov length
 - `transport` Type of transport (momentum or heat)
 - `uf` universal function family
```

src/Common/SurfaceFluxes/SurfaceFluxes.jl
```diff
@@ -271,6 +272,64 @@ function compute_physical_scale(
...
+"""
+function recover_profile(
```
charleskawczynski, line 289: If we're exporting `recover_profile`, and knowing how much `monin_obukhov_length` will likely be used, can we write a method `recover_profile(:: SurfaceFluxConditions)` and `monin_obukhov_length(:: SurfaceFluxConditions)` so that users are encouraged to use a simpler interface? It might be easier, even if we need to add more properties to `SurfaceFluxConditions`.

ilopezgp: The SurfaceFluxConditions contains a Monin Obukhov Length, this is kind of a chicken and egg situation.

ilopezgp: And the SurfaceFluxConditions are built with the iterative solve. Both `monin_obuhov_length` and recover_profile are not iterative. In addition, the output of `recover_profile` is an input in `surface_conditions`.

charleskawczynski: Oh, I see. It's more acting as a helper for constructing the input. This is just for the ICs, though, right? It might still be a good idea to define
```julia
monin_obukhov_length(sc::SurfaceConditions) = sc.L_MO
```
In case we're able to eventually hide the ICs from the user. Maybe it'll be easier to know if this is useful after seeing it used, though. Up to you.

ilopezgp: I agree this is a good idea, which we can probably use in the EDMF closures. I will add this getter function.

ilopezgp:
bors try

ilopezgp:
bors try

ilopezgp:
bors try

ilopezgp:
bors try

ilopezgp:
There we go! @charleskawczynski

charleskawczynski (approved):

PR #2060, opened 2021-02-25
Equidistant cubed-sphere

### LINKED ISSUE #1778: Explore a different type of cubed-sphere
### Description

This is probably not an urgent matter, considering the more pressing issues on the Numerics side we are all working on right now (e.g., split discretizations), but I'd like to write this down here more as food for thoughts for the people who are not familiar with this topic and would like to learn more.

There are several ways in which one can define a cubed-sphere. 

Among the most popular ones for the solution of PDEs are the equi-distant and the equi-anglular ones (but there is also one that is said to [best preserve Jacobians](https://journals.ametsoc.org/mwr/article/145/3/1083/103369/A-Nonhydrostatic-Multiscale-Model-on-the-Uniform)). See this nice interactive [tool](http://acmg.seas.harvard.edu/geos/cubed_sphere/CubeSphere_comparision.html) to observe the differences between them. 

Mainly, the equi-angular one is the most commonly found in the literature, because most papers that treat solutions of PDEs on cubed-spheres, for instance for the shallow-water equations, utilize Finite Differences approximations, for which the important aspect is that edges of elements that belong to different patches (or panels) of the cube are straight. But for Finite Element discretizations (either CG or DG), this requirement is not important, since elements are disjoint and there is no preference or requirement to have co-planar elements. Moreover, this type of cubed-sphere generates smaller elements around the corners of the cube, which might exacerbate the CFL condition more than you might expect, because smaller elements means closer nearest-neighboring nodes.  

On the other hand, the equi-distant cubed-sphere allows for a better ratio of smallest element to largest element, thus allows further apart nearest-neighboring nodes, and might help a little with CFL (see [this reference](https://rmets.onlinelibrary.wiley.com/doi/abs/10.1002/qj.49712253209)). 

Currently in the ClimateMachine, we have the [equi-angular](https://github.com/CliMA/ClimateMachine.jl/blob/3154952d43bf7048f3c633bcff8cd03f304fa6f6/src/Numerics/Mesh/Topologies.jl#L916-L971) one.

valeriabarra:
### Description

This PR adds support for an equidistant cubed-sphere (as defined in [Rancic at al](https://doi.org/10.1002/qj.49712253209) and [Nair et al](https://doi.org/10.1175/MWR2890.1) ). 

Things done in this PR so far: 
- [x] Renamed existing `cubedshellwarp` -> `cubed_sphere_warp` (changed shell to sphere and used snake case, for better readability and accessibility). Similarly, renamed `cubedshellunwarp` -> `cubed_sphere_unwarp`. For assignments of functors, I then needed to create different specialized wrappers: `equiangular_cubed_sphere_warp`, `equiangular_cubed_sphere_unwarp`, `equidistant_cubed_sphere_warp`, and `equidistant_cubed_sphere_unwarp`.
- [x] Tests for the equidistant cubed-sphere warp/unwarp functions have been added (the same as for the equiangular case). 

This will close #1778. 

Things to think about/do: 
- [x] How to change the interface to give the user the possibility to choose between one or the other cubed-sphere? Perhaps @charleskawczynski can give me a hint.
~~- [ ] Confirm that using the equidistant cubed-sphere in an experiment should improve the CFL condition.~~ <-- Edit: This is actually not the case with the equidistant cubed sphere, my bad for misinterpreting Rancic's paper.

### DIFF
--- a/src/Driver/driver_configs.jl
+++ b/src/Driver/driver_configs.jl
@@ -355,7 +355,7 @@ function AtmosGCMConfiguration(
     ),
     mpicomm = MPI.COMM_WORLD,
     grid_stretching = (nothing,),
-    meshwarp::Function = cubedshellwarp,
+    meshwarp::Function = equiangular_cubed_shell_warp,
     numerical_flux_first_order = RusanovNumericalFlux(),
     numerical_flux_second_order = CentralNumericalFluxSecondOrder(),
     numerical_flux_gradient = CentralNumericalFluxGradient(),
--- a/src/Driver/driver_configs.jl
+++ b/src/Driver/driver_configs.jl
@@ -355,7 +355,7 @@ function AtmosGCMConfiguration(
     ),
     mpicomm = MPI.COMM_WORLD,
     grid_stretching = (nothing,),
-    meshwarp::Function = equiangular_cubed_shell_warp,
+    meshwarp::Function = equiangular_cubed_sphere_warp,
     numerical_flux_first_order = RusanovNumericalFlux(),
     numerical_flux_second_order = CentralNumericalFluxSecondOrder(),
     numerical_flux_gradient = CentralNumericalFluxGradient(),
--- a/src/Numerics/Mesh/Topologies.jl
+++ b/src/Numerics/Mesh/Topologies.jl
@@ -12,11 +12,11 @@ export AbstractTopology,
     NoTopography,
     DCMIPMountain,
     isstacked,
-    cubedshellwarp,
+    equiangular_cubed_shell_warp,
     cubedshelltopowarp,
     compute_lat_long,
     compute_analytical_topography,
-    cubedshellunwarp
+    equiangular_cubed_shell_unwarp
 
 export grid1d, SingleExponentialStretching, InteriorStretching
 
@@ -761,7 +761,7 @@ end
 
 Generate a cubed shell mesh with the number of elements along each dimension of
 the cubes being `Nelem`. This topology actual creates a cube mesh, and the
-warping should be done after the grid is created using the `cubedshellwarp`
+warping should be done after the grid is created using the `equiangular_cubed_shell_warp`
 function. The coordinates of the points will be of type `T`.
 
 The elements of the shell are partitioned equally across the MPI ranks based
@@ -785,13 +785,13 @@ topology = CubedShellTopology(MPI.COMM_SELF, 10, Float64)
 # Shell radius = 1
 x1, x2, x3 = ntuple(j->topology.elemtocoord[j, :, :], 3)
 for n = 1:length(x1)
-   x1[n], x2[n], x3[n] = Topologies.cubedshellwarp(x1[n], x2[n], x3[n])
+   x1[n], x2[n], x3[n] = Topologies.equiangular_cubed_shell_warp(x1[n], x2[n], x3[n])
 end
 
 # Shell radius = 10
 x1, x2, x3 = ntuple(j->topology.elemtocoord[j, :, :], 3)
 for n = 1:length(x1)
-  x1[n], x2[n], x3[n] = Topologies.cubedshellwarp(x1[n], x2[n], x3[n], 10)
+  x1[n], x2[n], x3[n] = Topologies.equiangular_cubed_shell_warp(x1[n], x2[n], x3[n], 10)
 end
 ```
 """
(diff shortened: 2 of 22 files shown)


valeriabarra:
bors try

valeriabarra:
The mesh for the generated equidistant cubed-sphere looks like this. Makie code can be found [here](https://gist.github.com/valeriabarra/e92f165008492b3208e2110672a3cef3).

image

smarras79:
@valeriabarra is it really equidistant? It doesn't seem to be the case from that plot.

jkozdon:
Thanks @valeriabarra! I'll take a look in the next couple days. Feel free to ping me again if I forget.

> How to change the interface to give the user the possibility to choose between one or the other cubed-sphere? Perhaps @charleskawczynski can give me a hint.

Possibly we should be using the type system for defining the warping. Since the way the interpolation is done we need to unwarp, so its possible that the mesh warping needs to be type that get carried around with the mesh.

Side question: Does the interpolation machinery in the diagnostic currently work with orography? Or is a perfect sphere assumed? I don't see any non-linear iterations on a warping function, so I would guess that topography is not support. Doing this all through the type space could help simplify the interface.

src/Numerics/Mesh/Topologies.jl
```diff
@@ -1020,38 +1026,83 @@ function cubedshellunwarp(x1, x2, x3)
...
+
+    x1 = R * a / r
+    x2 = R * b / r
+    x3 = R * c / r
+
+    return x1, x2, x3
+end
+
+"""
+equidistant_cubed_shell_unwarp(x1, x2, x3)
```
charleskawczynski, line 1078:
```suggestion
    equidistant_cubed_shell_unwarp(x1, x2, x3[, R])
```
Can we document the arguments here?

test/Numerics/Mesh/topology.jl
```diff
@@ -4,69 +4,209 @@ using Combinatorics, MPI
...
+    import ClimateMachine.Mesh.Topologies: equiangular_cubed_shell_warp
 
     @testset "check radius" begin
-        @test hypot(cubedshellwarp(3.0, -2.2, 1.3)...) ≈ 3.0 rtol = eps()
-        @test hypot(cubedshellwarp(-3.0, -2.2, 1.3)...) ≈ 3.0 rtol = eps()
-        @test hypot(cubedshellwarp(1.1, -2.2, 3.0)...) ≈ 3.0 rtol = eps()
-        @test hypot(cubedshellwarp(1.1, -2.2, -3.0)...) ≈ 3.0 rtol = eps()
-        @test hypot(cubedshellwarp(1.1, 3.0, 0.0)...) ≈ 3.0 rtol = eps()
-        @test hypot(cubedshellwarp(1.1, -3.0, 0.0)...) ≈ 3.0 rtol = eps()
+        @test hypot(equiangular_cubed_shell_warp(3.0, -2.2, 1.3)...) ≈ 3.0 rtol =
```
charleskawczynski, line 11: Just a thought, it might be nice to make a short alias for `equiangular_cubed_shell_warp` so that we don't get hammered by the formatter

valeriabarra:
bors try

valeriabarra:
bors try

test/Numerics/Mesh/topology.jl
valeriabarra: Thanks for the idea. What alias would you have in mind in this case? I specifically wanted to use snake case for accessibility and readability though

src/Numerics/Mesh/Topologies.jl
valeriabarra: Thanks @charleskawczynski I believe I improved the string. Let me know if it is not clear enough. 
I'm not quite sure I understand your suggested snippet of code. Would in that case `x3` be an array?

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
@smarras79 it is equidistant. However, I am analyzing Rancic's paper again, and it turns out that actually the equiangular gnomonic one (what we have already in CliMA) is the one that has the best ratio of smallest to largest element in the grid. Hence, this new cubed-sphere shouldn't help with the CFL at all, as I was hoping. 

Good thing it didn't take me much to try this out. We can all decide if we don't want to add it then.

smarras79:
> @smarras79 it is equidistant. However, I am analyzing Rancic's paper again, and it turns out that actually the equiangular gnomonic one (what we have already in CliMA) is the one that has the best ratio of smallest to largest element in the grid. Hence, this new cubed-sphere shouldn't help with the CFL at all, as I was hoping.
> 
> Good thing it didn't take me much to try this out. We can all decide if we don't want to add it then.

Right, no matter what cubed sphere we have, none of them will really help with the Courant value because we don't have that constraining in the horizontal direction. Our limiting factor is the vertical grid distance. We can gain a little, but really nothing significant. Thanks for putting this in though and making it easier to add new ones as we please.

src/Numerics/Mesh/Topologies.jl
charleskawczynski: The `[, R]` is just to indicate an optional argument.

test/Numerics/Mesh/topology.jl
charleskawczynski: 🤷🏻‍♂️ `warp_ecs`? Again, just a thought. Up to you.

src/Numerics/Mesh/Topologies.jl
charleskawczynski: The 4 spaces before `equidistant_cubed_shell_unwarp` is more important, since it's needed for Documenter.jl to recognise and format `equidistant_cubed_shell_unwarp` as code.

valeriabarra: Thanks Charlie!

valeriabarra: Added the indentation

valeriabarra:
bors try

src/Driver/driver_configs.jl
```diff
@@ -355,7 +355,7 @@ function AtmosGCMConfiguration(
     ),
     mpicomm = MPI.COMM_WORLD,
     grid_stretching = (nothing,),
-    meshwarp::Function = cubedshellwarp,
+    meshwarp::Function = equiangular_cubed_shell_warp,
```
valeriabarra, line 358: @charleskawczynski in the case of an assignment like this one, how would it look with the new interface you showed me? 

Like 
```
meshwarp::Function = cubed_sphere_warp(EquiangularCubedSphere()),
```
with no additional arguments?

valeriabarra: Or maybe I figured! 
Here I would just keep the assignment as is (just renaming the function)

```
meshwarp::Function = cubed_sphere_warp,
```
And then later, when `meshwarp` is used, I would used it with the specialization?
But later on in the file I see it is only used as `meshwarp = meshwarp`

valeriabarra: I think this question might turn out to be a bit more complex than I initially thought. Since I have seen that this `meshwarp` is used in every occurrence of `DiscontinuousSpectralElementGrid` , and this is used a gazillion times. 
Can I have a pointer on how to solve this in a Julian way?
cc: @charleskawczynski @jkozdon

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
This is ready for review! 
It passes tests locally for me. 
At the moment, this cubed-sphere is not set up to be used in any test/experiment, since it turned out that it is not advantageous for what I was hoping. 
But this PR adds capability to support different types of cubed-sphere in the future. Thanks for any feedback.

valeriabarra:
When approved, I will take care of reducing commits/squashing

valeriabarra:
bors try

valeriabarra:
@jkozdon this is ready for review. 

My only question is: should I have also included a new version for the [cubedshelltopowarp](https://github.com/CliMA/ClimateMachine.jl/blob/b95346123ca8b856cbf545386d039f694689a6f6/src/Numerics/Mesh/Topologies.jl#L1478) function? I would think so, since it uses the equiangular cubed-sphere formulas, but since we haven't discussed this one, I wanted to make sure. Thank you!

PR #2074, opened 2021-03-03
Add `--checkpoint-on-crash`

### LINKED ISSUE #1468: Save diagnostics at the timestep before crashing
# Description

It would be useful for debugging to output the interpolated diagnostics (and/or dump un-interpolated state/aux variables) for the timestep before the model crashes. We have one crude version that runs on 1 rank in [this branch](https://github.com/CliMA/ClimateMachine.jl/tree/ln/save_crashed_tstep_1rank) (the functionality for ntasks=1 should be obtainable by just making the modifications in `src/Driver/Driver.jl` and `src/Diagnostics/atmos_gcm_default.jl` ) , but this takes a while. 

# Additional context

# For CLIMA Developers

### LINKED ISSUE #1510: Heartbeat with synchronised error handling

kpamnany:
### Description

If specified, `--checkpoint-on-crash` adds exception checking to the `wait` call in `MPIStateArrays` which is where most such crashes will be detected. An `Allreduce` call is added here which notifies all MPI ranks of the crash; a checkpoint is then written.

I've tested this using the baroclinic wave experiment (with the [filter](https://github.com/CliMA/ClimateMachine.jl/blob/master/experiments/TestCase/baroclinic_wave.jl#L288) removed to force a crash) on CPU and GPU with 1, 2 and 32 MPI ranks and it seems to work fine.

Closes #1468 and #1510.

### DIFF
--- a/src/Arrays/MPIStateArrays.jl
+++ b/src/Arrays/MPIStateArrays.jl
@@ -14,6 +14,8 @@ using ..VariableTemplates:
 
 using Base.Broadcast: Broadcasted, BroadcastStyle, ArrayStyle
 
+struct ErrorOnRemoteNode <: Exception end
+
 # This is so we can do things like
 #   similar(Array{Float64}, Int, 3, 4)
 Base.similar(::Type{A}, ::Type{FT}, dims...) where {A <: Array, FT} =
@@ -29,7 +31,13 @@ cpuify(x::AbstractArray) = convert(Array, x)
 cpuify(x::Real) = x
 
 export MPIStateArray,
-    euclidean_distance, weightedsum, array_device, vars, show_not_finite_fields
+    euclidean_distance,
+    weightedsum,
+    array_device,
+    vars,
+    show_not_finite_fields,
+    ErrorOnRemoteNode,
+    checked_wait
 
 """
     MPIStateArray{FT, DATN<:AbstractArray{FT,3}, DAI1, DAV,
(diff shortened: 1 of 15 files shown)


LenkaNovak:
Works like a charm! 🏆

LenkaNovak (approved):

src/Arrays/MPIStateArrays.jl
```diff
@@ -888,4 +900,38 @@ function show_not_finite_fields(Q::MPIStateArray)
...
+    end
+end
+function checked_wait(
+    device::CUDADevice,
+    event,
+    progress = nothing,
+    check = false,
+)
+    wait(device, event, progress)
+end
```
jkozdon, line 935: Where does the GPU exception get caught? (if this works on the GPU?)

jkozdon (commented):
Overall this seems reasonable. I haven't had a chance to run the code yet, but how does the GPU error get caught? Also, I assume that when this is used parallel performance takes a big hit? So should that be noted?

kpamnany:
> Overall this seems reasonable. I haven't had a chance to run the code yet, but how does the GPU error get caught? Also, I assume that when this is used parallel performance takes a big hit? So should that be noted?

`wait`s on `CUDADevice` are asynchronous with respect to the host, so they cannot `throw`; any exceptions are gathered and carried along until a `wait` on `CPU` is issued (which for us happens in `end_ghost_exchange`). At this point, those exceptions are thrown to us. So yes, this does work on the GPU.

Also yes, the `Allreduce` will hurt performance though I haven't quantified how much. I'll add help text about this though, good point.

PR #2076, opened 2021-03-03
Remove parameter type in Impenetrable

charleskawczynski:
### Description

When trying to refactor the balance law boundary conditions, it's difficult to work with `Impenetrable` because, without the parameter, it only specifies BCs on a single component of the momentum. @blallen, mind if I spread these changes to the ocean, too?

### DIFF
--- a/docs/src/APIs/Atmos/AtmosModel.md
+++ b/docs/src/APIs/Atmos/AtmosModel.md
@@ -72,11 +72,9 @@ ClimateMachine.Atmos.RayleighSponge
 
 ```@docs
 ClimateMachine.Atmos.AtmosBC
-ClimateMachine.Atmos.DragLaw
 ClimateMachine.Atmos.Impermeable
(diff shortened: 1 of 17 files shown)


charleskawczynski:
We may not need this. Leaving on back burner for the moment.

blallen:
From our side, @sandreza and I would prefer if we don't remove the flexible parameter approach. We think it makes more sense to have these be composable since will the impenetrable/penetrable BC constraint is applied on momentum, it really is a boundary condition on the mass and mass flux. Also, we actually use both Penetrable and Impenetrable BCs in the Ocean models, so it is nice to be able to freely compose those with FreeSlip, NoSlip, and MomentumFlux BCs on the momentum... even if it is only a savings of one type it is much less code redundancy.

charleskawczynski:
Ok, I don't think this is a big issue. Closing.

PR #2083, opened 2021-03-04
Fix call to f_h

ilopezgp:
### Description

This PR fixes a small bug regarding the universal function `Psi_h`. `f_h` is only needed in conditions of `ζ<0`, and can be ill-defined otherwise. This PR moves the call to that function to the appropriate conditional block.

### DIFF
--- a/src/Common/SurfaceFluxes/UniversalFunctions.jl
+++ b/src/Common/SurfaceFluxes/UniversalFunctions.jl
@@ -207,11 +207,11 @@ function Psi(uf::Businger, ζ, tt::HeatTransport)
             return -9 * ζ / 4
         end
     else
-        f_h = f_heat(uf, ζ)
-        if uf.L >= 0
+        if ζ >= 0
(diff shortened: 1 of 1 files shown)


charleskawczynski (commented):
Moving `f_h` looks good. This is interesting, I'm not sure why I used `ζ` above-- the paper uses `L` (Nishizawa, eq. A5, A6). I'm fine with these changes if they make sense / fix things.

ilopezgp:
> Moving `f_h` looks good. This is interesting, I'm not sure why I used `ζ` above-- the paper uses `L` (Nishizawa, eq. A5, A6). I'm fine with these changes if they make sense / fix things.

Yes, the paper uses `L` but I changed it to `ζ` consistent with the rest of the code. Both are equivalent since `ζ=z/L` and z is positive.

charleskawczynski (approved):

charleskawczynski:
Sounds good. bors r+

PR #2084, opened 2021-03-04
enable netcdf diagnostics for stable boundary layer case in EDMF set up

jiahe23:
### Description

This PR adds a single stack diagnostic configuration for the `stable_bl_edmf.jl` driver that enables the netcdf output for the experiment.

```
CLIMA_OUTPUT='/central/scratch/jiahe/edmf_sbl'
julia --project --diagnostics default --output-dir $CLIMA_OUTPUT
```
dumps the following two data
```
SBL_EDMF_DumpState.nc
SBL_EDMF_DumpTendencies.nc
```

vars in `SBL_EDMF_DumpState.nc`
```
netcdf SBL_EDMF_DumpState {
dimensions:
	x = 1 ;
	y = 1 ;
	z = 9 ;
	time = UNLIMITED ; // (1 currently)
variables:
	double x(x) ;
	double y(y) ;
	double z(z) ;
	double time(time) ;
		time:units = "seconds since 1900-01-01 00:00:00" ;
		time:long_name = "time" ;
	double ρ(time, z, y, x) ;
	double ρu\[1\](time, z, y, x) ;
	double ρu\[2\](time, z, y, x) ;
	double ρu\[3\](time, z, y, x) ;
	double energy.ρe(time, z, y, x) ;
	double turbconv.environment.ρatke(time, z, y, x) ;
	double turbconv.environment.ρaθ_liq_cv(time, z, y, x) ;
	double turbconv.environment.ρaq_tot_cv(time, z, y, x) ;
	double turbconv.environment.ρaθ_liq_q_tot_cv(time, z, y, x) ;
	double turbconv.updraft\[1\].ρa(time, z, y, x) ;
	double turbconv.updraft\[1\].ρaw(time, z, y, x) ;
	double turbconv.updraft\[1\].ρaθ_liq(time, z, y, x) ;
	double turbconv.updraft\[1\].ρaq_tot(time, z, y, x) ;
}
```

### DIFF
--- a/test/Atmos/EDMF/stable_bl_edmf.jl
+++ b/test/Atmos/EDMF/stable_bl_edmf.jl
@@ -154,7 +154,7 @@ function main(::Type{FT}, cl_args) where {FT}
     )
     # ---
 
-    dgn_config = config_diagnostics(driver_config)
+    dgn_config = config_diagnostics(driver_config, timeend)
 
     cbtmarfilter = GenericCallbacks.EveryXSimulationSteps(1) do
         Filters.apply!(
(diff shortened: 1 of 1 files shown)


jiahe23:
@charleskawczynski

Here I only added the `config_diagnostics` inside `stable_bl_edmf.jl` and modified the main function so that it calls it (instead of the one inside the `stable_bl_model.jl`).

I think we should also delete the `config_diagnostics` from `stable_bl_model.jl` and add it directly to `stable_bl_les.jl` in a separate PR. What do you think?

charleskawczynski (commented):
Looks fine to me. Should we remove `config_diagnostics(driver_config)` from `experiments/AtmosLES/stable_bl_model.jl` and fix this problem everywhere?

jiahe23:
> Looks fine to me. Should we remove `config_diagnostics(driver_config)` from `experiments/AtmosLES/stable_bl_model.jl` and fix this problem everywhere?

I was think we should do it in a separate PR, because this will require to add `config_diagnostics()` in `experiments/AtmosLES/stable_bl_les.jl`. It looks like the one in bomex_les is modified to take into more input arguments, that we could follow for the stable_bl_les.

ilopezgp:
The new function looks pretty general. Could we have the `config_diagnostics` defined in some `helper_funcs` file, so it is accessible to all EDMF simulations? If so, it would be a good idea to expose the dump intervals as an argument to the function instead of having it hardcoded.

PR #2088, opened 2021-03-04
Fix identity operation in reduce_nodal_stack

jkozdon:

### DIFF
--- a/src/Utilities/SingleStackUtils/SingleStackUtils.jl
+++ b/src/Utilities/SingleStackUtils/SingleStackUtils.jl
@@ -315,7 +315,21 @@ function reduce_nodal_stack(
 
     state_data = array_device(Q) isa CPU ? Q.realdata : Array(Q.realdata)
     z = vrange.start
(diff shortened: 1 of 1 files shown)


charleskawczynski (approved):
I see what you mean, @jkozdon, about the double counting on the first nodal stack. This looks like a bug fix.

kpamnany (approved):
Good catch!

jkozdon:
Looking at this test more, it seems to me that using `+` or `*` in the `reduce_nodal_stack` is really weird. I'll leave them for now, but it's just a bit odd (since the returned `z` value doesn't really mean anything in this case)

jkozdon:
> Looking at this test more, it seems to me that using `+` or `*` in the `reduce_nodal_stack` is really weird. I'll leave them for now, but it's just a bit odd (since the returned `z` value doesn't really mean anything in this case)

To clarify, I think the code still does the right thing for the result, it's just the `z` value doesn't have a well defined meaning.

PR #2089, opened 2021-03-05
improve diagnostic configuration interface for edmf

jiahe23:
This PR adds diagnostics_configuration in edmf helper_function and enables netcdf diagnostics in bomex and sbl edmf driver.

### Description

### DIFF
--- a/test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
+++ b/test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
@@ -1,0 +1,42 @@
+"""
+    config_diagnostics(driver_config, timeend; interval=nothing)
+
+Returns the state and tendency diagnostic groups
+"""
+
+function config_diagnostics(driver_config, timeend; interval = nothing)
+    FT = eltype(driver_config.grid)
+    info = driver_config.config_info
+    if interval == nothing
+        interval = "$(cld(timeend, 2) + 10)ssecs"
+        #interval = "10steps"
+    end
+
+    boundaries = [
+        FT(0) FT(0) FT(0)
+        FT(info.hmax) FT(info.hmax) FT(info.zmax)
+    ]
+    axes = (
+        [FT(1)],
+        [FT(1)],
+        collect(range(boundaries[1, 3], boundaries[2, 3], step = FT(50)),),
+    )
+    interpol = ClimateMachine.InterpolationConfiguration(
+        driver_config,
+        boundaries;
+        axes = axes,
+    )
+    ds_dgngrp = setup_dump_state_diagnostics(
+        SingleStackConfigType(),
+        interval,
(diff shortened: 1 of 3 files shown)


test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    boundaries = [
+        FT(0) FT(0) FT(0)
+        FT(info.hmax) FT(info.hmax) FT(info.zmax)
+    ]
+    axes = (
+        [FT(1)],
+        [FT(1)],
+        collect(range(boundaries[1, 3], boundaries[2, 3], step = FT(50)),),
```
ilopezgp, line 22: What does the `step` represent here?

test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
```diff
@@ -0,0 +1,42 @@
...
+    config_diagnostics(driver_config, timeend; interval=nothing)
+
+Returns the state and tendency diagnostic groups
+"""
+
+function config_diagnostics(driver_config, timeend; interval = nothing)
+    FT = eltype(driver_config.grid)
+    info = driver_config.config_info
+    if interval == nothing
+        interval = "$(cld(timeend, 2) + 10)ssecs"
```
ilopezgp, line 11: Does this mean that the default interval is ~half of the simulation? Could we do 1/10th at least?

test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
charleskawczynski: `step` is an optional kwarg to Julia's `range`.

test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
```diff
@@ -0,0 +1,42 @@
+"""
+    config_diagnostics(driver_config, timeend; interval=nothing)
+
+Returns the state and tendency diagnostic groups
+"""
+
```
charleskawczynski, line 6:
```suggestion
```
We can't have spaces between doc strings and functions

test/Atmos/EDMF/helper_funcs/diagnostics_configuration.jl
charleskawczynski: This was @kpamnany's suggestion. It might be nice if we made a struct and internalize this logic (it's just not immediately intuitive). Basically, IIUC, this code sets the interval to output once during, roughly, the middle of the simulation. I believe diagnostics are automatically dumped at the first and last steps, so it should result in ~3 outputs total.

ilopezgp (approved):

PR #2100, opened 2021-03-09
Add AtmosPhysics constructor, and use in drivers

charleskawczynski:
### Description

This PR
 - Adds a `AtmosPhysics` outer constructor, which takes over most of the `AtmosModel` constructor
 - Constructs `AtmosPhysics` before `AtmosModel in the drivers
 - Defines `prognostic_vars(::AtmosPhysics)`, so that we can use it for pre-processing BCs.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -1,6 +1,7 @@
 module Atmos
 
 export AtmosModel,
+    AtmosPhysics,
     AtmosAcousticLinearModel,
     AtmosAcousticGravityLinearModel,
     HLLCNumericalFlux,
@@ -174,6 +175,46 @@ struct AtmosPhysics{FT, PS, RS, E, M, C, T, TC, HD, VS, P, R, TR, LF}
 end
 
 """
+    AtmosPhysics{FT}()
+
+Constructor for `AtmosPhysics`.
+"""
+function AtmosPhysics{FT}(
+    param_set::AbstractParameterSet;
+    energy = TotalEnergyModel(),
+    ref_state = HydrostaticState(DecayingTemperatureProfile{FT}(param_set),),
+    turbulence = SmagorinskyLilly{FT}(C_smag(param_set)),
+    turbconv = NoTurbConv(),
+    hyperdiffusion = NoHyperDiffusion(),
+    viscoussponge = NoViscousSponge(),
+    moisture = EquilMoist(),
+    precipitation = NoPrecipitation(),
+    radiation = NoRadiation(),
+    tracers = NoTracers(),
+    lsforcing = NoLSForcing(),
+    compressibility = Compressible(),
+) where {FT <: AbstractFloat}
+
+    args = (
+        param_set,
+        ref_state,
+        energy,
+        moisture,
+        compressibility,
+        turbulence,
+        turbconv,
+        hyperdiffusion,
+        viscoussponge,
+        precipitation,
+        radiation,
+        tracers,
+        lsforcing,
+    )
+    return AtmosPhysics{FT, typeof.(args)...}(args...)
+end
+
+
+"""
     AtmosModel <: BalanceLaw
 
 A `BalanceLaw` for atmosphere modeling. Users may over-ride prescribed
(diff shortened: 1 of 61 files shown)


charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -213,49 +302,38 @@ Constructor for `AtmosModel` (where `AtmosModel <: BalanceLaw`).
 """
 function AtmosModel{FT}(
     orientation::Orientation,
-    param_set::AbstractParameterSet;
+    physics::AtmosPhysics;
```
ilopezgp, line 305: Can we have the same order of args in the AtmosModel function and the structure? e.g. physics, problem, orientation here

src/Atmos/Model/AtmosModel.jl
```diff
@@ -213,49 +302,38 @@ Constructor for `AtmosModel` (where `AtmosModel <: BalanceLaw`).
...
+        physics.turbulence,
+        physics.turbconv,
+        physics.hyperdiffusion,
+        physics.viscoussponge,
+        physics.moisture,
+        physics.precipitation,
+        physics.radiation,
+        physics.tracers,
+        physics.lsforcing,
+        physics.compressibility,
```
ilopezgp, line 336: I suppose this is the part that will change in the future?

charleskawczynski: Yes

src/Atmos/Model/AtmosModel.jl
charleskawczynski: I don't think so because the physics/orientation are non-optional at this point, and the problem/source etc. here are the defaults for all configs (GCM, LES etc.).

ilopezgp (approved):

src/Atmos/Model/AtmosModel.jl
charleskawczynski: We could get rid of the duplicate fields, and define
```julia
const PhysicsSyms = fieldnames(AtmosPhysics)
const PhysicsProps = Union{typeof.(Val.(PhysicsSyms))...}
unval(::Type{Val{S}}) where {S} = S
Base.getproperty(atmos::AtmosModel, sym::Symbol) = getproperty(atmos, Val(sym))
Base.getproperty(atmos::AtmosModel, ::T) where {T <: PhysicsProps} = getfield(atmos.physics, unval(T))
Base.getproperty(atmos::AtmosModel, ::T) where {T} = getfield(atmos, unval(T))
```
but I think there are reservations about overloading `Base.getproperty(::AtmosModel, ::Symbol)`. @simonbyrne ?

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Actually, that's not GPU compatible, but this works:
```julia
function Base.getproperty(atmos::AtmosModel, sym::Symbol)
    if sym === :problem
        return Base.getfield(atmos, sym)
    elseif sym === :orientation
        return Base.getfield(atmos, sym)
    elseif sym === :source
        return Base.getfield(atmos, sym)
    elseif sym === :data_config
        return Base.getfield(atmos, sym)
    elseif sym === :physics
        return Base.getfield(atmos, sym)
    else
        physics = Base.getfield(atmos, :physics)
        return Base.getproperty(physics, sym)
    end
end
```

charleskawczynski: I'm still not quite happy with the duplication, nor with this solution. I've opened #2105 (and we'll need several similar PRs for the other properties) in hopes to avoid the duplication and not need this hack.

charleskawczynski:
I'm going to break this up into several PRs to avoid code duplication.

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
charleskawczynski: This has now been resolved as we have moved all of the properties into `AtmosPhysics` without duplicating.

charleskawczynski:
bors try

src/Atmos/Model/AtmosModel.jl
```diff
@@ -205,19 +246,33 @@ struct AtmosModel{FT, PH, PR, O, S, DC} <: BalanceLaw
...
+turbulence_model(atmos::AtmosModel) = turbulence_model(atmos.physics)
+turbconv_model(atmos::AtmosModel) = turbconv_model(atmos.physics)
+hyperdiffusion_model(atmos::AtmosModel) = hyperdiffusion_model(atmos.physics)
+viscoussponge_model(atmos::AtmosModel) = viscoussponge_model(atmos.physics)
+precipitation_model(atmos::AtmosModel) = precipitation_model(atmos.physics)
+radiation_model(atmos::AtmosModel) = radiation_model(atmos.physics)
+tracer_model(atmos::AtmosModel) = tracer_model(atmos.physics)
+lsforcing_model(atmos::AtmosModel) = lsforcing_model(atmos.physics)
+
+parameter_set(physics::AtmosPhysics) = physics.param_set
```
charleskawczynski, line 263: @kpamnany what do you think about defining the same method for `AtmosPhysics`? Does this look okay?

charleskawczynski:
Once this merges, we should be able to revive #2122 with a much cleaner approach.

src/Atmos/Model/AtmosModel.jl
kpamnany: Looks fine to me.

PR #2105, opened 2021-03-11
Add parameter_set method, and remove .param_set calls

charleskawczynski:
### Description

This PR adds `parameter_set` to `BalanceLaws` and uses this in favor of `.param_set`. If we create/use enough getter wrappers, then we won't need to duplicate code in #2100.

### DIFF
--- a/experiments/AtmosLES/dycoms.jl
+++ b/experiments/AtmosLES/dycoms.jl
@@ -108,9 +108,10 @@ function flux(::DYCOMSRadiation{Energy}, atmos, args)
     # Constants
     upward_flux_from_cloud = m.F_0 * exp(-aux.∫dnz.radiation.attenuation_coeff)
     upward_flux_from_sfc = m.F_1 * exp(-aux.∫dz.radiation.attenuation_coeff)
+    param_set = parameter_set(atmos)
     free_troposphere_flux =
         m.ρ_i *
-        FT(cp_d(atmos.param_set)) *
+        FT(cp_d(param_set)) *
         m.D_subsidence *
         m.α_z *
         cbrt(Δz_i) *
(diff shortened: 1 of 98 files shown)


experiments/AtmosLES/dycoms.jl
```diff
@@ -108,9 +108,10 @@ function flux(::DYCOMSRadiation{Energy}, atmos, args)
     # Constants
     upward_flux_from_cloud = m.F_0 * exp(-aux.∫dnz.radiation.attenuation_coeff)
     upward_flux_from_sfc = m.F_1 * exp(-aux.∫dz.radiation.attenuation_coeff)
+    param_set
```
ilopezgp, line 111: param_set = parameter_set(atmos) ?

src/Atmos/Model/bc_energy.jl
```diff
@@ -35,8 +35,9 @@ function atmos_energy_boundary_state!(
 )
     @unpack aux⁻, state⁻, t = args
     FT = eltype(aux⁻)
-    _T_0::FT = T_0(atmos.param_set)
-    _cv_d::FT = cv_d(atmos.param_set)
+    param_set
```
ilopezgp, line 38: I see this line several times. Is it supposed to define the param_set from the getter, but the RHS is missing?

charleskawczynski: Thanks, fixed.

experiments/AtmosLES/dycoms.jl
charleskawczynski: Thanks, fixed.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

kpamnany (commented):
I'm not seeing what this accomplishes?

charleskawczynski:
> I'm not seeing what this accomplishes?

It's still a WIP, but the basic idea is that we'd like to be able to incrementally assemble the atmos model (see #2100). We'll need to change the internal structure to do that without duplicating fields. This PR will allow us to restructure the `AtmosModel` to have `atmos.physics.param_set` (still haven't made that change yet). We'll need to do the same with all other atmos properties, I figure this is the most wide-spread one, though.

kpamnany (approved):
With the caveat that many getter functions will make the interface more wall-like, this does address the problem you're trying to solve.

Issue #1648, opened 2020-10-21
Make aux state BC-aware


charleskawczynski:
### Description

Right now, recover_thermo_state is an optimization that potentially leads to thermodynamic inconsistencies at the boundaries for PhaseEquil models. Concretely, the update_auxiliary_state! is unaware of how state is updated in boundary_state!. This is a more general issue, and pertains to code outside of thermodynamics. I think the right way to handle this to ensure that all RHS evaluations are in sync. This can be done, for example, by updating the BC non aux variables, then call update_auxiliary_state! with the BC-aware state.

charleskawczynski:
Once this is fixed, we should add the recover thermo state implementations back in. I've recorded what we had here:
```julia
function recover_thermo_state(
    atmos::AtmosModel,
    moist::DryModel,
    state::Vars,
    aux::Vars,
)
    e_int = internal_energy(atmos, state, aux)
    return PhaseDry(atmos.param_set, e_int, state.ρ)
end

function recover_thermo_state(
    atmos::AtmosModel,
    moist::EquilMoist,
    state::Vars,
    aux::Vars,
)
    e_int = internal_energy(atmos, state, aux)
    return PhaseEquil{eltype(state), typeof(atmos.param_set)}(
        atmos.param_set,
        e_int,
        state.ρ,
        state.moisture.ρq_tot / state.ρ,
        aux.moisture.temperature,
    )
end

function recover_thermo_state(
    atmos::AtmosModel,
    moist::NonEquilMoist,
    state::Vars,
    aux::Vars,
)
    e_int = internal_energy(atmos, state, aux)
    q = PhasePartition(
        state.moisture.ρq_tot / state.ρ,
        state.moisture.ρq_liq / state.ρ,
        state.moisture.ρq_ice / state.ρ,
    )

    return PhaseNonEquil{eltype(state), typeof(atmos.param_set)}(
        atmos.param_set,
        e_int,
        state.ρ,
        q,
    )
end
```
Back into `src/Atmos/Model/thermo_states.jl` and 

```julia
function recover_thermo_state_up(
    m::AtmosModel,
    moist::Union{DryModel, EquilMoist},
    state::Vars,
    aux::Vars,
    ts::ThermodynamicState = recover_thermo_state(m, state, aux),
)
    N_up = n_updrafts(m.turbconv)
    ts_up = vuntuple(N_up) do i
        recover_thermo_state_up_i(m, state, aux, i, ts)
    end
[...]

PR #2107, opened 2021-03-12
Update FV reconstruction with anelastic1D capabilities

ilopezgp:
### Description

This PR updates `prognostic_to_primitive!` and `primitive_to_prognostic!` to be consistent with the Anelastic1D and Compressible formulations. Inner thermo state constructors are now handled through dispatchers when possible. Recovery tests are added for the Anelastic1D() `prognostic_to_primitive!` and `primitive_to_prognostic!` constructors.

### DIFF
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -227,8 +227,18 @@ struct Compressible <: Compressibilty end
 
 Dispatch on Anelastic1D model
 
- - Density is constant in time
- - Remove momentum z-component tendencies
+ - The state density is taken constant in time and equal to the reference density. This
+    constant density profile is used in all equations and conversions from conservative to specific
+    variables per unit mass. The density can be accessed using the dispatch function
+    `density(atmos, state, aux)`.
+ - The thermodynamic state is constructed from the reference pressure (constant in time),
+    and the internal energy (which evolves in time).
+ - The state density is not consistent with the thermodynamic state, since we neglect 
+    buoyancy perturbations on all equations except in the vertical buoyancy flux.
+ - The density obtained from the thermodynamic state, `air_density(ts)`, recovers the full density, which
+    should only be used to compute buoyancy and buoyancy fluxes, and in the FV reconstruction.
+ - Removes momentum z-component tendencies, assuming balance between the pressure gradient and buoyancy
+    forces.
 """
 struct Anelastic1D <: Compressibilty end
 
@@ -479,18 +489,27 @@ turbulence_tensors(atmos::AtmosModel, args...) =
 """
     density(atmos::AtmosModel, state::Vars, aux::Vars)
 
-In the Anelastic1D state, `state.ρ` that is used to
-extract intrinsic values (i.e. `e=state.energy.ρe/state.ρ`)
-is time invariant (by eliminating the tendencies) while
-`ref_state.ρ` is used to construct a thermodynamic state.
-`density` will get the `ref_state.ρ` for thermodynamic state
-when the model is Anelastic1D.
+Density used in the conservative form of the prognostic equations.
+In the Compressible case, it is equal to the prognostic density,
+whereas in the Anelastic1D case it is the reference density,
+which is constant in time.
 """
 density(atmos::AtmosModel, state::Vars, aux::Vars) =
     density(atmos.compressibility, state, aux)
 density(::Compressible, state, aux) = state.ρ
 density(::Anelastic1D, state, aux) = aux.ref_state.ρ
 
+"""
+    pressure(atmos::AtmosModel, ts, aux::Vars)
+
+Diagnostic pressure consistent with the given thermodynamic state ts.
+In the Anelastic1D case it is the reference pressure,
+which is constant in time.
+"""
+pressure(atmos::AtmosModel, ts, aux::Vars) =
+    pressure(atmos.compressibility, ts, aux)
+pressure(::Compressible, ts, aux) = air_pressure(ts)
+pressure(::Anelastic1D, ts, aux) = aux.ref_state.p
 
 include("declare_prognostic_vars.jl") # declare prognostic variables
 include("multiphysics_types.jl")      # types for multi-physics tendencies
--- a/src/Atmos/Model/prog_prim_conversion.jl
+++ b/src/Atmos/Model/prog_prim_conversion.jl
@@ -23,18 +23,7 @@ function prognostic_to_primitive!(
     aux,
 )
     atmos.energy isa EnergyModel || error("EnergyModel only supported")
-    prognostic_to_primitive!(
-        atmos,
-        atmos.moisture,
-        prim,
-        prog,
-        Thermodynamics.internal_energy(
-            prog.ρ,
-            prog.energy.ρe,
-            prog.ρu,
-            gravitational_potential(atmos.orientation, aux),
-        ),
-    )
+    prognostic_to_primitive!(atmos, atmos.moisture, prim, prog, aux)
     prognostic_to_primitive!(atmos.turbconv, atmos, atmos.moisture, prim, prog)
 end
 
@@ -54,13 +43,7 @@ function primitive_to_prognostic!(
     prim::Vars,
     aux,
 )
-    primitive_to_prognostic!(
-        atmos,
-        atmos.moisture,
-        prog,
-        prim,
-        gravitational_potential(atmos.orientation, aux),
-    )
+    primitive_to_prognostic!(atmos, atmos.moisture, prog, prim, aux)
     primitive_to_prognostic!(atmos.turbconv, atmos, atmos.moisture, prog, prim)
 end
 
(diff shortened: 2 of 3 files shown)


src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -74,11 +76,12 @@ function prognostic_to_primitive!(
...
     prog::Vars,
     e_int::AbstractFloat,
+    aux::Vars,
 )
-    ts = PhaseDry(parameter_set(atmos), e_int, prog.ρ)
-    prim.ρ = prog.ρ
-    prim.u = prog.ρu ./ prog.ρ
-    prim.p = air_pressure(ts)
+    ts = recover_thermo_state(atmos, prog, aux)
+    prim.ρ = air_density(ts) # Needed for recovery of energy
```
charleskawczynski, line 82: If this is called on the boundary (@Zhengyu-Huang can you confirm?), I would recommend calling `new_thermo_state` instead (it's equivalent same at the moment).

src/Atmos/Model/AtmosModel.jl
```diff
@@ -489,6 +489,14 @@ density(atmos::AtmosModel, state::Vars, aux::Vars) =
 density(::Compressible, state, aux) = state.ρ
 density(::Anelastic1D, state, aux) = aux.ref_state.ρ
 
+pressure(atmos::AtmosModel, state::Vars, aux::Vars) =
+    pressure(atmos.compressibility, atmos, state, aux)
+
+function pressure(::Compressible, atmos, state, aux)
+    ts = recover_thermo_state(atmos, state, aux)
```
charleskawczynski, line 496: Based on the way that this is used, it might be better to pass `ts` as an argument.

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -74,11 +76,12 @@ function prognostic_to_primitive!(
...
     e_int::AbstractFloat,
+    aux::Vars,
 )
-    ts = PhaseDry(parameter_set(atmos), e_int, prog.ρ)
-    prim.ρ = prog.ρ
-    prim.u = prog.ρu ./ prog.ρ
-    prim.p = air_pressure(ts)
+    ts = new_thermo_state(atmos, prog, aux)
+    prim.ρ = air_density(ts) # Needed for recovery of energy
+    prim.u = prog.ρu ./ density(atmos, prog, aux)
```
charleskawczynski, line 83: What are the two different densities?

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -134,14 +139,15 @@ function primitive_to_prognostic!(
...
 )
     atmos.energy isa EnergyModel || error("EnergyModel only supported")
     ts = PhaseDry_ρp(parameter_set(atmos), prim.ρ, prim.p)
     e_kin = prim.u' * prim.u / 2
 
-    prog.ρ = prim.ρ
-    prog.ρu = prim.ρ .* prim.u
-    prog.energy.ρe = prim.ρ * total_energy(e_kin, e_pot, ts)
+    prog.ρ = density(atmos, prim, aux)
+    prog.ρu = density(atmos, prim, aux) .* prim.u
```
yairchn, line 149: maybe use 'prog.ρ' in 149 instead of re-calling 'density' ?

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -134,14 +139,15 @@ function primitive_to_prognostic!(
...
+    prog.energy.ρe = density(atmos, prim, aux) * total_energy(e_kin, e_pot, ts)
```
yairchn, line 150: same here maybe use 'prog.ρ' in 150 instead of re-calling 'density' ?

src/Atmos/Model/prog_prim_conversion.jl
ilopezgp: In compressible they are the same. In anelastic, air_density(ts) returns the density from the thermodynamic state constructed by (p_ref, prog.e_int), which is not prog.ρ = aux.ref_state.ρ, but the thermodynamically consistent density.

We need this density to recover the thermo state from the primitive variables.

src/Atmos/Model/AtmosModel.jl
ilopezgp: the anelastic1D version does not use `ts`

ilopezgp:
bors try

src/Atmos/Model/AtmosModel.jl
charleskawczynski: Fair, but we typically know what `ts` is anyway. Right now, `recover_thermo_state` performs SA, so my suggestion is more of a performance related

test/Atmos/prog_prim_conversion/runtests.jl
```diff
@@ -84,79 +78,109 @@ function assign!(state, bl, moisture::NonEquilMoist, nt, ::Primitive)
...
-    bl = TestBL(param_set, DryModel(), EnergyModel())
-    vs_prog = vars_state(bl, Prognostic(), FT)
-    vs_prim = vars_state(bl, Primitive(), FT)
-    prog_arr = zeros(varsize(vs_prog))
-    prim_arr = zeros(varsize(vs_prim))
-    prog = Vars{vs_prog}(prog_arr)
-    prim = Vars{vs_prim}(prim_arr)
-    for nt in PhaseDryProfiles(param_set, ArrayType)
-        @unpack e_int, e_pot = nt
+    compressibility = Tuple((Anelastic1D(), Compressible()))
```
charleskawczynski, line 89:
```suggestion
    compressibility = (Anelastic1D(), Compressible())
```

ilopezgp:
bors try

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -74,11 +76,12 @@ function prognostic_to_primitive!(
     prim::Vars,
     prog::Vars,
     e_int::AbstractFloat,
```
charleskawczynski, line 78: Can/should we remove this from the arg list?

ilopezgp: Happy you mentioned this! I have now moved this low-level functions to be computed within the functions for each type, in terms of the aux vars. This is a tad more verbose, but avoids possible inputs that are incoherent with respect to the auxiliary state, since this function is accessible outside here. As an example, we were actually using the functions `(bl, bl.moisture, ...) ` in the test. I have made it now so that the test calls the high-level dispatcher, as we should do everywhere in the model.

src/Atmos/Model/prog_prim_conversion.jl
```diff
@@ -73,33 +56,25 @@ function prognostic_to_primitive!(
...
     prog::Vars,
-    e_int::AbstractFloat,
+    aux::Vars,
 )
-    ts = PhaseDry(parameter_set(atmos), e_int, prog.ρ)
-    prim.ρ = prog.ρ
-    prim.u = prog.ρu ./ prog.ρ
-    prim.p = air_pressure(ts)
+    ts = new_thermo_state(atmos, prog, aux)
+    prim.ρ = air_density(ts) # Needed for recovery of energy, not prog.ρ in anelastic1d
```
yairchn, line 62: In the anelastic case (where we initiate state.ρ with aux_ref_state.ρ and state.ρ has no tendencies) do you expect to have any difference between state.ρ and aux_ref_state.ρ?

ilopezgp: The initialization state.ρ = aux_ref_state.ρ is not hardcoded for now, and so could be problematic, see issue #2111 . If it is initialized, I do not expect that there would be a change, but choosing the ref is safer since that one is not updated. We can later on ensure that the the state.ρ stays constant.

Zhengyu-Huang: I think there might be a issue that your density is not the reference density but you use reference density to compute the flux

yairchn (approved):

src/Atmos/Model/prog_prim_conversion.jl
charleskawczynski: Maybe we should clearly document what the two different densities are, in the `Compressibility` docs. I think it's clear from the code _how_/_where_ things are computed, but the _why_ is less clear to me. Specifically, what equations are these two different equations satisfying?

src/Atmos/Model/prog_prim_conversion.jl
charleskawczynski: I worry a bit about using the `aux` state here (#1648), but we should be okay if we only use `new_thermo_state`

charleskawczynski (commented):
I'm fine with merging once [this comment](https://github.com/CliMA/ClimateMachine.jl/pull/2107#discussion_r594551241) is addressed, and the remaining discussions are resolved. Overall, these changes look like a net +

src/Atmos/Model/prog_prim_conversion.jl
ilopezgp: @Zhengyu-Huang can you define what you mean by _your density_ in terms of variables in the code? I do not understand what you mean. Please clarify the possible problem so that we can resolve it.

src/Atmos/Model/prog_prim_conversion.jl
ilopezgp: Done!

ilopezgp:
bors try

ilopezgp:
I think this is ready!

ilopezgp:
bors try

ilopezgp:
bors try

ilopezgp:
bors try

PR #2109, opened 2021-03-12
Remove prognostic variable type parameter from tendency definition

charleskawczynski:
### Description

After looking into refactoring the BCs, I realized that there are some limitations with applying the same pattern used in the tendency specification. In addition, these limitations can be overcome by removing the prognostic variable type from the tendency definitions (which is similar to how they used to be defined).

This PR looks big, but all it really does is removes the prognostic variable from the tendency definition. In the context of the physics kernels, it changes the `flux` (and `source`) arguments from
```julia
flux(::DYCOMSRadiation{Energy}, atmos, args)
```
to
```julia
flux(::Energy, ::DYCOMSRadiation, atmos, args)
```

This PR also adds [DispatchedTuples.jl](https://github.com/charleskawczynski/DispatchedTuples.jl) as a dependency to handle coupling the tendency definitions and prognostic variables in a more elegant way than how it's currently working. Specifically, these changes are:

Master:
```julia
# Diagonalize sources: only add sources that correspond to correct equation
diag_source(pv::PV, m::AtmosModel, s::TendencyDef{Source, PV}) where {PV} = s
diag_source(pv, m, s) = nothing

# Filter sources / empty elements
filter_sources(t::Tuple) = filter(x -> !(x == nothing), t)
filter_sources(pv::PrognosticVariable, m, srcs) =
    filter_sources(map(s -> diag_source(pv, m, s), srcs))

# Entry point
eq_tends(pv::PrognosticVariable, m::AtmosModel, tt::Source) =
    (filter_sources(pv, m, m.source)..., eq_tends(pv, m.turbconv, tt)...)
```

This PR:
```julia

eq_tends(pv::PrognosticVariable, m::AtmosModel, tt::Source) =
    (dispatch(m.source, pv)..., eq_tends(pv, m.turbconv, tt)...)
```

Here, `m.source` is now a `DispatchedTuple`, which, when called with the prognostic variable, picks the correct index. The dispatched tuple is constructed at the driver level via

```julia
function dispatched_tuple(driver_sources::Tuple)
    tup = map(driver_sources) do t
        map(prognostic_vars(t)) do pv
            (pv, t)
        end
    end
    tup = tuple_of_tuples(tup) # Flattens "tuple of tuple of tuples" to "tuple of tuples"
    return DispatchedTuple(tup)
end
```

This same pattern can be used for BCs.

Also, I'm not sold on the name `dispatched_tuple`, open to bikeshedding here


charleskawczynski:
bors try

charleskawczynski:
bors try-

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
cc @simonbyrne

kpamnany (approved):
Simplifies things; always good to see. Definitely need a better name than `dispatched_tuple`; maybe something like `prognosticvar_source_map`?

PR #2115, opened 2021-03-15
Synchronize maxiter and temperature tolerance between EquilMoist and PhaseEquil

### LINKED ISSUE #2108: Inconsistent maxiter values in thermo dispatchers and inner constructors
### Description

The dispatcher of Atmos thermodynamic states (`new_thermo_state`) uses moist thermodynamic constructors with maxiter and tolerances given by the `moisture` model. However, the inner constructors may have different maxiter and tolerances. This can lead to accuracy inconsistencies throughout the model.

Example:

In `states.jl` the maxiter is 8,
```
function PhaseEquil(
    param_set::APS,
    e_int::FT,
    ρ::FT,
    q_tot::FT,
    maxiter::Int = 8,
    temperature_tol::FT = FT(1e-1),
    ::Type{sat_adjust_method} = NewtonsMethod,
) where {FT <: Real, sat_adjust_method}
    ...
    return PhaseEquil{FT, typeof(param_set)}(param_set, e_int, ρ, q_tot_safe, T)
end
```

But in the dispatcher in `thermo_states.jl`,

```
function new_thermo_state(
    atmos::AtmosModel,
    energy::EnergyModel,
    moist::EquilMoist,
    state::Vars,
    aux::Vars,
)
    e_int = internal_energy(atmos, state, aux)
    param_set = parameter_set(atmos)
    return PhaseEquil(
        param_set,
        e_int,
        state.ρ,
        state.moisture.ρq_tot / state.ρ,
        moist.maxiter,
        moist.tolerance,
    )
end
```

where moist.maxiter has a default value of 3:

```
struct EquilMoist{FT} <: MoistureModel
    maxiter::Int
    tolerance::FT
end
EquilMoist{FT}(;
    maxiter::IT = 3,
    tolerance::FT = FT(1e-1),
) where {FT <: AbstractFloat, IT <: Int} = EquilMoist{FT}(maxiter, tolerance)
```

charleskawczynski:
### Description

Closes #2108.

This PR technically changes the behavior of the drivers, so I'm wondering if we should adjust the drivers so that this PR is a no-op.

### DIFF
--- a/src/Common/Thermodynamics/states.jl
+++ b/src/Common/Thermodynamics/states.jl
@@ -68,6 +68,9 @@ PhasePartition(q_tot::FT, q_liq::FT) where {FT <: Real} =
 PhasePartition(q_tot::FT) where {FT <: Real} =
     PhasePartition(q_tot, zero(FT), zero(FT))
 
+const ITERTYPE = Union{Int, Nothing}
+TOLTYPE(::Type{FT}) where {FT <: AbstractFloat} = Union{FT, Nothing}
+
 #####
 ##### Dry states
 #####
@@ -236,10 +239,12 @@ function PhaseEquil(
     e_int::FT,
     ρ::FT,
     q_tot::FT,
-    maxiter::Int = 8,
-    temperature_tol::FT = FT(1e-1),
+    maxiter::IT = nothing,
+    temperature_tol::FTT = nothing,
     ::Type{sat_adjust_method} = NewtonsMethod,
-) where {FT <: Real, sat_adjust_method}
+) where {FT <: Real, sat_adjust_method, IT <: ITERTYPE, FTT <: TOLTYPE(FT)}
+    maxiter === nothing && (maxiter = 8)
+    temperature_tol === nothing && (temperature_tol = FT(1e-1))
     phase_type = PhaseEquil
     q_tot_safe = clamp(q_tot, FT(0), FT(1))
     T = saturation_adjustment(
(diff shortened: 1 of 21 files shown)


src/Common/Thermodynamics/states.jl
```diff
@@ -296,9 +301,11 @@ function PhaseEquil_ρθq(
     ρ::FT,
     θ_liq_ice::FT,
     q_tot::FT,
-    maxiter::Int = 36,
-    temperature_tol::FT = FT(1e-1),
-) where {FT <: Real}
+    maxiter::IT = nothing,
+    temperature_tol::FTT = nothing,
+) where {FT <: Real, IT <: ITERTYPE, FTT <: TOLTYPE(FT)}
+    maxiter === nothing && (maxiter = 36)
```
ilopezgp, line 307: Should we bump 36 to 40? Just to have the minimum amount of numbers (PhaseEquil_peq uses 40).

charleskawczynski: We can, I forget exactly why the number is so large considering this is now used at runtime with the Anelastic1D model. I imagine that the ref profiles were not converging without such a large number.

ilopezgp: Yes, I know for a fact the `_peq` constructor is very slow and for some of the tests cases it fails to converge.I looked into this with @trontrytel but we could not find a reason. I think a large number for this constructor is fine since we use it only in the anelastic setup, and that one is not operational.

ilopezgp: That said, this applies to `_peq`, I don't know what's the situation with `_ρθq`, which I think is used for the Compressible, theta model.

kpamnany (approved):
Not sure what you mean by changing the behavior of the drivers? Looks like a NOP to me?

charleskawczynski:
> Not sure what you mean by changing the behavior of the drivers? Looks like a NOP to me?

It's not a NOP at the moment because any drivers that don't specify `maxiter`/`tolerance` in `EquilMoist`, currently fall back to 
```julia
    maxiter::IT = 3,
    tolerance::FT = FT(1e-1),
```
But with this PR, they change to `nothing`, which then falls back on the new default in Thermodynamics.jl:
```julia
    maxiter = 8
    temperature_tol = FT(1e-1)
```
Hm.. Actually, technically, it _is_ a NOP (for all "working drivers") based on how `maxiter`/`temperature_tol` works (the non-linear solver exit condition depends on the tolerance _only_). So, maybe we're in the clear, then.

PR #2121, opened 2021-03-16
Do not create example blocks on executed code

charleskawczynski:
### Description

We were running tutorials in parallel via `Literate.notebook()`, and running them all again in serial in `makedocs`. This PR removes the `@example` blocks from the tutorial markdown so that they're only executed in parallel in `Literate.notebook()`. One side-effect of this is that suppressing output is more important because truncation seems to work differently when executing markdown in Literate compared to `makedocs`, so some output in the tutorials have also been suppressed (which we probably should have done anyway). In addition, the resolution on some of the animations have been reduced.

### DIFF
--- a/docs/list_of_tutorials.jl
+++ b/docs/list_of_tutorials.jl
@@ -101,14 +101,15 @@ if generate_tutorials
     println("Building literate tutorials...")
 
     @everywhere function generate_tutorial(tutorials_dir, tutorial)
-        gen_dir =
-            joinpath(GENERATED_DIR, relpath(dirname(tutorial), tutorials_dir))
-        input = abspath(tutorial)
-        script = Literate.script(input, gen_dir)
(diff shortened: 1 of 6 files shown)


charleskawczynski:
bors try

jakebolewski:
It looks like something is blowing up building the search index :expressionless: https://buildkite.com/clima/climatemachine-docs/builds/1087#80424eed-59e2-41c0-8ad8-7bb1f1f4a928/6

I think the fail to deploy error doesn't get bubbled up correctly, I'll look to see if there is some option we are not setting.

jakebolewski:
Ah it is trying to build a search index across all open PR's, not sure if that means it should be > 150 mb but might be the cause, especially looking at the number of assets in the search index. 

 The build times are actually in-parallel and fast again though!

PR #2065, opened 2021-02-26
Refactor BCs - iteration 2

charleskawczynski:
### Description

Building on-top of #2056, this PR further refactors the BCs.

- [ ] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [ ] Unit tests are included OR N/A.
- [ ] Code is exercised in an integration test OR N/A.
- [ ] Documentation has been added/updated OR N/A.

PR #2056, opened 2021-02-23
Refactor BCs - Add `BCDef` and `bc_val` to BalanceLaws

charleskawczynski:
### Description

This is the first PR of many to refactor the boundary conditions. This PR:

 - Changes the Atmos BC interface to accept tuples of `BCDef` types, so that we can loop over the BCs / prognostic variables
 - Individual `bc_val` function kernels will (eventually) be DG-agnostic
 - Add a BC table (similar to the tendency table)

It'll be much easier to do this incrementally, and this PR takes a small step and handles:

 - `boundary_state!` for momentum
 - some shared infrastructure in BalanceLaws

Similar to the tendency layer specification, this PR (and those more to come) will add types and hooks to add a layer on-top of the existing BC interface.

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #2122, opened 2021-03-17
Refactor BCs

charleskawczynski:
### Description

This PR:

 - Changes the Atmos BC interface to accept tuples of `BCDef` types, so that we can loop over the BCs / prognostic variables
 - Individual `bc_val` function kernels will (eventually) be DG-agnostic
 - Add a BC table (similar to the tendency table)

It'll be much easier to do this incrementally, and this PR takes a small step and handles:

 - `boundary_state!` for momentum
 - some shared infrastructure in BalanceLaws

Similar to the tendency layer specification, this PR (and those more to come) will add types and hooks to add a layer on-top of the existing BC interface.

Supercedes #2056, #2065.


charleskawczynski:
I don't think this will work for all drivers yet but this works for the bomex EDMF experiment-- we may want to add `physics` to `AtmosBC` in all the drivers in a separate PR so that this PR impacts fewer files.

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

charleskawczynski:
bors try

PR #2131, opened 2021-03-19
Cut 3D bickley jet test runtime in half

blallen:
### Description

See title. Previous tests took 15 + 10 minutes to run, which is a long time when you want to make sure you didn't break anything locally. Now takes 7.5 + 5 minutes, which is much more acceptable (generally only need to wait for one test offline). Will also speed up things when running with bors and slurm, so double benefits.

No code was changed, just the time for the test and the ref values.


blallen:
bors try

jakebolewski (approved):

PR #2133, opened 2021-03-19
Fix EDMF covariance dissipations

ilopezgp:
### Description

This PR

- Fixes a bug in the definition of the covariance dissipations in the EDMF kernels, which were previously multiplied by density and area fraction twice. This has likely not affected our simulations, since we are still working with dry models mostly.

- Eliminates the use of enforce_positivity in the computation of TKE dissipation, and on `compute_gradient_argument!`. Enforcing positivity here does more harm than good. In the case of dissipation, not enforcing positivity on the argument makes the dissipation positive, which acts to regularize the problem. In the case of flux computation, enforcing bounds artificially weakens flux divergences near regions of negative covariance, which also increases discontinuities and noise. We should enforce bounds for the covariances (if needed) in a single place, preferably through a source/sink?

I also cleaned up some functions that have unused arguments.


ilopezgp:
bors try

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -681,56 +680,50 @@ function source(::en_ρatke, ::BuoySource, atmos, args)
...
 end
 
 function source(::en_ρaθ_liq_cv, ::DissSource, atmos, args)
-    @unpack env, K_h, Diss₀ = args.precomputed.turbconv
-    gm = args.state
+    @unpack Diss₀ = args.precomputed.turbconv
     en = args.state.turbconv.environment
-    ρa₀ = gm.ρ * env.a
-    return -ρa₀ * Diss₀ * en.ρaθ_liq_cv
+    return -Diss₀ * en.ρaθ_liq_cv
```
yairchn, line 694: should we use enforce_positivity here as well? 
'' θ_liq_cv  =  enforce_positivity(en.ρaθ_liq_cv) / gm.ρ / env.a ''

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -681,56 +680,50 @@ function source(::en_ρatke, ::BuoySource, atmos, args)
...
 end
 
 function source(::en_ρaq_tot_cv, ::DissSource, atmos, args)
-    @unpack env, K_h, Diss₀ = args.precomputed.turbconv
-    gm = args.state
+    @unpack Diss₀ = args.precomputed.turbconv
     en = args.state.turbconv.environment
-    ρa₀ = gm.ρ * env.a
-    return -ρa₀ * Diss₀ * en.ρaq_tot_cv
+    return -Diss₀ * en.ρaq_tot_cv
```
yairchn, line 700: same here

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -681,56 +680,50 @@ function source(::en_ρatke, ::BuoySource, atmos, args)
...
 end
 
 function source(::en_ρaθ_liq_q_tot_cv, ::DissSource, atmos, args)
-    @unpack env, K_h, Diss₀ = args.precomputed.turbconv
-    gm = args.state
+    @unpack Diss₀ = args.precomputed.turbconv
     en = args.state.turbconv.environment
-    ρa₀ = gm.ρ * env.a
-    return -ρa₀ * Diss₀ * en.ρaθ_liq_q_tot_cv
+    return -Diss₀ * en.ρaθ_liq_q_tot_cv
```
yairchn, line 706: we could limit θ_liq_q_tot_cv between +/- sort(θ_liq_cv*q_tot_cv) as well

test/Atmos/EDMF/edmf_kernels.jl
ilopezgp: I don't think we should do it here. This is the dissipation function. If `θ_liq_cv` hasn't been clipped before, having a _positive_ dissipation actually helps bring it to zero.

test/Atmos/EDMF/edmf_kernels.jl
ilopezgp: See comment above.

test/Atmos/EDMF/edmf_kernels.jl
ilopezgp: I don't think we should clip it here. If it is above those variables, it means we did not clip when we should have beforehand.

charleskawczynski (approved):
I'm fine with this, but I haven't looked closely at the equations. Can @yairchn confirm this looks okay?

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -681,56 +680,50 @@ function source(::en_ρatke, ::BuoySource, atmos, args)
...
 
 function source(::en_ρatke, ::DissSource, atmos, args)
-    @unpack env, l_mix, Diss₀ = args.precomputed.turbconv
+    @unpack env, Diss₀ = args.precomputed.turbconv
     gm = args.state
     en = args.state.turbconv.environment
     ρa₀ = gm.ρ * env.a
-    tke_en = enforce_positivity(en.ρatke) / gm.ρ / env.a
+    tke_en = enforce_positivity(en.ρatke) / ρa₀
     return -ρa₀ * Diss₀ * tke_en  # tke Dissipation
```
yairchn, line 688: in that case (i.e. below) this 
-ρa₀ * Diss₀ * tke_en
should be 
-en.ρatke* Diss₀ 
as well

yairchn:
> I'm fine with this, but I haven't looked closely at the equations. Can @yairchn confirm this looks okay?

Yes the changes to the equations are correct  - bottom line from all the changes is that a term like this in the previous version:
-ρa₀ * Diss₀ * en.ρaθ_liq_cv
has a₀ twice (i.e. a₀*a₀) once in 'en.ρaθ_liq_cv' and once in 'ρa₀'.

test/Atmos/EDMF/edmf_kernels.jl
ilopezgp: Yes, I will change this. It is good to have rectifying damping. Thanks!

ilopezgp:
@yairchn I made the change to the TKE dissipation that you proposed. I also eliminated the enforce_positivity bound in the `compute_gradient_argument!`. 

Enforcing bounds here is inconsistent, because it will avoid diffusive homogenization of the covariance fluxes when the covariance turns negative! This might have exacerbated the problems you observed of noise amplification when the covariances go negative.

ilopezgp:
bors try

ilopezgp:
bors try

ilopezgp:
It should be fine, the CI fail was unrelated to EDMF.

PR #2136, opened 2021-03-22
Add tendency specification layer to land model

charleskawczynski:
### Description

This PR adds the tendency specification layer to the land model. This is basically how the atmosphere does it.

### DIFF
--- a/src/Land/Model/LandModel.jl
+++ b/src/Land/Model/LandModel.jl
@@ -1,6 +1,8 @@
 module Land
 
 using DocStringExtensions
+using UnPack
+using DispatchedTuples
 using LinearAlgebra, StaticArrays
 
 using CLIMAParameters
@@ -12,6 +14,11 @@ using ..MPIStateArrays
 using ..BalanceLaws
 import ..BalanceLaws:
     BalanceLaw,
+    prognostic_vars,
+    flux,
+    source,
+    eq_tends,
+    precompute,
     vars_state,
     flux_first_order!,
     flux_second_order!,
@@ -47,7 +54,7 @@ Users may over-ride prescribed default values for each field.
 # Fields
 $(DocStringExtensions.FIELDS)
 """
-struct LandModel{PS, S, LBC, SRC, IS} <: BalanceLaw
+struct LandModel{PS, S, LBC, SRC, SRCDT, IS} <: BalanceLaw
     "Parameter set"
     param_set::PS
     "Soil model"
@@ -56,6 +63,8 @@ struct LandModel{PS, S, LBC, SRC, IS} <: BalanceLaw
     boundary_conditions::LBC
     "Source Terms (Problem specific source terms)"
     source::SRC
+    "DispatchedTuple of sources"
+    source_dt::SRCDT
     "Initial Condition (Function to assign initial values of state variables)"
     init_state_prognostic::IS
 end
@@ -81,7 +90,15 @@ function LandModel(
     init_state_prognostic::IS = nothing,
 ) where {SRC, IS, LBC}
     @assert init_state_prognostic ≠ nothing
-    land = (param_set, soil, boundary_conditions, source, init_state_prognostic)
+    source_dt = prognostic_var_source_map(source)
+    land = (
+        param_set,
+        soil,
+        boundary_conditions,
+        source,
+        source_dt,
+        init_state_prognostic,
+    )
     return LandModel{typeof.(land)...}(land...)
 end
 
@@ -173,16 +190,16 @@ function flux_second_order!(
     aux::Vars,
     t::Real,
 )
-    flux_second_order!(
-        land,
-        land.soil,
-        flux,
-        state,
-        diffusive,
-        hyperdiffusive,
-        aux,
-        t,
-    )
+    tend = Flux{SecondOrder}()
+    _args = (; state, aux, t, diffusive, hyperdiffusive)
+    args = merge(_args, (precomputed = precompute(land, _args, tend),))
+
+    map(prognostic_vars(land)) do prog
+        var, name = get_prog_state(flux, prog)
+        val = Σfluxes(prog, eq_tends(prog, land, tend), land, args)
+        setproperty!(var, name, val)
+    end
+    nothing
 
 end
 
@@ -205,7 +222,16 @@ function source!(
     t::Real,
     direction,
 )
-    land_source!(land.source, land, source, state, diffusive, aux, t, direction)
+    tend = Source()
+    _args = (; state, aux, t, direction, diffusive)
+    args = merge(_args, (precomputed = precompute(land, _args, tend),))
+
+    map(prognostic_vars(land)) do prog
+        var, name = get_prog_state(source, prog)
+        val = Σsources(prog, eq_tends(prog, land, tend), land, args)
+        setproperty!(var, name, val)
+    end
+    nothing
 end
 
 
@@ -220,6 +246,8 @@ function init_state_prognostic!(
     land.init_state_prognostic(land, state, aux, coords, t, args...)
 end
 
+include("prog_types.jl")
+
 include("RadiativeEnergyFlux.jl")
 using .RadiativeEnergyFlux
 include("SoilWaterParameterizations.jl")
@@ -233,5 +261,10 @@ include("Runoff.jl")
 using .Runoff
 include("land_bc.jl")
 include("soil_bc.jl")
+
+include("prognostic_vars.jl")
+
 include("source.jl")
+include("land_tendencies.jl")
+
 end # Module
--- a/src/Land/Model/land_tendencies.jl
+++ b/src/Land/Model/land_tendencies.jl
@@ -1,0 +1,50 @@
+#####
+##### Sources
+#####
+
+eq_tends(pv::AbstractPrognosticVariable, m::LandModel, tt::Source) =
+    (m.source_dt[pv]...,)
+
+#####
+##### First order fluxes
+#####
+
+eq_tends(::PV, ::LandModel, ::Flux{FirstOrder}) where {PV} = ()
+
+
+#####
+##### Second order fluxes
+#####
+
+# Empty by default
+eq_tends(pv::PV, ::AbstractSoilComponentModel, ::Flux{SecondOrder}) where {PV} =
+    ()
+
+eq_tends(pv::PV, land::LandModel, tt::Flux{SecondOrder}) where {PV} =
+    (eq_tends(pv, land.soil.heat, tt)..., eq_tends(pv, land.soil.water, tt)...)
+
+eq_tends(
+    pv::PV,
+    land::LandModel,
+    tt::Flux{SecondOrder},
+) where {PV <: VolumetricInternalEnergy} =
+    (eq_tends(pv, land.soil.heat, land.soil.water, tt)...,)
+
+# TODO: move to soi_heat.jl?
+eq_tends(
+    ::VolumetricInternalEnergy,
+    ::SoilHeatModel,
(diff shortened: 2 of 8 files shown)


kmdeck:
exciting!!

src/Land/Model/source.jl
```diff
@@ -82,35 +69,21 @@ function land_source!(
...
-    source.soil.water.ϑ_l -= freeze_thaw / _ρliq
-    source.soil.water.θ_i += freeze_thaw / _ρice
+    return (; freeze_thaw)
 end
 
+function source(::ϑLiquid, s::PhaseChange, land::LandModel, args)
+    @unpack state = args
+    @unpack freeze_thaw = args.precomputed.dtup[s]
+    FT = eltype(state)
+    _ρliq = FT(ρ_cloud_liq(land.param_set))
```
jakebolewski, line 79: `p_cloud_liq()` looks an accessor  from the atmos model

src/Land/Model/source.jl
```diff
@@ -82,35 +69,21 @@ function land_source!(
...
-            t,
-            direction,
-        )
-        return nothing
-    end
+function source(::θIce, s::PhaseChange, land::LandModel, args)
+    @unpack state = args
+    @unpack freeze_thaw = args.precomputed.dtup[s]
+    FT = eltype(state)
+    _ρice = FT(ρ_cloud_ice(land.param_set))
```
jakebolewski, line 87: `p_cloud_liq`

jakebolewski (dismissed):

src/Land/Model/source.jl
charleskawczynski: It's used above in the file (in the now `precompute(source_type::PhaseChange, land::LandModel, args, tt::Source)`)

charleskawczynski: The same goes with `ρ_cloud_ice`

kmdeck: it is the density of liquid water or frozen water. maybe a candidate for a rename..

src/Land/Model/source.jl
kmdeck: this one is OK - we want rho_i d(\theta_i)/dt = - rho_l d(\theta_l)/dt . theta is proportional to the volume in the phase, so this makes it so freezing/thawing conserves the mass of water, while the volume changes by the right amount.

src/Land/Model/prog_types.jl
```diff
@@ -0,0 +1,7 @@
+#####
+##### Prognostic Variable types
+#####
+
+struct ϑLiquid <: AbstractPrognosticVariable end
+struct InternalEnergy <: AbstractPrognosticVariable end
```
kmdeck, line 6: Should we be careful that this is a volumetric internal energy, as opposed to a specific internal energy?

src/Land/Model/soil_heat.jl
```diff
@@ -194,22 +194,15 @@ function compute_gradient_flux!(
...
+
+function flux(::InternalEnergy, ::DiffWaterFlux, ::LandModel, args)
     param_set = parameter_set(land)
     ρe_int_l = volumetric_internal_energy_liq(aux.soil.heat.T, param_set)
-    diffusive_water_flux =
-        -ρe_int_l .* get_diffusive_water_term(soil.water, diffusive)
-    diffusive_heat_flux = -diffusive.soil.heat.κ∇T
-    flux.soil.heat.ρe_int += diffusive_heat_flux + diffusive_water_flux
-
+    return -ρe_int_l * diffusive.soil.water.K∇h
```
kmdeck, line 207: does this only evaluate when we have a SoilWaterModel? The variable is not defined for PrescribedWaterModel

src/Land/Model/soil_heat.jl
```diff
@@ -194,22 +194,15 @@ function compute_gradient_flux!(
...
-    t::Real,
-)
+struct DiffHeatFlux <: TendencyDef{Flux{SecondOrder}} end
+struct DiffWaterFlux <: TendencyDef{Flux{SecondOrder}} end
+
+function flux(::InternalEnergy, ::DiffHeatFlux, ::LandModel, args)
+    return -args.diffusive.soil.heat.κ∇T
+end
+
+function flux(::InternalEnergy, ::DiffWaterFlux, ::LandModel, args)
```
kmdeck, line 204: I think the arguments here should be:  (prog::InternalEnergy, tend::DiffusiveWaterFlux, land::LandModel, args)

how does the line immediately below this (205) know that land is the third argument?

src/Land/Model/soil_heat.jl
kmdeck: nvm, I think that is taken care of here:
```
eq_tends(::InternalEnergy, ::SoilWaterModel, ::Flux{SecondOrder}) =
    (WaterDiffusion(),)
    ```
    ?

File: src/Land/Model/prog_types.jl (hunk 3 above)
Comment (charleskawczynski) [reply src/Land/Model/prog_types.jl:6]: This is basically a mirror from
```julia
vars_state(heat::SoilHeatModel, st::Prognostic, FT) = @vars(ρe_int::FT)
```
and
```julia
vars_state(water::SoilWaterModel, st::Prognostic, FT) = @vars(ϑ_l::FT, θ_i::FT)
```
I'm not sure if there are better names out there, but that's the basic idea.

src/Land/Model/soil_heat.jl
charleskawczynski: Yes, this is handled by that and `prognostic_vars`:
```julia
prognostic_vars(water::PrescribedWaterModel) = ()
prognostic_vars(water::SoilWaterModel) = (ϑLiquid(), θIce())
prognostic_vars(heat::PrescribedTemperatureModel) = ()
prognostic_vars(heat::SoilHeatModel) = (InternalEnergy(),)

prognostic_vars(land::LandModel) =
    (prognostic_vars(land.soil.water)..., prognostic_vars(land.soil.heat)...)
```
So, if we don't have `SoilWaterModel` then

```julia
    map(prognostic_vars(land)) do prog
        var, name = get_prog_state(flux, prog)
        val = Σfluxes(prog, eq_tends(prog, land, tend), land, args)
        setproperty!(var, name, val)
    end
```
doesn't traverse the `SoilWaterModel` variables 🙂

src/Land/Model/soil_heat.jl
charleskawczynski: Yeah, we need `land::LandModel`. It looks like `tend` is not used, and `prog` is generally not used. I'll update to include the `land` argument.

charleskawczynski: Fixed... Hopefully that makes more sense

src/Land/Model/land_tendencies.jl
```diff
@@ -0,0 +1,32 @@
...
+
+eq_tends(pv::PV, land::LandModel, tt::Flux{SecondOrder}) where {PV} =
+    (eq_tends(pv, land.soil.heat, tt)..., eq_tends(pv, land.soil.water, tt)...)
+
+# TODO: move to soi_heat.jl?
+eq_tends(::InternalEnergy, ::SoilHeatModel, ::Flux{SecondOrder}) =
+    (DiffHeatFlux(), DiffWaterFlux())
+
+# TODO: move to soi_water.jl?
+eq_tends(::InternalEnergy, ::SoilWaterModel, ::Flux{SecondOrder}) =
```
kmdeck, line 31: Shouldnt be internalenergy here, but theta liquid.

src/Land/Model/prog_types.jl
kmdeck: I was thinking back to the discussion we had when we first added this variable. variables with name `e_` denote a specific energy per unit volume, `E_` denotes energy, and here we have `rhoe_`, denoting energy per unit volume. Since we took the care there to be precise, should this be named something like `VolumetricInternalEnergy` or `InternalEnergyDensity` instead?

src/Land/Model/land_tendencies.jl
kmdeck: (the coupled tutorial did not work without this, im surprised the water integration test do?)

src/Land/Model/soil_heat.jl
kmdeck: thanks! makes sense!

src/Land/Model/prog_types.jl
charleskawczynski: Either one seems to me

src/Land/Model/land_tendencies.jl
charleskawczynski: Yeah, I think that some of these methods need some tweaking. I'll try to take another look soon

src/Land/Model/soil_heat.jl
```diff
@@ -194,22 +194,17 @@ function compute_gradient_flux!(
...
-    soil::SoilModel,
-    heat::SoilHeatModel,
-    flux::Grad,
-    state::Vars,
-    diffusive::Vars,
-    hyperdiffusive::Vars,
-    aux::Vars,
-    t::Real,
-)
+struct DiffHeatFlux <: TendencyDef{Flux{SecondOrder}} end
```
kmdeck, line 197: I wonder if we might rename these.. These arent fluxes, but divergences of fluxes.  I have also seen people refer to the water driven part as heat advection...will look around the literature and see what names people use for this. I guess the first is `HeatConduction`.

charleskawczynski: I think the phenomenon in general is an acceptable way to look at it. I think `HeatConduction` is probably fine, too. FWIW, I think it's not too difficult to change these names after the fact. Ultimately there will be a function (e.g.,)
```julia
flux(::InternalEnergy, ::DiffHeatFlux, land::LandModel, args)
```
That will return the flux (the divergence of which is taken by the framework and added to all other tendencies).

src/Land/Model/LandModel.jl
```diff
@@ -205,7 +222,16 @@ function source!(
     t::Real,
     direction,
 )
-    land_source!(land.source, land, source, state, diffusive, aux, t, direction)
+    tend = Source()
+    _args = (; state, aux, t, direction, diffusive)
+    args = merge(_args, (precomputed = precompute(land, _args, tend),))
+
+    map(prognostic_vars(land)) do prog
```
kmdeck, line 229: FWIW, this is a little opaque (though the comments in the src file for `get_prog_state` help a lot)...
we have `prog` referring to e.g. an instance of type  of Prognostic variable, e.g. InternalEnergy(), 
`var` referring to `state`, `source`, or `flux`, etc.
`name` something else
as a physicist, I would say the variable is rhoe_int, and that would be it. so maybe this is some of the "under the hood" code leaking through?


In any case, I think this accumulates all sources for that variable and then sets source.soil.heat.rho_eint = val, and someone adding in a new source or flux term probably doesnt need to parse it.

kmdeck (commented):
I think this mainly looks good! It's a little abstract. Some minor things re: naming to address, and it would be good to see the tutorials and other tests pass, but then I think we can incorporate it? @jakebolewski do you see any issue why adding in the overland flow model to this would be difficult or cause issues?

src/Land/Model/LandModel.jl
charleskawczynski: Yeah, this is pretty generic code. You'll see that the Atmos implementation is exactly the same with `atmos` instead of `land`. This could eventually be moved to `BalanceLaws`.

> In any case, I think this accumulates all sources for that variable and then sets source.soil.heat.rho_eint = val, and someone adding in a new source or flux term probably doesnt need to parse it.

Correct.

kmdeck (approved):

PR #2138, opened 2021-03-23
Land/soil/add residual water

kmdeck:
### Description
This adds in the residual water content parameter. In our planned LSMs, we will set this to zero, but it is useful to have for comparisons to other benchmarks and experiments (which do not set it to zero). Because of this, we added it in as a soil model parameter, but the default is zero.

There was an issue with the formatter and this file: .dev/clima_formatter_default_image.jl. We turned the formatter off for that file.

### DIFF
--- a/.dev/clima_formatter_default_image.jl
+++ b/.dev/clima_formatter_default_image.jl
@@ -1,4 +1,5 @@
 #!/usr/bin/env julia
+#! format: off
 #
 # Called with no arguments will replace the default system image with one that
 # includes the JuliaFormatter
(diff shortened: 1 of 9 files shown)


kmdeck:
will compress commits when approved

gagnelandmanna (approved):
Hi Katherine! I read all the changes, also checked that the functions changed don't affect the SoilHeatParameterizations.jl or the soil_heat.jl files. Looks great :-)!

kmdeck:
bors try

PR #2159, opened 2021-04-01
Simplify Ekman BCs, set a_min=0

ilopezgp:
### Description

This PR simplifies the Ekman BCs, so that the BC of TKE equates the initial TKE, preventing sharp initial gradients. The new EDMF BCs are given by a new SurfaceModel type, as suggested by @charleskawczynski .

I also set `a_min=0` for all EDMF simulations since this has already been tested to work, and this is the physical limit. We might want to remove `a_min` soon when we transition to convective cases.

### DIFF
--- a/test/Atmos/EDMF/closures/surface_functions.jl
+++ b/test/Atmos/EDMF/closures/surface_functions.jl
@@ -40,7 +40,6 @@ function subdomain_surface_values(
     zLL::FT,
 ) where {FT}
 
-    turbconv = turbconv_model(atmos)
     N_up = n_updrafts(turbconv)
     gm = state
     # TODO: change to new_thermo_state
@@ -58,7 +57,7 @@ function subdomain_surface_values(
     oblength = turbconv.surface.obukhov_length
     ustar = turbconv.surface.ustar
 
-    unstable = oblength < 0
+    unstable = oblength < -eps(FT)
     fact = unstable ? (1 - surf.ψϕ_stab * zLL / oblength)^(-FT(2 // 3)) : 1
     tke_fact = unstable ? cbrt(zLL / oblength * zLL / oblength) : 0
     ustar² = ustar^2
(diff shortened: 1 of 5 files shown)


test/Atmos/EDMF/ekman_layer.jl
```diff
@@ -85,6 +85,48 @@ function init_state_prognostic!(
...
+        atmos::AtmosModel{FT},
+        state::Vars,
+        aux::Vars,
+        zLL::FT,
+    ) where {FT}
+
+Returns the surface value of EDMF variables for the
+Ekman layer simulation, overriding general method.
+"""
+function subdomain_surface_values(
```
charleskawczynski, line 99: Should we define this next to the existing one and just define and use a new surface model? Simply overriding the existing method seems risky (and will require more work to merge later when we want to move all of this to `src/`).

ilopezgp: Done!

test/Atmos/EDMF/closures/surface_functions.jl
```diff
@@ -58,7 +57,7 @@ function subdomain_surface_values(
     oblength = turbconv.surface.obukhov_length
     ustar = turbconv.surface.ustar
 
-    unstable = oblength < 0
+    unstable = oblength < -eps(FT)
```
yairchn, line 60: do we consider 'oblength=0' a stable case or do we have a neutral case as well ?

yairchn (approved):

test/Atmos/EDMF/closures/surface_functions.jl
ilopezgp: In reality, oblength-> 0 is not possible because the numerator is nonzero. We use oblength=0 for now because we only modify the EDMF scheme with oblength <0. We will have to pass this from the SurfaceFluxes.jl module in the end.

ilopezgp:
bors try

ilopezgp:
bors try

test/Atmos/EDMF/ekman_layer.jl
```diff
@@ -98,7 +98,13 @@ function main(::Type{FT}, cl_args) where {FT}
...
     # N_updrafts = 1
     # N_quad = 3
-    # turbconv = EDMF(FT, N_updrafts, N_quad, param_set)
+    # turbconv = EDMF(
+    #     FT,
+    #     N_updrafts,
+    #     N_quad,
+    #     param_set,
+    #     surface = NeutralDrySurfaceModel{FT}(param_set),
+    # )
```
charleskawczynski, line 107: Don't we want this uncommented?

ilopezgp: No, the driver should be grid-mean only, anelastic1d for now. When the coupled EDMF works we can split this into two separate tests.

ilopezgp: The new behavior has been tested locally and works

PR #2162, opened 2021-04-03
Geometric factors data structures

valeriabarra:
### Description

This ~~WIP~~ PR provides a clean-up of the interface to store and access geometric factors data at quadrature points. It adds two structures `VolumeGeometry` and `SurfaceGeometry` for easy access of the metric terms data in different modules. These structures are defined in an independent module, called `GeomData.jl` to avoid the issue of circular dependency between the `Metrics.jl` and the `Grids.jl` modules. 

Things left TODO:
- [x] Modify also the `computegeometry_fvm` function ~~?~~
- [x]  ~~Clean-up existing definitions of loose const variables in place of structs ?~~ Edit: This can be left for a subsequent PR since there is a more compelling Core revamp in the making
- [x] Update the `Base.getproperty` and `Base.propertynames` functions ~~?~~ 
- [x] Re-compute inverse of discrete Jacobian, `∂x/∂ξ`, for the 3D case, since the discrete curl-invariant form of `∂ξ/∂x` is *not* equal the inverse of `∂x/∂ξ` (credits to @jkozdon for this observation)  
- [x] Debug and Test

### DIFF
--- a/src/Numerics/Mesh/Grids.jl
+++ b/src/Numerics/Mesh/Grids.jl
@@ -829,7 +829,21 @@ function computegeometry_fvm(elemtocoord, D, ξ, ω, meshwarp)
     (vgeo, sgeo, x_vtk)
 end
 
+"""
+    computegeometry(elemtocoord, D, ξ, ω, meshwarp)
+
+Compute the geometric factors data needed to define metric terms at
+each quadrature point. First, compute so called "topology coordinates" from
+reference coordinates ξ. Then map these topology coordinate
+to physical coordinates. Then setup the geometric factors data necessary at
+each quadrature point by calling [`setup_geom_factors_data!`](@ref). Then
+compute the Jacobian of the mapping from reference coordinates to physical
+coordinates, i.e., ∂x/∂ξ. Finally, compute the metric terms by calling the
+function [`computemetric!`](@ref).
+"""
+
 function computegeometry(elemtocoord, D, ξ, ω, meshwarp)
+    FT = eltype(D[1])
     dim = length(D)
     nface = 2dim
     nelem = size(elemtocoord, 3)
@@ -843,8 +857,6 @@ function computegeometry(elemtocoord, D, ξ, ω, meshwarp)
 
     Np = prod(Nq)
     Nfp = div.(Np, Nq)
-
-    FT = eltype(D[1])
 
     vgeo = VolumeGeometry(FT, Nq, nelem)
 
@@ -866,9 +878,8 @@ function computegeometry(elemtocoord, D, ξ, ω, meshwarp)
 
     setup_geom_factors_data!(vgeo, D...)
 
-    # c) computes partial derivatives ∂x/∂ξ
+    # c) computes Jacobian matrix, ∂x/∂ξ
     compute_dxdxi_jacobian!(vgeo, D...)
-    # Compute Jacobian matrix, ∂x/∂ξ
 
     # d) computes the metric terms
     Metrics.computemetric!(vgeo, sgeo, D...)
--- a/src/Numerics/Mesh/Grids.jl
+++ b/src/Numerics/Mesh/Grids.jl
@@ -1,5 +1,5 @@
 module Grids
-using ..Topologies
+using ..Topologies, ..GeomData
 import ..Metrics, ..Elements
 import ..BrickMesh
 
@@ -829,75 +829,6 @@ function computegeometry_fvm(elemtocoord, D, ξ, ω, meshwarp)
     (vgeo, sgeo, x_vtk)
 end
 
-struct VolumeGeometry{Nq, A<:AbstractArray}
-    ξ1x1::A
-    ξ2x1::A
-    ξ3x1::A
-    ξ1x2::A
-    ξ2x2::A
-    ξ3x2::A
-    ξ1x3::A
-    ξ2x3::A
-    ξ3x3::A
-    x1ξ1::A
-    x2ξ1::A
-    x3ξ1::A
-    x1ξ2::A
-    x2ξ2::A
-    x3ξ2::A
-    x1ξ3::A
-    x2ξ3::A
-    x3ξ3::A
-    ωJ::A
-    ωJI::A
-    ωJH::A
-    x1::A
-    x2::A
-    x3::A
-    JcV::A
-    function VolumeGeometry{Nq}(args::A...) where {Nq, A<:AbstractArray}
-        new{Nq,A}(args...)
-    end
-end
-
-"""
-    VolumeGeometry(FT, Nq::Tuple, nelems::Integer)
-
-Construct an empty `VolumeGeometry` object, in `FT` precision.
-- `Nq` is a tuple containing the number of quadrature points in each direction.
-- `nelem` is the number of elements.
-"""
-function VolumeGeometry(FT, Nq::NTuple{N,Int}, nelem::Int) where {N}
-    array = zeros(FT, prod(Nq), fieldcount(VolumeGeometry), nelem)
-    VolumeGeometry{Nq}(ntuple(j -> @view(array[:, j, :]), fieldcount(VolumeGeometry))...)
-end
-
-struct SurfaceGeometry{Nq, A<:AbstractArray}
-    n1::A
-    n2::A
-    n3::A
-    sωJ::A
-    vωJI::A
-    function SurfaceGeometry{Nq}(args::A...) where {Nq, A<:AbstractArray}
-        new{Nq,A}(args...)
-    end
-end
-
-"""
-SurfaceGeometry(FT, Nq::Tuple, nface::Integer, nelem::Integer)
-
-Construct an empty `SurfaceGeometry` object, in `FT` precision.
-- `Nq` is a tuple containing the number of quadrature points in each direction.
-- `nface` is the number of faces.
-- `nelem` is the number of elements.
-"""
-function SurfaceGeometry(FT, Nq::NTuple{N,Int}, nface::Int, nelem::Int) where {N}
-    Np = prod(Nq)
-    Nfp = div.(Np, Nq)
-    array = zeros(FT, maximum(Nfp), fieldcount(SurfaceGeometry), nface, nelem)
-    SurfaceGeometry{Nfp}(ntuple(j -> @view(array[:, j, :, :]), fieldcount(SurfaceGeometry))...)
-end
-
 function computegeometry(elemtocoord, D, ξ, ω, meshwarp)
     dim = length(D)
     nface = 2dim
@@ -994,7 +925,7 @@ function computegeometry(elemtocoord, D, ξ, ω, meshwarp)
 end
 
 """
-    horizontal_metrics!(vgeo::VerticalGeometry, Nq, ω)
+    horizontal_metrics!(vgeo::VolumeGeometry, Nq, ω)
 
 Compute the horizontal mass matrix `ωJH` field of `vgeo`
 ```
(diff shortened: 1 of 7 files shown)


src/Numerics/Mesh/Metrics.jl
```diff
@@ -98,127 +100,298 @@ function creategrid!(x1, x2, x3, e2c, ξ1, ξ2, ξ3)
...
+function setup_geom_factors_data!(
+    vgeo::VolumeGeometry{NTuple{3,Int}, <:AbstractArray},
+    D1,
+    D2,
+    D3,
+)
+    Nq = (size(D1, 1), size(D2, 1), size(D3, 1))
+    nelem = div(length(vgeo.ωJ), prod(Nq))
+
+    vgeo.x1 = reshape(vgeo.x1, (Nq..., nelem))
```
simonbyrne, line 174: we can't modify the object a field of an immutable object refers to

src/Numerics/Mesh/Metrics.jl
```diff
@@ -98,127 +100,298 @@ function creategrid!(x1, x2, x3, e2c, ξ1, ξ2, ξ3)
...
+function compute_dxdxi_jacobian!(
+    vgeo::VolumeGeometry{NTuple{3,Int}, <:AbstractArray},
+    D1,
+    D2,
+    D3
+)
+
+    T = eltype(vgeo.x1)
+    Nq = (size(D1, 1), size(D2, 1), size(D3, 1))
+
```
simonbyrne, line 252: define local variables here:
```suggestion
r_x1 = reshape(vgeo.x1, (Nq..., nelem))
```
use a consistent convention so that e.g. `r_X` is `X` reshaped.

src/Numerics/Mesh/Metrics.jl
```diff
@@ -98,127 +100,298 @@ function creategrid!(x1, x2, x3, e2c, ξ1, ξ2, ξ3)
...
+    vgeo.ξ2x2 = reshape(vgeo.ξ2x2, (Nq..., nelem))
+    vgeo.ξ3x2 = reshape(vgeo.ξ3x2, (Nq..., nelem))
+    vgeo.ξ1x3 = reshape(vgeo.ξ1x3, (Nq..., nelem))
+    vgeo.ξ2x3 = reshape(vgeo.ξ2x3, (Nq..., nelem))
+    vgeo.ξ3x3 = reshape(vgeo.ξ3x3, (Nq..., nelem))
+
+end
+
+"""
+    compute_dxdxi_jacobian!(vgeo, D1, D2)
```
simonbyrne, line 192: might as well use unicode in the function name.

simonbyrne: If there a standard name in the finite element literature, that would be ideal.

simonbyrne: Might be better to make the function signature `compute_dxdxi_jacobian!(vgeo, (D1, D2))`, as then they're consistent.

src/Numerics/Mesh/Metrics.jl
```diff
@@ -98,127 +100,298 @@ function creategrid!(x1, x2, x3, e2c, ξ1, ξ2, ξ3)
...
-    x1 = reshape(x1, (Nq, nelem))
-    J = reshape(J, (Nq, nelem))
-    JcV = reshape(JcV, (Nq, nelem))
-    ξ1x1 = reshape(ξ1x1, (Nq, nelem))
+    nelem = div(length(vgeo.ωJ), Nq)
+
     nface = 2
-    n1 = reshape(n1, (1, nface, nelem))
-    sJ = reshape(sJ, (1, nface, nelem))
+    sgeo.n1 = reshape(sgeo.n1, (1, nface, nelem))
```
simonbyrne, line 308: again
```suggestion
    r_n1 = reshape(sgeo.n1, (1, nface, nelem))
```
etc.

src/Numerics/Mesh/Metrics.jl
```diff
@@ -98,127 +100,298 @@ function creategrid!(x1, x2, x3, e2c, ξ1, ξ2, ξ3)
...
-    sJ .= 1
+    sgeo.n1[1, 1, :] .= -sign.(vgeo.ωJ[1, :])
+    sgeo.n1[1, 2, :] .= sign.(vgeo.ωJ[Nq, :])
+    sgeo.sωJ .= 1
     nothing
 end
 
 """
-    computemetric!(x1, x2, J, JcV, ξ1x1, ξ2x1, ξ1x2, ξ2x2, sJ, n1, n2, D1, D2)
+    computemetric!(vgeo, sgeo, D1, D2)
```
simonbyrne, line 323: Again, would be good to make this a tuple.
```suggestion
    computemetric!(vgeo, sgeo, (D1, D2))
```

src/Numerics/Mesh/Metrics.jl
valeriabarra: Thanks! I think it's taken care of now.

src/Numerics/Mesh/Metrics.jl
valeriabarra: I decided to call the local variables without the prefix. I thought it was clearer.

src/Numerics/Mesh/Metrics.jl
valeriabarra: In commit [964adc8](https://github.com/CliMA/ClimateMachine.jl/pull/2162/commits/964adc8885a0172c740eaee49c01a209ff5b40a1) I decided to rename this function `compute_reference_to_physical_coord_jacobian` so that it doesn't need to use unicode and it is clear which coordinate transformation we are computing the Jacobian matrix of. Since soon we might have a longer chain rule of coordinate transformations, I thought it was better to be explicit in the name. Let me know what you think

src/Numerics/Mesh/Metrics.jl
valeriabarra: Thanks. I went ahead and changed the definition of this function, as well as the `creategrid` one, to accept a tuple as argument, as suggested

valeriabarra:
Hey @simonbyrne , I added a temporary `array` to the `VolumeGeometry` struct. This suggestion came from @jkozdon actually (thanks to both for your useful suggestions), so that I could have quickly tested my changes in the metric kernels, before having to change all instances of the access to the `vgeo` arguments in all tests and parts of the code that use it. 

I now need some help in debugging this. :( 

It would be good if this got merged into `master` before other PRs that also touch on the geom factors data (e.g., I've noticed  #2170 - but I don't know if there are more)

bischtob:
Why not call the module Geometry?

valeriabarra:
> Why not call the module Geometry?

We could... Thanks for the suggestion. I was thinking to add the "Data" word, because it pertains the data structures needed to store data needed for the geometric factors. But `GeometryData` sounded a bit too long, so I went for the shortened version :)

blallen:
> > Why not call the module Geometry?
> 
> We could... Thanks for the suggestion. I was thinking to add the "Data" word, because it pertains the data structures needed to store data needed for the geometric factors. But `GeometryData` sounded a bit too long, so I went for the shortened version :)

👍 for `Geometry`, especially if the two structures we actually use are called `VolumeGeometry` and `SurfaceGeometry`

reading the rest of this in more detail now, but I already like how it looks just because of the `vgeo -> VolumeGeometry` and `sgeo -> SurfaceGeometry` change 😉

valeriabarra:
> > > Why not call the module Geometry?
> > 
> > 
> > We could... Thanks for the suggestion. I was thinking to add the "Data" word, because it pertains the data structures needed to store data needed for the geometric factors. But `GeometryData` sounded a bit too long, so I went for the shortened version :)
> 
> 👍 for `Geometry`, especially if the two structures we actually use are called `VolumeGeometry` and `SurfaceGeometry`
> 
> reading the rest of this in more detail now, but I already like how it looks just because of the `vgeo -> VolumeGeometry` and `sgeo -> SurfaceGeometry` change 😉

Thank you! Just be aware that I'm still in the debugging/testing phase. But hopefully this could go in pretty quickly 😊
I will change the module name as suggested

valeriabarra:
@blallen , @bischtob we already have a module called `Geometry.jl` and it pertains all the geometry (including resolution and length scales). 

What do you think if we called this new one `GeometricFactors.jl` ? Since it really contains the geometric factor data?

bischtob:
> @blallen , @bischtob we already have a module called `Geometry.jl` and it pertains all the geometry (including resolution and length scales).
> 
> What do you think if we called this new one `GeometricFactors.jl` ? Since it really contains the geometric factor data?

Sounds good :)!

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

src/Numerics/Mesh/Grids.jl
```diff
@@ -301,6 +301,8 @@ struct DiscontinuousSpectralElementGrid{
         (vgeo, sgeo, x_vtk) =
             computegeometry(topology.elemtocoord, D, ξ, ω, meshwarp)
 
+        vgeo = vgeo.array
+        sgeo = sgeo.array
```
valeriabarra, line 305: @simonbyrne and @jkozdon  I put here these assignments so that tests that use the driver configs could pass. But now I still have other tests (see CI failure) that do not pass. I don't know what's the best way to fix these issues without having to change everywhere in the code where `vgeo` and `sgeo` are used assuming the old array type. Any help would be much appreciated, thanks.

valeriabarra:
Note that the 1D, 2D, and 3D  `test/Numerics/Mesh/Metrics.jl` kernel tests all pass now with the new data structures.

valeriabarra:
bors try

src/Numerics/Mesh/Grids.jl
valeriabarra: Thanks, Jeremy! He helped me out identify where the issue was

jkozdon: No problem. Glad I could be a second second set of eyes for you.

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
bors try

valeriabarra:
This PR is ready for review, thanks! 
Sorry about the large number of commits and bors invocations. I was struggling once again with debugging some `Documenter.jl` warnings. If things look good, I will reduce the number of commits after revision. Thanks!

src/Numerics/Mesh/Metrics.jl
```diff
@@ -391,247 +574,151 @@ function computemetric!(
...
+                    a31[i, j, k] * a21[i, j, k] +
+                    a32[i, j, k] * a22[i, j, k] +
+                    a33[i, j, k] * a23[i, j, k]
+                )
+            x3ξ3[i, j, k, e] =
+                1.0 / det * (
+                    a31[i, j, k] * a31[i, j, k] +
+                    a32[i, j, k] * a32[i, j, k] +
+                    a33[i, j, k] * a33[i, j, k]
+                )
```
valeriabarra, line 667: @jkozdon @simonbyrne  I know that for a 3x3 we could have used a built-in method to invert the matrix, but I needed to do it for every quadrature point, in each element, so I went for the inversion of the matrix by hand. Let me know if you have any feedback, thanks!

simonbyrne: `inv` with StaticArrays should be equivalent.

valeriabarra: Thanks. Do you want me to change it?

simonbyrne: It's fine for now. We can revisit this in ClimateMachineCore.jl

simonbyrne (approved):

simonbyrne:
Thanks, that looks great.

valeriabarra:
Thanks @simonbyrne for the approval. I would clean up the number of commits before merging. I'll stop bors if it's not too late

valeriabarra:
bors try

valeriabarra:
bors try-

valeriabarra:
bors try

PR #2157, opened 2021-04-01
Land/soil/phase change tutorial

kmdeck:
### Description
This adds in two tutorials on phase changes in soil. One compares to an analytic solution, which applies under certain circumstances (no water diffusion, e.g.), and the other carries out a more complex example of a freezing front (with water diffusion). 

The source.jl code is also updated to reflect a change in how we model freezing and thawing.

This also required an update in the integration test.
- [X] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [X] Unit tests are included OR N/A.
- [X] Code is exercised in an integration test OR N/A.
- [X] Documentation has been added/updated OR N/A.

PR #2165, opened 2021-04-05
do not render notebook output as the default is to execute it

jakebolewski:
### Description

Looking at the generation of https://github.com/CliMA/ClimateMachine.jl/pull/2157 we are actually (still) executing the tutorials multiple times because the default for notebook generation is to execute the notebook and we already do that for the markdown output.

### DIFF
--- a/docs/list_of_tutorials.jl
+++ b/docs/list_of_tutorials.jl
@@ -122,7 +122,6 @@ if generate_tutorials
                 documenter = false,
                 preprocess = mdpre,
             )
-            Literate.notebook(input)
         end
     end
 


jakebolewski:
@charleskawczynski  I think this was just overlooked when we were updating the tutorial generation.

charleskawczynski (approved):
Good catch

charleskawczynski:
Yeah, it looks like the [default](https://github.com/fredrikekre/Literate.jl/blob/master/src/Literate.jl#L298) is execute for notebook.

Issue #2116, opened 2021-03-15
Grid imprinting in specific variables (per unit mass) due to filtering, and energy-momentum leakage


ilopezgp:
### Description

The application of DG filters to conservative variables, e.g. `"energy.ρe"`, can lead to grid imprinting in specific variables per unit mass, such as the potential temperature θ. This issue is demonstrated with an Anelastic1D simulation of a dry neutrally stratified Ekman layer in DG, using polynomial order `N=4` and filtering in `"energy.ρe"`. The setup to reproduce these results can be found in PR #2112.

In addition, there seems to exist leakage from momentum into the energy field. The potential temperature in this simulation should stay constant, since there is no buoyancy forcing of any kind. We will see below that this is not the case.

Finally, the horizontal velocity should develop the classical Ekman spiral over time, starting from an initial constant velocity profile. The vertical extent of the Ekman layer depends on the strength of SGS diffusion. Here, I consider two cases: 1) Constant kinematic viscosity of 0.1, 2) Smagorinsky-Lilly diffusion with parameter equal to 0.21.

To showcase the problem, I first perform a simulation using the aforementioned configuration for 9 simulation hours and **NO FILTERING**. Results are shown in the figure below for constant kinematic viscosity of 0.1.

<img width="1258" alt="Screen Shot 2021-03-15 at 4 05 45 PM" src="https://user-images.githubusercontent.com/38085953/111232424-557a2400-85a8-11eb-9543-080cba067b40.png">

Note that in the previous figure, the potential temperature field experiences perturbations that mimic the horizontal velocity structure. These perturbations are not physical, to my knowledge. 

Grid imprinting is not an issue in the previous simulation. However, it appears when applying filtering. I consider now the same simulation, using Boyd Vandeven filtering with `Nc=1, s=4`, following the [tutorial](https://clima.github.io/ClimateMachine.jl/latest/generated/Numerics/DGMethods/showcase_filters/). The results are shown below,

<img width="1254" alt="Screen Shot 2021-03-15 at 4 13 03 PM" src="https://user-images.githubusercontent.com/38085953/111232960-58294900-85a9-11eb-9a74-82ad28bbbdcc.png">

Note that there is grid imprinting throughout the θ profile caused by the filter. Continuing through the examples in the [tutorial](https://clima.github.io/ClimateMachine.jl/latest/generated/Numerics/DGMethods/showcase_filters/), I use now `Nc=1, s=8`,

[...]

PR #2167, opened 2021-04-06
Add an intrinsic filter to EDMF model

yairchn:
### Description
This PR adds an IntrinsicEDMFFilter to edmf_model.jl with the filter kernels in edmf_kernels.jl and a list of boolean arguments for which of the EDMF variable should be filtered 

### DIFF
--- a/test/Atmos/EDMF/edmf_kernels.jl
+++ b/test/Atmos/EDMF/edmf_kernels.jl
@@ -31,7 +31,8 @@ import ClimateMachine.TurbulenceConvection:
     turbconv_normal_boundary_flux_second_order!
 
 using ClimateMachine.Thermodynamics: air_pressure, air_density
-
+import ClimateMachine.Mesh.Filters:
+    vars_state_filtered, compute_filter_argument!, compute_filter_result!
 
 include(joinpath("helper_funcs", "nondimensional_exchange_functions.jl"))
 include(joinpath("helper_funcs", "lamb_smooth_minimum.jl"))
@@ -76,6 +77,9 @@ end
 function vars_state(::Environment, ::Primitive, FT)
     @vars(atke::FT, aθ_liq_cv::FT, aq_tot_cv::FT, aθ_liq_q_tot_cv::FT,)
 end
+
+vars_state_filtered(target::SpecificEDMFFilter, FT) =
+    vars_state(target.atmos, Primitive(), FT)
 
 function vars_state(
     m::NTuple{N, Updraft},
@@ -154,6 +158,9 @@ function vars_state(m::EDMF, st::GradientFlux, FT)
         ∇v::SVector{3, FT}
     )
 end
+
+
+
 
 abstract type EDMFPrognosticVariable <: AbstractPrognosticVariable end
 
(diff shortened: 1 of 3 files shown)


test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -372,6 +379,61 @@ function primitive_to_prognostic!(
 
 end
 
+function compute_filter_argument!(
+    target::IntrinsicEDMFFilter,
```
ilopezgp, line 383: I think the notation we use throughout the model is "specific" for intensive variables per unit mass. See #2116 . Intrinsic does not have a well established meaning in this context.

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -372,6 +379,61 @@ function primitive_to_prognostic!(
...
+    gm = state
+    up = state.turbconv.updraft
+    en = state.turbconv.environment
+    fs_up = filter_state.turbconv.updraft
+    fs_en = filter_state.turbconv.environment
+    ρ_inv = 1 / gm.ρ
+    N_up = n_updrafts(turbconv_model(target.atmos))
+
+    @unroll_map(N_up) do i
+        target.ρa && (fs_up[i].a = up[i].ρa * ρ_inv)
```
ilopezgp, line 397: I am not familiar with `&&` in Julia unroll maps, what is this doing?

test/Atmos/EDMF/edmf_kernels.jl
```diff
@@ -77,6 +78,9 @@ function vars_state(::Environment, ::Primitive, FT)
     @vars(atke::FT, aθ_liq_cv::FT, aq_tot_cv::FT, aθ_liq_q_tot_cv::FT,)
 end
 
+vars_state_filtered(target::SpecificEDMFFilter, FT) =
+    vars_state(target.atmos, Primitive(), FT)
```
ilopezgp, line 82: Is this supposed to be Primitive() or Prognostic() ?

yairchn: its a shortcut to get definitions of the density free variables as defined in Primitive

charleskawczynski: `Primitive` is correct-- we want the `NamedTuple` type to exclude `ρ`

test/Atmos/EDMF/edmf_kernels.jl
charleskawczynski: `unroll_map` wraps `i` in `Val` and manually unrolls the call to `map` ([here](https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Utilities/VariableTemplates/VariableTemplates.jl#L369-L389)). This was some magic @simonbyrne wrote.

The syntax
```julia
foo() && bar()
```
is Julia short hand for
```julia
if foo()
bar()
end
```

ilopezgp: I get it now, pretty neat!

ilopezgp (approved):

test/Atmos/EDMF/edmf_model.jl
```diff
@@ -483,6 +483,26 @@ function EDMF(
     return EDMF{FT, N_up, typeof.(args)...}(args...)
 end
 
+Base.@kwdef struct SpecificEDMFFilter{A, B <: Bool} <: AbstractFilterTarget
+    "The AtmosModel"
+    atmos::A
+    "filter edmf updraft a"
+    ρa::B = false
```
ilopezgp, line 490: Can we set all of these to true?

test/Atmos/EDMF/ekman_layer.jl
```diff
@@ -104,19 +106,21 @@ function main(::Type{FT}, cl_args) where {FT}
         N_quad,
         param_set,
         surface = NeutralDrySurfaceModel{FT}(param_set),
+        coupling = Coupled(),
```
ilopezgp, line 109: I would keep this decoupled for now. I would like this to test the Anelastic1D with grid mean viscosity

test/Atmos/EDMF/ekman_layer.jl
```diff
@@ -104,19 +106,21 @@ function main(::Type{FT}, cl_args) where {FT}
...
     )
 
     C_smag_ = C_smag(param_set)
-    turbulence = ConstantKinematicViscosity(FT(0.1))
+    turbulence = ConstantKinematicViscosity(FT(0))
```
ilopezgp, line 113: Following the comment above, I would keep this as it was. We can introduce another test for coupled EDMF.

ilopezgp:
bors try

test/Atmos/EDMF/edmf_kernels.jl
charleskawczynski: We've discussed this since and, actually, we need to spell this out because `Primitive` has pressure (instead of energy?), and we want specific total internal energy.

ilopezgp (approved):

PR #2169, opened 2021-04-06
Add cutoff filter params in single stack config

charleskawczynski:
### Description

I think this is all that's needed. Can someone check that this works for a particular driver?


yairchn (approved):
LGTM

src/Driver/driver_configs.jl
```diff
@@ -583,9 +583,12 @@ function SingleStackConfiguration(
     numerical_flux_second_order = CentralNumericalFluxSecondOrder(),
     numerical_flux_gradient = CentralNumericalFluxGradient(),
     fv_reconstruction = nothing,
+    Ncutoff = get_polyorders(N),
```
ilopezgp, line 586: I suppose `Ncutoff` is defined so we can set the cutoff manually, but can't we do that already through `ClimateMachine.Settings.cutoff_degree`?

src/Driver/driver_configs.jl
```diff
@@ -622,12 +625,24 @@ function SingleStackConfiguration(
         meshwarp = meshwarp,
     )
 
+    if cutofforder_horz < polyorder_horz && cutofforder_vert < polyorder_vert
+        filter = CutoffFilter(
+            grid,
+            (cutofforder_horz + 1, cutofforder_horz + 1, cutofforder_vert + 1),
+        )
+    else
```
ilopezgp, line 633: would it be useful to add an else if for cases where we only filter horizontal or vertical?

charleskawczynski: Yeah, that's a good point, we might want to carefully handle that in the SingleStackConfiguration.

src/Driver/driver_configs.jl
charleskawczynski: I kind of just followed the existing pattern from the LES/GCM configs 🤷🏻‍♂️

ilopezgp: That's fine then. If anything, it's redundant but it doesn't harm

src/Driver/driver_configs.jl
valeriabarra: We had  a similar [conversation](https://github.com/CliMA/ClimateMachine.jl/pull/1953#issuecomment-773493460) in PR #1953 . I put the `&&` in that condition, because a different condition for horizontal/vertical is currently not supported at the grid level, where we have an `@assert` everywhere, checking for cutoff degree to be less than polynomial degree component-wise. So having only one of the two, would require refactoring that.

PR #2170, opened 2021-04-07
3D Fast Tensor Product kernels are added.

skandalaCLIMA:
3D Fast Tensor Product kernels are added.
Author: Sriharsha Kandala <skandala@caltech.edu>

### Description


src/Numerics/Mesh/Grids.jl
```diff
@@ -263,6 +270,8 @@ struct DiscontinuousSpectralElementGrid{
     """
     facemap::Union{DAI3, Nothing}
 
+    "Temporary Storage for on spectral element mesh"
+    scratch::DAT3
```
valeriabarra, line 274: I am not fully understanding this docstring. Also, would you consider giving this variable a more explanatory name? Even though it is temporary only, it would make the code easier for others to follow

skandalaCLIMA: These arrays are used for intermediate storage when contructing various terms needed for CG kernels and post-processing. Its size is similar to state variables and lives on the DG/CG grid. I will expand on the name and documentation.

valeriabarra: Thanks!

PR #2135, opened 2021-03-22
add mass preserving filter

sandreza:
This PR adds a cutoff filter that preserves the cell average for curved elements. The current filters in CLIMA do not have this property

- [x] Code follows the [style guidelines](https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/) OR N/A.
- [x] Unit tests are included OR N/A.
- [x] Code is exercised in an integration test OR N/A.
- [x] Documentation has been added/updated OR N/A.

PR #2171, opened 2021-04-07
Add spherical domains for CNSE3D model

blallen:
### Description

This PR adds everything necessary for running a hydrostatic balance test on the sphere with the CNSE3D model, as well as testing tracer conservation. Some interface changes were made to the FluidBC to make them GPU friendly and additional domains were added to handle the sphere. The core numerics change to get these tests running properly were added in #2135.


blallen:
bors try

blallen:
bors try

test/Numerics/DGMethods/compressible_navier_stokes_equations/three_dimensional/ThreeDimensionalCompressibleNavierStokesEquations.jl
```diff
@@ -279,8 +279,8 @@ end
     ρu = state.ρu
     ρθ = state.ρθ
 
-    u = ρu / ρ
-    θ = ρθ / ρ
+    u = ρu
+    θ = ρθ
```
blallen, line 283: @sandreza why did you need to do this?

blallen:
bors try

test/Numerics/DGMethods/compressible_navier_stokes_equations/three_dimensional/ThreeDimensionalCompressibleNavierStokesEquations.jl
sandreza: I was playing around with momentum diffusion, (which seems like the right thing to do in this case)

sandreza (approved):

test/Numerics/DGMethods/compressible_navier_stokes_equations/shared_source/grids.jl
```diff
@@ -190,3 +190,116 @@ function DiscretizedDomain(
...
+        x3, x2, x1 = f(R, b / c, a / c)
+    elseif fdim == 3 && c < 0
+        # ( *, *,-R) : Face VI from Ronchi, Iacono, Paolucci (1996)
+        x3, x2, x1 = f(-R, b / c, a / c)
+    else
+        error("invalid case for cubedshellwarp: $a, $b, $c")
+    end
+
+    return x1, x2, x3
+end
```
charleskawczynski, line 236: This exact code snippet already exists exists [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Numerics/Mesh/filter.jl#L401-L434). Can we somehow share this instead?

sandreza: > This exact code snippet already exists exists [here](https://github.com/CliMA/ClimateMachine.jl/blob/master/test/Numerics/Mesh/filter.jl#L401-L434). Can we somehow share this instead?

We should definitely share it. The code snippet is already in https://github.com/CliMA/ClimateMachine.jl/blob/master/src/Numerics/Mesh/Topologies.jl#L2022. The keyword arguments made it so that the function could not be called in a way that repeated previous behavior. I'm open to suggestions on the best way of moving forward on that

test/Numerics/DGMethods/compressible_navier_stokes_equations/sphere/grid_notes.jl
```diff
@@ -0,0 +1,38 @@
...
+# second index is ids
+# third index is elements
+# the are are 16 ids
+# 1-9 are metric terms
+# 10 is the mass matrix
+# 11 is the inverse mass matrix
+# 12 is MHid (horizontal mass matrix?)
+# 13-15 are x1 x2 x3
+# 16 is the vertical volume jacobian
+# grid.vgeo[:, grid.ξ1x1id, :]
```
blallen, line 38: @sandreza @charleskawczynski we should probably add this to official documentation somewhere

jkozdon: @valeriabarra is reworking all this as well in #2162

blallen:
bors try

src/Numerics/Mesh/Topologies.jl
```diff
@@ -2024,27 +2024,12 @@ function cubedshelltopowarp(
     b,
     c,
     R = max(abs(a), abs(b), abs(c));
-    r_inner = _planet_radius,
-    r_outer = _planet_radius + domain_height,
+    r_inner = 0,
+    r_outer = 1,
     topography::AnalyticalTopography = NoTopography(),
 )
+    f = get_topography_function(topography, r_inner, r_outer)
```
blallen, line 2031: @charleskawczynski here's how I handled the code duplication

test/Numerics/DGMethods/compressible_navier_stokes_equations/three_dimensional/ThreeDimensionalCompressibleNavierStokesEquations.jl
blallen: @jkozdon any thoughts?

src/Numerics/Mesh/Topologies.jl
charleskawczynski: Why couldn't we call the existing function? Wasn't it exactly the same?

test/Numerics/DGMethods/compressible_navier_stokes_equations/shared_source/grids.jl
valeriabarra: @sandreza , @blallen In PR #2060 I had renamed `cubedshellwarp` as  `cubed_sphere_warp` and so on for the unwarp version and added two specialized case wrappers for the multiple dispatch (I preferred using `snake_case` because it is more readable/accessible). I reckon that I didn't update the `cubedshelltopowarp` function because it was not clear to us in the atmos group how/where that was used. It would be nice if now that you are touching this, the interface and function names could be unified a bit. Thanks

bischtob (approved):

src/Numerics/Mesh/Topologies.jl
blallen: it relied on planet_radius and domain_height and explicitly called analytical topography... which is unnecessary if you don't have topography. also causes the code to crash if planet_radius and domain_height are undefined, which is pretty common in the numerics tests. 

but also see this comment by @valeriabarra: https://github.com/CliMA/ClimateMachine.jl/pull/2171#discussion_r611946215

test/Numerics/DGMethods/compressible_navier_stokes_equations/shared_source/grids.jl
blallen: oh! is that where it went? in that case I can probably remove this commit https://github.com/CliMA/ClimateMachine.jl/pull/2171/commits/6d4c7eaba50683d11eabfda81cd9057d4536f3b4 and just call `cubed_sphere_warp`

blallen:
bors try

src/Numerics/Mesh/Topologies.jl
blallen: dropped this change based on @valeriabarra's comment

test/Numerics/DGMethods/compressible_navier_stokes_equations/shared_source/grids.jl
blallen: it worked! yay fewer changes!

blallen:
bors try

test/Numerics/DGMethods/compressible_navier_stokes_equations/shared_source/grids.jl
valeriabarra: Thanks @blallen . So, the `cubedshelltopowarp` and `cubedshelltopounwarp` are not needed at all then? Should they be renamed at least to follow the current convention of the other ones?

valeriabarra: @blallen not sure if you saw this. I can rename them in a separate quick PR

valeriabarra: Added #2181

blallen: sorry for the delay; gmail has decided everything from github is spam :'(

I think the `topowarp` versions are different functions? so #2181 is probably the way to go

PR #2172, opened 2021-04-07
Add mutating flux-source knl fallbacks in BL

charleskawczynski:
### Description

This PR adds

 - `flux_first_order!`
 - `flux_second_order!`
 - and `source!`

fallbacks in `BalanceLaws/`, and removes the implementation in `AtmosModel.jl`. We can also remove the implementations in the land model once #2136 lands.

I guess one benefit of keeping these implementations would be for simpler debugging when we start coupling models, so I'd like to hear feedback about this idea.

### DIFF
--- a/docs/src/APIs/BalanceLaws/BalanceLaws.md
+++ b/docs/src/APIs/BalanceLaws/BalanceLaws.md
@@ -23,6 +23,7 @@ Source
 TendencyDef
 eq_tends
 prognostic_vars
+get_prog_state
 projection
 precompute
 prognostic_var_source_map
--- a/src/Atmos/Model/AtmosModel.jl
+++ b/src/Atmos/Model/AtmosModel.jl
@@ -599,39 +599,6 @@ function precompute(atmos::AtmosModel, args, tt::Flux{FirstOrder})
     return (; ts, turbconv)
 end
 
-"""
-    flux_first_order!(
-        m::AtmosModel,
-        flux::Grad,
-        state::Vars,
-        aux::Vars,
-        t::Real
-    )
-
-Computes and assembles non-diffusive fluxes in the model
-equations.
-"""
-@inline function flux_first_order!(
-    atmos::AtmosModel,
-    flux::Grad,
-    state::Vars,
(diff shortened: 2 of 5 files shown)


kpamnany (approved):
I think specialized implementations can be added back in if needed, but this should be the default.

kmdeck:
Hi Charlie, 

Im answering this from the point of view of not understanding the whole picture and pros/cons :) but I like the idea of this being out of view - it's like how the original implementation of the accumulation of sources over the source tuple was out of view, and the users just added in methods of `source` for each possible source type. 

I apologize, I think you already addressed this, but I think it is important, as tendency types are developed more, to keep the option of defining simple balance laws and problems, like that of the heat tutorial, without having to use them. That tutorial was such an useful introduction to the balance law structure for us on the land side, and it's probably complex enough for a new user as is. If possible, I think it is better to introduce new users to these abstractions one at a time - i.e. preserving the option not to use the tendency spec.

anyways, PR looks fine to me 👍

kmdeck (approved):

charleskawczynski:
> I apologize, I think you already addressed this, but I think it is important, as tendency types are developed more, to keep the option of defining simple balance laws and problems, like that of the heat tutorial, without having to use them. That tutorial was such a useful introduction to the balance law structure for us on the land side, and it's probably complex enough for a new user as is. If possible, I think it is better to introduce new users to these abstractions one at a time - i.e. preserving the option not to use the tendency spec.

We still have the option of `import`ing the `flux_first_order!`, `flux_second_order!`, and `source!` methods and configuring problems in the old way as done in the heat tutorial.

That said, I guess this PR does create a kind of disconnect with the heat tutorial and the Atmos/Land implementations. I've thought about applying the same changes to the heat tutorial, but I think we have room to improve the required boilerplate that the tendency specification layer requires. I don't think it's a blocker though.

@kpamnany one thing I noticed is that we're not really methodical w.r.t using `@inline` (we use it for all but `source!` in Atmos, and it doesn't look like it was used at all in Land). Thoughts on this?

PR #2173, opened 2021-04-07
Add physics to AtmosBC constructor

charleskawczynski:
### Description

In preparation for #2122, this PR adds `physics` to the `AtmosBC` constructor, so that we can loop over the prognostic variables and slowly remove all other `AtmosBC` (e.g., `momentum`, `energy`, etc.).

### DIFF
--- a/docs/src/GettingStarted/Atmos.md
+++ b/docs/src/GettingStarted/Atmos.md
@@ -7,25 +7,35 @@ below. In this implementation of the `AtmosModel` we concern ourselves
 with the conservative form of the compressible equations of moist fluid
 motion given a set of initial, boundary and forcing(source) conditions.
 
+First, we construct the atmospheric physics via `AtmosPhysics`:
+
+```
+physics = AtmosPhysics{FT}(;
+    param_set::AbstractParameterSet;
+    ref_state = HydrostaticState(DecayingTemperatureProfile{FT}(param_set),)
+    turbulence = SmagorinskyLilly{FT}(0.21),
+    hyperdiffusion = NoHyperDiffusion(),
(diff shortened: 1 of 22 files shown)


charleskawczynski:
There is an alternative to this-- we could allow `AtmosProblem`/`AtmosBC` to be constructed with `physics = nothing`. Then, we could 
 - Check for `nothing` when constructing the `AtmosModel`
 - If the `AtmosBC` was constructed with `nothing`, then reconstruct `AtmosProblem` in `AtmosModel` with `physics` (which already exists).

The pros / cons:
 - Disallowing `AtmosBC(nothing)` constructor (what the PR currently does): more driver changes, requires specific construction order of `AtmosProblem` and `AtmosPhysics`
 - Allowing `AtmosBC(nothing)` constructor (alternative): fewer to no driver changes, requires no specific construction order of `AtmosProblem` and `AtmosPhysics`, but the `AtmosModel` will overwrite the input `AtmosBC`/`AtmosProblem` when `AtmosBC` is constructed with `AtmosBC(physics = nothing)`.

jakebolewski:
As an alternative couldn't you pass through the method / type constructor in the AtmosModel definition and then construct and instance of the BC object as late as possible when it is actually needed, passing though the physics defined in the AtmoModel at that point?

jakebolewski (approved):

charleskawczynski:
> As an alternative couldn't you pass through the method / type constructor in the AtmosModel definition

Yes, but then how could users customize `AtmosBC` at the driver level?

charleskawczynski:
Let's move forward with this-- I think it'll be easier to undo if we want.

PR #2174, opened 2021-04-08
add in surface flow component; overland flow model

kmdeck:
### Description
add in a surface flow component to the Land Model. Currently, it supports a NoSurfaceFlow model, which is the default, and also overland flow by solving the shallow water equations under a number of assumptions (St Venant Equations).


src/Land/Model/SurfaceFlow.jl
```diff
@@ -0,0 +1,256 @@
...
+    state⁺.surface.area = bc_function(aux⁻, t)
+end
+
+"""
+    Precip <: TendencyDef{Source}
+
+A source term for overland flow where a prescribed net precipitation
+drives the overland flow.
+"""
+struct Precip{FT, F} <: TendencyDef{Source}
```
kmdeck, line 231: I stuck this in here b/c it is only used with the SurfaceFlow model. downside is now that sources are split across multiple files..

src/Land/Model/SurfaceFlow.jl
```diff
@@ -0,0 +1,256 @@
...
+    end
+end
+
+function (p::Precip{FT})(x, y, t) where {FT}
+    FT(p.precip(x, y, t))
+end
+
+
+prognostic_vars(::Precip) = (SurfaceWaterArea(),)
+#I thought there was a default so I wouldnt have to define this?
```
kmdeck, line 245: I thought there was a default so I wouldnt have to define this?

src/Land/Model/land_tendencies.jl
```diff
@@ -9,19 +9,36 @@ eq_tends(pv::AbstractPrognosticVariable, m::LandModel, tt::Source) =
 ##### First order fluxes
 #####
 
-eq_tends(::PV, ::LandModel, ::Flux{FirstOrder}) where {PV} = ()
+eq_tends(pv::PV, land::LandModel, tt::Flux{FirstOrder}) where {PV} = (
+    eq_tends(pv, land.soil.heat, tt)...,
+    eq_tends(pv, land.soil.water, tt)...,
+    eq_tends(pv, land.surface, tt)...,
+)
+# can we replace these empty ones with a method for ::BalanceLaw?
```
kmdeck, line 17: # can we replace these empty ones with a method for ::BalanceLaw?

src/Land/Model/SurfaceFlow.jl
kmdeck, line 231: also, does it matter that TendencyDef{Source} doesnt know what FT, F are?

src/Land/Model/SurfaceFlow.jl
```diff
@@ -0,0 +1,256 @@
...
+        new{FT, F}(precip)
+    end
+end
+
+function (p::Precip{FT})(x, y, t) where {FT}
+    FT(p.precip(x, y, t))
+end
+
+
+prognostic_vars(::Precip) = (SurfaceWaterArea(),)
```
kmdeck, line 244: also - Im still calling the prognostic variable an area,and we divide by a parameter width(x,y) to get a height. actually, in 2D overland flow, the PV should just be the height. But I left this alone for now.

src/Land/Model/source.jl
```diff
@@ -29,7 +30,39 @@ function precompute(land::LandModel, args, tt::Source)
     end)
     return (; dtup)
 end
+#=
```
kmdeck, line 33: will remove commented stuff when we're ready

.gitignore
```diff
@@ -5,6 +5,7 @@
 *.jl.*.cov
 *.jl.mem
 *.DS_Store
+*.vscode
```
kmdeck, line 8: not sure why this creeped in or why it is changed..

jakebolewski:
bors try

src/Land/Model/SurfaceFlow.jl
charleskawczynski: Yeah, that is one downside 🙁. Ultimately, it's because the types need to be defined before they are referred to.

charleskawczynski: No, `TendencyDef{Source}` doesn't need `FT`, but the subtypes can have them. Sometimes we have them in the atmosphere, e.g. `BomexGeostrophic{FT} <: TendencyDef{Source}`.

src/Land/Model/SurfaceFlow.jl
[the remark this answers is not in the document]
charleskawczynski: We need this definition to provide information as to which prognostic variables each source should be applied to only because sources are passed in as a `Tuple` at the driver level. Fluxes are handled differently since they're all handled in `eq_tends` and not passed in from the driver.

src/Land/Model/land_tendencies.jl
charleskawczynski: Yes, I think we could define a highly generic fallback. I think I didn't at first just to be careful, but since everything shows up in the table we should know if we're accidentally calling the fallback. Probably better to do this in a separate PR, though.

jakebolewski:
bors try

.gitignore
kmdeck: changing this back

src/Land/Model/SurfaceFlow.jl
kmdeck: Hm - I actually stuck it in here because it is only used by this model, and the default for this model is to not have any prognostic variables. In the future I think it might make sense to rename `source.jl` to `soil_source.jl` unless we decide to put all sources for every component in that file. we'll see when we get more components

src/Land/Model/land_tendencies.jl
kmdeck: ok! ill open up an issue so we dont forget

src/Land/Model/land_tendencies.jl
```diff
@@ -9,19 +9,36 @@ eq_tends(pv::AbstractPrognosticVariable, m::LandModel, tt::Source) =
...
+eq_tends(::SurfaceWaterArea, ::OverlandFlowModel, ::Flux{FirstOrder}) =
+    (VolumeAdvection(),)
 
 #####
 ##### Second order fluxes
 #####
 
 # Empty by default
+
+# can we replace these empty ones with a method for ::BalanceLaw?
```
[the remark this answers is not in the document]
kmdeck: issue #2182

src/Land/Model/SurfaceFlow.jl
```diff
@@ -0,0 +1,248 @@
...
+
+
+function precompute(source_type::Precip, land::LandModel, args, tt::Source)
+    nothing
+end
```
charleskawczynski, line 240: `precompute` should always return a `NamedTuple`

charleskawczynski: Also, there's a fallback, so we generally don't need to define this unless it returns something (unless you need a fallback with a different number of arguments). I think we could probably define a generic fallback for 4 arguments like this also (i.e., make this definition generic and move to `BalanceLaws/`)

kmdeck: ah ok. Ill change that! also, look like there is no fallback method for a `source_type` that does nothing? i.e. if not all sources need to precompute something.

kmdeck: (I see a fallback for `precompute(bl, args, source)` only)

charleskawczynski: Yeah, fine to leave it here for now, but instead return a `NamedTuple`. We can make it generic and move to BalanceLaws layer

kmdeck: sorry, some crosstalk while the page was not reloaded :) thanks for the feedback!

kmdeck:
@jakebolewski I think this looks pretty good except for the difference between the CSV file and the output, and I still am unsure if we should change the variable name to `height` instead of `area` and remove the `width` factor. It doesnt really make sense for this, but I think we would need that if we wanted to model rivers using this type of model. I guess the idea was that a possible river model (using clima solvers) would define a width function, but overland flow would set width = 1, so it doesnt affect the balance law.

kmdeck:
bors try

kmdeck:
bors try

jakebolewski (approved):

PR #2177, opened 2021-04-12
Add `AtmosSpecificFilterPerturbations` filter

charleskawczynski:
### Description

This PR adds `AtmosSpecificFilterPerturbations` which filters the specific variable (prognostic variables divided by ρ) perturbations.

At the moment, the driver that implements this (`test/Atmos/EDMF/ekman_layer.jl`) crashes after a while of running. I'm not 100% sure about implementation details (@yairchn, @ilopezgp please check to make sure this is doing what we intend), but it's at least running now.


yairchn (approved):
LGTM

src/Atmos/Model/filters.jl
```diff
@@ -51,3 +46,78 @@ function compute_filter_result!(
...
+
+function compute_filter_argument!(
+    target::AtmosSpecificFilterPerturbations,
+    filter_state::Vars,
+    state::Vars,
+    aux::Vars,
+)
+
+    ρ_inv = 1 / state.ρ
+    filter_state.ρ = state.ρ - aux.ref_state.ρ
```
ilopezgp, line 65: I don't think we should filter density in the specific filter.

src/Atmos/Model/filters.jl
```diff
@@ -51,3 +46,78 @@ function compute_filter_result!(
...
+
+function compute_filter_result!(
+    target::AtmosSpecificFilterPerturbations,
+    state::Vars,
+    filter_state::Vars,
+    aux::Vars,
+)
+
+    state.ρ = filter_state.ρ + aux.ref_state.ρ
+    ρ = state.ρ # TODO: double check
```
ilopezgp, line 98: The density that we use to reconstruct the filtered state should not be filtered. Otherwise this is equivalent to the full filtering approach.

ilopezgp (changes requested):

src/Atmos/Model/filters.jl
charleskawczynski: What should this be, then?

src/Atmos/Model/filters.jl
charleskawczynski: I'm not sure how we can get `state.ρ` in `compute_filter_result!` if it's not filtered. Might need to rope in @mwarusz.

mwarusz: In `compute_filter_result!` the `state` output argument comes in filled with the all state variables so it should contain `state.ρ`.

ilopezgp (approved):

yairchn (approved):
LGTM

PR #2184, opened 2021-04-20
Add single tutorial mode

charleskawczynski:
### Description

This PR adds a hack in the tutorials so that we can flip a switch (`run_single_tutorial`) to run single tutorials without executing all of the other tutorials / documentation. This is quite a bit of a hack, but it does significantly reduce the single tutorial generation time:

```julia
include("docs/make.jl") # compile first, takes ~ 70 seconds
...
@time include("docs/make.jl")
....
1.310631 seconds (2.47 M allocations: 139.898 MiB, 2.17% gc time)
```

This may help alleviate tutorial development pain until Documenter.jl offers a way to build a single doc page--I don't see an open issue, but I also don't see this feature anywhere in the docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ docs/build/
 docs/site/
 docs/src/tutorials/
 docs/src/generated/
+docs/transient_dir/generated/
 !docs/src/assets/*.png
 !docs/src/assets/*.svg
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -7,6 +7,14 @@ ENV["CLIMATEMACHINE_SETTINGS_DISABLE_CUSTOM_LOGGER"] = true
 
 using Distributed
 using DocumenterCitations
+
+run_single_tutorial = false && isempty(get(ENV, "CI", "")) # never use remotely
+@everywhere source_dir = joinpath(@__DIR__, "src")
+if run_single_tutorial
+    tutorial_string_match = "<tutorial_name>"
+    single_tutorial_dir = joinpath(@__DIR__, "transient_dir")
+    source_dir = single_tutorial_dir
+end
 bib = CitationBibliography(joinpath(@__DIR__, "bibliography.bib"))
 
 @everywhere push!(LOAD_PATH, joinpath(@__DIR__, ".."))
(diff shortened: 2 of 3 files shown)


docs/make.jl
```diff
@@ -5,6 +5,13 @@ ENV["CLIMATEMACHINE_SETTINGS_DISABLE_GPU"] = true
 # avoid problems with Documenter/Literate when using `global_logger()`
 ENV["CLIMATEMACHINE_SETTINGS_DISABLE_CUSTOM_LOGGER"] = true
 
+run_single_tutorial = false
+run_single_tutorial = run_single_tutorial && isempty(get(ENV, "CI", "")) # never use remotely
+run_single_tutorial && (tutorial_string_match = "tendency_specification_layer")
```
kpamnany, line 10: Is this a particular tutorial? If so, it might be too hacky for me.

.gitignore
```diff
@@ -13,6 +13,7 @@ docs/build/
 docs/site/
 docs/src/tutorials/
 docs/src/generated/
+docs/DUMMY_DIR/generated/
```
kpamnany, line 16: Let's not call this `DUMMY_DIR`? Maybe `transient` or something?

docs/make.jl
```diff
@@ -5,6 +5,13 @@ ENV["CLIMATEMACHINE_SETTINGS_DISABLE_GPU"] = true
...
+run_single_tutorial && (single_tutorial_dir = joinpath(@__DIR__, "DUMMY_DIR"))
+source_dir =
+    run_single_tutorial ? single_tutorial_dir : joinpath(@__DIR__, "src")
+
```
kpamnany, line 14:
```suggestion
run_single_tutorial = false && isempty(get(ENV, "CI", "")) # never use remotely
source_dir = joinpath(@__DIR__, "src")
if run_single_tutorial
    tutorial_string_match = "tendency_specification_layer"
    single_tutorial_dir = joinpath(@__DIR__, "DUMMY_DIR")
    source_dir = single_tutorial_dir
end
```

kpamnany (approved):
It _is_ hacky, but I assume there's a need.

docs/make.jl
charleskawczynski: Yes, it is, this is what users can change to grep the tutorial they're testing

.gitignore
charleskawczynski: that's fine by me

docs/make.jl
charleskawczynski: Should we set `tutorial_string_match = ""` by default or something?

.gitignore
charleskawczynski: Changed to `transient_dir`

charleskawczynski:
> It is hacky, but I assume there's a need.

Yeah, I just started working on a new tutorial and I ended up commenting out tens to hundreds of lines of code-- all of which need to be undone for the PR-- just to make the iteration process quicker.

docs/make.jl
kpamnany: Maybe `"<tutorial_name>"`? Your choice. :)

PR #2189, opened 2021-04-22
soil water params functions of aux

kmdeck:
### Description
This PR

- moves Ksat, specific storage (S_s) and the residual water content from `soil.param_functions` into `soil.water.param_functions`. 
- Allows for Ksat, S_s, theta_r, and hydraulic model params to be functions or floats when defined, converts them internally into functions (i.e. either the function that was originally passed or else something like `(aux)->defined_value`, and then in source code evaluates those functions of aux in the balance law functions when they are needed (i.e. at a point). This will be useful when modeling soil with different properties as a function of space (aux.x, aux.y, aux.z).
- updates tests to reflect this.


kmdeck:
Jake's advice: 
dropping the union for scalars, providing a scalar constructor that passes through a constant function for all values, and lifting the functions into the type parameter should work

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
bors cancel

kmdeck:
bors try

kmdeck:
to do: check documentation and hydraulic functions tutorial, make sure cluster tests pass.

kmdeck:
bors try

kmdeck:
bors try

kmdeck:
PS - lots of files changed, but the main ones to look at are `src/Land/Model/soil_water.jl` and `src/Land/Model/SoilWaterParameterizations.jl`. Everything else is updating tests and tutorials and docs.

kmdeck:
> If possible, it could be clearer to include all the water parameters and functions in the SoilWaterParameterizations.jl (like the WaterParamFunctions which are in soil_model). Not sure if this would take a lot of time to change, but could make it more organized, easier to understand and change the model for an outside, if all the parameters are in one place, and the model in another. It could also be good maybe to have one script which converts all the input data into the right type, instead of converting each parameter individually inside their functions. Not sure if possible? These could be changes for future, I think this is a good pr and will allow us to do benchmark

Hey Anna,

Good questions - we had talked about this briefly before, and how there are kind of two options (SoilParams only, or SoilParams for shared parameters, HeatParams for heat only, and WaterParams for water only). Originally all the params (except vG) were in SoilParams, this PR changed it by creating WaterParams. I could definitely change it back! this would be the place to do it instead of making it here only to undo it later.

src/Land/Model/soil_water.jl
```diff
@@ -171,29 +214,27 @@ function soil_init_aux!(
...
-        soil.param_functions.θ_r,
-    )
-    ψ = pressure_head(
-        water.hydraulics,
-        soil.param_functions,
-        water.initialϑ_l(aux),
-        water.initialθ_i(aux),
-    )
+    θ_r = soil.water.param_functions.θ_r(aux)
+    ν = soil.param_functions.porosity #(aux)
```
jakebolewski, line 218: Is there no need to soil change porosity as a function of soil depth / space?

jakebolewski (approved):

src/Land/Model/soil_water.jl
kmdeck: no, youre right that we need that too. We need to change all the soil parameters, but I started with just the ones that are used by the hydrology model. Heat model parameters and porosity (used by both heat and water) will come after. 

on a related note, prior to this PR, all soil parameters were in a single object (`SoilParamFunctions`), but we thought it might be cleaner to split them into heat parameters, water parameters, and joint parameters (which I started to do here). But maybe it is simpler to keep them all in one spot (undo that particular change before merging), and change all params to be functions of aux, all at once, by extending this PR to heat params and porosity too 🤷‍♀️

PR #2218, opened 2021-07-17
A draft of the aerosol activation parameterization.

trontrytel:
### Description

This parameterization computes the number of aerosol particles to be turned into cloud droplets. Implemented during SURF by @Montu12345 and @imdula. 

We are still working on docs and testing but @charleskawczynski could you take a look and tell us what you think about the code structure?

### DIFF
--- a/src/Atmos/Parameterizations/CloudPhysics/AerosolModel.jl
+++ b/src/Atmos/Parameterizations/CloudPhysics/AerosolModel.jl
@@ -1,0 +1,39 @@
+"""
+
+
+"""
+
+module AerosolModel
+
+export mode
+export aerosol_model
+
+# individual aerosol mode struct
+# TODO:
+# the first 3 lines will probably model variables
+# the rest are constants depending on aerosol type and
+# could be packed into something like chemical composition struct
+# TODO - should we name the elements?
+struct mode{T}
+    r_dry::Real
+    stdev::Real
+    N::Real
+    mass_mix_ratio::T
+    soluble_mass_frac::T
+    osmotic_coeff::T
+    molar_mass::T
+    dissoc::T
+    aerosol_density::T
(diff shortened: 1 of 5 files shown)


src/Atmos/Parameterizations/CloudPhysics/AerosolModel.jl
```diff
@@ -0,0 +1,39 @@
+"""
+
+
+"""
+
```
charleskawczynski, line 5:
```suggestion
```

test/Atmos/Parameterizations/AerosolActivation/runtests.jl
```diff
@@ -0,0 +1,368 @@
...
+
+    println("----------")
+    println("mean_hygroscopicity: ")
+    println(tp_mean_hygroscopicity(param_set, AM_1))
+    println(mean_hygroscopicity(param_set, AM_1))
+
+    for AM in AM_test_cases
+        @test all(
+            tp_mean_hygroscopicity(param_set, AM) .≈
+            mean_hygroscopicity(param_set, AM),
```
charleskawczynski, line 328: It looks like these tests involve copies of the functions in `tests/` and comparing against `src/`. Is there a better way we can test this?

trontrytel: We are working on some sanity checks and tests comparing with the original paper. 

The functions in tests were implemented independently from the source. But I agree we need better tests than just that.

src/Atmos/Parameterizations/CloudPhysics/AerosolModel.jl
charleskawczynski: This will break documenter…

PR #2227, opened 2021-09-13
Upgrade climaparameters, thermodynamics, and microphysics

charleskawczynski:
### Description

This PR upgrades CLIMAParameters, Thermodynamics.jl, and Microphysics, which need to be done together.


Manifest.toml
```diff
@@ -8,20 +8,18 @@ version = "1.0.1"
```
trontrytel, line 8: Why do we track Manifest.toml files here?

Project.toml
```diff
@@ -63,7 +63,7 @@ Adapt = "2.0.2, 3.2"
 ArgParse = "1.1"
 ArtifactWrappers = "0.1.1"
 BenchmarkTools = "0.5"
-CLIMAParameters = "0.1"
+CLIMAParameters = "0.2"
```
trontrytel, line 66: Shouldnt we also have CloudMicrophysics = "0.2" in the Project.toml?

docs/src/HowToGuides/Atmos/MoistureAndPrecip.md
```diff
@@ -69,27 +69,8 @@ using CLIMAParameters
 using CLIMAParameters.Planet
 using CLIMAParameters.Atmos.Microphysics
```
trontrytel, line 71: We should probably also update the documentation text above, since our structure is much simpler now. Maybe something like:

The microphysics parameterizations used to compute the sources and sinks for
moisture and precipitation variables  depends on 
[CLIMAParameters.jl](https://github.com/CliMA/CLIMAParameters.jl).
A set with all the parameter values needs to be created and passed in as one
of the arguments.

trontrytel (approved):
Thank you for updating this! Left some minor comments.

Manifest.toml
charleskawczynski: This is a [good thread](https://discourse.julialang.org/t/does-manifest-toml-belong-in-the-repository/12029) on when it should be tracked or not.

Project.toml
charleskawczynski: This _is_ the Project.toml. Do you mean the docs? If so, then yes 😂.

trontrytel: I thought that ClimateMachine.jl now depends on CloudMicrophysics.jl. So I thought it should actually be in the dependencies here?

Manifest.toml
trontrytel: Oh - makes sense. Thanks!

Project.toml
trontrytel: i.e. we need CloudMicrophysics.jl to specify some of the multiphysics tendencies?

charleskawczynski: https://github.com/CliMA/ClimateMachine.jl/blob/ck/up_climaparams/Project.toml#L13

charleskawczynski:
Ugh, I forgot that this needs to be done on Julia 1.5.4. Updating...

charleskawczynski:
I realized that I needed to add the thermo interface updates...

